PR #1203, opened 2014-12-17
REP-1272 HERP Filter

dmnjohns:


repose-aggregator/components/filters/herp/src/main/scala/org.openrepose.filters.herp/HerpFilter.scala
```diff
@@ -0,0 +1,135 @@
...
+      "tenantID" -> httpServletRequest.getHeader(OpenStackServiceHeader.TENANT_ID.toString),
+      "roles" -> httpServletRequest.getHeaders(OpenStackServiceHeader.ROLES.toString).asScala.map(Role(_)).toArray,
+      "userAgent" -> httpServletRequest.getHeader(CommonHttpHeader.USER_AGENT.toString),
+      "requestMethod" -> httpServletRequest.getMethod,
+      "requestURL" -> Option(httpServletRequest.getAttribute("http://openrepose.org/requestUrl")).map(_.asInstanceOf[String]).orNull,
+      "parameters" -> translateParameters().asJava.entrySet(),
+      "timestamp" -> System.currentTimeMillis(),
+      "responseCode" -> httpServletResponse.getStatus,
+      "responseMessage" -> HttpStatusCode.fromInt(httpServletResponse.getStatus).name(),
+      // todo: "responseBody" -> ??? // wrap the request in mutable request to get the body
```
dmnjohns: Kind of? It wasn't a part of the acceptance criteria for this story, but it is something we *may* want to log. I mostly left the line in there for the comment since things get a little messier when you have to read the response body.

We can kill the line if desired.

adrianjgeorge:
retest this please

repose-aggregator/components/filters/herp/src/test/scala/org.openrepose.filters.herp/HerpFilterTest.scala
```diff
@@ -0,0 +1,335 @@
...
+      """.stripMargin
+    val template = new Template
+    template.setValue(templateText)
+    template.setCrush(true)
+    herpConfig.setTemplate(template)
+  }
+
+  describe("the doFilter method") {
+    it("should be empty if field data is not present") {
+      // when:
```
dmnjohns: Hah, I was looking for a ScalaTest trait that supported parameterized testing is a nice way, but what I found was kind of ugly. What would these tests look like using the FeatureSpec trait?

repose-aggregator/components/filters/herp/src/main/scala/org.openrepose.filters.herp/HerpFilter.scala
```diff
@@ -0,0 +1,135 @@
...
+  override def destroy(): Unit = {
+    logger.trace("HERP filter destroying ...")
+    configurationService.unsubscribeFrom(config, this.asInstanceOf[UpdateListener[_]])
+
+    logger.trace("HERP filter destroyed.")
+  }
+
+  override def configurationUpdated(config: HerpConfig): Unit = {
+    def template :StringReader = {
+      var  templateText = config.getTemplate.getValue.trim
```
dmnjohns: What happened with the formatting here? Double space, colon prepended to the type...

${repose.herp.log.name}
```diff
@@ -0,0 +1,448 @@
...
+
+2014-12-18 10:36:10,554 5375 [repose:node] WARN  org.openrepose.commons.config.parser.jaxb.JaxbConfigurationParser  - This *MIGHT* be important! Unable to read configuration file: /home/jennyvo/repose/repose-aggregator/functional-tests/spock-functional-test/target/repose_home/configs/dist-datastore.cfg.xml (No such file or directory)
+2014-12-18 10:36:10,554 5375 [repose:node] ERROR org.openrepose.core.services.config.impl.PowerApiConfigurationManager  - Configuration update error. Reason: Parsed object from XML does not match the expected configuration class. Expected: org.openrepose.core.services.datastore.distributed.config.DistributedDatastoreConfiguration  -  Actual: null
+2014-12-18 10:36:10,559 5380 [repose:node] INFO  org.openrepose.core.services.datastore.distributed.impl.distributed.jetty.DistributedDatastoreLauncherServiceImpl  - Launching Datastore servlet on port: 0
+2014-12-18 10:36:10,560 5381 [repose:node] INFO  org.eclipse.jetty.server.Server  - jetty-9.2.z-SNAPSHOT
+2014-12-18 10:36:10,562 5383 [repose:node] INFO  org.openrepose.services.healthcheck.impl.HealthCheckServiceImpl  - HealthCheckService.reportIssue: DistDatastoreCo
```
dmnjohns: Get rid of this file, please.

dkowis: Yeah this really needs to be rebased out so we don't clutter the history.

dmnjohns:
It looks like tests failed because the filter-bundle project could not find the default-router snapshot dependency in nexus. That seems like an expected issue at the moment.

dkowis:
On 01/09/2015 07:44 AM, Damien Johnson wrote:
> It looks like tests failed because the filter-bundle project could not
> find the default-router snapshot dependency in nexus. That seems like an
> expected issue at the moment.

Not really. It probably means that we didn't remove a dependency from a
something...

dmnjohns:
Oh, is default router something we removed from the project altogether? Yeah, you're probably right then. I'll check it out.

adrianjgeorge:
we removed it recently, i thought this branch was based before that removal though?

${repose.herp.log.name}
```diff
@@ -0,0 +1,448 @@
+2014-12-18 10:32:03,469 841  [main] INFO  org.eclipse.jetty.util.log  - Logging initialized @831ms
```
adrianjgeorge: Pretty sure this isn't supposed to be here.

dkowis:
Perhaps. But any snapshot dependencies get installed into maven local, during a multi-module build, not to the nexus server....

adrianjgeorge:
yes, but the reactor should be smart enough to order them in such a way that, a new one gets generated for this build before an old one would get picked up in a later part.

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/filters/herp/HerpSimpleTest.groovy
```diff
@@ -0,0 +1,182 @@
...
+    def cleanupSpec() {
+        if (deproxy) {
+            deproxy.shutdown()
+        }
+
+        if (repose) {
+            repose.stop()
+        }
+    }
+    @Ignore
```
adrianjgeorge: intentionally still here?

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/filters/herp/HerpSimpleTest.groovy
```diff
@@ -0,0 +1,182 @@
...
+        messageChain.receivedResponse.code == "200"
+        checkAttribute(jsonpart, listattr)
+        result.ServiceCode == "repose"
+        result.Region == "USA"
+        result.DataCenter == "DFW"
+        result.Request.Method == "GET"
+        result.Response.Code == 200
+        result.Response.Message == "OK"
+    }
+    @Ignore
```
adrianjgeorge: also still ignored

jennyvo: oops, will remove it.

repose-aggregator/components/filters/herp/src/main/scala/org.openrepose.filters.herp/HerpFilter.scala
```diff
@@ -0,0 +1,134 @@
...
+          parametersMap.mapValues((values :Array[String]) => values.map(Parameter(_))).toMap
+        }
+        case None => Map[String, Array[Parameter]]()
+      }
+    }
+
+    val templateValues = Map(
+      "userName" -> httpServletRequest.getHeader(OpenStackServiceHeader.USER_NAME.toString),
+      "impersonatorName" -> httpServletRequest.getHeader(OpenStackServiceHeader.IMPERSONATOR_NAME.toString),
+      "tenantID" -> httpServletRequest.getHeader(OpenStackServiceHeader.TENANT_ID.toString),
```
dmnjohns: Shouldn't the projectID be mapped here in case we're logging Keystone V3 requests?

Also, to parrot @adrianjgeorge, we should probably log all tenant/project IDs and not just the first to be returned.

adrianjgeorge: i'd say it should populate the map under one title with one or the other whichever it gets a hit on, so the people writing the template only have one field to worry about.

adrianjgeorge: given that reasoning go with projectid, so we don't have to change it a couple months down the line.

repose-aggregator/components/filters/herp/src/main/scala/org.openrepose.filters.herp/HerpFilter.scala
```diff
@@ -0,0 +1,134 @@
...
+      "roles" -> httpServletRequest.getHeaders(OpenStackServiceHeader.ROLES.toString).asScala.map(Role(_)).toArray,
+      "userAgent" -> httpServletRequest.getHeader(CommonHttpHeader.USER_AGENT.toString),
+      "requestMethod" -> httpServletRequest.getMethod,
+      "requestURL" -> Option(httpServletRequest.getAttribute("http://openrepose.org/requestUrl")).map(_.toString).orNull,
+      "parameters" -> translateParameters().asJava.entrySet(),
```
dmnjohns: We should, at some point, break this up into encoded/decoded parameters.

Jorge's suggests we provide the raw query string as one key, and the decoded query map as another.

repose-aggregator/components/filters/herp/src/test/scala/org.openrepose.filters.herp/HerpFilterTest.scala
```diff
@@ -0,0 +1,335 @@
...
+      servletRequest.setAttribute("http://openrepose.org/queryParams", Map("foo" -> Array("bar", "baz")).asJava)
+
+      // when:
+      herpFilter.configurationUpdated(herpConfig)
+      herpFilter.doFilter(servletRequest, servletResponse, filterChain)
+
+      // then:
+      def logEvents = listAppender.getEvents
+      logEvents.size shouldBe 1
+      logEvents.get(0).getMessage.getFormattedMessage should include("\"Parameters\" : {  \"foo\" : [  \"bar\",  \"baz\",  ],  }")
```
dmnjohns: It this valid JSON? The last value of the array still has a ',' appended to it.

adrianjgeorge: json allows trailing commas, because awesome. or at least i recall it allowing them

adrianjgeorge: never do annoying special cases for the first/last case ever again

dmnjohns: That would be lovely, but it doesn't look like the JSON spec itself allows for that.

http://stackoverflow.com/a/201856
http://www.json.org/

adrianjgeorge: crap, stack overflow disagrees with me

dmnjohns: There isn't a sad face emoticon available in Github. I am extremely disappointed.

adrianjgeorge: yeah, i just found that. I wonder why i though it did... That makes things much harder, because we don't have that kind of fine control in mustache, at least not in a way i can think of.

adrianjgeorge: Ugh, the other solution is to wrap anything like this in an object that does a to string with the correct values, but that limits us, because then we are enforcing json syntax into it.

adrianjgeorge:
may need to switch templating engines to something with more flexibilty

repose-aggregator/components/filters/herp/src/test/scala/org.openrepose.filters.herp/HerpFilterTest.scala
dkowis: http://stackoverflow.com/questions/201782/can-you-use-a-trailing-comma-in-a-json-object

the JSON spec does not allow trailing commas. I would not leave the trailing comma.

oh, you just found all this. /me reads up DERP

adrianjgeorge:
you should change the docs in the xsd too, it lists the available fields

wdschei:
Retest this please.

dmnjohns:
@jennyvo It looks like the failing test is expecting an encoded parameter, but parameters are now decoded.

The raw query string can be parsed from the requestQueryString field. We may want to add that to the functional verification?

adrianjgeorge:
retest this please

adrianjgeorge:
cant be merged

PR #1210, opened 2014-12-31
Rep 1527 - Forwarded Proto Header Filter

InblEric:
From the acceptance criteria:

X-FORWARDED-PROTO: http 
X-FORWARDED-PROTO: https

Looks like http actually comes in as HTTP/1.1 but I don't think this should be a problem as long as whoever uses the filter knows what they're looking for. I assume https would come in as HTTPS but I'm not entirely sure.


wdschei:
Retest this please.

adrianjgeorge:
why would their snapshots be in central, but not the actual releases? that doesn't seem right

adrianjgeorge:
there's no config to speak of, so lets not do the factory and all that other crap. Justa plain simple filter should do the job. also we should check the output and make sure it's consistent with nginx?

dkowis:
@adrianjgeorge this is why: https://github.com/rackerlabs/repose/pull/1124#discussion_r18177949 They're not publishing releases there yet.

dmnjohns:
Closing this. Refer to #1215

PR #1215, opened 2015-01-05
REP-1527 Forwarded Proto Filter

dmnjohns:
I refactored @InblEric's code a bit to drop some unnecessary scaffolding and generally clean things up a bit.

The original PR is at #1210


repose-aggregator/components/filters/forwarded-proto/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+            <artifactId>utilities</artifactId>
+        </dependency>
+
+        <!-- Mockrunner for mock HttpServletRequest -->
+        <dependency>
+            <groupId>com.mockrunner</groupId>
+            <artifactId>mockrunner-servlet</artifactId>
+            <version>1.0.0</version>
+            <scope>test</scope>
+        </dependency>
```
dmnjohns: I bet you can kill this dependency and just use a Mockito mock for the simple interactions necessary with the HttpServletRequest.

dmnjohns: Actually, I wouldn't worry about this. We pull in mockrunner for testing elsewhere, and removing it makes things a little messier. I guess it's just a matter of preference.

repose-aggregator/components/filters/forwarded-proto/src/main/scala/org/openrepose/filters/forwardedproto/ForwardedProtoFilter.scala
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  override def doFilter(servletRequest: ServletRequest, servletResponse: ServletResponse, filterChain: FilterChain): Unit = {
+    val httpServletRequest = servletRequest.asInstanceOf[HttpServletRequest]
+
+    if (Option(httpServletRequest.getHeader(X_FORWARDED_PROTO)).isEmpty) {
+      logger.debug(s"Adding the $X_FORWARDED_PROTO header")
+
+      val mutableHttpServletRequest = MutableHttpServletRequest.wrap(httpServletRequest)
+      mutableHttpServletRequest.addHeader(X_FORWARDED_PROTO, servletRequest.getProtocol)
```
adrianjgeorge: so we are assuming that's going to give us the appropriate value?

dmnjohns: Dunno. So what we get from that call is detailed here: http://docs.oracle.com/javaee/6/api/javax/servlet/ServletRequest.html#getProtocol()

X-Forwarded-Proto is an extension header, so I don't believe that there is a formal specification for it. There may be standards for it, though. It appears that the "Forwarded" header has been specified in http://tools.ietf.org/html/rfc7239 and is meant as a substitute for the X-Forwarded-Proto header.

I don't expect the value overly matters as long as the origin service knows what to expect. With that said, I'm fine changing the header name, value, or both.

adrianjgeorge: So my understanding is we are trying to mimic the behavior of nginx in this regard.

dmnjohns: Ah, that's doable. So just straight "HTTP" or "HTTPS", then?

adrianjgeorge: sounds right

PR #1216, opened 2015-01-06
REP-1264 Classloader tests

dkowis:
Can be merged into the spring branch when completed.


dkowis:
I merged this already, so that we can move forward. it's done

PR #1219, opened 2015-01-08
REP-1560 Header quality to differentiate default tenant/project ID

dmnjohns:


repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/openstack/OpenStackAuthenticationHeaderManager.java
```diff
@@ -144,14 +146,22 @@ private void setImpersonator() {
     private void setTenant() {
         filterDirector.requestHeaderManager().putHeader(OpenStackServiceHeader.TENANT_NAME.toString(), cachableToken.getTenantName());
         if(!this.isDelagable && !this.isTenanted) {
-            filterDirector.requestHeaderManager().putHeader(OpenStackServiceHeader.TENANT_ID.toString(), cachableToken.getTenantId());
+            if (sendTenantIdQuality) {
```
adrianjgeorge: no unit tests to prove this change works or was even neccessary? Someone isn't doing tdd.

repose-aggregator/components/filters/openstack-identity-v3/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3Handler.scala
```diff
@@ -134,36 +134,38 @@ class OpenStackIdentityV3Handler(identityConfig: OpenstackIdentityV3Config, iden
...
+          project.name.foreach { projectName =>
+            requestHeaderManager.putHeader(OpenStackIdentityV3Headers.X_PROJECT_NAME.toString, projectName)
           }
         }
-        token.get.rax_impersonator.map { impersonator =>
-          impersonator.id.map(requestHeaderManager.putHeader(OpenStackIdentityV3Headers.X_IMPERSONATOR_ID.toString, _))
-          impersonator.name.map(requestHeaderManager.putHeader(OpenStackIdentityV3Headers.X_IMPERSONATOR_NAME.toString, _))
+        projectIdUriRegex match {
+          case Some(regex) =>
+            writeProjectHeader(token.flatMap(_.project.flatMap(_.id)), token.flatMap(_.roles).getOrElse(List[Role]()),
```
adrianjgeorge: Nope, this is officially to big of a mess to read

repose-aggregator/components/filters/openstack-identity-v3/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3Handler.scala
```diff
@@ -251,17 +253,38 @@ class OpenStackIdentityV3Handler(identityConfig: OpenstackIdentityV3Config, iden
...
-      } else {
-        Set.empty
+  private def writeProjectHeader(defaultProject: Option[String], roles: List[Role], projectFromUri: Option[String],
+                                 writeAll: Boolean, sendQuality: Boolean, headerManager: HeaderManager) {
+    lazy val projectsFromRoles = (roles.collect { case Role(_, _, Some(projectId), _, _, _) => projectId} :::
+      roles.collect { case Role(_, _, _, Some(raxId), _, _) => raxId}).toSet
+
+    if (writeAll && sendQuality) {
+      defaultProject.foreach(headerManager.appendHeader(OpenStackIdentityV3Headers.X_PROJECT_ID, _, 1.0))
+      projectsFromRoles foreach { rolePid =>
```
adrianjgeorge: You have the same code here twice, the difference is the quality, one is 0.5, the other is default (1.0). Maybe extract into a method that takes the quality? The rule of thumb is the third time you write a piece of code, but i always start looking at two. I'm not saying you have to do this, just something to think about.

repose-aggregator/components/filters/openstack-identity-v3/src/test/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3HandlerTest.scala
```diff
@@ -291,35 +291,56 @@ class OpenStackIdentityV3HandlerTest extends FunSpec with BeforeAndAfter with Ma
...
     val roles = List(Role(null, null, Option("12345"), null, null), Role(null, null, Option("67890"), null, null))
 
     it("should only provide the url project when the flag says to not write all") {
-      val randomness = java.util.UUID.randomUUID.toString
-      identityV3Handler invokePrivate writeProjectHeader(randomness, roles, false, filterDirector)
-      verify(headerManager).appendHeader(org.mockito.Matchers.eq("X-Project-ID"), org.mockito.Matchers.eq(randomness))
+      val defaultPid = java.util.UUID.randomUUID.toString
+      val uriPid = java.util.UUID.randomUUID.toString
+      identityV3Handler invokePrivate writeProjectHeader(Some(defaultPid), roles, Some(uriPid), false, false, headerManager)
+      verify(headerManager).appendHeader(org.mockito.Matchers.eq("X-Project-ID"), org.mockito.Matchers.eq(uriPid))
```
adrianjgeorge: Couldn't statically import for some reason? Already one in scope?

repose-aggregator/components/filters/openstack-identity-v3/src/test/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3HandlerTest.scala
```diff
@@ -291,35 +291,56 @@ class OpenStackIdentityV3HandlerTest extends FunSpec with BeforeAndAfter with Ma
...
+      val uriPid = java.util.UUID.randomUUID.toString
       val roleList = List(Role(null, null, None, None, None), Role(null, null, Option("foo"), None, None), Role(null, null, None, None, None))
-      identityV3Handler invokePrivate writeProjectHeader(randomness, roleList, true, filterDirector)
-      verify(headerManager).appendHeader(org.mockito.Matchers.eq("X-Project-ID"), org.mockito.Matchers.eq("foo"), org.mockito.Matchers.eq(randomness))
+      identityV3Handler invokePrivate writeProjectHeader(Some(defaultPid), roleList, Some(uriPid), true, false, headerManager)
+      verify(headerManager).appendHeader(org.mockito.Matchers.eq("X-Project-ID"), org.mockito.Matchers.eq(defaultPid))
+      verify(headerManager).appendHeader(org.mockito.Matchers.eq("X-Project-ID"), org.mockito.Matchers.eq("foo"))
+    }
+
+    it("should add qualities when flagged and not writing all") {
```
adrianjgeorge: I'm not sure we want this flag to be usable without the write all, it has no real value. I think it might be worth restricting so customers don't shoot themselves in the foot somehow.

dmnjohns: I think we discussed this point while you were out of the room on Tuesday and decided that the two flags in question could be independent.

I don't mind restricting usage, but in my opinion, adding a quality shouldn't break anything to begin with, but even if it would break something, we default to false, so a user would have to do work to break themselves. Could that happen? Absolutely. The alternative is more restrictive, though, and requires that we could attributes in some way (XSD assert, which only works with XML 1.1 validators, or in the code... both ways are a bit messy).

repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/openstack/OpenStackAuthenticationHeaderManager.java
dmnjohns: Busted. I started to write tests then wept at the current state of testing around this class.

I'll go add these now.

adrianjgeorge: I will also accept altering the functional tests to look for it if it's to big of a pile.

repose-aggregator/components/filters/openstack-identity-v3/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3Handler.scala
dmnjohns: It was kind of like that before... though I did have to add another method argument.

I can pull the flapMap calls out and stick them into vals for readability. Same with the extractProjectIdFromUri(...) call. Would that be enough?

Part of the problem is that we pass config values to make testing easier, but there really isn't a requirement that they be passed. The config is a field, after all.

dmnjohns: Honestly, there are a bunch of cases to handle here. That's why I had trouble figuring out what was going on when I went to make changes.

repose-aggregator/components/filters/openstack-identity-v3/src/test/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3HandlerTest.scala
adrianjgeorge: So currently it won't break anything, but that doesn't mean it'll always be the case, and i've seen many customers have random crap turned on in there config because they are having problems getting something to do what they expect and just start doing random crap without understanding the consequences. If we can safegaurd against i'm all for that. Again a decision i don't agree with while i'm not present.

repose-aggregator/components/filters/openstack-identity-v3/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3Handler.scala
dmnjohns: Fair enough. I'll create a nested method to use.

repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/openstack/OpenStackAuthenticationHeaderManager.java
dmnjohns: Then I will take the easy way out and write functional tests. :+1: 

I'm just waiting for the day where we can nuke the Keystone V2 code.

repose-aggregator/components/filters/openstack-identity-v3/src/test/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3HandlerTest.scala
dmnjohns: So I don't exactly know what the issue is. My guess is that Scala's AnyRef eq is interfering. When I try to find the source of the unqualified eq() function, I just find that it's a synthetic function.

http://www.scala-lang.org/api/current/index.html#scala.AnyRef

repose-aggregator/components/filters/openstack-identity-v3/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3Handler.scala
adrianjgeorge: Yeah making seperate well named values would probably go a long way towards making it more readable

repose-aggregator/components/filters/openstack-identity-v3/src/test/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3HandlerTest.scala
adrianjgeorge: about what i expected, super ugly that we have to do the fully qualified name.

dmnjohns: Hm, actually, I wonder if we could just assign a variable to point to that function and use the variable? Something like:

val mockitoEq = org.mockito.Matchers.eq

adrianjgeorge: ohh, that's hot, and would probably work, first class functions for the win. Although you'll probably have to do def not val

dmnjohns: So that doesn't quite work, but there is a better alternative.

import org.mockito.Matchers.{eq => mockitoEq}

repose-aggregator/components/filters/openstack-identity-v3/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3Handler.scala
dmnjohns: Actually, I kind of don't like how it looks once I started doing it. If you're fine with leaving it as-is for now, then I think that's what I'll do.

repose-aggregator/components/filters/openstack-identity-v3/src/test/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3HandlerTest.scala
```diff
@@ -3,12 +3,13 @@ package org.openrepose.filters.openstackidentityv3
 import java.util
 
 import com.mockrunner.mock.web.{MockHttpServletRequest, MockHttpServletResponse}
-import org.openrepose.core.filter.logic.{FilterAction, FilterDirector, HeaderManager}
 import org.junit.runner.RunWith
+import org.mockito.Matchers.{eq => mockitoEq}
```
adrianjgeorge: reason #205 why scala is kind of awesome: aliasing of function names in the import

repose-aggregator/external/service-clients/auth/src/main/java/org/openrepose/common/auth/openstack/OpenStackToken.java
```diff
@@ -31,7 +31,7 @@
     private final String defaultRegion;
     private static final Logger LOG = LoggerFactory.getLogger(OpenStackToken.class);
     private static final QName REGION_QNAME = new QName("http://docs.rackspace.com/identity/api/ext/RAX-AUTH/v1.0", "defaultRegion", "prefix");
-    private final List<String> tenantIds;
+    private final Set<String> tenantIds;
```
adrianjgeorge: This seems sketch, does the contract say it's a set?

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/filters/identityv3/multiProjectIdswQuality/MultiProjectIdsWQualityTest.groovy
```diff
@@ -0,0 +1,117 @@
...
+        where:
+        defaultProject  | secondProject   | reqProject      | clientToken       | serviceRespCode   | numberProjects
+        "123456"        | "test-project"  | "123456"        |UUID.randomUUID()  | "200"             | 2
+        "test-project"  | "12345"         | "12345"         |UUID.randomUUID()  | "200"             | 2
+        "test-project"  | "12345"         | "test-project"  |UUID.randomUUID()  | "200"             | 2
+        "123456"        | "123456"        | "test-proj-id"  |UUID.randomUUID()  | "401"             | 1
+        "123456"        | "test-project"  | "openstack"     |UUID.randomUUID()  | "401"             | 2
+    }
+
+    @Unroll ("No project id form token object: request project #reqProject")
```
adrianjgeorge: from not form?

PR #1220, opened 2015-01-08
Removing redundant groupIds and dependencies from poms

dmnjohns:


repose-aggregator/components/filters/translation/pom.xml
```diff
@@ -87,11 +86,6 @@
             <artifactId>xalan</artifactId>
             <version>2.7.1</version>
         </dependency>
-        <dependency>
```
adrianjgeorge: this one seems sketch? doesn't it use commons pool?

dmnjohns: It does, but commons pool is an inherited dependency.

adrianjgeorge: if it's using it directly, it keeps it's dependency. we do not count on transitive dependencies. ever.

adrianjgeorge: for the record from a parent pom is different, because that is effectively a direct dependency. but a dependency of our dependency is not kosher

dmnjohns: I understand that.

I lied anyway, though. It's actually dependency managed in a parent pom, but the dependency is duplicated in this pom. So I just removed the duplicate dependency.

adrianjgeorge: dependency managed doesn't make it our dependency, it just says what one we'll get if we do count on the dependency

dmnjohns: Again, I understand that.

Just take a look at the file, eh?

adrianjgeorge: ahh, i see it was in the same pom twice. well that's a totally legit removal then.

PR #1224, opened 2015-01-16
Fixing transitive dependencies and removing unused dependency

dmnjohns:


dmnjohns:
Oh wait, this depends on core-service-api which is only in the spring branch. I'll take that out tomorrow.

PR #1225, opened 2015-01-16
Wire-tap std-out and std-err

dkowis:
Using this dependency, we can easily build a wiretapper that will grab
anything that goes to stderr or stdout and also log it to a named
logger. This is very useful.

Because I cherry-picked it from the spring branch, it got mad. It seems most of the conflicts were whitespace conflicts, but the patch here is good.

Conflicts:
	repose-aggregator/core/core-lib/pom.xml
	repose-aggregator/core/core-lib/src/main/java/org/openrepose/core/services/logging/LoggingServiceImpl.java
	repose-aggregator/pom.xml

### DIFF
--- a/repose-aggregator/core/core-lib/pom.xml
+++ b/repose-aggregator/core/core-lib/pom.xml
@@ -210,6 +210,12 @@
             <artifactId>log4j-core</artifactId>
         </dependency>
 
+        <!-- this guy gets me some dependencies that let me log STDERR and STDOUT 
+        while still letting them go to the console -->
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-iostreams</artifactId>
+        </dependency>
         <dependency>
             <groupId>org.scala-lang</groupId>
             <artifactId>scala-library</artifactId>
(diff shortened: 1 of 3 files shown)


dkowis:
So interestingly enough, on master, this doesn't work very well, because it'll get double wrapped, because we do spring wrong there. I'll have to find a better place to put it on the master branch, but we don't want to merge that into spring. 

So maybe we don't put this on master, and I just use it for my debugging.....

dkowis:
Yeah I don't need this any more in master, unless we want to do it, at which point the solution is to: 
```java
static {
        //Do some log wrapping only one time, not every time we configure
        //Wiretap Standard Error to the STDERR logger
        PrintStream stdErr = IoBuilder.forLogger("STDERR")
                .setLevel(Level.WARN)
                .filter(System.err) //Also output to standard err, but I want it in my logs!
                .buildPrintStream();
        System.setErr(stdErr);

        PrintStream stdOut = IoBuilder.forLogger("STDOUT")
                .setLevel(Level.INFO)
                .filter(System.out)
                .buildPrintStream();
        System.setOut(stdOut);
}
```
in the logger impl class (or any class really), so it's only run once per JVM.

adrianjgeorge:
so this is already on spring branch? if so, i'll just close this in it'll come in correctly in a couple of days

dkowis:
Yeah. That's a good plan. Do that. 

On January 16, 2015 7:59:07 AM CST, Adrian J George III <notifications@github.com> wrote:
>so this is already on spring branch? if so, i'll just close this in
>it'll come in correctly in a couple of days
>
>---
>https://github.com/rackerlabs/repose/pull/1225#issuecomment-70256633

-- 

PR #1226, opened 2015-01-16
REP-1526 Akka Client NPE

wdschei:
There are a couple of notes still in the code that are questions as to how and where we want to handle some of these new bubbled up TimeoutExceptions in place of the original NPE.


repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/ClientAuthenticationHandlerFactory.java
```diff
@@ -77,7 +78,12 @@ public void configurationUpdated(ClientAuthConfig modifiedConfig) {
                     accountRegexExtractor.addPattern(clientMapping.getIdRegex());
                 }
                 if (modifiedConfig.getAtomFeeds() != null) {
-                    activateOpenstackAtomFeedListener(modifiedConfig);
+                    try {
+                        activateOpenstackAtomFeedListener(modifiedConfig);
+                    } catch (TimeoutException e) {
```
adrianjgeorge: Should this just get swallowed here? What was happening before with the potential NPE? or was there no potential before?

repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/atomfeed/FeedCacheInvalidator.java
```diff
@@ -65,7 +66,13 @@ public void run() {
...
-                CacheKeys keys = rdr.getCacheKeys();
+                CacheKeys keys = null;
+                try {
+                    keys = rdr.getCacheKeys();
+                } catch (TimeoutException e) {
+                    LOG.error("Unable get Cached Keys.");
+                    LOG.trace("", e);
+                }
 
                 userKeys.addAll(keys.getUserKeys());
```
adrianjgeorge: you just created a potential NPE, if the call ti get cache keys throws a timeout exception, this line will throw an npe.

repose-aggregator/components/filters/client-auth/src/test/java/org/openrepose/filters/clientauth/openstack/OpenStackAuthenticationHandlerTest.java
```diff
@@ -242,7 +243,7 @@ public void standUp() throws Exception {
         }
 
         @Test
-        public void tenantIdFromTokenMatchesURI() {
+        public void tenantIdFromTokenMatchesURI() throws TimeoutException {
```
adrianjgeorge: it's odd to me that theses tests didn't already say throws exception, that's what the intellij template will do, so you don't have to worry about this sort of thing. le sigh.

repose-aggregator/services/service-client/impl/src/main/java/org/openrepose/services/serviceclient/akka/impl/AkkaServiceClientImpl.java
```diff
@@ -60,29 +61,39 @@ public UntypedActor create() {
 
 
     @Override
-    public ServiceClientResponse get(String hashKey, String uri, Map<String, String> headers) {
+    public ServiceClientResponse get(String hashKey, String uri, Map<String, String> headers) throws TimeoutException {
```
adrianjgeorge: this doesn't solve the npe, it only handles it in the case of a timeout exception. if there is something else exceptional you are still going to potentially return null. This where we do what i was talking about during sprint planning (maybe it didn't get captured?). Instead of only dealing with the timeout exception, we catch all exceptions, and wrap them in an akka service client specific exception. Should the caller care to take different action depending on what the underlying exception was they can dig it out.

repose-aggregator/components/filters/client-auth/src/test/java/org/openrepose/filters/clientauth/openstack/OpenStackAuthenticationHandlerTest.java
wdschei: These have all been changed to just throws Exception.

repose-aggregator/services/service-client/impl/src/main/java/org/openrepose/services/serviceclient/akka/impl/AkkaServiceClientImpl.java
wdschei: This has now been wrapped in a new `AkkServiceClientException`. In the couple of places that are setting the Response status code, the cause is checked to see if it is an `instanceof TimeoutException`. If it is a `TimeoutException`, then the Response status code is set to `GATEWAY_TIMEOUT`(504); else it is set to `INTERNAL_SERVER_ERROR`(500).

repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/atomfeed/FeedCacheInvalidator.java
wdschei: Expanded the try/catch block to encompass the potential NPE that was created.

repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/ClientAuthenticationHandlerFactory.java
wdschei: This was originally just bubbling up the NPE as an unchecked `Exception`. If we don't consume it here, then we have to add it to the `UpdateListener` interface. It just seems kind of strange to burden all `UpdateListener`'s for the behavior of this one implementor. Should we re-throw an unchecked exception here? That seems kind of odd too.

wdschei: Upon further review of this, the NPE was being caught at `PowerApiConfigurationManager.subscribeTo()` (ln:123) and `PowerApiUpdateManagerEventListener.onEvent()` (ln:49) depending on exactly when in the Repose life-cycle it was thrown. In both cases the catching of the NPE is what triggers marking the filter as Failed To Load before eventually logging the NPE.

To consume it here runs the risk of the filter not being marked as Failed To Load. If we don't throw an un-checked exception, then we have to burden all of the existing implementations with guarding against the possibility of a new checked exception being thrown.

I can add a new `UncheckedAkkaServiceClientException` that extends `RuntimeException` and have it wrap the `AkkaServiceClientException` or it's `Cause` in order to let it bubble to the top?

I heard you like wrappers for your wrappers, because wrapping is good!?!?

repose-aggregator/services/service-client/api/src/main/java/org/openrepose/services/serviceclient/akka/AkkaServiceClientException.java
```diff
@@ -0,0 +1,19 @@
+package org.openrepose.services.serviceclient.akka;
+
+public class AkkaServiceClientException extends Exception {
```
adrianjgeorge: Do you use all these constructors?

wdschei: No. The only one used is the AkkaServiceClientException(String, Throwable). It was just the standard custom Exception template.

Do you want me to remove the other three?

adrianjgeorge: Yeah, we'll add the others as we need them, because they are probably not getting coverage right now either.

adrianjgeorge:
Changes look good

repose-aggregator/services/service-client/api/src/main/java/org/openrepose/services/serviceclient/akka/AkkaServiceClientException.java
wdschei: They are removed.

repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/ClientAuthenticationHandlerFactory.java
wdschei: The caught `AkkaServiceClientException` is now wrapped in an `UpdateFailedException` and re-thrown. The `UpdateListener` interface was updated so `configurationUpdated()` now throws `UpdateFailedException`. This still bubbles it way up to the `PowerApiConfigurationManager` and `PowerApiUpdateManagerEventListener` just like the NPE used to, only this new exception is logged in a more appropriate manner.

repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/atomfeed/AuthFeedReader.java
```diff
@@ -1,7 +1,8 @@
 package org.openrepose.filters.clientauth.atomfeed;
 
+import org.openrepose.services.serviceclient.akka.AkkaServiceClientException;
+
 public interface AuthFeedReader {
-   
-   CacheKeys getCacheKeys();
-   String getFeedId();
```
adrianjgeorge: why was the other method removed?

repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/atomfeed/AuthFeedReader.java
```diff
@@ -1,7 +1,8 @@
...
+
+   CacheKeys getCacheKeys() throws AkkaServiceClientException;
```
adrianjgeorge: Also this is another case where if you need to pass out an exception it should be a localized one. I'm pretty sure that anyone using something of this type of interface isn't going to care that the implementer underneath is using akka.

repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/atomfeed/AuthFeedReader.java
wdschei: It was annotated by IntelliJ as not being used, so while I was there I cleaned it up.

repose-aggregator/external/service-clients/auth/src/main/java/org/openrepose/common/auth/openstack/AuthenticationService.java
```diff
@@ -12,11 +13,11 @@
  */
 public interface AuthenticationService {
 
-    AuthenticateResponse validateToken(String tenant, String userToken);
+    AuthenticateResponse validateToken(String tenant, String userToken) throws AkkaServiceClientException;
```
adrianjgeorge: Not sure how i missed this before, but the same deal as the other one, interfaces should be throwing other modules exceptions as a general rule. If they need to throw an exception, it should be one they define.

repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/atomfeed/AuthFeedReader.java
wdschei: The AuthFeedReader now throws its own FeedException.

repose-aggregator/external/service-clients/auth/src/main/java/org/openrepose/common/auth/openstack/AuthenticationService.java
wdschei: The AuthenticationService now throws its own AuthenticationServiceException.

adrianjgeorge:
retest this please

PR #1227, opened 2015-01-20
Updated the two locations we use .execute()

dkowis:
This now creates a new BasicHttpContext() so that we don't reuse cookie
stores. It's probably been wrong since the HttpPooling stuff was added.


dkowis:
This isn't the right way to do it, so I'm going to make a new branch anyway.

PR #1228, opened 2015-01-21
Rep 1673 client state

dkowis:
This is a test, and then a solution for the session bugs we've been having. Maybe there's more to it, but I think I've got it covered. In a thousand different requests, with a very small pool size (5) I didn't ever encounter a duplicated session. I make two requests with each http client to ensure that the same client gets the same session, and that was reliable as well.


adrianjgeorge:
It wasn't set before. It's the main reason for this change.

adrianjgeorge:
@dkowis if you rebase this to the 6.2.2.o release branch we can get it out sooner

dkowis:
@adrianjgeorge I can probably do that... lemme figure out how to do it heh.

dkowis:
So there's another pull request #1229 that contains these 4 commits rebased on top of the 6.2.2.0 branch. I think we're not merging that branch back into master, because we cherry picked things off for it?

adrianjgeorge:
No, that one will get merged into master as soon as it's cut

dkowis:
Okay, then we can close this one, as it'll get into master off the other branch. Yes?

PR #1236, opened 2015-01-29
Fixed the expected error message in a test that doesn't run in the normal Jenkins build.

wdschei:
This error message was normalized in REP-1526 Akka Client NPE. Since this test doesn't execute as part of the normal build cycle, it did not get caught until the nightly after the merge.


wdschei:
Retest this please.

PR #1241, opened 2015-02-06
REP-1542 Filter UAE

wdschei:


repose-aggregator/components/filters/herp/src/test/scala/org/openrepose/filters/herp/HerpFilterTest.scala
```diff
@@ -293,65 +296,209 @@ class HerpFilterTest extends FunSpec with BeforeAndAfterAll with BeforeAndAfter
...
+          val matcherOne = new Match
+          matcherOne.setField("userName")
+          matcherOne.setRegex(condition._1) // Conditionally matches
+          matchersOne.add(matcherOne)
+          val matcherTwo = new Match        // AND'd
+          matcherTwo.setField("roles")
+          matcherTwo.setRegex(".*BUZ.*")    // Always matches
+          matchersOne.add(matcherTwo)
+          filtersOut.add(filterOne)
+          val filterTwo = new FilterOut     // OR'd
```
adrianjgeorge: I don't think you end up exercising the or really, because you match on the first one every time. You should call it twice with a different request each time, and setup your criteria such that the two calls get caught on the different filterOuts.

wdschei: This doesn't test the match criteria in both positions, true, but due to the nature of how Scala filter'ing works it shouldn't matter. Just to avoid any potential issues if the data structure or filter mechanism change, I updated the test to run it through in both positions.

repose-aggregator/components/filters/herp/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
```diff
@@ -68,42 +72,97 @@ class HerpFilter @Inject()(configurationService: ConfigurationService) extends F
...
+    case class MapValue(value: Option[Map[String, Array[String]]]) extends Value {
+      override def toJava: Any = {
+        val x = value.orNull
+        if (x == null) {
+          null
+        } else {
+          x.asJava.entrySet()
+        }
+      }
+    }
```
dmnjohns: This can probably be reduced to:

value.map(x.asJava.entrySet()).orNull

wdschei: Why yes it can. Updated.

adrianjgeorge:
Fix the failing test, and it looks good.

PR #1242, opened 2015-02-06
Rep 1749 xslt namespaces

dkowis:
Here's a different implementation that uses the XSLT to translate the configuration instead of code.

It contains a pile of tests to run that translation across all of them, and a couple cases where it's already correct, or it's just super wrong.

Additionally, this exposed a method "for testing" because the design of the JAXB stuff couples validation and unmarshalling in a rather terrible way. There are TODOs around this so when we get to refactoring how our JAXB configuration is loaded, we can do it in a non-terrible way.


dkowis:
@tylerroyal checked this over my shoulder as well.

dkowis:
I am really hating xml today. If I include the saxon dependencies in core, I have to include saxon ee, because if the license is there, it tries to use saxon-ee specific stuff, blowing up. If I include saxon-ee in core, the container tests are blowing up, because for some reason they're trying to find/startup saxon-ee without a license file.

Is there any other XSLT 2.0 compliant transformer thing? I'm about to say we don't do this and we keep the hack code because it's less gross than doing XSLT and dealing with it's dependency hell. I'm worried about valid XML that the code doesn't deal with.

RackerWilliams:
No you're doing something wrong.  SaxonHE should never try to read the license file.  SaxonEE works great without a license file.  I have tests in API checker that prove it.  It shouldn't try to read the license file at all unless you're trying to do something that needs it.

dkowis:
On 02/09/2015 05:07 PM, Jorge Williams wrote:
> No you're doing something wrong. SaxonHE should never try to read the
> license file. SaxonEE works great without a license file. I have tests
> in API checker that prove it. It shouldn't try to read the license file
> at all unless you're trying to do something that needs it.

The container loaded repose are trying to read the license file. I
didn't change anything there, nor did I do really anything other than
create a transformer factory: "net.sf.saxon" not the saxonica one. I'm
doing this part right.

Repose core expects it to at least be on the class path, so if the
license file is present, core has the necessary classes that saxon will
generate apparently. Without saxon-ee being on the classpath and if
saxon_home is set, it will bomb.

These two scenarios are the mutually exclusive problem.

RackerWilliams:
I can take a look at it tomorrow with you if you like.  I'm pretty sure API checker would be breaking all over the place if this couldn't work.

dkowis:
retest this please

repose-aggregator/commons/configuration/src/main/java/org/openrepose/commons/config/parser/jaxb/UnmarshallerValidator.java
```diff
@@ -39,137 +40,90 @@
     private Schema schema;
     private Unmarshaller unmarshaller;
     private DocumentBuilder db;
+    //The transformer we'll use for translating our configuration xmls
+    public static final String SAXON_HE_FACTORY_NAME = "net.sf.saxon.TransformerFactoryImpl";
```
adrianjgeorge: Thought you dropped the saxon one?

repose-aggregator/commons/configuration/src/main/java/org/openrepose/commons/config/parser/jaxb/UnmarshallerValidator.java
```diff
@@ -39,137 +40,90 @@
...
+    public DOMSource validate(Document doc) throws IOException, SAXException {
+        if (schema != null) {
+            try {
+                schema.newValidator().validate(new DOMSource(doc));
+                //If we can't parse it, try translating it and validating it again
+            } catch (SAXException saxParseException) {
+                if (saxParseException.getLocalizedMessage().contains("Cannot find the declaration of element")) {
+                    //Run a quick XSLT
+                    try {
+                        //Why does this die a horrible death when the saxon EE license is specified?
```
adrianjgeorge: out of date comment?

repose-aggregator/commons/configuration/src/test/resources/unmarshallerValidator/limits/limits.xsd
```diff
@@ -0,0 +1,283 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
adrianjgeorge: What? Why does this exist?

adrianjgeorge: Nevermind. That's annoying to have to do, but i understand why.

repose-aggregator/pom.xml
```diff
@@ -31,6 +31,8 @@
         <jackson.version>1.9.13</jackson.version>
         <yammer.version>2.2.0</yammer.version>
         <jetty.version>9.2.0.v20140526</jetty.version>
+        <!-- NOTE: the saxon version should match that of API-checker, because there is saxon in core :( -->
+        <saxon.version>9.4.0.9</saxon.version>
```
adrianjgeorge: I thought you went with xalan instead of saxon?

dkowis: We might only have saxon in one other place... In the translation filter, which uses both saxon dependencies. I'd like to keep this, since we manage many other versions.

repose-aggregator/commons/configuration/src/main/java/org/openrepose/commons/config/parser/jaxb/UnmarshallerValidator.java
dkowis: Yep, I'll clean up my saxon leavings.

repose-aggregator/commons/configuration/src/test/resources/unmarshallerValidator/limits/limits.xsd
dkowis: Yeah I hate this super hard. But we're supposed to drop all this in a later release, so it's not as bad.

PR #1244, opened 2015-02-10
Update the sax parsing and the cache key to support TRR_USER events

dkowis:


adrianjgeorge:
well, that was stupid easy

dkowis:
Force pushed to rebase off master...

PR #1247, opened 2015-02-15
Flume Appender Fix

dmnjohns:
So the problem seems to have arisen from the Log4j2Plugins.dat file which was generated and made its way into our fat jar. I am not 100% certain why that broke things (maybe the plugin was generated for the flume appender jar and moving it into our fat jar messed everything up?) but changing the mechanism log4j2 uses to load the plugin at least got our logs back.

I have also not yet tested this. I'm going to let the PR builder test our current system test suite, and I imagine our new Repose to Flume tests will make sure the appender is working.

A couple of other things to consider:
1. We may be able to prevent the Log4j2Plugins.dat file from being generated rather than excluding it in the shader plugin.
2. There are a couple of other ways we could add the plugin without changing the log4j2 configuration file. See: http://logging.apache.org/log4j/2.x/manual/plugins.html

### DIFF
--- a/repose-aggregator/core/valve/pom.xml
+++ b/repose-aggregator/core/valve/pom.xml
@@ -204,6 +204,11 @@
                                         <exclude>META-INF/*.SF</exclude>
                                         <exclude>META-INF/*.DSA</exclude>
                                         <exclude>META-INF/*.RSA</exclude>
+                                    </excludes>
+                                </filter>
+                                <filter>
+                                    <artifact>org.apache.logging.log4j:log4j-flume-ng</artifact>
+                                    <excludes>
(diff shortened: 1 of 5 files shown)


dkowis:
Looks like the actual problem is this: https://issues.apache.org/jira/browse/LOG4J2-673

Seems to be a bug in log4j2 and shaded jars...

repose-aggregator/core/valve/pom.xml
```diff
@@ -206,6 +206,12 @@
                                         <exclude>META-INF/*.RSA</exclude>
                                     </excludes>
                                 </filter>
+                                <filter>
```
dkowis: This should have some comments providing context as to why we did this:
https://issues.apache.org/jira/browse/LOG4J2-673 is probably a good place to start

Reasoning why we excluded just the one dependencies plugin.dat is also useful.

PR #1248, opened 2015-02-16
Rep 1435 absolute wadl paths

dkowis:
This includes #1247 because otherwise none of the tests will pass, because they need log output :( I'm pretty confident that #1247 is the proper fix.

This just does a check on the wadl path if it's absolute or relative, and makes the proper decision.

An absolute path would look like:
```
wadlPath="/home/dkowis/wat-wadl.xml"
```
Relative path (relative to the repose configuration root):
```
wadlPath="wadls/wat-wadl.xml"
```


repose-aggregator/extensions/api-validator/src/main/java/org/openrepose/filters/apivalidator/ValidatorConfigurator.java
```diff
@@ -135,6 +136,11 @@ private String getPath(String path, String configRoot) {
     }
 
     private String getWadlPath(String uri, String configRoot) {
-        return new File(configRoot, uri).toURI().toString();
+        //If the wadl path is already absolute, just return that rather than prepending the config root
+        if(Paths.get(uri).isAbsolute()) {
```
adrianjgeorge: what happened to using [this](http://docs.oracle.com/javase/7/docs/api/java/net/URI.html#resolve(java.lang.String))?

dkowis: Either works I suppose. I didn't parse the documentation for URI correctly, and ended up down the `Paths` file path. Do I need to change it?

```
scala> import java.net.URI
import java.net.URI

scala> val configRoot = new URI("file:///etc/repose")
configRoot: java.net.URI = file:///etc/repose

scala> configRoot.resolve("/something")
res0: java.net.URI = file:/something

scala> configRoot.resolve("something")
res1: java.net.URI = file:/etc/something

scala> val configRoot2 = new URI("file:///etc/repose/")
configRoot2: java.net.URI = file:///etc/repose/

scala> configRoot2.resolve("/something")
res2: java.net.URI = file:/something

scala> configRoot2.resolve("something")
res3: java.net.URI = file:/etc/repose/something
```

adrianjgeorge: this works

PR #1251, opened 2015-02-23
REP-1772 Provide Default CF CADF Template

dmnjohns:


repose-aggregator/components/filters/herp/src/test/scala/org/openrepose/filters/herp/HerpFilterTest.scala
```diff
@@ -154,6 +213,19 @@ class HerpFilterTest extends FunSpec with BeforeAndAfterAll with BeforeAndAfter
...
+
+      // when:
+      herpFilter.configurationUpdated(herpConfig)
+      herpFilter.doFilter(servletRequest, servletResponse, filterChain)
+
+      // then:
+      def logEvents = listAppenderPre.getEvents
+      logEvents.size shouldBe 1
+      logEvents.get(0).getMessage.getFormattedMessage should include("\"CadfMethod\" : \"update/post\"")
+    }
```
wdschei: Should this be in foreach loop to test all of the conditions and increase code coverage?

wdschei: Never mind. It has it's own test. These could be collapsed into one and test it would go all the way through.

repose-aggregator/components/filters/herp/src/test/scala/org/openrepose/filters/herp/HerpFilterTest.scala
```diff
@@ -561,4 +673,50 @@ class HerpFilterTest extends FunSpec with BeforeAndAfterAll with BeforeAndAfter
...
+      "put"    -> "update/put",
+      "delete" -> "update/delete",
+      "patch"  -> "update/patch"
+    )
+    methods.foreach { method =>
+      it(s"should translate ${method._1} into ${method._2}") {
+        methodFormatter(method._1, null) should equal (method._2)
+      }
+    }
+  }
```
wdschei: This could be collapsed with the other test.

repose-aggregator/components/filters/herp/src/test/scala/org/openrepose/filters/herp/HerpFilterTest.scala
adrianjgeorge: They are testing two different things. One is testing the formatter, the other is testing to make sure it's exposed to a template. Both things that need to happen. They are testing the individual unit that i care about in each case. A unit test shouldn't be checking on two different things.

PR #1254, opened 2015-03-02
Consolidated the RackspaceIdentityBasicAuthHandler/Factory into the RackspaceIdentityBasicAuthFilter.

wdschei:
Consolidated the tests in a similar fashion.

Removed one layer of abstraction from this filter for readability and in turn maintainability.


adrianjgeorge:
If it passes, we'll ship it

PR #1255, opened 2015-03-02
Clearing TechDebt for the "Either log or rethrow this exception" rule in SonarQube (squid:S1166).

wdschei:


repose-aggregator/functional-tests/mocks-servlet/src/main/java/org/openrepose/commons/utils/test/mocks/providers/MockServiceProvider.java
```diff
@@ -12,6 +13,7 @@
 
 public class MockServiceProvider {
```
adrianjgeorge: I'm curious as to why this exists and what it's providing us.

PR #1257, opened 2015-03-02
Clearing TechDebt in the CommandDriver class for Sonar rules squid:S106 and squid:S1147.

wdschei:


adrianjgeorge:
What are these rules?

wdschei:
The `squid:S106` is the Sonar variant of the `PMD.SystemPrintln` that was already there.

The `squid:S1147` is titled **System.exit(...) and Runtime.getRuntime().exit(...) should not be called**  (https://dev.eclipse.org/sonar/rules/show/squid:S1147?modal=true&layout=false) and since this is a legitimate use of `System.exit()` so it should be suppressed.

PR #1261, opened 2015-03-04
Added a functional test to confirm that Repose will start with no filters in the system model.

wdschei:


adrianjgeorge:
Simple as they come, lets hope the tests pass.

PR #1268, opened 2015-03-13
config service fix for double eventing

adrianjgeorge:
Changed the order in which things happen when subscribing to stop accidental double event triggers


dmnjohns:
:+1: I'll pull it in as soon as tests pass.

PR #1274, opened 2015-03-23
REP-1865 Copyright/License Take 2

wdschei:
This is the checking of the Copyright/License info using the Maven Plugin.


http-delegation/pom.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
adrianjgeorge: Let's not do this. This is pretty terrible.

pom.xml
```diff
@@ -35,6 +36,8 @@
     </licenses>
 
     <modules>
+        <module>http-delegation</module>
```
adrianjgeorge: we don't want to bastardize these into the maven build

http-delegation/src/main/scala/com/rackspace/httpdelegation/HttpDelegationHeader.scala
```diff
@@ -1,3 +1,22 @@
+/*
+ * #%L
```
adrianjgeorge: What are these escape sequences for?

repose-aggregator/commons/configuration/src/test/resources/unmarshallerValidator/xsd/httpx-headers.xsd
```diff
@@ -1,4 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!--
```
adrianjgeorge: why ignored here, but not elsewhere?

wdschei: This is the flag for the license-maven-plugin to neither add a header to this file nor fail a build because it doesn't have one. Since this is a file not maintained by Repose, it is ignored for the purpose of this plugin by the build.

http-delegation/src/main/scala/com/rackspace/httpdelegation/HttpDelegationHeader.scala
wdschei: They are the open (#%L), section (%%), and close (#L%) markers used by the license-maven-plugin for identifying the headers it has generated.

pom.xml
wdschei: Since the two Gradle projects now have a POM with packing type pom, only the license-maven-plugin is executed.

http-delegation/pom.xml
wdschei: This allows us to ensure the two Gradle support projects are compliant with the Copyright/License header.

pom.xml
wdschei: Do we want to add a Gradle equivalent to these two projects? Since they don't get built as part of the normal Jenkins run, they won't fail the build is a header is missing or incorrect.

adrianjgeorge:
Egad, it won't show me the entire diff, but it looks like you did what we talked about, and this much more palatable.

adrianjgeorge:
I saw you excluded the files in our spock functional tests, but what about the ones in the external project?

wdschei:
Since all of the files in the *JavaEE 6 - XML Schema Library* (jee6-schemas) and *OpenStack Auth XML Schemas* (os-auth-schemas) modules were all stock with no modifications, the default file roots in the *Repose* (repose) module did not require modification to exclude them. Previously they had been identified as schemas (XSD's) that were not maintained by this project and therefore had the `Ignore-License` header added even though it was not strictly required.

wdschei:
I'm currently almost net 0 with  `21,712 ++ / 19,661 --`. There are so many changes in this one (copy/paste'ish) it will push me way positive. I need to find more to get rid of. ;-)

PR #1275, opened 2015-03-27
REP-1869 Update POM's to sign RPM and DEB artifacts.

wdschei:
This requires the Repose Signing Keys and Passphrase ver. 3 dated Mar 26, 2015 to be on the Jenkins Master and Slaves before the final test can be executed.

### DIFF
--- a/repose-aggregator/installation/deb/pom.xml
+++ b/repose-aggregator/installation/deb/pom.xml
@@ -46,6 +46,16 @@
                             <goals>
                                 <goal>jdeb</goal>
                             </goals>
+                            <configuration>
+                                <signPackage>true</signPackage>
+                                <!-- Old Debian based distros. -->
+                                <!--<signMethod>debsig-verify</signMethod>-->
+                                <!-- New Debian based distros. -->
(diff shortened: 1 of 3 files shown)


adrianjgeorge:
i like it when it's easy

wdschei:
This failure was due to the new keys not being on the Jenkins boxes.

wdschei:
I couldn't find a way to only sign the DEB's during the release phase. This would alleviate this too.

repose-aggregator/installation/deb/pom.xml
```diff
@@ -23,6 +23,43 @@
         <module>repose-valve</module>
     </modules>
 
+    <profiles>
+        <profile>
+            <!-- this profile works with gnupg 2.0.x NOT with 2.1.x -->
+            <id>release-sign-artifacts</id>
```
adrianjgeorge: Is this profile name the same as the one used for signing everything else?

wdschei: Yes. This is the id used in the top level to extract the environment variables into maven properties and in the DEB and RPM POM's to indicate to actually use them.

PR #1279, opened 2015-04-06
REP-1953 Licensing Round Two

wdschei:
Added the new DEPENDENCIES.txt file, normalized our dependency versions where possible, and consolidated dependency management where smart.

A lot of the versions were normalized to the highest version in use unless there was a reason not too.

### DIFF
--- a/repose-aggregator/components/filters/versioning/pom.xml
+++ b/repose-aggregator/components/filters/versioning/pom.xml
@@ -63,14 +63,20 @@
             <version>1.3</version>
             <scope>test</scope>
         </dependency>
+
         <dependency>
-            <groupId>org.codehaus.jackson</groupId>
-            <artifactId>jackson-core-lgpl</artifactId>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.codehaus.jackson</groupId>
-            <artifactId>jackson-mapper-lgpl</artifactId>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
             <scope>test</scope>
         </dependency>
 
--- a/repose-aggregator/core/core-lib/pom.xml
+++ b/repose-aggregator/core/core-lib/pom.xml
@@ -21,9 +21,6 @@
         <reposeVersionName>Dirk</reposeVersionName>
         <sonar.jacoco.itReportPath>${project.basedir}/../../target/jacoco-it.exec</sonar.jacoco.itReportPath>
         <test.artifact.version>2.1</test.artifact.version>
-
-        <!-- The fasterXML Jackson deps aren't in managed dependencies, because they're only in core right now -->
-        <fasterxml.jackson.version>2.2.2</fasterxml.jackson.version>
     </properties>
 
     <dependencies>
@@ -240,19 +237,16 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-core</artifactId>
-            <version>${fasterxml.jackson.version}</version>
         </dependency>
 
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>${fasterxml.jackson.version}</version>
         </dependency>
 
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-annotations</artifactId>
-            <version>${fasterxml.jackson.version}</version>
         </dependency>
         <!-- END OF json Support for log4j2 configuration files -->
 
--- a/repose-aggregator/core/core-lib/pom.xml
+++ b/repose-aggregator/core/core-lib/pom.xml
@@ -33,7 +33,6 @@
             <groupId>javax.inject</groupId>
             <artifactId>javax.inject</artifactId>
             <version>1</version>
-            <scope>provided</scope>
         </dependency>
 
         <!-- stuff pulled in from the classloader project -->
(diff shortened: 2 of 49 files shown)


adrianjgeorge:
I think you should leave test dependencies off the list, since they are not shipped with repose.

repose-aggregator/components/filters/versioning/pom.xml
```diff
@@ -63,14 +63,20 @@
...
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
+            <scope>test</scope>
+        </dependency>
         <dependency>
-            <groupId>org.codehaus.jackson</groupId>
-            <artifactId>jackson-core-lgpl</artifactId>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
             <scope>test</scope>
```
adrianjgeorge: this isn't going to be respected unless it's being dependency managed at test in the parent chain. If it's truly meant to be test only here, it needs to also be dependency managed to test here.

repose-aggregator/core/core-lib/pom.xml
```diff
@@ -243,26 +237,22 @@
...
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-annotations</artifactId>
-            <version>${fasterxml.jackson.version}</version>
         </dependency>
         <!-- END OF json Support for log4j2 configuration files -->
 
         <dependency>
             <groupId>com.typesafe</groupId>
             <artifactId>config</artifactId>
-            <version>1.0.2</version>
```
adrianjgeorge: is this used more than one place?

repose-aggregator/installation/deb/repose-cli-utils/pom.xml
```diff
@@ -142,6 +142,21 @@
...
+                                                <type>perm</type>
+                                                <prefix>
+                                                    /usr/share/doc/${artifactName}
+                                                </prefix>
+                                                <filemode>444</filemode>
+                                            </mapper>
+                                            <src>
+                                                ${basedir}/../../../../DEPENDENCIES.txt
+                                            </src>
+                                        </data>
```
adrianjgeorge: Shouldn't we be adding the contributors file as well?

wdschei: Good catch. Done.

repose-aggregator/core/core-lib/pom.xml
wdschei: `com.typesafe.config` is being used in 
 - _profile-support_
    - _core-support_
       - **core-lib**
       - **valve**
       - **web-application**
    - _services-pom_
       - _service-client-support_
          - **service-client-akka**

It was originally being defined in each of those bottom levels. Since it was not being dependency managed from a common parent, multiple versions were being brought in.

repose-aggregator/components/filters/versioning/pom.xml
wdschei: The `dependencyManagement` definition of the **Jackson** artifacts is in the parental lineage module `profile-support` POM. Since it does not include a **scope** element, the default of `compile` is assumed. If a child module re-defines the scope, it will override the default. In this particular case, the scope of the **Jackson** dependencies will be `test`.

```
$ cd <REPOSE_DIR>/repose-aggregator/components/filters/versioning
$ mvn dependency:tree
[INFO] Scanning for projects...
[INFO]                                                                         
[INFO] ------------------------------------------------------------------------
[INFO] Building Repose Components - Versioning 7.0.1.2-SNAPSHOT
[INFO] ------------------------------------------------------------------------
[INFO] 
[INFO] --- maven-dependency-plugin:2.1:tree (default-cli) @ versioning ---
[INFO] org.openrepose:versioning:jar:7.0.1.2-SNAPSHOT
...
[INFO] +- com.fasterxml.jackson.core:jackson-annotations:jar:2.4.0:test (scope not updated to compile)
[INFO] +- com.fasterxml.jackson.core:jackson-core:jar:2.4.0:test (scope not updated to compile)
[INFO] +- com.fasterxml.jackson.core:jackson-databind:jar:2.4.0:test (scope not updated to compile)
...
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
```

wdschei:
Any of the dependencies I have documented that need to be limited I can change, but none of them are scoped to `test`.

repose-aggregator/components/filters/versioning/pom.xml
adrianjgeorge: That's weird, normally anything that's dependency manages the only way to change something about it is to dependency manage it again. Maybe it's because the scope wasn't defined in the original dependency management section.

repose-aggregator/core/core-lib/pom.xml
adrianjgeorge: Fair enough

adrianjgeorge:
About those dependencies, i'm wondering if they are transitively test. In this case defined in a module that gets pulled into another as a test dependency. Because glassfish and tomcat should not be hard dependencies.

repose-aggregator/components/filters/versioning/pom.xml
wdschei: I couldn't find a definitive answer on this one way or the other. There was just a bunch of folks saying "I think it ...". So the final test was to run`mvn dependency:tree`. ;-)

james49:
The Dependencies.txt file is missing this required (according to the story) blub at the top:
"Repose has a number of dependencies that are covered by separate licenses, listed below. By using, modifying, or distributing Repose you may also be subject to the terms of those licenses."

james49:
Not seeing this change, nor evidence of the original URL it refers to:
"
Change the LICENSE.txt file to replace
    https://repose.atlassian.net/wiki/display/REPOSE/License
with
    DEPENDENCIES.txt
and add
    Contributors should also review CONTRIBUTORS.txt.
"

wdschei:
@james49 I updated the text which is evidently supposed to be slightly different from that on the GitHub site (e.g. README.md).

wdschei:
With the core-lib POM having the scope of `javax.inject` dependency changed from default (e.g. compile) to provided it causes an error in ValveRunner building.

https://jenkins.openrepose.org/job/repose-pull_requests/12/org.openrepose$valve/console

Do we want to take it back to default and add it to the DEPENDENCIES.txt?

adrianjgeorge:
looks good, hopefully it passes

wdschei:
They passed the the last run before changing that dependency's scope, so it should. Fingers crossed.

PR #1281, opened 2015-04-08
Rep 1973 ssl ciphers

dkowis:
This resolves (REP-1973)[https://repose.atlassian.net/browse/REP-1973] by enabling the users to configure in the container.cfg.xml.

There's an example in the example container.cfg.xml, but it's just a basic example. I couldn't find an obvious "this is the best way to do everything SSL" except here: https://www.playframework.com/documentation/2.3.x/CipherSuites and that's not quite perfectly related to what we're doing.

Some of it is better, because using the properties makes some things easier, but it might also be JDK8 dependent.

Also this resolves a bug with SSL not working **at all** in Repose 7.x until this is merged.

As a side effect of testing this, and because the maven shade plugin isn't very smart, we have updated to the latest version of Apache HTTPclient. If all the integration tests pass, we're probably using lots of deprecated things. This could affect the behavior of repose, but I have no way of testing it otherwise, and no way of keeping it out of the prod scope.


dkowis:
Wat da hek. I'll look at this, it's hopefully something stupid...

adrianjgeorge:
looks good

PR #1282, opened 2015-04-13
Rep 1820

adrianjgeorge:
I don't know if @dkowis is done with the new filter yet, but I'd like to get the tests running so that any that might be behaving weirdly can be fixed.


repose-aggregator/components/filters/merge-header/src/main/resources/META-INF/config/examples/merge-header.cfg.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<merge-header xmlns='http://docs.openrepose.org/repose/merge-header/v1.0'>
+    <!-- both response/request configs are optional, if neither are specified, this filter is a no-op -->
+    <response>
+        <header>your-mother</header>
```
dmnjohns: Leave the mothers out of this!

Dorothy Mantooth is a saint!

repose-aggregator/components/filters/merge-header/src/main/scala/org/openrepose/filters/mergeheader/MergeHeaderFilter.scala
```diff
@@ -0,0 +1,39 @@
...
+    // TODO: Clean up the asInstanceOf below, if possible?
+    configurationService.subscribeTo(filterConfig.getFilterName,
+      config,
+      xsdURL,
+      handlerFactory.asInstanceOf[UpdateListener[MergeHeaderConfig]],
+      classOf[MergeHeaderConfig])
+  }
+
+  override def doFilter(servletRequest: ServletRequest, servletResponse: ServletResponse, filterChain: FilterChain): Unit = {
+    new FilterLogicHandlerDelegate(servletRequest, servletResponse, filterChain).doFilter(handlerFactory.newHandler)
```
dmnjohns: Is there a reason to use the FilterLogicHandlerDelegate? Would it be simpler to write this filter like we wrote the HERP filter?

dmnjohns: Note that this kind of ties us to the FilterDirector that everyone hates.

dkowis: It's also the pattern that all the other filters are using, and I know this works. And when we replace the thing, we'll fix it. 

I don't care for it either, but I know this works....

repose-aggregator/components/filters/rate-limiting/src/test/groovy/org/openrepose/filters/ratelimiting/RateLimitingServiceHelperTest.groovy
```diff
@@ -100,6 +102,7 @@ public class RateLimitingServiceHelperTest extends Specification {
         user == MOST_QUALIFIED_USER
     }
 
+    @Ignore("another one that isn't going to work right because where the processing of things happens is in a different spot")
```
dmnjohns: Should we add a timebomb for these since they broke due to changes in the wrapper, but are actually trying to prove functionality of the helper?

repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/PowerFilterChain.java
```diff
@@ -100,8 +107,35 @@ public void startFilterChain(ServletRequest servletRequest, ServletResponse serv
...
         servletRequest.setAttribute("http://openrepose.org/requestUrl", ((HttpServletRequest) servletRequest).getRequestURL().toString());
         servletRequest.setAttribute("http://openrepose.org/queryParams", servletRequest.getParameterMap());
+        MutableHttpServletRequest wrappedRequest = MutableHttpServletRequest.wrap((HttpServletRequest) servletRequest);
+        splitRequestHeaders(wrappedRequest);
 
-        doFilter(servletRequest, servletResponse);
+        doFilter(wrappedRequest, servletResponse);
+    }
+
+    private void splitRequestHeaders(MutableHttpServletRequest request) {
```
dmnjohns: Does it make more sense for this method (and associated methods) to be inside the MutableHttpServletRequest wrapper instead? Then we could still split the headers in this class, but anyone using our wrappers could also split headers at will. We still get the explicit splitHeaders() call in the PowerFilterChain, but the PowerFilterChain won't reuse this logic, and it's semi-verbose in al already crowded class.

adrianjgeorge: i'd really rather not, exposing that functionality is just giving more rope in to what is already an insanely complicated thing.

repose-aggregator/components/filters/merge-header/src/main/scala/org/openrepose/filters/mergeheader/MergeHeaderFilter.scala
dmnjohns: I mean... I know dropping this pattern works. They both work.

But whatever. We will be forced to remove it when we replace the thing anyway, so I guess I don't super care.

dmnjohns:
If you guys are happy with it, it looks fine to me. I'll merge once tests pass.

dkowis:
I have to leave for the day, I'd like to examine the deproxy test that was tacked on to the end of the HeaderNormalizationTest tomorrow before we merge this.

dkowis:
Sure enough, deproxy can do the header counting, I was just doing it wrong. We should separate the tests from the `HeaderNormalizationTest`, since I don't think they're connected in any way... But we should keep only one of the tests, and probably the deproxy one, because it's less complex.

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/filters/headerNormalization/HeaderNormalizationTest.groovy
```diff
@@ -242,4 +244,52 @@ class HeaderNormalizationTest extends ReposeValveTest {
         //"content-type" | "application/xMl"
         //"Content-Type" | "APPLICATION/xml"
     }
+
+    def "Should not split request headers when configured as such with merge-header filter"() {
```
adrianjgeorge: This test shouldn't be part of this test class. It should be in it's own test class with it's own configuration, not tacked onto a test for a different filter.

dmnjohns:
retest this please

dmnjohns:
It looks like the CombineLimitsTransformTest is failing because of namespace prefixes. The test is trying to match the pattern:
```.*(<rates xmlns.*>.*</rates>).*(<absolute>.*</absolute>).*```
against:
```<ns2:rates xmlns:ns2="http://docs.openstack.org/common/api/v1.0" xmlns="http://docs.openrepose.org/repose/rate-limiting/v1.0">...</ns2:rates><absolute>...</absolute>```

PR #1285, opened 2015-04-15
Rep 1980 osv3 domain

dkowis:
This sets up the Openstack V3 API class to be able to append a 
```json
"domain":{
  "id": "someConfiguredId"
}
```
in the request for an admin token for the Openstack V3 filter.


adrianjgeorge:
Hooray easy ones. ship it

PR #1287, opened 2015-04-17
if the tenant does not start with hybrid: then return a 403/404 if maske...

tylerroyal:
...d


adrianjgeorge:
easy peasy, lets see it pass

adrianjgeorge:
@jennyvo you going to write a masked version of the test as well?

adrianjgeorge:
The pdf doesn't belong in the source code repository. If we want to keep it lets put it on the wiki somewhere.

adrianjgeorge:
@tylerroyal looks good, rebase out the commit with the pdf and i'll merge it in

tylerroyal:
Rebase accomplished.

PR #1292, opened 2015-04-23
REP-1704 Local and Process Tracing initial implementation

dmnjohns:
We still need to update the header name to whatever it's supposed to be, but otherwise I think we're done here.


repose-aggregator/components/filters/herp/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
```diff
@@ -144,6 +144,8 @@ class HerpFilter @Inject()(configurationService: ConfigurationService,
       "responseCode" -> httpServletResponse.getStatus,
       "responseMessage" -> Try(HttpStatus.valueOf(httpServletResponse.getStatus).name).getOrElse("UNKNOWN"),
       "guid" -> java.util.UUID.randomUUID.toString,
+      "requestGuid" -> stripHeaderParams(httpServletRequest.getHeader(CommonHttpHeader.REQUEST_GUID.toString)),
+      "responseGuid" -> stripHeaderParams(httpServletRequest.getHeader(CommonHttpHeader.RESPONSE_GUID.toString)),
```
dmnjohns: Should these be used in the default HERP configuration rather than {{guid}}?

repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/http/CommonHttpHeader.java
```diff
@@ -30,6 +30,10 @@
     AUTHORIZATION("Authorization"),
     WWW_AUTHENTICATE("WWW-Authenticate"),
 
+    //Tracing
+    REQUEST_GUID("X-Request-Guid"), //todo: correct name
+    RESPONSE_GUID("X-Response-Guid"), //todo: this may also be unnecessary if we tie the response to the request guid
```
dmnjohns: ^

repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/logging/apache/LogFormatArgument.java
```diff
@@ -37,6 +37,8 @@
     STATUS_CODE("s"),
     REMOTE_USER("u"),
     URL_REQUESTED("U"),
+    REQUEST_GUID("R"), //todo: make parsing take more than one character, would it be better to document that this can be extracted using ${X-REQUEST-GUID}i?
+    //todo: insert RESPONSE_GUID if ${X-RESPONSE-GUID}i is not acceptable
```
dmnjohns: ^

repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/PowerFilterChain.java
```diff
@@ -277,6 +283,20 @@ private void splitResponseHeaders(MutableHttpServletResponse mutableHttpResponse
         }
     }
 
+    //todo: is it safe to assume that if a request guid header does not exist on the request, we should not add a
+    //response guid header to the response? if so, the addTraceHeader flag is unnecessary, as is the UUID generation.
+    private void addResponseGuid(MutableHttpServletRequest request, MutableHttpServletResponse response) {
```
dmnjohns: ^

repose-aggregator/components/filters/herp/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
adrianjgeorge: why are there two of them? they should be the same value

repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/http/CommonHttpHeader.java
adrianjgeorge: Yeah, they should be a single value applied on both request and response

repose-aggregator/components/filters/herp/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
dmnjohns: Dealio.

repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/openstack/OpenStackAuthenticationHandler.java
```diff
@@ -107,14 +107,19 @@ private AuthToken validateTenant(AuthenticateResponse resp, String tenantID) {
 
     @Override
     public AuthToken validateToken(ExtractorResult<String> account, String token) throws AuthServiceException {
+        return validateToken(account, token, null);
```
adrianjgeorge: This doesn't seem right. Shouldn't you be passing along the guid?

repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/common/AuthenticationHandler.java
```diff
@@ -96,10 +97,16 @@ protected AuthenticationHandler(Configurables configurables, AuthTokenCache cach
 
     protected abstract AuthToken validateToken(ExtractorResult<String> account, String token) throws AuthServiceException;
 
+    protected abstract AuthToken validateToken(ExtractorResult<String> account, String token, String requestGuid) throws AuthServiceException;
```
adrianjgeorge: Why duplicate all the methods? I'm assuming it was with an eye towards backwards compatibility, but i don't like have the vestigial methods. Unless you've got a good reason why to keep them?

repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/openstack/OpenStackAuthenticationHandler.java
adrianjgeorge: Ohh, i see these are the extra ones that i was talking about above

repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/common/AuthenticationHandler.java
dmnjohns: The only reason would be backwards compatibility. Internally, I don't think we call the old methods anymore. In this instance, I thought it best to preserve the old interface for the time being simply because soon enough the entire filter will be re-written.

Excision of the old methods should be trivial.

repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/openstack/OpenStackAuthenticationHandler.java
dmnjohns: Yup. Just an attempt not to repeat myself.

repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/common/AuthenticationHandler.java
adrianjgeorge: Yeah, just remove them

adrianjgeorge:
guid is only being applied to responses that get to the origin service, it's not being added to any response we create because we've rejected the request

PR #1295, opened 2015-04-27
Cache verification stuff

dkowis:
This actually makes the cache keys for stuff delete properly.
Fixes: https://repose.atlassian.net/browse/REP-2130


adrianjgeorge:
looks good

PR #1299, opened 2015-05-03
REP-2162 Response Status Code BUG

wdschei:
Added tests for several filters to ensure the Response Status Code is being set appropriately and fixed those were it wasn't.

Squashed the OpenStackResponseHandler into the only use of the class to alleviate unnecessary abstractions.

Updated the TranslationHandler to not have duplicated logic with an embedded return.


repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/openstack/OpenStackAuthenticationHandler.java
```diff
@@ -26,15 +26,20 @@
 import org.openrepose.common.auth.openstack.AuthenticationService;
 import org.openrepose.common.auth.openstack.AuthenticationServiceClient;
 import org.openrepose.common.auth.openstack.OpenStackToken;
+import org.openrepose.commons.utils.StringUtilities;
```
adrianjgeorge: Don't use this string utilities use the one from commons lang

adrianjgeorge: Oh, I see you just collapsed something in, eh, fix it anyway.

repose-aggregator/components/filters/rate-limiting/src/test/java/org/openrepose/filters/ratelimiting/RateLimitingHandlerTest.java
```diff
@@ -81,7 +82,7 @@ public void shouldReturnUnauthorizedWhenUserInformationIsMissing() {
 
         @Before
         public void setup() {
-            List<String> headerNames = new ArrayList<String>();
+            final List<String> headerNames = new ArrayList<>();
```
adrianjgeorge: why the addition of final?

repose-aggregator/components/filters/rate-limiting/src/test/java/org/openrepose/filters/ratelimiting/RateLimitingHandlerTest.java
```diff
@@ -93,7 +94,12 @@ public void setup() {
             defaultConfig.setValue(10);
             defaultConfig.setUnit(org.openrepose.core.services.ratelimit.config.TimeUnit.MINUTE);
 
-            when(mockedRequest.getHeaderNames()).thenReturn(Collections.enumeration(headerNames));
+            when(mockedRequest.getMethod()).thenReturn("GET");
+            when(mockedRequest.getHeaderNames()).thenAnswer(new Answer<Object>() {
```
adrianjgeorge: Why this change?

repose-aggregator/components/filters/rate-limiting/src/test/java/org/openrepose/filters/ratelimiting/RateLimitingHandlerTest.java
wdschei: Since the `headerNames` variable is now accessed directly in the anonymous `Answer` class it has to be declared `final`.

repose-aggregator/components/filters/rate-limiting/src/test/java/org/openrepose/filters/ratelimiting/RateLimitingHandlerTest.java
wdschei: The original declaration of the mock'd when/thenReturn of the `headerNames` in the `setup()` was actually non-functional. This is why at the top of each method there was another re-declaration of this when/thenReturn clause. I just pulled working when/thenReturn from the top of each method and placed it in the `setup()` in turn forcing the `final` declaration of the `headerNames`.

adrianjgeorge: Why is answer necessary at all? you should be able to get then return to work. what error is it giving?

wdschei: It is not giving an error, the returned enumeration is just empty.

wdschei: This seems to be by it was copy/pasted into each of the tests.

adrianjgeorge: that's super weird

wdschei:
Retest this please.

adrianjgeorge:
seems to be legit

PR #1303, opened 2015-05-08
Raise event from rate limiting filter when limit is breached + Use Percent encoding in uri's 

ganeshs:
This pull has two changes,

1. I have made changes in the RateLimit module to raise an event whenever a breach happens. This would allow users to tap into the limit breach event and do custom functionalities
2. The proxy request URI constructed by repose is not using percent encoding. This results in spaces to be encoded as '+' instead of '%20'. I have provided a fix for that as well.


dmnjohns:
Retest this please.

dmnjohns:
To start, thanks for the pull request! I know the code you touched isn't particularly user-friendly, but hopefully it didn't give you too much trouble.

I will note that all source files will need a license header. I'd also like to see tests for the two changes you outlined (the character encoding and event raising).

Otherwise, this is simple enough and looks good to me.

ganeshs:
Added license headers and tests for the two changes

ganeshs:
The failures were due to check style violations. this is fixed in the latest commit.

ganeshs:
Can you retest this.

dmnjohns:
Retest this please.

@ganeshs As an update, we've been discussing this PR a bit. It looks good, but we've thought about splitting the two changes apart to keep things tidy. I promise we're looking into this!

dmnjohns:
Retest this please

ganeshs:
I have reverted the change for "Use Percent encoding in uri's". Will file a separate pull after ensuring all scenarios are tested. Tests are passing now on my local. Please merge

ganeshs:
Retest this please

adrianjgeorge:
retest this please

adrianjgeorge:
@dimtruck any chance you could performance test this change? at this point that's really my only concern

dimtruck:
This is running right now.

dimtruck:
Ran two performance tests against the PR.  

Test 1: At throughput of 500 requests per second, response time was at 7 ms and 60 errors (at startup) 
Test 2: At throughput of 500 requests per second, response time was at 8 ms and 63 errors (at startup) 
Baseline: At throughput of 500 requests per second, response time is at 7 ms 

The real concern, however, is the 10% used heap memory increase in both tests (25% to 28%).  I am running another baseline against current repose master to validate that his memory increase is not just due to this pull request.

ganeshs:
This change shouldn't have any impact on the heap as long as the events are consumed. I see PowerProxyEventKernel dispatching the events to the listener. I assume it's getting started at the startup.

dimtruck:
After running 2 more tests against master branch, the difference between the PR and current master is around 5% used heap memory.  Most of those are in new EventService objects and most of that is in Eden (so gets cleaned up pretty quickly).  I'm ok with pulling it in.

PR #1304, opened 2015-05-10
adding facility for core service extension

vishal-kedia:
adding facility for core service extension independent of package nam…


adrianjgeorge:
@vishal-kedia I like it, but could you add soem tests?

vishal-kedia:
@adrianjgeorge  I have added the test cases as well.

adrianjgeorge:
ok to test

repose-aggregator/core/core-lib/src/main/java/org/openrepose/core/spring/CoreSpringProvider.java
```diff
@@ -148,6 +150,15 @@ public void initializeCoreContext(String configRoot,
             }
 
             coreContext.scan(coreScanPackage);
+            
+            final String extendedServicePackageNameKey = conf.getString("extendedServicePackageNameKey");
```
dkowis: Typesafe config can read items from system properties, or rather, configuration entries can be overridden by system properties.

I highly recommend using the typesafe config as the primary way to set the extended service package, and then overriding that package with a system property.

What I'd do, is set the extended service name to blank in the default config file.

Then you can override that using a system property at runtime, and for testing, you can easily override the config using a configuration file in the test resources.

dkowis:
@vishal-kedia I've made a slightly different pull request incorporating your work, but using the typesafe config library to provide the system property integration, let me know what you think.

PR #1306, opened 2015-05-20
Rep 1828 tracing

dkowis:
Do all the tracing!

### DIFF
--- a/repose-aggregator/services/datastore/impl/src/main/java/org/openrepose/nodeservice/distributed/servlet/DistributedDatastoreServlet.java
+++ b/repose-aggregator/services/datastore/impl/src/main/java/org/openrepose/nodeservice/distributed/servlet/DistributedDatastoreServlet.java
@@ -19,7 +19,10 @@
  */
 package org.openrepose.nodeservice.distributed.servlet;
 
+import org.apache.logging.log4j.ThreadContext;
+import org.openrepose.commons.utils.http.CommonHttpHeader;
 import org.openrepose.commons.utils.io.ObjectSerializer;
+import org.openrepose.core.logging.TracingKey;
 import org.openrepose.core.services.datastore.*;
 import org.openrepose.core.services.datastore.distributed.ClusterConfiguration;
 import org.openrepose.core.services.datastore.distributed.ClusterView;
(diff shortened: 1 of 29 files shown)


dmnjohns:
Should we update the default log4j2 config shipped with Repose to include the new GUID tag?

dkowis:
Yeah that's a good idea! 

On May 20, 2015 3:20:37 PM CDT, Damien Johnson <notifications@github.com> wrote:
>Should we update the default log4j2 config shipped with Repose to
>include the new GUID tag?
>
>---
>https://github.com/rackerlabs/repose/pull/1306#issuecomment-104023666

-- 

repose-aggregator/services/datastore/impl/src/main/java/org/openrepose/nodeservice/distributed/servlet/DistributedDatastoreServlet.java
```diff
@@ -97,7 +100,12 @@ protected void service(HttpServletRequest request, HttpServletResponse response)
             if ("PATCH".equals(request.getMethod())) {
                 doPatch(request, response);
             } else {
+                String traceGUID = request.getHeader(CommonHttpHeader.TRACE_GUID.toString());
```
adrianjgeorge: we use patch a lot, why did you not include it?

dkowis: Probably just because derp. This is easy to solve. I fix. 

On May 21, 2015 7:52:26 AM CDT, Adrian J George III <notifications@github.com> wrote:
>> @@ -97,7 +100,12 @@ protected void service(HttpServletRequest
>request, HttpServletResponse response)
>>              if ("PATCH".equals(request.getMethod())) {
>>                  doPatch(request, response);
>>              } else {
>> +                String traceGUID =
>request.getHeader(CommonHttpHeader.TRACE_GUID.toString());
>
>we use patch a lot, why did you not include it?
>
>---
>https://github.com/rackerlabs/repose/pull/1306/files#r30798192

-- 

adrianjgeorge:
super happy this worked, i don't suppose slf4j has a similiar concept, so that we aren't tying ourselves into log4j? If it's not there, it's not there and we live with the lock in.

dkowis:
it might: http://www.slf4j.org/api/org/slf4j/MDC.html

I'll see if I can use that as a drop in.

PR #1307, opened 2015-05-21
new dependency files

tylerroyal:
FYI, Valve and War end up being the same since the DD uses jetty and that was the only difference.


dkowis:
I removed the "test" file that shouldn't have been committed without just adding another delete.

PR #1311, opened 2015-05-29
REP-2097 Simple RBAC Filter

wdschei:


wdschei:
Retest this please.

wdschei:
Retest this please.

repose-aggregator/extensions/simple-rbac/pom.xml
```diff
@@ -0,0 +1,127 @@
...
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.openrepose</groupId>
+            <artifactId>utilities</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.openrepose</groupId>
+            <artifactId>api-validator</artifactId>
```
adrianjgeorge: you depend on the validator filter?

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/config/simple-rbac.xsd
```diff
@@ -0,0 +1,115 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
adrianjgeorge: no license?

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/config/simple-rbac.xsd
```diff
@@ -0,0 +1,115 @@
...
+            </xs:annotation>
+        </xs:attribute>
+        <xs:attribute name="enable-masking-403s" type="xs:boolean" use="optional" default="false">
+            <xs:annotation>
+                <xs:documentation>
+                    <html:p>Instead of returning a FORBIDDEN (403), return a NOT FOUND (404) or METHOD NOT ALLOWED (405) as appropriate.</html:p>
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+        <xs:attribute name="dot-output" type="xs:string" use="optional">
```
adrianjgeorge: not sure we need this option. what's it's prpose? i think in the other filter it exists so that they can be compiled and then used in future configs instead of the wadl.

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/config/simple-rbac.xsd
```diff
@@ -0,0 +1,115 @@
...
+                <xs:documentation>
+                    <html:p>
+                        dot output file for this validator
+                        DOT is a plain text graph description language.
+                        It is a simple way of describing graphs that both humans and computer programs can use.
+                    </html:p>
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+        <xs:attribute name="wadl-output" type="xs:string" use="optional">
```
adrianjgeorge: why?

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/config/simple-rbac.xsd
```diff
@@ -0,0 +1,115 @@
...
+                   saxon:message="Exactly one of the resources attribute or resources element can be specified, but not both."/>
+    </xs:complexType>
+
+    <xs:complexType name="DelegatingType">
+        <xs:annotation>
+            <xs:documentation>
+                <html:p>Whether or not you would like this filter to populate the delegation headers. Inclusion means you do.</html:p>
+            </xs:documentation>
+        </xs:annotation>
+        <xs:attribute name="quality" type="QualityType" use="optional" default="0.1">
```
adrianjgeorge: this seems like a low default. it should be the same value as the validation filter, because it's basically fulfilling the same  role

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/examples/simple-rbac.cfg.xml
```diff
@@ -0,0 +1,16 @@
...
+<simple-rbac xmlns="http://docs.openrepose.org/repose/simple-rbac/v1.0"
+             roles-header-name="X-ROLES"
+             enable-masking-403s="false">
+    <resources>
+        [<![CDATA[
+/path/to/this  GET       role1,role2,role3,role4
+/path/to/this  PUT       role1,role2,role3
+/path/to/this  POST      role1,role2
+/path/to/this  DELETE    role1
+/path/to/that  GET,PUT   ALL
```
adrianjgeorge: I'm assuming all is a special value

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
```diff
@@ -0,0 +1,268 @@
...
+    configurationService.unsubscribeFrom(configurationFile, this)
+  }
+
+  override def doFilter(servletRequest: ServletRequest, servletResponse: ServletResponse, filterChain: FilterChain): Unit = {
+    if (!initialized) {
+      logger.error("Simple RBAC filter has not yet initialized...")
+      servletResponse.asInstanceOf[HttpServletResponse].sendError(500)
+    } else {
+      val mutableHttpRequest = MutableHttpServletRequest.wrap(servletRequest.asInstanceOf[HttpServletRequest])
+      val mutableHttpResponse = MutableHttpServletResponse.wrap(mutableHttpRequest, servletResponse.asInstanceOf[HttpServletResponse])
```
adrianjgeorge: why the response? are you changing altering it after you call chain.doFilter?

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
```diff
@@ -0,0 +1,268 @@
...
+
+      logger.trace("Simple RBAC filter processing request...")
+      validator.validate(mutableHttpRequest,mutableHttpResponse,filterChain)
```
adrianjgeorge: ugh, i just remembered that validator takes care of the chain and everything. I super hate that.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
```diff
@@ -0,0 +1,268 @@
...
+    }
+    logger.trace("Simple RBAC filter returning response...")
+  }
+
+  override def configurationUpdated(configurationObject: SimpleRbacConfig): Unit = {
```
adrianjgeorge: break this up its too large

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
```diff
@@ -0,0 +1,268 @@
...
+    configuration = configurationObject
+    val isDelegating = configuration.getDelegating != null
+    val delegationQuality = if (isDelegating) configuration.getDelegating.getQuality else 0.0
```
adrianjgeorge: these two lines can definitely become an option

adrianjgeorge: Option(configuration.getDelegating).map(_.getQuality).getOrElse(0.0)

adrianjgeorge: ugh you need the values below. id still wrap the first value in an option, and use said option to get the quality, and do a get or else on the option down below.

repose-aggregator/extensions/simple-rbac/pom.xml
wdschei: There is a helper class in it for talking to API-Checker. It is fairly small and could be duplicated.

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/config/simple-rbac.xsd
wdschei: The dot is the visualization. It is good for debugging what is going on in API-Checker. It is not produced by default, but essential to figure out what is broke when things go wrong. The WADL option is nice if someone wants to transfer to the full API Validation filter after their API grows beyond what the Simple RBAC filter supports.

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/config/simple-rbac.xsd
wdschei: Good catch. It is not set to 0.3; same as the validator-configuration.xsd.

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/examples/simple-rbac.cfg.xml
wdschei: The CDATA block is the new RBAC DSL that is defined in REP-2097. It is basically fully qualified path [whitespace] HTTP Method(s)(CSV) [whitespace] rax:role(s)(CSV). The special wildcard words are ANY and ALL which are interchangeable, but were intended to mean ANY HTTP Method and ALL rax:roles.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: Nope, but the API-Checker's validator needs it.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: Yep.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: I didn't like that when I wrote it, but I couldn't make it look good. With your suggestions and a fresh set of eyes, I think I may have cleaned this spot up fairly well.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: Does that work better after implementing the suggestions below, consolidating, and then general clean up?

repose-aggregator/extensions/simple-rbac/pom.xml
wdschei: This is in the new branch SimpleRbacFilter_DispatchHandler and put as a pull request to this branch.

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/config/simple-rbac.xsd
adrianjgeorge: Fair arguements.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
```diff
@@ -0,0 +1,261 @@
...
+        }
+        val uuid = UUID.randomUUID
+        validator = Validator.apply(
+          s"SimpleRbacValidator_$uuid",
+          new StreamSource(new ByteArrayInputStream(r.getBytes), "file://simple-rbac.wadl"),
+          config
+        )
+        initialized = true
+      case _ =>
+        logger.warn("Unable to generate the WADL; check the provided resources.")
```
adrianjgeorge: what's the behavior of the system, should a wadl fail to load?

wdschei: The filter is never initialized, so Repose will never reach a Ready state.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
```diff
@@ -0,0 +1,261 @@
...
+        case Some(n) =>
+          rbacToWadl(readResource(
+              configurationService.getResourceResolver.resolve(n).newInputStream()
+          ))
+        case _ =>
+          None
+      }
+    )
+    rbacWadl match {
+      case Some(r) =>
```
adrianjgeorge: more descriptive name please.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
```diff
@@ -0,0 +1,261 @@
...
+        validator = Validator.apply(
+          s"SimpleRbacValidator_$uuid",
+          new StreamSource(new ByteArrayInputStream(r.getBytes), "file://simple-rbac.wadl"),
+          config
+        )
+        initialized = true
+      case _ =>
+        logger.warn("Unable to generate the WADL; check the provided resources.")
+    }
+  }
```
adrianjgeorge: i didn't see you tear down the old validator. IS that something you need to do? I know we have issues around that ni the validator filter.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
adrianjgeorge: go ahead and make this log message an error then, we want to make it real easy to find the problem.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
```diff
@@ -0,0 +1,261 @@
...
+    val handlers: util.List[ResultHandler] = new util.ArrayList[ResultHandler]
+    Option(configuration.getDelegating) match {
+      case Some(d) =>
+        handlers.add(new MethodLabelHandler)
+        handlers.add(new DelegationHandler(d.getQuality))
+      case _ =>
+        handlers.add(new ServletResultHandler)
+    }
+    handlers.add(new InstrumentedHandler)
+    handlers.add(new ApiCoverageHandler)
```
adrianjgeorge: shouldn't this one be conditional? i thought it was a feature that had to be turned on in the other filter. is that not the case?

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
```diff
@@ -0,0 +1,261 @@
...
+    }
+    new DispatchHandler(handlers.toList:_*)
+  }
+
+  private def getPath(path: String, configRoot: String): String = {
+    val file: File = new File(path)
+    if (file.isAbsolute) {
+      file.getAbsolutePath
+    } else {
+      new File(configRoot, path).getAbsolutePath
```
adrianjgeorge: pretty sure this call works by itself and you don't have to do the check. Or maybe that was URIs....

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
```diff
@@ -0,0 +1,261 @@
...
+  def readResource(resourceStream: InputStream): Option[String] = {
+    Try(Some(Source.fromInputStream(resourceStream).getLines().mkString("\n"))).getOrElse(None)
+  }
+
+  private def rbacToWadl(rbac: Option[String]): Option[String] = {
+
+    case class Resource(path: String, methods: Set[String], roles: Set[String])
+
+    def parseLine(line: String): Option[Resource] = {
+      val values = line.split("\\s+")
```
adrianjgeorge: I assume this says any whitespace

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
```diff
@@ -0,0 +1,261 @@
...
+        ))
+      } else {
+        logger.warn(s"Malformed RBAC Resource: $line")
+        None
+      }
+    }
+
+    val parsed = rbac match {
+      case Some(l) =>
+        Some(l.replaceAll("[\r?\n?]", "\n").split('\n').toList.flatMap(parseLine))
```
adrianjgeorge: did this have to be flat map? seems like regular map would have worked?

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: Yes. s is whitespace and + is one or more.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
adrianjgeorge: Also Option(stuff seems more correct than Some(stuff

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
```diff
@@ -0,0 +1,261 @@
...
+          Try(values(2).split(',').toSet[String].map(_.trim)).getOrElse(Set.empty)
+        ))
+      } else {
+        logger.warn(s"Malformed RBAC Resource: $line")
+        None
+      }
+    }
+
+    val parsed = rbac match {
+      case Some(l) =>
```
adrianjgeorge: name again

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    parsed match {
+      case Some(_) =>
+        val header = s"""<application xmlns:rax="http://docs.rackspace.com/api"
+                        |         xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+                        |         xmlns="http://wadl.dev.java.net/2009/02"
+                        |    >
+                        |  <resources base="http://localhost">""".stripMargin
+        val resources = parsed.getOrElse(List.empty[Resource]).map { r =>
```
adrianjgeorge: your or else is never going to trigger, because you are in a case that says explicitly that you have a value. Maybe catch the value and use it directly instead?

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
```diff
@@ -0,0 +1,261 @@
...
+          def toMethods(resource: Resource, uuid: UUID) = {
+            val roles = resource.roles.mkString(" ")
+            val raxRoles = roles match {
```
adrianjgeorge: i feel like there is a better way of expressing this, but i'm not seeing it right now. also single line the second case

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: Done.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
```diff
@@ -0,0 +1,261 @@
...
+            }
+            resource.methods.map {
+              _ match {
+                case a if a.equalsIgnoreCase("ANY") || a.equalsIgnoreCase("ALL") =>
+                  s"""      <method name="GET"    id="_$uuid-GET"    $raxRoles/>
+                     |      <method name="PUT"    id="_$uuid-PUT"    $raxRoles/>
+                     |      <method name="POST"   id="_$uuid-POST"   $raxRoles/>
+                     |      <method name="DELETE" id="_$uuid-DELETE" $raxRoles/>""".stripMargin
+                case m =>
+                  s"""      <method name="$m"   id="_$uuid-$m"    $raxRoles/>"""
```
adrianjgeorge: how case sensitive is api-checker? will it accept get or only GET?

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
adrianjgeorge: name on r

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: API-Checker is not thread-safe? :-(
Adding the tread-safety wrappers now.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: It was, but it was always true in the conversion. When I just went back over it, it is a hidden value deep in the guts of the Validation filter even though it is a top-level attribute in the schema definition. I will put it in and bring it to the top of this filter. May require some rework on a few tests.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: This was ensuring that the config root is used if the reference is not originally absolute, but that internally it is always reference absolutely.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: The flatMap removes all of the None elements.

wdschei: rbac is already an Option coming into the method, so the Some only matches if it isn't a None.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
adrianjgeorge: I don't remember the exact problem, but we were getting memory leaks based on reloads. maybe @dimtruck or @dmnjohns remember the exact issue

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
adrianjgeorge: yeah, in a uri maybe a file, if you provide a root and a path it'll behave smart. if it's relative it appends to the root , if it's absolute it'll give you the absolute without the root.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
adrianjgeorge: I was refering to the new one you were creating

wdschei: Got it. Head on straight now.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: Done.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: Done.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: Done.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: The second case is done.

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/config/simple-rbac.xsd
wdschei: Added, but why didn't the check catch this before now?

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: API Checker is case sensitive on both methods and roles. This was updated to always put uppercase HTTP methods in the generated WADL.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
```diff
@@ -0,0 +1,294 @@
...
+      logger.error("Simple RBAC filter has not yet initialized...")
+      servletResponse.asInstanceOf[HttpServletResponse].sendError(500)
+    } else {
+      val mutableHttpRequest = MutableHttpServletRequest.wrap(servletRequest.asInstanceOf[HttpServletRequest])
+      val mutableHttpResponse = MutableHttpServletResponse.wrap(mutableHttpRequest, servletResponse.asInstanceOf[HttpServletResponse])
+
+      logger.trace("Simple RBAC filter processing request...")
+      validatorLock synchronized {
+        validator.validate(mutableHttpRequest, mutableHttpResponse, filterChain)
+      }
```
wdschei: @adrianjgeorge @dmnjohns @dimtruck Does the access to the API Checker's Validator need to be synchronized during the validate or just during the setup and teardown?

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/config/simple-rbac.xsd
RackerWilliams: Yea this seems wrong to me too. I guess for debugging?

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/examples/simple-rbac.cfg.xml
RackerWilliams: Why is this in a CDATA section?  You don't need it right?

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/config/simple-rbac.xsd
```diff
@@ -0,0 +1,147 @@
...
+    <xs:complexType name="SimpleRbacConfig">
+        <xs:annotation>
+            <xs:documentation>
+                <html:p>The root config type for the Simple RBAC filter configuration file.</html:p>
+            </xs:documentation>
+        </xs:annotation>
+
+        <xs:sequence>
+            <xs:element name="delegating" type="DelegatingType" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="resources" type="xs:string" minOccurs="0" maxOccurs="1">
```
RackerWilliams: If you want to be really safe you might want to create a special type to preserve whitespace in resources.  This guarantees that whitespace will be preserved.  The parser is not *required* to preserve whitespace unless you explicitly tell it to. 

Instead of saying that "resources" is of type xs:string...you can create a special type for resources, something like this:

````xml
<xs:simpleType name="ResourcesType">
    <xs:restriction base="xs:string">
      <xs:whiteSpace value="preserve"/>
    </xs:restriction>
  </xs:simpleType>
````

You can use "collapse" instead of "preserve" which will collapse existing whitespace elements into single spaces (and replace tabs, carriage returns etc, that may make thing easier to parse.

File: repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala (hunk 22 above)
Comment (RackerWilliams) [inline repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala]: validate is thread safe...don't lock!!

File: repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala [hunk 24]
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    parsed match {
+      case Some(values) =>
+        val header = s"""<application xmlns:rax="http://docs.rackspace.com/api"
+                        |         xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+                        |         xmlns="http://wadl.dev.java.net/2009/02"
+                        |    >
+                        |  <resources base="http://localhost">""".stripMargin
+        val resources = values.map { value =>
```
RackerWilliams: You should probably be using scala's built in XML support instead of putting string together the way you are. If anything it will always gurantee that you produce well formed XML -- it'll throw a compile error if you don't.  May may things a bit easier to read too.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
```diff
@@ -0,0 +1,294 @@
...
+              case anyAll if roles.equalsIgnoreCase("ANY") || roles.equalsIgnoreCase("ALL") => ""
+              case _ => s"""rax:roles="$roles""""
+            }
+            resource.methods.map {
+              _ match {
+                case anyAll if anyAll.equalsIgnoreCase("ANY") || anyAll.equalsIgnoreCase("ALL") =>
+                  s"""      <method name="GET"    id="_$uuid-GET"    $raxRoles/>
+                     |      <method name="PUT"    id="_$uuid-PUT"    $raxRoles/>
+                     |      <method name="POST"   id="_$uuid-POST"   $raxRoles/>
+                     |      <method name="DELETE" id="_$uuid-DELETE" $raxRoles/>""".stripMargin
```
RackerWilliams: Hmmm... I have a problem with ANY on the method name.  That implies that all methods are supported by the backed service...but there's no guarantee that the service actually supports all of them.  So for example you'll have to include the PATCH method -- something you're not currently doing -- what if they use a non-standard method name -- which is quite possible.  If you add PATCH and the backend service doesn't support it, will the backend generate the right error message?

A better method may be include multiple methods in the simple RBAC format on a single line -- the way you do roles.

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/config/simple-rbac.xsd
```diff
@@ -0,0 +1,147 @@
...
+                        <html:p>
+                            This is a list of resources, HTTP methods, and roles allowed to access them.
+                        </html:p>
+                    </xs:documentation>
+                </xs:annotation>
+            </xs:element>
+        </xs:sequence>
+
+        <xs:attribute name="roles-header-name" type="xs:string" use="optional" default="X-ROLES">
+            <xs:annotation>
```
RackerWilliams: My inclination here is to always assume X-ROLES...if someone wants a different header they can rename headers.  Let's not add this option...if we do add it then let's add options to use alternate Header names for all of our existing filters.

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/config/simple-rbac.xsd
```diff
@@ -0,0 +1,147 @@
...
+                <xs:documentation>
+                    <html:p>
+                        This is the name of the header where that roles are passed in.
+                    </html:p>
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+        <xs:attribute name="resources-file-name" type="xs:string" use="optional">
+            <xs:annotation>
```
RackerWilliams: The way we've been doing this is with other filters is to have an href attribute.  For example 

<resources href="path/to/file"/>

or 

<resources>
.
.
.
.
</resources>

Then have an assertion that makes sure you can have one form and not the other.  I think this is a much better approach.  I think the original validator filter works this way..so does translation..and the RMS...you can find examples there.

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/config/simple-rbac.xsd
adrianjgeorge: He's got another comment in here discussing it. Mostly it comes down to it gives them an easy path of transition to moving to a full wadl later if they need further features. They just turn this on and it spits out a wadl we know works, and they can use that as the basis for any further changes.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
adrianjgeorge: It does support many methods currently, via a comma separated list. This is a convenience to allow them to let anything through, it does exactly what it says. If they don't want to let anything through they shouldn't use it.

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/config/simple-rbac.xsd
adrianjgeorge: Commonality of name is a good thing, then they don't stumble on it while trying to configure.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: Got it. Updated to not be synchronized.

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/examples/simple-rbac.cfg.xml
wdschei: Correct. The CDATA section is not needed and has been removed.

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/config/simple-rbac.xsd
wdschei: The preserve restriction has been implemented.

repose-aggregator/extensions/simple-rbac/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: The documentation annotation was added/updated in the XSD to explain this is intended for simple RBAC, hence the name of the filter, and that ANY and ALL are used as a shorthand for the four standard REST methods.

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/config/simple-rbac.xsd
wdschei: I agree, we should add this capability to all of the existing filters keeping the current hard-coded values as the defaults.

repose-aggregator/extensions/simple-rbac/src/main/resources/META-INF/schema/config/simple-rbac.xsd
wdschei: Updated.

PR #1315, opened 2015-06-03
Ordering Rather Than Wrapping, Auto-Formatting, Test

dmnjohns:
This PR includes a couple of suggestions I have. Sorry about the formatting mess. If you want to format your branch, I'll rebase off of it to clean this PR up. I believe using IntelliJ's auto-formatting is the right call since it provides consistency.

One of the big changes here is to use Scala TreeMap and TreeSet so that we can provide a case-insensitive ordering. The prevents the need to wrap header names. The idea was first implemented in api-checker, as you can see at: https://github.com/rackerlabs/api-checker/blob/master/core/src/main/scala/com/rackspace/com/papi/components/checker/util/HeaderMap.scala

I also reworked getSplittablePreferredHeader. In my opinion, it's much cleaner now. If you like it, we could do the same for getPreferredHeader.

I also added a test to show that getPreferredHeader does not work when there are multiple values on a single line. EDIT: Realized that this test was redundant with the appendHeader getPreferredHeader test, so I removed it and added a note to that test.

### DIFF
--- a/repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
+++ b/repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
@@ -44,17 +44,26 @@ class HttpServletRequestWrapper(originalRequest: HttpServletRequest)
     override def compare(x: String, y: String): Int = x compareToIgnoreCase y
   }
 
-  def getHeaderNamesSet: Set[String] = {
+  def getHeaderNamesScala: Set[String] = {
     super.getHeaderNames.asScala.toSet.filterNot(removedHeaders.contains) ++ headerMap.keySet
   }
 
-  override def getHeaderNames: util.Enumeration[String] = getHeaderNamesSet.toIterator.asJavaEnumeration
+  override def getHeaderNames: util.Enumeration[String] = getHeaderNamesScala.toIterator.asJavaEnumeration
 
-  override def getHeaderNamesList: util.List[String] = getHeaderNamesSet.toList.asJava
+  override def getHeaderNamesList: util.List[String] = getHeaderNamesScala.toList.asJava
 
   override def getIntHeader(headerName: String): Int = Option(getHeader(headerName)).getOrElse("-1").toInt
 
-  override def getHeaders(headerName: String): util.Enumeration[String] = getHeadersScalaList(headerName).toIterator.asJavaEnumeration
+  def getHeadersScala(headerName: String): List[String] = {
+    if (removedHeaders.contains(headerName)) {
+      List[String]()
+    }
+    else {
+      headerMap.getOrElse(headerName, super.getHeaders(headerName).asScala.toList)
+    }
+  }
+
+  override def getHeaders(headerName: String): util.Enumeration[String] = getHeadersScala(headerName).toIterator.asJavaEnumeration
 
   override def getDateHeader(headerName: String): Long = {
     Option(getHeader(headerName)) match {
@@ -67,22 +76,13 @@ class HttpServletRequestWrapper(originalRequest: HttpServletRequest)
     }
   }
 
-  override def getHeader(headerName: String): String = getHeadersScalaList(headerName).headOption.orNull
+  override def getHeader(headerName: String): String = getHeadersScala(headerName).headOption.orNull
 
-  def getHeadersScalaList(headerName: String): List[String] = {
-    if (removedHeaders.contains(headerName)) {
-      List[String]()
-    }
-    else {
-      headerMap.getOrElse(headerName, super.getHeaders(headerName).asScala.toList)
-    }
-  }
-
-  override def getHeadersList(headerName: String): util.List[String] = getHeadersScalaList(headerName).asJava
+  override def getHeadersList(headerName: String): util.List[String] = getHeadersScala(headerName).asJava
 
   override def addHeader(headerName: String, headerValue: String): Unit = {
-    val existingHeaders: List[String] = getHeadersScalaList(headerName) //this has to be done before we remove from the list,
-                                                                        // because getting this list is partially based on the contents of the removed list
+    val existingHeaders: List[String] = getHeadersScala(headerName) //this has to be done before we remove from the list,
+                                                                    // because getting this list is partially based on the contents of the removed list
     if (removedHeaders.contains(headerName)) {
       removedHeaders = removedHeaders.filterNot(_.equalsIgnoreCase(headerName))
     }
@@ -92,13 +92,12 @@ class HttpServletRequestWrapper(originalRequest: HttpServletRequest)
   override def addHeader(headerName: String, headerValue: String, quality: Double): Unit = addHeader(headerName, headerValue + ";q=" + quality)
 
   override def appendHeader(headerName: String, headerValue: String): Unit = {
-    val existingHeaders: List[String] = getHeadersScalaList(headerName)
+    val existingHeaders: List[String] = getHeadersScala(headerName)
     existingHeaders.headOption match {
-      case (Some(value)) => {
+      case Some(value) =>
         val newHeadValue: String = value + "," + headerValue
         headerMap = headerMap + (headerName -> (newHeadValue +: existingHeaders.tail))
-      }
-      case (None) => addHeader(headerName, headerValue)
+      case None => addHeader(headerName, headerValue)
     }
   }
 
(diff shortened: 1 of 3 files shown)


repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
```diff
@@ -63,97 +76,67 @@ class HttpServletRequestWrapper(originalRequest: HttpServletRequest)
...
-  case class HeaderWithParameters(value :String, parameters :Map[String, String])
-  case class HeaderWithQuailty(value :String, quality :Double)
-
-  def getValueWithQuality(headerValues :List[HeaderWithParameters]) :List[HeaderWithQuailty] = {
-    headerValues.map { header =>
-      HeaderWithQuailty(header.value, Option(header.parameters.getOrElse("q", "1.0")).map(_.toDouble).getOrElse(0.0))
+  def getPreferredHeader(headerName: String, getFun: String => List[String]): String = {
+    def parseQuality(headerValue: String): Double = {
+      //todo: what if the parameter value is not a number? Try and default to 1.0?
+      headerValue.split(";").tail.find(param => "q".equalsIgnoreCase(param.split("=")(0).trim))
```
adrianjgeorge: if it's not a number but it was provided i make it 0, might be why you think that it didn't work.
The whole thing is pretty jacked if there are multiple values on a single line, because this method works under the assumption that there are only single values per line.

adrianjgeorge: also this is way less clear

repose-aggregator/commons/utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapperTest.scala
```diff
@@ -93,6 +94,7 @@ class HttpServletRequestWrapperTest extends FunSpec with BeforeAndAfter with Mat
       wrappedRequest.getHeaderNames.asScala.toList should contain theSameElementsAs headerMap.keys.filterNot( _ == "foo")
     }
 
+    //todo: fix this test -- it should not be case sensitive
```
adrianjgeorge: this test is right, your changes are the problem. header names are case insensitive and looks like your changes have broken that, and are returning multiples. plus you've forced us to handle the case insensitivity everywhere else in our code, which someone will forget to do eventually on a change. that's why we made the header name wrapper in the first place.

repose-aggregator/commons/utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapperTest.scala
```diff
@@ -521,6 +523,7 @@ class HttpServletRequestWrapperTest extends FunSpec with BeforeAndAfter with Mat
       wrappedRequest.getPreferredHeader("ornament") shouldBe "santa"
     }
 
+    //todo: fix this test -- shouldn't the higher quality value be returned?
     it("should return the appropriate value when a higher quality is appended onto a line") {
```
adrianjgeorge: this one is the squirrely one. the behavior that it should display is not very clear, and is more as we decide it.

repose-aggregator/commons/utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapperTest.scala
dmnjohns: Right, header names are case insensitive. The header is still there, but add header works by using the name passed as the key in the map. In other words, the casing of the header name passed into addHeader is used rather than the original casing. We have to store it with one casing or another; my changes used the casing provided to the addHeader method, your code uses the casing of the header that existed before calling addHeader.

In what way is case checking needed elsewhere? We never returned the wrapped name from the interface anyway. Or are you talking about internal to this wrapper class? Wrapping made things messy and harder to follow for me, so I proposed this alternative.

repose-aggregator/commons/utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapperTest.scala
dmnjohns: Agreed. Not too hard to fix either way.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
dmnjohns: Alright. Making it 0 is easy enough to do using Try. Also, the assumption that we're looking at a non-splittable header is a fair point. Leave that to the user to decide which method to call. In which case I think this code still works as best it can.

As for clarity, I thought this was better than the three function and two case classes being used before. So. Much. Parsing. Logic. Still, you can write the parseQuality function however you'd like as long as it satisfies the signature, so if using substrings or something else is cleared, that's an easy change.

repose-aggregator/commons/utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapperTest.scala
dmnjohns: I'm kind of leaning toward parsing from the end of the value. Header value parameters always following the value itself, so if a value is not splittable, it kind of makes sense to return everything before the last ;q=<blargh>. This test expects everything before the first ;q=<blargh>.

To clarify, given the header value:
"Foo: bar;q=0.9,baz;q=0.8"

I would expect the return value to be:
"bar;q=0.9,baz"

Is that reasonable?

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
adrianjgeorge: the other thing was a hot mess, but this is a little too obtuse. this might be a case where splitting the single line into multiples with named values provides clarity.

dmnjohns: Fair enough. The neat thing isn't the parsing logic -- that's just frustrating. The neat thing was parameterizing the function to get header values, the matching on that, and using .maxBy passing in the parsing function.

If you decide to adapt this code, you can make it parse however you'd like.

repose-aggregator/commons/utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapperTest.scala
adrianjgeorge: i haven't pulled this down yet, because crap connection, but i assumed this test started failing because foo started showing up twice. If that's not the case, meh. also we should always lean towards keeping the original value. In a number of places in your version you are having to do equalsIgnoreCase versus a standard equals. When i don't have a crap connection, i'll add a couple of tests.

repose-aggregator/commons/utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapperTest.scala
adrianjgeorge: So i went that way at first, but headers read left to right which means the first semi-colon indicates the end of the value and the beginning of parameters.

repose-aggregator/commons/utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapperTest.scala
dmnjohns: Nope, not twice, once with a different case key. As for keeping the original... I don't know that it matters to me. If you're explicitly passing a differently cased header to addHeader, I would lean toward using that since it was the last change made, and it was made explicitly.

Yeah, I wasn't thrilled about the equalsIgnoreCase things. There is probably a better way to do that which leveraged the Ordering, but I'll have to look at it when I get to the office.

repose-aggregator/commons/utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapperTest.scala
dmnjohns: Yeah, that's kind of what I meant by that last block of text in my last comment. Still, going that route means the function breaks if we're given a supposedly non-splittable header that is actually multiple headers the earlier of which has parameters. I guess it doesn't matter because that use case kind of breaks the contract, but parsing from the right doesn't hurt as (as far as I know) so I figured we may as well do that.

Alas, do no care. Please don't break the contract when using the wrapper.

dmnjohns:
I am a tad concerned that, while there is a way to add a header with arbitrary parameters, there is not currently a way to get those parameters back. We only ever return the value with the parameters stripped. Should we perhaps add methods to the interface which allow a user to view parameters? Something like getFullHeader(...).

repose-aggregator/commons/utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapperTest.scala
dmnjohns: Sure enough, getting rid of equalsIgnoreCase is super easy. Rather than:
removedHeaders.filterNot(_.equalsIgnoreCase(headerName))

We could use:
removedHeaders - headerName

adrianjgeorge: that going to work with different casing?

adrianjgeorge:
we actually never strip parameters unless you've asked for the preferred value

PR #1318, opened 2015-06-04
REP-2212 Http Client >4.3 Upgrade

dmnjohns:
Mostly just to run tests...


Mario-Lopez:
Can we get a rebase? 😜

dmnjohns:
Never ever.

But maybe. If we ever get back around to doing this work.

PR #1320, opened 2015-06-08
Rep 2136

adrianjgeorge:


repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/servlet/http/QualityFormatException.java
```diff
@@ -0,0 +1,13 @@
+package org.openrepose.commons.utils.servlet.http;
```
wdschei: Needs a license header.

repose-aggregator/commons/utilities/pom.xml
```diff
@@ -50,6 +50,38 @@
...
+
+        <!-- Mockrunner for mock HttpServletRequest -->
+        <dependency>
+            <groupId>com.mockrunner</groupId>
+            <artifactId>mockrunner-servlet</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.apache.tomcat</groupId>
+                    <artifactId>servlet-api</artifactId>
+                </exclusion>
```
dmnjohns: Should this exclusion be added to the profile-support pom alongside the commons-logging exclusion?

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
```diff
@@ -0,0 +1,150 @@
...
+    if (status == RequestBodyStatus.Reader) throw new IllegalStateException else status = RequestBodyStatus.InputStream
+    inputStream
+  }
+
+  override def getReader: BufferedReader = {
+    if (status == RequestBodyStatus.InputStream) throw new IllegalStateException else status = RequestBodyStatus.Reader
+    new BufferedReader(new InputStreamReader(inputStream))
+  }
+
+  def getHeaderNamesScala: Set[String] = headerMap.keySet ++ super.getHeaderNames.asScala.toSet.filterNot(removedHeaders.contains)
```
dmnjohns: What happens if super.getHeaderNames returns null? As per the HttpServletRequest JavaDoc, the return type of that call is:
"an enumeration of all the header names sent with this request; if the request has no headers, an empty enumeration; if the servlet container does not allow servlets to use this method, null"

I'm pretty sure a NPE will be thrown at present.

This brings up a general concern that our wrapper does not abide by the security contract when getting a header is prevented by the container rather than the contents of the request.

dmnjohns: Oops, I meant this comment to be attached to the getHeaderNames overridden method, not this one...

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
```diff
@@ -0,0 +1,150 @@
...
+  override def getHeader(headerName: String): String = getHeadersScala(headerName).headOption.orNull
+
+  override def getHeadersList(headerName: String): util.List[String] = getHeadersScala(headerName).asJava
+
+  override def addHeader(headerName: String, headerValue: String): Unit = {
+    val existingHeaders: List[String] = getHeadersScala(headerName) //this has to be done before we remove from the list,
+                                                                    // because getting this list is partially based on the contents of the removed list
+    if (removedHeaders.contains(headerName)) {
+      removedHeaders = removedHeaders - headerName
+    }
```
dmnjohns: Thinking about it, this conditional is unnecessary. We can always attempt to remove headerName, although it will only ever have an effect if headerName is in the set.

That is, we can simplify the three lines here to just the middle line.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
```diff
@@ -0,0 +1,150 @@
...
+    headerMap = headerMap + (headerName -> (existingHeaders :+ headerValue))
+  }
+
+  override def addHeader(headerName: String, headerValue: String, quality: Double): Unit = addHeader(headerName, headerValue + ";q=" + quality)
```
dmnjohns: Is appending ";q=blargh" enough here? That is, what if the provided headerValue already contains a quality and this method is being called to override that quality? Will we support that?

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
```diff
@@ -0,0 +1,150 @@
...
+
+  override def appendHeader(headerName: String, headerValue: String): Unit = {
+    val existingHeaders: List[String] = getHeadersScala(headerName)
+    existingHeaders.headOption match {
+      case Some(value) =>
+        val newHeadValue: String = value + "," + headerValue
+        headerMap = headerMap + (headerName -> (newHeadValue +: existingHeaders.tail))
```
dmnjohns: Appending the new header value to the end of the first header line messes with the ordering of headers. Would it be better to append the new value to the end of the last header line? Note that ordering DOES matter with headers.

repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/servlet/http/QualityFormatException.java
adrianjgeorge: Not sure why it didn't fail without it

repose-aggregator/commons/utilities/pom.xml
adrianjgeorge: Probably

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
adrianjgeorge: I'm not overly concerned with the security contract. If we get prevented by the container on our super call that's fine. If they've added one under that name they probably want to be able to access it, so it should work as is. The null thing is freaking stupid and i hate the damn contract. We'll end up wrapping the reques tin an option and accessing it out that way.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
adrianjgeorge: Yeah, this is a holdover from a prior version of this code that the nested code wouldn't have worked without the check before.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
adrianjgeorge: If they've provided a quality already they screwed up and that's not my problem.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
adrianjgeorge: I know it matters, and appending to the first value seemed the most right. Honestly I need to go back in and javadoc the interface really well, so that it's clear what the behavior will be in weird cases.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
dmnjohns: So the case I'm thinking of is when a user calls getHeader or getHeaders and gets back a value with a quality in it.

Imagine you have a request with the following headers:
x-auth-header: panda;q=0.7
x-keystone-header: rat;q=0.7

And you want to modify your request so that those headers are combined like so:
x-auth-header: panda;q=0.7, rat;q=0.8

Is there a way to achieve that with the current code without having to parse the value outside of the wrapper? What I would try and do is get the value of x-keystone-header, append it to x-auth-header with a new quality, then remove x-keystone-header.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
dmnjohns: Fair enough.

dimtruck:
:+1:

dimtruck:
:+1:

dmnjohns:
Looks good, I'll pull it in now.

PR #1323, opened 2015-06-24
Updated the HERP filter to not split the User-Agent header value on ';'.

wdschei:


adrianjgeorge:
@dimtruck you didn't appear to check that the correct value was being logged to herp which is what the bug was. we knew we weren't splitting it elsewhere.

PR #1329, opened 2015-06-30
Rep 2317

mknocke:
Opening a pull request to run my functional test since I cannot run them locally.

### DIFF
--- a/repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/PowerFilter.java
+++ b/repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/PowerFilter.java
@@ -50,6 +50,7 @@ import org.openrepose.core.services.reporting.metrics.MetricsService;
 import org.openrepose.core.services.rms.ResponseMessageService;
 import org.openrepose.core.spring.ReposeSpringProperties;
 import org.openrepose.core.systemmodel.*;
+import org.openrepose.nodeservice.httpcomponent.HttpComponentFactory;
 import org.openrepose.powerfilter.filtercontext.FilterContext;
 import org.openrepose.powerfilter.filtercontext.FilterContextFactory;
 import org.slf4j.Logger;
@@ -383,6 +384,21 @@ public class PowerFilter extends DelegatingFilterProxy {
             mutableHttpResponse.sendError(HttpServletResponse.SC_BAD_GATEWAY, "Error processing request");
             mutableHttpResponse.setLastException(ex);
         } finally {
+            try {
+                boolean isValidMethod = false;
+                for (HttpComponentFactory hcf : HttpComponentFactory.values()) {
+                    if (hcf.name().equalsIgnoreCase(mutableHttpRequest.getMethod())) {
+                        isValidMethod = true;
+                    }
+                }
+                if (!isValidMethod) {
+                    throw new Exception();
+                }
+            } catch (Exception e) {
+                LOG.debug("{}:{} -- Invalid HTTP method requested: {}", clusterId, nodeId, mutableHttpRequest.getMethod(), e);
+                mutableHttpResponse.sendError(HttpServletResponse.SC_BAD_REQUEST, "Error processing request");
+                mutableHttpResponse.setLastException(e);
+            }
             // In the case where we pass/route the request, there is a chance that
             // the response will be committed by an underlying service, outside of repose
             if (!mutableHttpResponse.isCommitted()) {
--- a/repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/PowerFilter.java
+++ b/repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/PowerFilter.java
@@ -362,8 +362,12 @@ public class PowerFilter extends DelegatingFilterProxy {
 
         MDC.put(TracingKey.TRACING_KEY, traceGUID);
         try {
-            // ensures that the method name exists
-            HttpComponentFactory.valueOf(mutableHttpRequest.getMethod().toUpperCase());
+            try {
+                // ensures that the method name exists
+                HttpComponentFactory.valueOf(mutableHttpRequest.getMethod().toUpperCase());
+            } catch (IllegalArgumentException iae) {
+                throw new InvalidMethodException("Request contained an unknown method.", iae);
+            }
             // ensures that the request URI is a valid URI
             new URI(mutableHttpRequest.getRequestURI());
             final PowerFilterChain requestFilterChain = getRequestFilterChain(mutableHttpResponse, chain);
@@ -378,10 +382,10 @@ public class PowerFilter extends DelegatingFilterProxy {
                 }
                 requestFilterChain.startFilterChain(mutableHttpRequest, mutableHttpResponse);
             }
-        } catch (IllegalArgumentException iae) {
-            LOG.debug("{}:{} -- Invalid HTTP method requested: {}", clusterId, nodeId, mutableHttpRequest.getMethod(), iae);
+        } catch (InvalidMethodException ime) {
+            LOG.debug("{}:{} -- Invalid HTTP method requested: {}", clusterId, nodeId, mutableHttpRequest.getMethod(), ime);
             mutableHttpResponse.sendError(HttpServletResponse.SC_BAD_REQUEST, "Error processing request");
-            mutableHttpResponse.setLastException(iae);
+            mutableHttpResponse.setLastException(ime);
         } catch (URISyntaxException use) {
             LOG.debug("{}:{} -- Invalid URI requested: {}", clusterId, nodeId, mutableHttpRequest.getRequestURI(), use);
             mutableHttpResponse.sendError(HttpServletResponse.SC_BAD_REQUEST, "Error processing request");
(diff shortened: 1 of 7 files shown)


adrianjgeorge:
ok to test

repose-aggregator/core/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
```diff
@@ -149,6 +154,20 @@ class ReposeJettyServer(val clusterId: String,
...
+    val security: ConstraintSecurityHandler = new ConstraintSecurityHandler()
+    s.setHandler(security)
+
+    //Adds a constraint that will throw 405 when method does not exist
+    val constraint = new Constraint()
+    constraint.setName("methods")
+
+    val mapping = new ConstraintMapping()
+    mapping.setPathSpec("/*")
+    mapping.setMethodOmissions(HttpComponentFactory.values.map(method => method.toString))
```
dmnjohns: method => method.toString could be replaced by _.toString

In this simple case, I'd suggest it. It kind of comes down to preference, though.

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/core/proxy/InvalidRequestMethodTest.groovy
```diff
@@ -0,0 +1,59 @@
...
+    def "Should return 405 when method name is invalid for request"() {
+        when:
+        MessageChain mc = deproxy.makeRequest([url: reposeEndpoint, method: method, requestBody: reqBody])
+
+        then:
+        mc.receivedResponse.code == "405"
+
+        where:
+        method  | reqBody
+        "pull"  | "blah"
```
dmnjohns: I know you probably aren't done, but you may want to test all of the methods that we allow as well. Just to be comprehensive.

dmnjohns: Also, I'm not sure that you care about parameterizing the request body. I'd probably just unroll a list of methods.

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/core/proxy/InvalidRequestMethodTest.groovy
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    def cleanupSpec() {
+        if (repose) {
+            repose.stop()
+        }
+        if (deproxy) {
+            deproxy.shutdown()
+        }
+    }
```
dmnjohns: This cleanupSpec isn't necessary. Your test inherits a cleanupSpec method from the ReposeValveTest class which does the same thing.

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/core/proxy/InvalidRequestMethodTest.groovy
```diff
@@ -0,0 +1,61 @@
...
+
+        then:
+        mc.receivedResponse.code == "405"
+
+        where:
+        method << ["pull", "derp", "invalid"]
+    }
+
+    @Unroll
+    def "Should return 200 when method name is valid for request"() {
```
dmnjohns: Not something you have to do, but something you could do that's kind of neat: use #<var> in the method name to replace it with parameterized values from your test.

In other words, you could have the test description:
"Should return 200 when method name (#method) is valid for request"

Or something cleaner, but you get the idea. It can be nice to see in the description what exactly is being tested.

repose-aggregator/core/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
```diff
@@ -35,6 +35,8 @@ import org.springframework.web.context.ContextLoaderListener
 import org.springframework.web.context.support.AnnotationConfigWebApplicationContext
 import org.springframework.web.filter.DelegatingFilterProxy
 
+import scala.collection.JavaConverters._
+
```
dmnjohns: Delete this.

repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/PowerFilter.java
```diff
@@ -383,6 +384,21 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
+                    }
+                }
+                if (!isValidMethod) {
+                    throw new Exception();
+                }
+            } catch (Exception e) {
+                LOG.debug("{}:{} -- Invalid HTTP method requested: {}", clusterId, nodeId, mutableHttpRequest.getMethod(), e);
+                mutableHttpResponse.sendError(HttpServletResponse.SC_BAD_REQUEST, "Error processing request");
+                mutableHttpResponse.setLastException(e);
+            }
```
dmnjohns: You don't want this in the finally block. It needs to run before the filter chain is run.

repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/PowerFilter.java
```diff
@@ -383,6 +384,21 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
         } finally {
+            try {
+                boolean isValidMethod = false;
+                for (HttpComponentFactory hcf : HttpComponentFactory.values()) {
+                    if (hcf.name().equalsIgnoreCase(mutableHttpRequest.getMethod())) {
+                        isValidMethod = true;
+                    }
+                }
+                if (!isValidMethod) {
+                    throw new Exception();
```
dmnjohns: You probably don't need to use a try catch with an exception here. You already know the case, so just do the work instead of handling it as an exception.

repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/PowerFilter.java
```diff
@@ -383,6 +384,21 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
             mutableHttpResponse.sendError(HttpServletResponse.SC_BAD_GATEWAY, "Error processing request");
             mutableHttpResponse.setLastException(ex);
         } finally {
+            try {
+                boolean isValidMethod = false;
+                for (HttpComponentFactory hcf : HttpComponentFactory.values()) {
+                    if (hcf.name().equalsIgnoreCase(mutableHttpRequest.getMethod())) {
```
dmnjohns: Arrays.asList(yourArray).contains(yourValue)

No need for the boolean variable. Just use a conditional based on the line above. If the method is not supported, reject the request.

repose-aggregator/core/valve/pom.xml
```diff
@@ -117,6 +117,12 @@
         </dependency>
 
         <dependency>
+            <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-security</artifactId>
+            <version>${jetty.version}</version>
+        </dependency>
```
dmnjohns: You can probably kill this dependency off too.

repose-aggregator/core/valve/src/test/scala/org/openrepose/valve/ReposeJettyServerTest.scala
```diff
@@ -199,7 +199,5 @@ class ReposeJettyServerTest extends FunSpec with Matchers with BeforeAndAfterAll
     expectedProperties.foreach { case (k, v) =>
       server.appContext.getEnvironment.getProperty(k) shouldBe v
     }
-
   }
-
```
dmnjohns: Might as well revert this file to master. With the route we're going, this change is unrelated.

repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/PowerFilter.java
```diff
@@ -374,6 +378,10 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
                 }
                 requestFilterChain.startFilterChain(mutableHttpRequest, mutableHttpResponse);
             }
+        } catch (IllegalArgumentException iae) {
```
dmnjohns: This is the only thing that makes me a little nervous. This is a fairly common exception, and I'd hate to log an invalid method when that isn't the cause.

One alternative would be to break the check into its own try block. That would let us fail earlier (not that we spend much time processing prior to this) and catch the exception on a very isolated portion of the code.

I'll defer to @adrianjgeorge

dimtruck:
This looks good to QE.  Good to merge whenever.

repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/PowerFilter.java
adrianjgeorge: I agree see if you can't scope it down, so that it's only catching the exception at the call level. Without looking in depth at the code you may end up catching it, and then wrapping it in a new exception and throwing that to catch where you are currently catching this one

repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/PowerFilter.java
```diff
@@ -374,6 +382,10 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
                 }
                 requestFilterChain.startFilterChain(mutableHttpRequest, mutableHttpResponse);
             }
+        } catch (InvalidMethodException ime) {
+            LOG.debug("{}:{} -- Invalid HTTP method requested: {}", clusterId, nodeId, mutableHttpRequest.getMethod(), ime);
+            mutableHttpResponse.sendError(HttpServletResponse.SC_BAD_REQUEST, "Error processing request");
+            mutableHttpResponse.setLastException(ime);
```
dmnjohns: It works, so I :+1: 

However, I wonder which is cleaner/clearer: this approach, or handling the check in its own try/catch block earlier in the method?

dmnjohns: Quickly looking at the code, if we used a try/catch earlier, we'd lose all of the work that happens in the finally.

The proposed approach gives us clarity and lets us capture some metrics by making use of the finally block.

Ship it.

PR #1332, opened 2015-07-03
Test and Fix for Auth/Rate Limiting Bug

dmnjohns:
This relates to https://repose.atlassian.net/browse/REP-2377

As you can see, the problem is the way the filter director and mutableservletrequest handle quality. Rather than treating quality as part of the header value, it is stored separately. See:
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/http/header/HeaderValueImpl.java#L71

So currently, we use add header to add the "107;q=1.0" value which we've constructed to the "x-pp-groups" header. When the header is actually added, it is given a hard coded quality of 1.0. See:
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/http/header/HeaderValueImpl.java#L44

When we try to get the preferred "x-pp-groups" header, we get "107;q=1.0" rather than simply "107" because the wrapper has made the distinction between value and quality, and treated the value we added as just the value.

~~This is also why manually passing the "x-pp-groups" header works. We don't use the filter director to add the header, it's just parsed from the request that we pass when we wrap it.~~

THIS IS WHY I REALLY, REALLY WANT TO SWITCH TO THE NEW WRAPPERS AND KILL THE OLD WRAPPERS AND FILTER DIRECTOR.

**EDIT**: An accurate explanation of the exact issue is in the PR comments below. Please read those.

### DIFF
--- a/repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/framework/mocks/MockIdentityService.groovy
+++ b/repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/framework/mocks/MockIdentityService.groovy
@@ -241,7 +241,7 @@ class MockIdentityService {
                 }
             }
 
-        } else if (nonQueryPath.startsWith("/users/")) {
+        } else if (nonQueryPath.startsWith("/v2.0/users/")) {
 
             if (isGetGroupsCallPath(nonQueryPath)) {
                 if (method == "GET") {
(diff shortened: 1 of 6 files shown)


repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/framework/mocks/MockIdentityService.groovy
```diff
@@ -241,7 +241,7 @@ class MockIdentityService {
                 }
             }
 
-        } else if (nonQueryPath.startsWith("/users/")) {
+        } else if (nonQueryPath.startsWith("/v2.0/users/")) {
```
dmnjohns: I've added /v2.0 to this mock in a few places because... well, I'm not sure why it wasn't there or how our other tests pass. I imagine this may cause other tests to fail. Ugh.

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/framework/mocks/MockIdentityService.groovy
```diff
@@ -587,7 +587,7 @@ class MockIdentityService {
             """{
   "RAX-KSGRP:groups": [
     {
-        "id": "0",
+        "id": "107",
```
dmnjohns: I probably should have left this as 0 and configured that to be used for the test. This may break other tests. Again, this was mostly just used to diagnose what was going on.

dmnjohns:
So I lied. The wrapper totally doesn't handle qualities, that part is true. If a header value with a quality is passed to the wrapper, it stores the quality and value together and returns the full string when getHeader or getPreferredHeader is used. So you're probably wondering how this was working before. Well that was my lie. It is header splitting. Well, technically it's not the splitting that makes a difference, but the parsing that is done when splitting is attempted.

Take a look at the PowerFilterChain: 
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/PowerFilterChain.java#L124

Notice that we clear all of the header values parsed by the wrapper, and replace them with the "split" header values that have undergone additional parsing by way of the HeaderFieldParser and HeaderValueParser classes. In contrast, the MutableServletRequest uses HeaderValuesImpl.extract to parse header fields when a request is wrapper.

dmnjohns:
So while this fix is still valid, since headers values have a default quality of 1.0 anyway, perhaps the better fix would be to make the MutableServletRequest/Response wrappers use the HeaderFieldParser to parse headers.

Of course, there probably isn't a point since we're working on switching over to the new wrappers which will also solve this problem.

dmnjohns:
This test may be useful later, but I'm going to close this PR since the code should not be pulled in.

This PR will serve to define the bug and the information we have around it.

PR #1333, opened 2015-07-03
Removing quality as part of the value

dmnjohns:
Here's the quick and dirty fix.

If all tests pass, this is probably good to go.

Refer to #1332 for more info.


dmnjohns:
This would still need to be done for Identity v3. It's likely other filters suffer from the same bug.

PR #1334, opened 2015-07-06
Rep 2356: No longer sends empty header for trace id

mknocke:


adrianjgeorge:
ok to test

dmnjohns:
The story (REP-1356) mentions this bug happening when sending requests to Identity. Were you able to confirm that that no longer happens? I don't believe the RequestProxyServiceImpl class is used by the client auth filter.

mknocke:
Removed the unit tests using mockrunner since they bring an unwanted dependency that causes SaxAuthFeedReaderTest to fail. The functional test covers the cases the unit tests covered.

repose-aggregator/external/service-clients/auth/src/main/java/org/openrepose/common/auth/openstack/AuthenticationServiceClient.java
```diff
@@ -376,7 +376,9 @@ private String getAdminToken(String requestGuid, boolean force) throws AuthServi
         try {
             if (adminToken == null) {
                 Map<String, String> headerMap = new HashMap<>();
-                headerMap.put(CommonHttpHeader.TRACE_GUID.toString(), requestGuid);
+                if (requestGuid != null) {
```
adrianjgeorge: Should probably do the not empty check instead of a null check

PR #1335, opened 2015-07-07
REP-2335 Prevent malformed return values when the API Validator filter returns a Saxon error.

wdschei:
This has gone from an API-Checker modification to add an option for only using the WADL rax:message in order to prevent malformed values being returned, to making the ResponseMessageHandler a bit more capable.

There were several places were the redundant <> generics definitions were removed.
The HttpLogFormatterTest.WhenParsingSimpleArguments had a one line static method replaced by the actual code to increase readability.
Several tests were reordered in HttpLogFormatterTest.WhenSettingLogic and the couple that were missing were added.

### DIFF
--- a/repose-aggregator/core/core-lib/src/main/java/org/openrepose/core/services/rms/ImmutableFormatTemplates.java
+++ b/repose-aggregator/core/core-lib/src/main/java/org/openrepose/core/services/rms/ImmutableFormatTemplates.java
@@ -21,6 +21,7 @@ package org.openrepose.core.services.rms;
 
 import org.openrepose.commons.utils.StringUtilities;
 import org.openrepose.commons.utils.logging.apache.HttpLogFormatter;
+import org.openrepose.commons.utils.logging.apache.HttpLogFormatterState;
 import org.openrepose.core.services.rms.config.Message;
 import org.openrepose.core.services.rms.config.StatusCodeMatcher;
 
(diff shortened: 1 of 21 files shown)


repose-aggregator/core/core-lib/src/main/java/org/openrepose/core/services/rms/ImmutableFormatTemplates.java
```diff
@@ -43,8 +44,10 @@ private ImmutableFormatTemplates(List<StatusCodeMatcher> statusCodes) {
                 final String statusCodeId = statusCode.getId();
                 final String href = message.getHref();
                 final String stringTemplate = !StringUtilities.isBlank(href) ? new HrefFileReader().read(href, statusCodeId) : message.getValue();
-
-                formatTemplates.put(statusCodeId + message.getMediaType(), new HttpLogFormatter(stringTemplate));
+                final String mediaType = message.getMediaType();
```
adrianjgeorge: this should be content type not media type. media type has to do with what is coming back from the origin service. We care about what kind of message they are trying to build which is content-type

repose-aggregator/core/core-lib/src/main/java/org/openrepose/core/services/rms/MessageFilter.java
```diff
@@ -34,52 +34,71 @@
...
+        if (messages != null && mediaType != null && contentTypes != null) {
+            for (MediaType contentType : contentTypes) {
                 for (Message message : messages) {
                     final String messageMediaType = message.getMediaType();
-
-
-                    if (StringUtilities.nullSafeEqualsIgnoreCase(messageMediaType, mediaType.getValue())) {
-                        return message;
+                    final String messageContentType = message.getContentType();
+                    if (StringUtils.equalsIgnoreCase(messageMediaType, mediaType)) {
```
adrianjgeorge: what's this change about?

adrianjgeorge:
You did a lot more changes here, what's the purpose?

adrianjgeorge:
once we have tests passing, it looks good.

wdschei:
retest this

wdschei:
retest this

wdschei:
retest this

wdschei:
Retest this please.

PR #1336, opened 2015-07-15
Rep 2378 intra filter logging nulls

Mario-Lopez:
When trace logging is on, PowerFilterChain.java will log the request and response along with the filter information.  Filters have an optional ID which was getting logged as null along side the filter name (e.g. "null-ip-identify" for a filter named "ip-identity" with no ID).

I added code to not log the filter ID when it's null or an empty string.  ResponseLog.java and RequestLog.java were updated to just use the filter name when the ID is empty (i.e. null or an empty string).

These classes didn't have unit tests, so I added new classes to unit test the updates I made (using Scala Test).  I also updated the functional test (IntrafilterLoggingTest.groovy which already had two filters configured) to make one of the filters have an ID and the other one not have an ID.  The test verifies that "context_1-ip-identity" and "client-auth" (no ID) gets logged respectively.  All unit tests pass, and the functional test that was updated passes as well.


wdschei:
Ok to test.

adrianjgeorge:
ok to test

repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/intrafilterLogging/RequestLog.java
```diff
@@ -41,33 +44,58 @@
...
+     * @param filterContext {@link FilterContext}
+     * @throws IOException if there's an issue converting the response body to a string
+     */
     public RequestLog(MutableHttpServletRequest mutableHttpServletRequest,
                       FilterContext filterContext) throws IOException {
 
         preamble = "Intrafilter Request Log";
         timestamp = new DateTime().toString();
-        currentFilter = filterContext.getFilterConfig().getId() + "-" + filterContext.getFilterConfig().getName();
+        currentFilter = getFilterDescription(filterContext.getFilterConfig());
```
adrianjgeorge: you shouldn't call instance methods in a constructor behavior can be erratic. Either make it a static method or better yet just do a simple ternary.

repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/intrafilterLogging/ResponseLog.java
```diff
@@ -39,12 +45,18 @@
...
+     * @param filterContext {@link FilterContext}
+     * @throws IOException if there's an issue converting the response body to a string
+     */
     public ResponseLog(MutableHttpServletResponse mutableHttpServletResponse,
                        FilterContext filterContext) throws IOException {
 
         preamble = "Intrafilter Response Log";
         timestamp = new DateTime().toString();
-        currentFilter = filterContext.getFilterConfig().getId() + "-" + filterContext.getFilterConfig().getName();
+        currentFilter = getFilterDescription(filterContext.getFilterConfig());
```
adrianjgeorge: same here

adrianjgeorge:
You added a lot of javadocs that don't really provide any value, because they basically say the exact same thing the method and variable names do. In general we don't add javadocs unless something is none obvious about usage.

Mario-Lopez:
I'll update it to a ternary op.  That JavaDoc, tho...  Most of the JavaDocs I added were for code that took me more than a minute to reason about.  I removed the docs for one of the methods.  Will the rest of the JavaDocs keep you up at night if I leave them in?

Mario-Lopez:
retest this please

Mario-Lopez:
retest this please

repose-aggregator/core/core-lib/src/test/scala/org/openrepose/powerfilter/intrafilterLogging/RequestLogTest.scala
```diff
@@ -0,0 +1,101 @@
+package org.openrepose.powerfilter.intrafilterLogging
```
wdschei: This file needs the License header to run on Jenkins.

repose-aggregator/core/core-lib/src/test/scala/org/openrepose/powerfilter/intrafilterLogging/ResponseLogTest.scala
```diff
@@ -0,0 +1,97 @@
+package org.openrepose.powerfilter.intrafilterLogging
```
wdschei: This file needs the License header to run on Jenkins.

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/core/intrafilterlogging/IntraFilterLoggingTest.groovy
```diff
@@ -0,0 +1,62 @@
+package features.core.intrafilterlogging
```
wdschei: This file needs the License header to run on Jenkins.

dmnjohns:
retest this please

repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/intrafilterLogging/RequestLog.java
```diff
@@ -41,17 +44,25 @@
     String requestBody;
     HashMap<String, String> headers;
 
+    /**
+     * Constructor populates all of the fields necessary for logging.
```
adrianjgeorge: It's not going to keep me up at night, but what value do you see in this? it tells me absolutely nothing that wasn't in already communicated by the names of the parameters

Mario-Lopez: There are a few things perplexing about this class.  I usually don't see package-level visibility for class fields.  To me that implies that there is another class in that same package that may be populating these fields (especially considering there are no setter methods).  I looked.  There is not.  Either it was a mistake or the serialization code requires it for some reason.  In any case, that particular JavaDoc addresses that.  The next time I have to go through this code, I don't have to worry about what else might be populating the fields.  I will know that at some point someone decided to tightly couple this POJO with MutableHttpServletRequest and FilterContext for some reason, and this JavaDoc allows me to remember this instantly.

repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/intrafilterLogging/RequestLog.java
```diff
@@ -20,17 +20,20 @@
...
 import javax.servlet.ServletInputStream;
 import java.io.IOException;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.List;
+import java.util.*;
 
+/**
+ * POJO that is used to log details about a response when the log level is set to TRACE.
```
adrianjgeorge: You've created an inaccuracy here. the behavior of this class doesn't care about the logging level, that's only relevant in where it's called.

Mario-Lopez: Potential inaccuracy*

Considering the constructor contains the logic for what gets logged (i.e. what goes into the fields), I thought the expected log level was relevant here.  I can change "when the log level is set to TRACE" to "(may not always be logged due to log level)" in order to decrease the likeliness of the documentation becoming stale.

Mario-Lopez: My apologies.  No documentation is better than poor documentation.  I'll push an update.

repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/intrafilterLogging/RequestLog.java
Mario-Lopez: My apologies.  No documentation is better than poor documentation.  I'll push an update.

adrianjgeorge:
you removed the line comments? those seemed to be the more valuable ones

PR #1343, opened 2015-07-21
Rep 2404 - DD Service not sending trace id

mknocke:
Cannot run functional tests on my computer, so opening pull request.

### DIFF
--- a/repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/core/tracing/TracingDistDatastoreTest.groovy
+++ b/repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/core/tracing/TracingDistDatastoreTest.groovy
@@ -18,6 +18,7 @@
  * =_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_=_
  */
 package features.core.tracing
+
 import framework.ReposeValveTest
 import framework.mocks.MockIdentityService
 import org.openrepose.commons.utils.io.ObjectSerializer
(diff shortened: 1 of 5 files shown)


wdschei:
ok to test

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/core/tracing/TracingDistDatastoreTest.groovy
```diff
@@ -96,4 +98,55 @@ class TracingDistDatastoreTest extends ReposeValveTest {
         lines.size() == 1
     }
 
+    def "DistDatastore service request should be added the tracing header"() {
```
adrianjgeorge: this test appears to have nothing to do with what this story is about?

repose-aggregator/services/datastore/impl/src/main/java/org/openrepose/core/services/datastore/impl/distributed/remote/command/AbstractRemoteCommand.java
```diff
@@ -57,8 +58,9 @@ public String getBaseUrl() {
     }
 
     protected Map<String, String> getHeaders(RemoteBehavior remoteBehavior) {
-        Map<String, String> headers = new HashMap<String, String>();
+        Map<String, String> headers = new HashMap<>();
         headers.put(DatastoreHeader.HOST_KEY.toString(), hostKey);
+        headers.put(CommonHttpHeader.TRACE_GUID.toString(), remoteEndpoint.getHostName());
```
adrianjgeorge: you should be using the same guid that is currently being run under on this node

adrianjgeorge:
Also, unlike the others you should always do this one, this is a case where we are internal within repose so we always want it

PR #1344, opened 2015-07-21
Updated the WAR pom to include the missing Log4J 2.x jars and to exclude the old commons-logging.

wdschei:
Also put the Spring dependencies in the same order as the Valve POM to make for easier comparisons.

### DIFF
--- a/repose-aggregator/core/web-application/pom.xml
+++ b/repose-aggregator/core/web-application/pom.xml
@@ -53,7 +53,13 @@
 
         <dependency>
             <groupId>org.springframework</groupId>
-            <artifactId>spring-beans</artifactId>
+            <artifactId>spring-web</artifactId>
+            <exclusions>
+                <exclusion>
+                    <artifactId>commons-logging</artifactId>
+                    <groupId>commons-logging</groupId>
+                </exclusion>
+            </exclusions>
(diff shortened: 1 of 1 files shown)


adrianjgeorge:
who was doing the exclusion or changing in scope?

wdschei:
These were not being excluded, they were just never being included. Since these are not needed at compile time they are not pulled in as dependencies. Therefore they have to be explicitly included like they were in the Valve POM. This is what the WAR POM change is.

This needs a test to confirm that all currently released 7.x versions fail and this branch forward pass. I just don't know what that would look like in our WAR testing framework. I will get with QE to figure this out.

dkowis:
I might be wrong, but doesn't the container provide the logging infrastructure in general with WAR file deployments?

wdschei:
The containers may supposed to be, but the internal Glassfish deployments were failing until they were added to the WAR.

adrianjgeorge:
the container is under no obligation to provide any sort of logging infrastructure to the app being run. It may have logging of its own and it's logs might include the apps depending on if they are fundamentally the same system, but it's not a guarantee. We need to be sure we provide the concrete implementations so that our logging works in either case.

PR #1345, opened 2015-07-22
Rep 2355 rewrite trans id header

Mario-Lopez:
Added "rewrite-tracing-header" attribute to the system model as a boolean to allow users to configure Repose to ignore inbound transaction ids but still have tracing enabled.  The default value is false.

Update was made to PowerFilter.java to remove the "X-Trans-Id" header when the "rewrite-tracing-header" attribute is set to true.  PowerFilter.java does not have an existing unit test, and the class is complex enough that adding a unit test would take a significant amount of time.

I added a functional test (RewriteTracingHeaderTest.groovy) to verify that the request and the response get a new "X-Trans-Id" header whether or not one was passed into the original request.  I also ran the TracingHeaderDisabledTest to regression test the original behavior.

I also added a unit test to verify that the Repose version number does not start with 8 (i.e. when we move to version 8, the unit test will fail).  As a part of making a major version number change, we should refactor the tracing attributes on the system model into their own element.  The unit test includes this information in a comment.

### DIFF
--- a/repose-aggregator/core/core-lib/src/test/scala/org/openrepose/core/spring/CoreSpringProviderTest.scala
+++ b/repose-aggregator/core/core-lib/src/test/scala/org/openrepose/core/spring/CoreSpringProviderTest.scala
@@ -102,4 +102,22 @@ class CoreSpringProviderTest extends FunSpec with Matchers with TestFilterBundle
     }
   }
 
+  describe("Repose Spring Properties") {
+    describe("Repose Version") {
+      it("is not 8") {
+        val reposeVersion = coreSpringProvider.getCoreContext.getEnvironment.getProperty(
+          ReposeSpringProperties.stripSpringValueStupidity(ReposeSpringProperties.CORE.REPOSE_VERSION))
+
+        reposeVersion should not startWith "8"
+
+        /*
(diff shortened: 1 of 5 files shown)


repose-aggregator/core/core-lib/src/test/scala/org/openrepose/core/spring/CoreSpringProviderTest.scala
```diff
@@ -102,4 +102,22 @@ class CoreSpringProviderTest extends FunSpec with Matchers with TestFilterBundle
     }
   }
 
+  describe("Repose Spring Properties") {
```
adrianjgeorge: when we talk about them, we call these timebomb tests. Could you add the word time bomb in the title or comment so that if we do a search for timebomb it'll pop up?

dimtruck:
Looks good to QE

repose-aggregator/core/core-lib/src/test/scala/org/openrepose/core/spring/CoreSpringProviderTest.scala
Mario-Lopez: Added!

PR #1346, opened 2015-07-24
New Response Wrapper Skeleton

adrianjgeorge:
Everyone take a look, tell me if there's any case you can think of that i haven't accounted for.
This isn't all the methods we'll have to write, these are just the ones that came with the interface (so i could document the fail state), and the ones we'd be adding.

### DIFF
--- a/repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
+++ b/repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
@@ -42,12 +42,14 @@ import javax.servlet.http.HttpServletResponse
  * stream, which in the case of one of the constructors is the original response's output stream. A mode of
  * ResponseMode.READONLY will see us wrapping the provided output stream in one that will record everything that goes
  * through but still passing the data along to the provided stream. This can be accessed by calling
- * getOutputStreamAsInputStream. If you desire to see the output after it went through a provided output stream you can
- * use our wrapping stream xxxxxxx yourself by first wrapping the original response's output stream in our wrapper and
- * then wrapping our stream with your own, and passing that in and setting the mode to ResponseMode.PASSTHROUGH.
- * ResponseMode.MUTABLE behaves similiarly, but will not write to the underlying output stream until commitToResponse is
- * called. Additionally, you can call setOutput with an input stream which we will wrap in an output stream which will
- * be used to write to the underlying output stream when commitToResponse is called.
+ * getOutputStreamAsInputStream. It should be noted that this will keep a secondary copy of the response in memory,
+ * which for large responses could be expensive. If you desire to see the output after it went through a provided output
+ * stream you can use our wrapping stream xxxxxxx yourself by first wrapping the original response's output stream in
+ * our wrapper and then wrapping our stream with your own, and passing that in and setting the mode to
+ * ResponseMode.PASSTHROUGH. ResponseMode.MUTABLE behaves similiarly, but will not write to the underlying output stream
+ * until commitToResponse is called. Additionally, you can call setOutput with an input stream which we will wrap in an
+ * output stream which will be used to write to the underlying output stream when commitToResponse is called. This mode
+ * will also break chunked encoding because the response can't be streamed directly through.
  *
  * @constructor the main constructor to be used for this class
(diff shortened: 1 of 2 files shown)


repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -0,0 +1,127 @@
...
+   * @param originalResponse the response to be wrapped
+   * @param headerMode the mode to use for header accessiblity
+   * @param bodyMode the mode to use for body accessibility
+   */
+  def this(originalResponse: HttpServletResponse, headerMode: ResponseMode, bodyMode: ResponseMode) =
+    this(originalResponse, headerMode, bodyMode, originalResponse.getOutputStream)
+
+  override def getHeaderNamesList: util.List[String] = ???
+
+  override def addHeader(headerName: String, headerValue: String, quality: Double): Unit = ???
```
dmnjohns: I think we talked about this on Friday, but that was days ago. What happened to the addHeader method without a quality?

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -0,0 +1,127 @@
...
+ * called. Additionally, you can call setOutput with an input stream which we will wrap in an output stream which will
+ * be used to write to the underlying output stream when commitToResponse is called.
+ *
+ * @constructor the main constructor to be used for this class
+ * @param originalResponse the response to be wrapped
+ * @param headerMode the read/write-ability mode of headers
+ * @param bodyMode the read/write-ability mode of the body
+ * @param desiredOutputStream the underlying output stream to be presented callers of getOutputStream, maybe wrapped
+ *                            depending on mode selection of the body
+ */
```
dmnjohns: What are the performance and memory usage implications of large body sizes? I'm actually not worried about performance, but it seems like any mode other than PASSTHROUGH for the body will require maintaining a buffer which could hit memory pretty hard. It is what it is, but should that be noted here?

We may also want to note that a mutable body mode will break chunked encoding.

dmnjohns:
So say I wanted to use this wrapper to inject a header value into the body (like really stupid templating). How would I do that?

It looks like I would have to wrap the response with this wrapper, pass it along, wait until doChain returns, call getOutputStreamAsInputStream, read the entire stream replacing the templated part and sticking the result into an input stream, then call setOutput on the resultant input stream. Is that correct?

If so, the servlet spec doesn't seem to handle that case very well...

dmnjohns:
Could we perhaps pass some context to the chained output stream that the user provides in the constructor? That way the user-provided output stream has access to header information (and possibly more) within the scope of the wrapper. That would make it way easier to conditionally mutate the body. I wouldn't even have to do any additional processing after calling doFilter.

dmnjohns:
This looks pretty good on the whole. I made some comments that are mostly just to help me understand this stuff better because even though I've worked with it a bit, it still seems pretty complex in my head. Still, I'm giving it a :+1: pending the addHeader without quality method.

Also, apologies for any stupid comments I made. It's a Monday or something.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
adrianjgeorge: It's part of the base class, so it didn't get stubbed out. We will do an implementation though.

dmnjohns: That's right. Gotcha.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
adrianjgeorge: You are correct, while it's obvious to me that's what's happening a comment to that effect is probably worthwhile

adrianjgeorge:
That is the process, and no the servlet spec doesn't handle this well.

PR #1347, opened 2015-07-27
Added support for filters that don't need spring annotations

Mario-Lopez:
I removed the Named annotation from DerpFilter and verified that the functional tests DerpAndClientAuthNDelegable.groovy, DerpAndDelegableQuality.groovy, and DerpResponseMessagingTest.groovy failed.  I also updated FilterContextFactoryTest.scala to remove the existing unit test that verifies that an unannotated filter should cause an exception to be thrown and added a test to verify it won't.  The unit and functional tests failed in the correct way.

I updated FilterContextFactory.java to first check if the Spring application context had any beans for the configured class, and if not then it'll try creating a new instance of that class using reflection.  I removed the exception handling for the case where we try to get the bean from the application context and it doesn't exist (since we first check now) and added exception handling for the case where we can't use reflection to create an instance of the class (e.g. if the class doesn't have a public no-arg constructor).

The check I added to see if the app context has any beans for the class is not ideal.  I didn't find a method to directly check if there were any beans for a class (as opposed to using a string).  Using exception handling as control flow seemed like a bad idea.  I resorted to getting the list of bean names for the class and checking whether or not I get back an empty array.

I verified that the functional and unit tests pass again after the update.

### DIFF
--- a/repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/filtercontext/FilterContextFactory.java
+++ b/repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/filtercontext/FilterContextFactory.java
@@ -27,7 +27,6 @@ import org.openrepose.core.systemmodel.Filter;
 import org.openrepose.powerfilter.FilterInitializationException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.support.AbstractApplicationContext;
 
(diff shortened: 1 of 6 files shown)


repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/filtercontext/FilterContextFactory.java
```diff
@@ -104,7 +103,9 @@ private FilterContext loadFilterContext(Filter filter, Collection<EarClassLoader
             //Get the specific class to load from the application context
             Class c = filterClassLoader.loadClass(filterType.getFilterClass().getValue());
 
-            final javax.servlet.Filter newFilterInstance = (javax.servlet.Filter) filterContext.getBean(c);
+            //If Spring didn't load the filter as a bean, then try manually creating a new instance of the class
+            final javax.servlet.Filter newFilterInstance = (javax.servlet.Filter)
+                    (filterContext.getBeanNamesForType(c).length > 0 ? filterContext.getBean(c) : c.newInstance());
```
adrianjgeorge: Go ahead and stick with handling it as an exception instead of by bean name. We don't have full control of the bean name so i wouldn't count on it.

Mario-Lopez: Update pushed!

adrianjgeorge:
Ship It

PR #1348, opened 2015-07-29
REP-2494 DerpFilter delegation message not making it to ResponseMessageHandler.

wdschei:
# tl;dr

If the initial `MutableHttpServletResponse` wrapper was re-wrapped by another type (e.g. `CheckerServletResponse`), then the delegation message set in the `DerpFilter` was never propagated to the `ResponseMessageHandler`.

# full
The `ResponseMessageHandler` is expecting to retrieve the message from `MutableHttpServletResponse.getMessage()`.

* https://github.com/rackerlabs/repose/blob/master/repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/logging/apache/format/stock/ResponseMessageHandler.java#L40

This message is set using the `HttpServletResponse.sendError(int, String)` on a `MutableHttpServletResponse`. The delegation message was originally only being set in the `DerpFilter` using `HttpServletResponse.getWriter.write(String)`.

* https://github.com/rackerlabs/repose/blob/master/repose-aggregator/components/filters/derp/src/main/scala/org/openrepose/filters/derp/DerpFilter.scala#L81

This could not be picked up by the `ResponseMessageHandler` since it was not being set as expected. The delegation message is now also being set in the `DerpFilter` using this overloaded `HttpServletResponse.sendError()`:

* https://github.com/rackerlabs/repose/blob/REP-2494_CloudFeedsIssue/repose-aggregator/components/filters/derp/src/main/scala/org/openrepose/filters/derp/DerpFilter.scala#L84

When *API-Checker* is in the `FilterChain` (e.g. `api-validator`, `simple-rbac`), the `ServletResponse` passed into `PowerFilterChain.doFilter()` at the following is a `CheckerServletResponse`:

* https://github.com/rackerlabs/api-checker/blob/master/core/src/main/scala/com/rackspace/com/papi/components/checker/handler/DelegationHandler.scala#L45

This is immediately wrapped in a `MutableHttpServletResponse` and sent on down the `FilterChain`:

* https://github.com/rackerlabs/repose/blob/master/repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/PowerFilterChain.java#L184-L185

The `HttpServletResponse` provides no mechanism to retrieve the message that was set using `sendError()`. The `HttpServletResponseWrapper`, is the base implementation of `HttpServletResponse`, and it's direct descendant `CheckerServletResponse` doesn't either.

So even though the second layer `MutableHttpServletResponse` had the delegation message set directly in the `DerpFilter` using the overloaded `HttpServletResponse.sendError()`,
that particular message was never passed on to the intermediary `CheckerServletResponse` or in turn the inner original  `MutableHttpServletResponse` for later retrieval by `ResponseMessageHandler`.

### DIFF
--- a/repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/servlet/http/MutableHttpServletResponse.java
+++ b/repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/servlet/http/MutableHttpServletResponse.java
@@ -302,9 +302,14 @@ public class MutableHttpServletResponse extends HttpServletResponseWrapper imple
 
     @Override
     public void sendError(int code, String message) {
-        super.setStatus(code);
-        this.message = message;
+        setStatus(code, message);
         error = true;
+    }
+
+    @Override
+    public void setStatus(int code, String message) {
+        super.setStatus(code, message);
+        this.message = message;
     }
 
     public boolean isError() {
--- a/repose-aggregator/components/filters/derp/src/main/scala/org/openrepose/filters/derp/DerpFilter.scala
+++ b/repose-aggregator/components/filters/derp/src/main/scala/org/openrepose/filters/derp/DerpFilter.scala
@@ -81,7 +81,7 @@ class DerpFilter extends Filter with HttpDelegationManager with LazyLogging {
     httpServletResponse.setContentLength(responseBody.length)
     httpServletResponse.setContentType(MediaType.TEXT_PLAIN)
     httpServletResponse.getWriter.write(responseBody)
-    httpServletResponse.sendError(statusCode)
+    httpServletResponse.sendError(statusCode, responseBody)
   }
 
   override def destroy(): Unit = {
--- a/repose-aggregator/components/filters/derp/src/test/scala/org/openrepose/filters/derp/DerpFilterTest.scala
+++ b/repose-aggregator/components/filters/derp/src/test/scala/org/openrepose/filters/derp/DerpFilterTest.scala
@@ -57,8 +57,7 @@ class DerpFilterTest extends FunSpec {
 
       derpFilter.doFilter(req, resp, null)
 
-      verify(resp).sendError(404)
-      verify(respWriter).write("not found")
+      verify(resp).sendError(404, "not found")
     }
 
     it("should send an error response corresponding to the delegation value with the highest quality") {
@@ -70,8 +69,7 @@ class DerpFilterTest extends FunSpec {
 
       derpFilter.doFilter(req, resp, null)
 
-      verify(resp).sendError(500)
-      verify(respWriter).write("bar")
+      verify(resp).sendError(500, "bar")
     }
 
     it("should send an error response corresponding to the delegation value with the highest quality, reverse order") {
@@ -83,8 +81,7 @@ class DerpFilterTest extends FunSpec {
 
       derpFilter.doFilter(req, resp, null)
 
-      verify(resp).sendError(500)
-      verify(respWriter).write("bar")
+      verify(resp).sendError(500, "bar")
     }
 
     it("should reject the request if no delegation value could be parsed") {
@@ -98,8 +95,10 @@ class DerpFilterTest extends FunSpec {
       derpFilter.doFilter(req, resp, fc)
 
       verify(fc, never()).doFilter(any(classOf[ServletRequest]), any(classOf[ServletResponse]))
-      verify(resp).sendError(500)
-      verify(respWriter).write(anyString())
+      //verify(resp).sendError(eq(500), anyString())                    // Won't compile; says to use type ascription
+      //verify(resp).sendError(eq(500:java.lang.Integer), anyString())  // Won't compile; type mismatch; found: Boolean; required: Int
+      //verify(resp).sendError(eq(500:Int), anyString())                // Won't compile; type mismatch; found: Int; required: AnyRef
+      verify(resp).sendError(500, anyString())                          // Runtime failure: Invalid use of argument matchers!
     }
 
     it("should treat a delegation value without an explicit quality as having a quality of 1") {
@@ -111,8 +110,7 @@ class DerpFilterTest extends FunSpec {
 
       derpFilter.doFilter(req, resp, null)
 
-      verify(resp).sendError(500)
-      verify(respWriter).write("bar")
+      verify(resp).sendError(500, "bar")
     }
   }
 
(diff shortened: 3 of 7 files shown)


repose-aggregator/components/filters/derp/src/main/scala/org/openrepose/filters/derp/DerpFilter.scala
```diff
@@ -81,7 +81,7 @@ class DerpFilter extends Filter with HttpDelegationManager with LazyLogging {
     httpServletResponse.setContentLength(responseBody.length)
     httpServletResponse.setContentType(MediaType.TEXT_PLAIN)
     httpServletResponse.getWriter.write(responseBody)
```
adrianjgeorge: this should probably go too

dmnjohns:
All the good things!

This is kind of what I suspected when you were asking about this stuff on IRC this morning. It's the only thing that made sense.

repose-aggregator/components/filters/derp/src/test/scala/org/openrepose/filters/derp/DerpFilterTest.scala
```diff
@@ -98,8 +95,10 @@ class DerpFilterTest extends FunSpec {
       derpFilter.doFilter(req, resp, fc)
 
       verify(fc, never()).doFilter(any(classOf[ServletRequest]), any(classOf[ServletResponse]))
-      verify(resp).sendError(500)
-      verify(respWriter).write(anyString())
+      //verify(resp).sendError(eq(500), anyString())                    // Won't compile; says to use type ascription
+      //verify(resp).sendError(eq(500:java.lang.Integer), anyString())  // Won't compile; type mismatch; found: Boolean; required: Int
+      //verify(resp).sendError(eq(500:Int), anyString())                // Won't compile; type mismatch; found: Int; required: AnyRef
+      verify(resp).sendError(500, anyString())                          // Runtime failure: Invalid use of argument matchers!
```
wdschei: Anyone have a clue of the right way to make Mokito work with Scala here?

repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/servlet/http/MutableHttpServletResponse.java
```diff
@@ -302,11 +302,16 @@ public void sendError(int code) throws IOException {
...
-        super.setStatus(code);
-        this.message = message;
+        setStatus(code, message);
         error = true;
     }
 
+    @Override
+    public void setStatus(int code, String message) {
+        super.setStatus(code, message);
+        this.message = message;
```
dmnjohns: I hadn't been paying that much attention to this until now, but this is kind of a weird thing to do.

First, the double parameter setStatus method has been deprecated (due to the lack of parameter clarity). See: https://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpServletResponse.html#setStatus(int, java.lang.String)

Second, we're now changing both the status message and the body of the response to be identical. That's kind of odd.

Third, sendError already has the correct behavior, I believe. Our sendError could just chain into super.sendError, store the message, and set the flag. Having our send error call super.setStatus(x, y) is changing the contracted behavior of sendError, which normally writes a body and not the status message. See:
https://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpServletResponse.html#sendError(int, java.lang.String)

Have I mentioned how much I dislike our mutable wrapper jank? =P

repose-aggregator/components/filters/derp/src/main/scala/org/openrepose/filters/derp/DerpFilter.scala
```diff
@@ -81,7 +81,7 @@ class DerpFilter extends Filter with HttpDelegationManager with LazyLogging {
...
-    httpServletResponse.sendError(statusCode)
+    httpServletResponse.sendError(statusCode, responseBody)
```
dmnjohns: If we fix sendError in the wrapper, we can get rid of the getWriter.write line again. That line is only used because using sendError on our wrapper is/was setting the status message rather than the body.

dimtruck:
@wdschei and I spoke and the tests that I was going to push are superfluous. We can merge whenever.

repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/servlet/http/MutableHttpServletResponse.java
wdschei: Since the original single parameter `sendError()` method was calling the single parameter `super.setStatus()` method instead of the single parameter `super.sendError()` method I opted to do the same for the two parameter flavor in an effort to minimize the behavior change as much as possible until we can get the new wrappers in place.

Yes, the two parameter flavor of the `setStatus()` method is deprecated *"due to ambiguous meaning of the message parameter"*, but it is this ambiguity that is biting us to begin with.

If we didn't do the write to the body, then the `ResponseMessageService` configuration could also use the *"IF_EMPTY"* override again. We would just have to provide a default catch-all `status-code` element in the `response-messaging.cfg.xml` that contained:
```
    <status-code id="Default" code-regex="\d\d\d" overwrite="IF_EMPTY">
        <message media-type="*/*">
            %M
        </message>
    </status-code>
```
But I will leave that for another story.

repose-aggregator/components/filters/derp/src/main/scala/org/openrepose/filters/derp/DerpFilter.scala
wdschei: Yep.

repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/servlet/http/MutableHttpServletResponse.java
wdschei: But hopefully all this is moot and the new wrappers will be in place and we don't have to worry about it.

dmnjohns: I expect the reason for sendError calling super.setStatus rather than super.sendError is that super.sendError commits the output stream, which we defer until the end with our wrapper. So using setStatus is probably the right call.

Even so, the second parameter of setStatus is not the same thing as the second parameter of sendError. Conflating the two leads to very non-obvious behavior, in my opinion.

Yes, the ambiguity is what is biting us. Which seems like a reason not to use it if we don't have to.

Your point about the ResponseMessagingService is true, but it seems like this is the wrong fix for it. In general, we don't write the status message. We send error messages as response bodies. For one, the error message may be fairly verbose, and so it doesn't belong on the status line. The fact of the matter is that delegating filters are trying to communicate some error. We shouldn't squelch that error, nor should we stick it in the status line. Perhaps a better alternative would be to send empty messages when delegating and a message is not beneficial. That was IF_EMPTY would still catch the response.

wdschei: Another way to do this would be to have `DerpFilter` add a *Delegation Message* header. This would propagate through all the wrappers. The `ResponseMessagingService` could retrieve it and removed it before being returned to the client.

PR #1349, opened 2015-08-03
Rep 2521 tenant id case sensitive

Mario-Lopez:
I updated Client Auth to be case sensitive when testing for Tenant ID equality.

Jenny added a functional test to verify that Client Auth is case sensitive when testing for Tenant ID equality.  Before my update, the functional test failed as expected.  After the update, the test passed.  Unit tests also correctly failed before the update and passed after the update.

Keystone v3 doesn't need to be updated:
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/components/filters/openstack-identity-v3/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3Handler.scala#L325


adrianjgeorge:
Did you check auth-z as well? I can't think of any checks it might do, but better safe than sorry.

Mario-Lopez:
I didn't see any Tenant ID checks in the auth-z code, but there are some Tenant Role checks.  I wrote REP-2541 to track it.

PR #1352, opened 2015-08-07
Rep 2379 logging timestamp formatting

Mario-Lopez:
The slf4j http logging string already had support for adding a {} between the %t characters, so I added support for putting a date format in there.  Not specifying a format defaults to the current behavior.  A timebomb was added to change the default format in version 8.  Details about the formatting are documented here (under %{format}t):
https://repose.atlassian.net/wiki/display/REPOSE/SLF4J+HTTP+Logging+filter#SLF4JHTTPLoggingfilter-Whatislogged

Functional tests were added to confirm the existing default behavior and to confirm the custom date format behavior.

### DIFF
--- a/repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/logging/apache/format/stock/TimeReceivedHandler.java
+++ b/repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/logging/apache/format/stock/TimeReceivedHandler.java
@@ -35,6 +35,9 @@ public class TimeReceivedHandler implements FormatterLogic {
 
     public TimeReceivedHandler(final String dateFormat) {
         this.dateFormat = StringUtils.isEmpty(dateFormat) ? DEFAULT_DATE_FORMAT : dateFormat;
+
+        // build it here so an exception will be thrown during application startup if the pattern is invalid
+        new SimpleDateFormat(dateFormat);
     }
 
     @Override
--- a/repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/logging/apache/format/stock/TimeReceivedHandler.java
+++ b/repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/logging/apache/format/stock/TimeReceivedHandler.java
@@ -19,6 +19,7 @@
  */
 package org.openrepose.commons.utils.logging.apache.format.stock;
 
+import org.apache.commons.lang3.StringUtils;
 import org.openrepose.commons.utils.logging.apache.format.FormatterLogic;
 
 import javax.servlet.http.HttpServletRequest;
@@ -28,8 +29,20 @@ import java.util.Calendar;
 
 public class TimeReceivedHandler implements FormatterLogic {
 
+    private final static String DEFAULT_DATE_FORMAT = "dd-MM-yyyy-HH:mm:ss.SSS";
+
+    private final String dateFormat;
+
+    public TimeReceivedHandler(final String dateFormat) {
+        this.dateFormat = StringUtils.isEmpty(dateFormat) ? DEFAULT_DATE_FORMAT : dateFormat;
+    }
+
     @Override
     public String handle(HttpServletRequest request, HttpServletResponse response) {
-        return new SimpleDateFormat("dd-MM-yyyy-HH:mm:ss.SSS").format(Calendar.getInstance().getTime());
+        return new SimpleDateFormat(dateFormat).format(Calendar.getInstance().getTime());
+    }
+
+    public String getDateFormat() {
+        return dateFormat;
     }
 }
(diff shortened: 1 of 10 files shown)


repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/logging/apache/format/stock/TimeReceivedHandler.java
```diff
@@ -28,8 +29,20 @@
...
+        this.dateFormat = StringUtils.isEmpty(dateFormat) ? DEFAULT_DATE_FORMAT : dateFormat;
+    }
+
     @Override
     public String handle(HttpServletRequest request, HttpServletResponse response) {
-        return new SimpleDateFormat("dd-MM-yyyy-HH:mm:ss.SSS").format(Calendar.getInstance().getTime());
+        return new SimpleDateFormat(dateFormat).format(Calendar.getInstance().getTime());
+    }
+
+    public String getDateFormat() {
```
adrianjgeorge: why is this here?

repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/logging/apache/format/stock/TimeReceivedHandler.java
```diff
@@ -28,8 +29,20 @@
...
+    private final String dateFormat;
+
+    public TimeReceivedHandler(final String dateFormat) {
+        this.dateFormat = StringUtils.isEmpty(dateFormat) ? DEFAULT_DATE_FORMAT : dateFormat;
+    }
+
     @Override
     public String handle(HttpServletRequest request, HttpServletResponse response) {
-        return new SimpleDateFormat("dd-MM-yyyy-HH:mm:ss.SSS").format(Calendar.getInstance().getTime());
+        return new SimpleDateFormat(dateFormat).format(Calendar.getInstance().getTime());
```
adrianjgeorge: the constructor errors out if it's a bad format right? if so i'd suggest building it in the constructor, that failure should chain up and prevent repose from starting, instead of bombing after the request has been successfully processed by the origin service.

repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/logging/apache/format/stock/TimeReceivedHandler.java
Mario-Lopez: HttpLogFormatterTest.java has some string parsing tests, and unfortunately it's forced to be more like an integration test than a unit test.  The existing tests just throw some data in and then verify it gets the correct data out, but the TimeReceivedHandler code is grabbing the current time instead of using some input data I can manipulate.  To verify the parsing code is working, I can either use this method to compare the the date format strings or I could grab the output of the class and use a regex to see if it looks like the correct date format was used.

Should I remove the method and go with the regex check?

repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/logging/apache/format/stock/TimeReceivedHandler.java
Mario-Lopez: Good call.  I'll add this in.

Mario-Lopez: Update pushed.

repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/logging/apache/format/stock/TimeReceivedHandler.java
adrianjgeorge: yeah, lets go with that. We try to avoid test code bleeding into prod.

repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/logging/apache/format/stock/TimeReceivedHandler.java
```diff
@@ -28,8 +29,23 @@
...
 
+    private final static String DEFAULT_DATE_FORMAT = "dd-MM-yyyy-HH:mm:ss.SSS";
+
+    private final String dateFormat;
+
+    public TimeReceivedHandler(final String dateFormat) {
+        this.dateFormat = StringUtils.isEmpty(dateFormat) ? DEFAULT_DATE_FORMAT : dateFormat;
+
+        // build it here so an exception will be thrown during application startup if the pattern is invalid
+        new SimpleDateFormat(dateFormat);
```
adrianjgeorge: Not sure why you didn't store it and use it?

Mario-Lopez: SimpleDateFormat is not thread-safe.  I thought about adding locking around it, but I wasn't sure what that would buy us over simply creating a new one each time.

repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/logging/apache/format/stock/TimeReceivedHandler.java
Mario-Lopez: kk, I'll remove the getter.

repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/logging/apache/format/stock/TimeReceivedHandler.java
adrianjgeorge: fair enough

PR #1353, opened 2015-08-07
Add test to ensure repose forward x-roles from auth to origin service…

jennyvo:
… through these filter client-auth, api-validator, herp and derp


adrianjgeorge:
ok to test

PR #1362, opened 2015-08-27
REP-2464 Auth Filters Are (Destructively) Replacing Roles Header

dmnjohns:


repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/openstack/OpenStackAuthenticationHeaderManager.java
```diff
@@ -189,6 +189,7 @@ private void setTenant() {
             }
         } else if (!sendAllTenantIds && sendTenantIdQuality) {
             if (!this.isDelagable && !this.isTenanted) {
+                //TODO: Should this, or any other, instance of putHeader be changed to appendHeader?
```
dmnjohns: I guess it would be beneficial to look at all the putHeader(...)s here.

wdschei: There should probably be a new story to review all uses of `putHeader()` and `appendHeader()` just to make sure the right one is being used.

dmnjohns: Thinking about it, I might add it to the agenda for a dev meeting. That way we can get input from the entire team. I'd be more comfortable with that then a spike or something where one person makes the decision.

wdschei: Good call. We just open a branch and do a group pairing session.

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/filters/clientauthn/ClientAuthFilterAddHeadersTest.groovy
```diff
@@ -0,0 +1,143 @@
...
+        (mc.handlings[0].request.headers.findAll("x-roles").toString()).contains("not-admin")
+        (mc.handlings[0].request.headers.findAll("x-roles").toString()).contains("compute:default")
+        (mc.handlings[0].request.headers.findAll("x-roles").toString()).contains("object-store:default")
+        (mc.handlings[0].request.headers.findAll("x-roles").toString()).contains("test-add-role")
+        (mc.handlings[0].request.headers.findAll("x-pp-user").toString()).contains("repose-test-add-user")
+        (mc.handlings[0].request.headers.findAll("x-pp-user").toString()).contains("username")
+    }
+
+    // REP-2464: Auth filter should add headers not replace headers
+    @Ignore ("We can turn on when impersonator role to header merge in to branch")
```
wdschei: Why is there an `Ignore`d test in this brand new file?

wdschei: Answered my own question.

repose-aggregator/components/filters/openstack-identity-v3/src/test/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3HandlerTest.scala
```diff
@@ -159,6 +159,23 @@ class OpenStackIdentityV3HandlerTest extends FunSpec with BeforeAndAfter with Ma
...
+      mockRequest.setHeader("X-Subject-Token", "123456")
+      identityConfig.setForwardGroups(false)
+      identityConfig.setValidateProjectIdInUri(null)
+      identityV3Handler = new OpenStackIdentityV3Handler(identityConfig, identityAPI)
+      identityV3Handler.handleRequest(mockRequest, mockServletResponse).requestHeaderManager.headersToAdd should contain(
+        Entry(
+          HeaderName.wrap(OpenStackServiceHeader.ROLES.toString),
+          JavaConversions.setAsJavaSet(Set("admin")))
+      )
+      identityV3Handler.handleRequest(mockRequest, mockServletResponse).requestHeaderManager.headersToRemove should not contain HeaderName.wrap(OpenStackServiceHeader.ROLES.toString)
```
wdschei: Wouldn't it be clearer if the `mockRequest` started with an `X-Roles` header that was not *admin* and then check to make sure it was still present and not marked for removal instead of calling `handleRequest()` twice to put the exact same header?

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/filters/clientauthn/ClientAuthFilterAddHeadersTest.groovy
dmnjohns: Not that I know of... I would think that should be of REP-2607? Not sure why 2464 is annotated here.

repose-aggregator/components/filters/openstack-identity-v3/src/test/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3HandlerTest.scala
dmnjohns: Yes. Good call.

Well, kind of. So since this filter is adhering to the old pattern which leverages the FilterDirector, no changes are actually made to the request itself. Plus I don't want to test the FilterDirector here. So I'll assume that the FilterDirector works. In that case, since the output of doFilter is a FilterDirector, I can only inspect that. So I just have to verify that the FilterDirector is not removing the header, but is still adding it.

Two calls to handleRequest should not have been called. I'll fix that.

dmnjohns:
All of these test failures are time-bombs.

I'll bump the Servlet spec ones which I know will not pass yet, and figure out what to do with the rest in a separate PR.

PR #1363, opened 2015-08-30
Rep 2517 session info in tracing header

Mario-Lopez:
The Tracing Header (x-trans-id) now supports containing JSON data that is Base64 encoded.

TracingHeaderHelper.scala was added to provide convenience methods for grabbing the Tracing UUID out of the encoded header (which is what the header used to contain), creating a new encoded header with a given Tracing UUID and VIA header (to signify that Repose created the header), and simply decoding the header but leaving the JSON data intact for logging purposes.  These methods are "safe" in that if the header provided is not properly encoded, they will return the original string so processing can continue.

PowerFilter.java was updated to use the helper and to log the JSON data so that it could be correlated with the Tracing UUID (i.e. Request ID) in other log lines.  Several other classes were updated to use the helper class.  The TracingKey class was moved to be in the same artifact and package as the helper class.  A new TraceGuidHandler was created to use the helper when logging the Trace GUID, and HttpLogFormatter was updated to use it.

PhoneHomeService.scala still needs to be updated to use the helper class.

The Tracing specific functional tests were run and pass, but I'm going to let Jenkins run the full suite of functional tests to see if they all pass.

### DIFF
--- a/repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/logging/TracingHeaderHelper.scala
+++ b/repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/logging/TracingHeaderHelper.scala
@@ -41,6 +41,7 @@ object TracingHeaderHelper extends LazyLogging {
   // JSON contents
   private val TraceGuidKey = "requestId"
   private val OriginKey = "origin"
+  private val UserKey = "user"
 
   def getTraceGuid(tracingHeader: String): String = {
     (Option(MDC.get(TracingKey.TRACING_KEY)), Option(tracingHeader)) match {
@@ -58,9 +59,17 @@ object TracingHeaderHelper extends LazyLogging {
   }
 
   def createTracingHeader(requestId: String, origin: String): String = {
+    createTracingHeader(requestId, origin, None)
+  }
+
+  def createTracingHeader(requestId: String, origin: String, user: String): String = {
+    createTracingHeader(requestId, origin, Option(user))
+  }
+
+  def createTracingHeader(requestId: String, origin: String, user: Option[String]): String = {
     Base64.encodeBase64String(
       ObjectMapper.writeValueAsBytes(
-        Map(TraceGuidKey -> requestId, OriginKey -> origin).asJava))
+        (Map(TraceGuidKey -> requestId, OriginKey -> origin) ++ user.map(UserKey -> _)).asJava))
   }
 
   def decode(tracingHeader: String): String = {
--- a/repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/logging/TracingHeaderHelper.scala
+++ b/repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/logging/TracingHeaderHelper.scala
@@ -1,0 +1,69 @@
+/*
+ * _=_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_=
+ * Repose
+ * _-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-
+ * Copyright (C) 2010 - 2015 Rackspace US, Inc.
+ * _-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-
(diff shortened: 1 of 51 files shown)


repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/logging/TracingHeaderHelper.scala
```diff
@@ -0,0 +1,69 @@
...
+        }
+    }
+  }
+
+  def createTracingHeader(requestId: String, origin: String): String = {
+    Base64.encodeBase64String(
+      ObjectMapper.writeValueAsBytes(
+        Map(TraceGuidKey -> requestId, OriginKey -> origin).asJava))
+  }
+
```
adrianjgeorge: should probably add one for when we have the username

repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/atomfeed/sax/SaxAuthFeedReader.java
```diff
@@ -122,7 +123,8 @@ private ServiceClientResponse getFeed(String traceID) throws FeedException {
         final Map<String, String> headers = new HashMap<>();
 
         if (isOutboundTracing) {
-            headers.put(CommonHttpHeader.TRACE_GUID.toString(), traceID);
+            headers.put(CommonHttpHeader.TRACE_GUID.toString(), TracingHeaderHelper.createTracingHeader(
+                    traceID, headers.get(CommonHttpHeader.VIA.toString())));
```
adrianjgeorge: nothing here to get

repose-aggregator/services/datastore/impl/src/main/java/org/openrepose/core/services/datastore/impl/distributed/remote/command/AbstractRemoteCommand.java
```diff
@@ -62,7 +61,8 @@ public String getBaseUrl() {
     protected Map<String, String> getHeaders(RemoteBehavior remoteBehavior) {
         Map<String, String> headers = new HashMap<>();
         headers.put(DatastoreHeader.HOST_KEY.toString(), hostKey);
-        headers.put(CommonHttpHeader.TRACE_GUID.toString(), MDC.get(TracingKey.TRACING_KEY));
+        headers.put(CommonHttpHeader.TRACE_GUID.toString(),
+                TracingHeaderHelper.createTracingHeader(TracingHeaderHelper.getTraceGuid(null), null));
```
adrianjgeorge: this one should be the original, not a new one

adrianjgeorge:
did i miss the part in the AkkaServiceClient where you take the header and get the trans id out to put in the logging context?

Mario-Lopez:
I took a look at everywhere we set the logging context, and they all seem to be accounted for.  Am I missing something?

https://github.com/rackerlabs/repose/search?utf8=%E2%9C%93&q=MDC

wdschei:
The `AkkaServiceClientImpl` uses the `AuthGetRequest` and `AuthPostRequest` which both extend `ActorRequest`. `ActorRequest` calls `MDC.getCopyOfContextMap()` which includes the `Trans ID` in the current *Mapped Diagnostic Context* and stores a local copy. The locally stored copy of the context map, including the `Trans ID`, is used to populate the active *Mapped Diagnostic Context* in `AuthTokenFutureActor`.

- https://github.com/rackerlabs/repose/blob/41c6f6a20962dd514361c4b2d390fa1b2d598122/repose-aggregator/services/service-client/impl/src/main/java/org/openrepose/core/services/serviceclient/akka/impl/ActorRequest.java#L13-L16
- https://github.com/rackerlabs/repose/blob/41c6f6a20962dd514361c4b2d390fa1b2d598122/repose-aggregator/services/service-client/impl/src/main/java/org/openrepose/core/services/serviceclient/akka/impl/AuthTokenFutureActor.java#L47

repose-aggregator/services/datastore/impl/src/main/java/org/openrepose/core/services/datastore/impl/distributed/remote/command/AbstractRemoteCommand.java
wdschei: This is an abstraction nightmare for the sake of abstracting abstractions. :-(

repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/atomfeed/sax/SaxAuthFeedReader.java
wdschei: Updated with a Repose templated Via header string.

PR #1368, opened 2015-09-02
Bumping time bombs

dmnjohns:


repose-aggregator/components/filters/rate-limiting/src/test/java/org/openrepose/filters/ratelimiting/RateLimitingHandlerTest.java
```diff
@@ -193,6 +193,9 @@ public void shouldRejectDescribeLimitsCallwith406() {
             assertEquals("On rejected media type, returned status code must be 406", HttpServletResponse.SC_NOT_ACCEPTABLE, director.getResponseStatusCode());
         }
 
+        // This test doesn't make sense. If the accept header is set to a media type that the rate limiting filter
+        // cannot handle, the rate limiting filter should return a 406, or a default media type (ignoring the accept
+        // header)
```
dmnjohns: Actually, I guess that could be what's going on here. We're adding an accept header of "application/xml" to fetch any limits from the origin service, then returning the combined limits in whichever format we decide. That is, we're ignoring the accept header value that is not supported. Except the code currently doesn't do this, so it would probably be a new story.

Is it time to rewrite rate limiting yet?

PR #1369, opened 2015-09-03
Updating bypass validation roles to pre authorized roles

dmnjohns:


dmnjohns:
The two lists in this configuration should just be one.

PR #1374, opened 2015-09-16
REP-2253 Servlet Response Wrapper

dmnjohns:


repose-aggregator/commons/utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapperTest.scala
```diff
@@ -0,0 +1,1411 @@
...
+import org.scalatest.junit.JUnitRunner
+import org.scalatest.mock.MockitoSugar
+import org.scalatest.{BeforeAndAfter, FunSpec, Matchers}
+
+import scala.io.Source
+
+@RunWith(classOf[JUnitRunner])
+class HttpServletResponseWrapperTest extends FunSpec with BeforeAndAfter with Matchers with MockitoSugar {
+
+  // Think of: casing, preceding/succeeding components, ordering, quality
```
dmnjohns: I should probably remove this note now.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -46,84 +53,253 @@ import javax.servlet.http.HttpServletResponse
...
+ * ResponseMode.PASSTHROUGH. ResponseMode.MUTABLE behaves similarly, but will not write to the underlying output stream
  * until commitToResponse is called. Additionally, you can call setOutput with an input stream which we will wrap in an
  * output stream which will be used to write to the underlying output stream when commitToResponse is called. This mode
  * will also break chunked encoding because the response can't be streamed directly through.
  *
+ * Note that mutation to the [[HttpServletResponse]] will be treated by this wrapper as permanent. Therefore, mutation
+ * should not be performed prior to downstream processing. In practice, mutation should not be performed on the response
+ * prior to calling the doFilter() method. The reason is that writing to a [[ServletOutputStream]] or
+ * [[PrintWriter]] can not be undone, so for the sake of consistency, headers are handled in the same manner. In
+ * fact, headers written prior to wrapping a request will not appear to exist within the wrapped response.
```
adrianjgeorge: response not request

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -46,84 +53,253 @@ import javax.servlet.http.HttpServletResponse
...
+  override def getPreferredSplittableHeaders(name: String): util.List[String] =
+    getPreferredHeaderValues(getSplittableHeaders(name))
+
+  override def getSplittableHeaders(name: String): util.List[String] =
+    getHeaderValues(name).foldLeft(List.empty[String])((list, value) => list ++ value.split(","))
+
+  /**
+   * @throws IllegalStateException when headerMode is anything other than ResponseMode.MUTABLE
+   */
+  override def addHeader(name: String, value: String): Unit = {
```
adrianjgeorge: This behavior isn't correct adds should always work, it's just a question of whether they immediately push through.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -46,84 +53,253 @@ import javax.servlet.http.HttpServletResponse
...
    */
-  override def appendHeader(headerName: String, headerValue: String): Unit = ???
+  override def addDateHeader(name: String, timeSinceEpoch: Long): Unit =
+    addHeader(name, DateUtils.formatDate(new Date(timeSinceEpoch)))
 
   /**
    * @throws IllegalStateException when headerMode is anything other than ResponseMode.MUTABLE
    */
-  override def appendHeader(headerName: String, headerValue: String, quality: Double): Unit = ???
+  override def appendHeader(name: String, value: String): Unit = {
```
adrianjgeorge: this looks like it would succeed regardless of mode if there was already a value.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -46,84 +53,253 @@ import javax.servlet.http.HttpServletResponse
...
    */
-  override def removeHeader(headerName: String): Unit = ???
+  override def appendHeader(name: String, value: String, quality: Double): Unit =
+    appendHeader(name, s"$value;q=$quality")
 
   /**
    * @throws IllegalStateException when headerMode is anything other than ResponseMode.MUTABLE
    */
-  override def replaceHeader(headerName: String, headerValue: String): Unit = ???
+  override def setHeader(name: String, value: String): Unit = {
```
adrianjgeorge: same problem as add

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -46,84 +53,253 @@ import javax.servlet.http.HttpServletResponse
...
+          }
+        }
+      }
+    )
+
+    //Write the body, if the body mode is set to mutable
+    val writeBody = Try(bodyOutputStream.commit())
+
+    //Throw a failure exception (both should be identical) if neither the header nor body mode is set to mutable
+    if (writeHeaders.isFailure && writeBody.isFailure) writeHeaders.get
```
adrianjgeorge: should be an or not an and, then you have to pay attention to which one failed

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -46,84 +53,253 @@ import javax.servlet.http.HttpServletResponse
...
+    responseBodyType match {
+      case ResponseBodyType.PrintWriter =>
+        throw new IllegalStateException("Cannot call getOutputStream after calling getWriter")
+      case _ =>
+        responseBodyType = ResponseBodyType.OutputStream
+        bodyOutputStream
+    }
+  }
+
+  override def flushBuffer(): Unit = {
```
adrianjgeorge: Have fun fixing this and implementing everything.

repose-aggregator/commons/utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapperTest.scala
```diff
@@ -0,0 +1,1408 @@
...
+      wrappedResponse.getPreferredHeaders("a") should be('empty)
+    }
+
+    it("should not return query parameters in the header value") {
+      val wrappedResponse = new HttpServletResponseWrapper(originalResponse, ResponseMode.MUTABLE, ResponseMode.PASSTHROUGH)
+
+      wrappedResponse.addHeader("A", "a;q=0.8;foo=bar")
+      wrappedResponse.addHeader("A", "b;bar=baz")
+
+      wrappedResponse.getPreferredSplittableHeaders("A") should contain only "b"
```
jennyvo: is this should be getPreferredHeaders ?
and should contain only ("a", "b") ?

repose-aggregator/commons/utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapperTest.scala
```diff
@@ -0,0 +1,1408 @@
...
+      wrappedResponse.getHeaderNamesList should contain only "b"
+    }
+
+    it("should return header names with the casing they were added with") {
+      val wrappedResponse = new HttpServletResponseWrapper(originalResponse, ResponseMode.MUTABLE, ResponseMode.PASSTHROUGH)
+
+      wrappedResponse.addHeader("A", "a")
+      wrappedResponse.addHeader("b", "b")
+
+      wrappedResponse.getHeaderNamesList should contain only("A", "b")
```
jennyvo: contain only("A", "a", "b")?

repose-aggregator/commons/utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapperTest.scala
```diff
@@ -0,0 +1,1408 @@
...
+
+      wrappedResponse.getPreferredSplittableHeaders("a") should be('empty)
+    }
+
+    it("should not return query parameters in the header value") {
+      val wrappedResponse = new HttpServletResponseWrapper(originalResponse, ResponseMode.MUTABLE, ResponseMode.PASSTHROUGH)
+
+      wrappedResponse.addHeader("A", "a;q=0.8;foo=bar,b;bar=baz")
+
+      wrappedResponse.getPreferredSplittableHeaders("A") should contain only "b"
```
jennyvo: should contain only("a","b") ?

repose-aggregator/commons/utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapperTest.scala
dmnjohns: It should absolutely be getPreferredHeaders, however it should still only contain "b" since it has a quality of 1.0 as opposed to "a"'s quality of 0.8.

repose-aggregator/commons/utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapperTest.scala
dmnjohns: Only "b" is correct for the same reason as above.

repose-aggregator/commons/utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapperTest.scala
```diff
@@ -0,0 +1,1408 @@
...
+      originalResponse.getHeaders("a").size() shouldEqual 2
+    }
+
+    it("should add a value to a header if it already exists, regardless of casing") {
+      val wrappedResponse = new HttpServletResponseWrapper(originalResponse, ResponseMode.MUTABLE, ResponseMode.PASSTHROUGH)
+      val now = System.currentTimeMillis()
+      val later = now + 1000
+
+      wrappedResponse.addDateHeader("a", now)
+      wrappedResponse.addDateHeader("a", later)
```
jennyvo: should be "A"?

dmnjohns: I think you meant to post this on the test below, but yes, good catch!

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -63,67 +76,275 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
...
+  private var responseBodyType = ResponseBodyType.Available
+
+  override def getResponse: ServletResponse = throw new UnsupportedOperationException("getResponse is not supported")
+
+  override def setResponse(servletResponse: ServletResponse): Unit =
+    throw new UnsupportedOperationException("setResponse is not supported")
+
+  override def getHeaderNamesList: util.List[String] = getHeaderNames.toList
+
+  override def getHeaderNames: util.Collection[String] = headerMap.keys
```
adrianjgeorge: Can these be swapped and potentially save a copying of the list? I guess my question really come down to is does .keys return a list? Or will it even work with the strength of scala typing?

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -63,67 +76,275 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
...
+  override def addHeader(name: String, value: String): Unit = {
+    headerMap = headerMap + (name -> (headerMap.getOrElse(name, Seq.empty[String]) :+ value))
+
+    if (headerMode != ResponseMode.MUTABLE) {
+      // Write through to the wrapped response immediately
+      super.addHeader(name, value)
+    }
+  }
+
+  override def addHeader(name: String, value: String, quality: Double): Unit =
```
adrianjgeorge: do we want to do some range checking on this?

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -63,67 +76,275 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
...
+    addHeader(name, DateUtils.formatDate(new Date(timeSinceEpoch)))
 
   /**
    * @throws IllegalStateException when headerMode is anything other than ResponseMode.MUTABLE
    */
-  override def removeHeader(headerName: String): Unit = ???
+  override def appendHeader(name: String, value: String): Unit = {
+    ifMutable(headerMode) {
+      val existingValues = getHeaders(name)
+      existingValues.lastOption match {
```
adrianjgeorge: i think this is opposite behavior of the request which appends it to the first value i believe

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -63,67 +76,275 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
...
+  }
 
-  override def getHeadersList(headerName: String): util.List[String] = ???
+  override def getContentType: String = getHeader(CommonHttpHeader.CONTENT_TYPE.toString)
 
-  override def getPreferredHeaders(headerName: String): util.List[String] = ???
+  override def setContentType(s: String): Unit = {
+    if (!isCommitted) {
+      val contentType =
+        if (responseBodyType != ResponseBodyType.PrintWriter) s
```
adrianjgeorge: what?

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -63,67 +76,275 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
...
+    Option(getHeader(CommonHttpHeader.CONTENT_TYPE.toString))
+      .map(_.split("charset=")(1))
+      .getOrElse("ISO-8859-1")
+  }
+
+  override def setCharacterEncoding(s: String): Unit = {
+    if (!isCommitted && responseBodyType != ResponseBodyType.PrintWriter) {
+      Option(getHeader(CommonHttpHeader.CONTENT_TYPE.toString))
+        .map(_.split(';')(0))
+        .foreach(contentType => setHeader(CommonHttpHeader.CONTENT_TYPE.toString, contentType + ";charset=" + s))
```
adrianjgeorge: why foreach?

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: .keys returns an Iterable which is actually a Set. I think you would end up writing headerMap.keys.toList for the getHeaderNamesList method anyway, so I don't think making that change buys us anything.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: I guess I'm leaning toward yes now, but you tell me.

I will say that the q value according to HTTP spec only allows up to three decimal points. That limits the quality domain to 10^3 + 1 distinct values. Not range checking obviously expands that domain, but honestly, there should never be a need.

On the usage side, do we care what a user stick heres? Our code always just takes the highest value. If we do range check, do we through a compile time or runtime exception? In the former case, do I want to deal with that every time I call this code? Probably not. In the latter case, if I haven't tested my code very well I start getting exceptions at runtime. Is an invalid quality an unrecoverable exception? The HTTP spec does define a quality value as being in the range of 0 to 1, but it does not preclude the existence of a parameter named 'q' being used to convey any value in the general case.

For reference:
http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.9

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: I should probably name that method parameter.

This just sets the content type to what it should be. That is, it either strips the character encoding or leaves it alone. According to the docs on ServletRequest, the character encoding should not be set if getWriter has been called.

http://docs.oracle.com/javaee/6/api/javax/servlet/ServletResponse.html#setContentType(java.lang.String)

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: You're right.

I still find that to be odd behavior, but I'll fix this to be consistent.

The reason that appending to the first value is odd is that header value ordering does not match method call ordering. While HTTP header ordering generally doesn't matter, the cases where it does are weird. It seems more intuitive to always add values to the end rather than somewhere in the middle, as would happen if you had multiple header values, called this appendHeader method, and then combined the header values into a single, comma-separated list of values.

As an alternative, I could update the request wrapper to be consistent with this. ;)

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: What else would you use?

foreach on an Option gave me the behavior I wanted. If the header exists, replace the character encoding. If it doesn't, do nothing.

Actually, it looks like I missed setting the character encoding on the PrintWriter. What's weird is that the PrintWriter doesn't take a character encoding parameter, it just uses the default encoding. That is, unless you construct the PrintWriter with a file rather than an OutputStream. I guess I'll have to create a Writer myself and wrap that in a PrintWriter.

dmnjohns: In fact, there's some wonkiness going on here. I will fix all of the character encoding stuff.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
adrianjgeorge: Oh, of course theres some stupid arcane contractual reason for making a super weird delineation

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
adrianjgeorge: I kind of suspected.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
adrianjgeorge: We didn't do it for the request, so i guess i'm not that worried about it.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
adrianjgeorge: We had good reason for doing it the other way before, i just don't remember it right now.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
adrianjgeorge: foreach seemed odd, because you were only ever going to have one element as written.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -63,67 +76,275 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
...
    * @throws IllegalStateException when bodyMode is ResponseMode.PASSTHROUGH
    */
-  def getOutputStreamAsInputStream(): InputStream = ???
+  def getOutputStreamAsInputStream: InputStream = bodyOutputStream.getOutputStreamAsInputStream
 
   /**
    * @throws IllegalStateException when bodyMode is anything other than ResponseMode.MUTABLE
    */
-  def setOutput(inputStream: InputStream): Unit = ???
+  def setOutput(inputStream: InputStream): Unit = bodyOutputStream.setOutput(inputStream)
```
adrianjgeorge: should we have this reset the ability to get the writer and output stream?

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
adrianjgeorge: wasn't this supposed to commit the response?

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -63,67 +76,275 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
...
+      resetBuffer()
+    }
+  }
 
   /**
-   * @throws IllegalStateException when neither headerMode nor bodyMode are ResponseMode.MUTABLE
+   * @throws IllegalStateException when neither headerMode nor bodyMode is ResponseMode.MUTABLE
    */
-  def commitToResponse(): Unit = ???
+  def commitToResponse(): Unit = {
```
adrianjgeorge: I thought we were going to have this commit the underlying if they had commited this one?

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
adrianjgeorge: You seem to be counting on the underlying isCommited, but it's not going to give you accurate answers especially since you don't chain this call (which you should not do if either mode is mutable).

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: I've made the change and will push it soon.

I'd still like to know the reason again, if you happen to remember.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: That makes sense. Will do.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: So the documentation says:
"Forces any content in the buffer to be written to the client. A call to this method automatically commits the response, meaning the status code and headers will be written."

That sounds a lot like a commit. However, if we do a full commit here, then the user who wrapped the request can not longer modify it. That seems contrary to the purpose of the wrapper.

I'm not exactly sure what you mean by counting on the underlying isCommitted.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: Fixed.

repose-aggregator/commons/utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: Fixed now.

PR #1381, opened 2015-09-25
Updated the message logged/sent in the PhoneHomeService to include the creation date as well as the JVM's version and name.

wdschei:


repose-aggregator/services/phone-home/src/main/scala/org/openrepose/core/services/phonehome/PhoneHomeService.scala
```diff
@@ -127,7 +128,23 @@ class PhoneHomeService @Inject()(@Value(ReposeSpringProperties.CORE.REPOSE_VERSI
...
+        if (formattedString.endsWith("Z")) {
+          formattedString.substring(0, formattedString.length() - 1) + "+00:00"
+        } else {
+          formattedString
+        }
+      }
+
       val updateMessage = Json.stringify(Json.obj(
+        "createdAt" -> getCurrentFormattedDateTime,
+        "createdAtMillis" -> System.currentTimeMillis,
```
dmnjohns: The createdAt and createdAtMillis values could differ here. You may want to record the time once by calling System.currentTimeMillis, then pass that value to the new Date(...). This would be an exceptionally rare case, but we could be 1ms off across formats.

repose-aggregator/services/phone-home/src/main/scala/org/openrepose/core/services/phonehome/PhoneHomeService.scala
```diff
@@ -127,7 +128,23 @@ class PhoneHomeService @Inject()(@Value(ReposeSpringProperties.CORE.REPOSE_VERSI
...
+      def getCurrentFormattedDateTime(): String = {
+        // From http://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html
+        // "For formatting, if the offset value from GMT is 0, 'Z' is produced."
+        // This only manipulates the Time Zones that produce a 'Z'.
+        val formattedString = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSXXX").format(new Date())
+        if (formattedString.endsWith("Z")) {
+          formattedString.substring(0, formattedString.length() - 1) + "+00:00"
+        } else {
+          formattedString
+        }
```
dmnjohns: If this service is primarily for internal consumption, should we just set the time zone to our time zone instead? That way we don't have to deal with user time zones, and we don't have to worry about converting the Z.

dmnjohns: If we do take the set-a-time-zone approach, you can probably inline this method. As long as it doesn't look too gross.

repose-aggregator/services/phone-home/src/test/scala/org/openrepose/core/services/phonehome/PhoneHomeServiceTest.scala
```diff
@@ -33,10 +35,12 @@ import org.openrepose.commons.utils.http.{CommonHttpHeader, ServiceClientRespons
 import org.openrepose.core.services.config.ConfigurationService
 import org.openrepose.core.services.serviceclient.akka.AkkaServiceClient
 import org.openrepose.core.systemmodel._
+import org.scalatest.junit.JUnitRunner
 import org.scalatest.mock.MockitoSugar
 import org.scalatest.{FunSpec, Matchers}
 import play.api.libs.json.{JsNull, Json}
 
+@RunWith(classOf[JUnitRunner])
```
dmnjohns: Oops. I totally missed this.

repose-aggregator/services/phone-home/src/main/scala/org/openrepose/core/services/phonehome/PhoneHomeService.scala
```diff
@@ -127,7 +128,23 @@ class PhoneHomeService @Inject()(@Value(ReposeSpringProperties.CORE.REPOSE_VERSI
...
+      }
+
       val updateMessage = Json.stringify(Json.obj(
+        "createdAt" -> getCurrentFormattedDateTime,
+        "createdAtMillis" -> System.currentTimeMillis,
+        "jreVersion" -> System.getProperty("java.version", "UNKNOWN"),
+        "jvmName" -> System.getProperty("java.vm.name", "UNKNOWN"),
         "serviceId" -> originServiceId,
```
dmnjohns: Thinking about it a little bit, my personal preference would be to keep serviceId as the first JSON field. That way, when looking at these reports, we can always find who the report is about at the top. It might make parsing these things ever so slightly easier.

Not a big deal either way, just my 2c.

repose-aggregator/services/phone-home/src/test/scala/org/openrepose/core/services/phonehome/PhoneHomeServiceTest.scala
```diff
@@ -426,12 +430,17 @@ class PhoneHomeServiceTest extends FunSpec with Matchers with MockitoSugar {
...
         anyString(),
         mockitoEq(collectionUri),
         anyMapOf(classOf[String], classOf[String]),
-        mockitoEq(expectedMessage),
+        stringArgumentCaptor.capture,
         mockitoEq(MediaType.APPLICATION_JSON_TYPE))
+      val postBody = stringArgumentCaptor.getValue.toString
+      assert("""\{"createdAt":"[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}.[0-9]{3}[+-][0-9]{2}:[0-9]{2}","createdAtMillis":[0-9]{13},"jreVersion":".*","jvmName":".*","serviceId"""".r
+        .findFirstIn(postBody).isDefined)
+      assert(postBody.endsWith(expectedMessage.substring(1)))
```
dmnjohns: ArgumentCaptors make me sad. Could we instead insert the regex into the expectedMessage and use mockito.Matchers.matches(...) instead of mockitoEq?

repose-aggregator/services/phone-home/src/main/scala/org/openrepose/core/services/phonehome/PhoneHomeService.scala
wdschei: Good point. Done.

repose-aggregator/services/phone-home/src/main/scala/org/openrepose/core/services/phonehome/PhoneHomeService.scala
wdschei: I made everything UTC/GMT/Zulu since that is a guaranteed present TimeZone.

repose-aggregator/services/phone-home/src/main/scala/org/openrepose/core/services/phonehome/PhoneHomeService.scala
wdschei: Makes no difference to me since they are going into a searchable backend (e.g. Mongo). So updated.

repose-aggregator/services/phone-home/src/test/scala/org/openrepose/core/services/phonehome/PhoneHomeServiceTest.scala
wdschei: I did this, but I'm not sure it made it any cleaner. On the other hand, it didn't make it any worse. ;-)

repose-aggregator/services/phone-home/src/main/scala/org/openrepose/core/services/phonehome/PhoneHomeService.scala
dmnjohns: True, but depending on your query, you may still have a bunch of results from different services. For example, say I found a bug in the api-validator filter. Now I want to see every service which uses the filter, so I query mongo. I get back a bunch of results, but I have to look at the 5th line of each to find what service it belongs to rather than the 1st. Or, more likely, I'm looking at compact JSON so I have to actually parse things with my eyeballs.

I'm actually not certain of the format the Mongo presents data in, or how good the tools are that will query the database, but this seemed like a simple thing to do that may or may not help.

repose-aggregator/services/phone-home/src/test/scala/org/openrepose/core/services/phonehome/PhoneHomeServiceTest.scala
dmnjohns: Yeah, all of the escaping makes it a bit gross... I didn't think about that. I wish Mockito took an actual Pattern or something. I am glad that we could cut out the ArgumentCaptor, but it didn't buy us as much as I had hoped.

Also, createdAtMillis kind of prevents us from just sticking all of those fields directly into the Json.obj(...) call. [0-9]{13} is not a valid type. Ugh.

dmnjohns:
I'll pull this in once tests pass if you're happy with it @wdschei. The only part that gives me pause is the test, but I can't think of a great way to do it. You may want to roll that change back, or leave it as is, either way is fine with me.

wdschei:
As long as we don't add anything else to what we send, then it will be fine. I think I will leave that for a lesson for the reader. ;-)

PR #1386, opened 2015-10-02
Rep 2729

dkowis:
There's a shell script that goes through a few iterations of JDKs on centos to ensure that the init script doesn't pretend to start up when jdk6 is installed.

I think the real cause of the problem is the dependency stuff for the RPM isn't working right. It's somehow selecting jdk6 when it shouldn't be. The debian package building stuff seems to work just fine, however. I did add the same logic to the initscripts to debian just in case.


dkowis:
aww, I'm not an admin any more.

wdschei:
Retest this please.

wdschei:
Looks good to me. Just waiting for QE to sign off before merging.

PR #1388, opened 2015-10-03
Rep 2387

dkowis:
Update the intrafilter logging to not be naieve about the header values and properly concatenate *all* headers. Tests are written using `x-pp-user` since it's known that repose splits that header.

PR #1389, opened 2015-10-06
Modifications to the test

dkowis:
Needed to grab the lines as it was going into the echo filter, coming
out of the translation filter, not going into the translation filter. We
don't see the results of a filter unless it's getting logged going INTO
another filter.

This should be applied to #1388 to get the build passing.


dkowis:
add to whitelist

dkowis:
test this please

wdschei:
Retest this please.

PR #1390, opened 2015-10-06
REP-2529 CORS

Mario-Lopez:
I added a new filter to support CORS requests.  In summary, the filter supports the CORS workflow and specifically filtering by origin (configurable) and HTTP method (configurable by resource).

Details are here:
https://repose.atlassian.net/wiki/display/REPOSE/CORS+Filter


repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/http/CommonHttpHeader.java
```diff
@@ -41,11 +41,22 @@
...
+
+    //CORS specific
+    ORIGIN("Origin"),
+    ACCESS_CONTROL_REQUEST_METHOD("Access-Control-Request-Method"),
+    ACCESS_CONTROL_REQUEST_HEADERS("Access-Control-Request-Headers"),
+    ACCESS_CONTROL_ALLOW_ORIGIN("Access-Control-Allow-Origin"),
+    ACCESS_CONTROL_ALLOW_CREDENTIALS("Access-Control-Allow-Credentials"),
+    ACCESS_CONTROL_EXPOSE_HEADERS("Access-Control-Expose-Headers"),
+    ACCESS_CONTROL_ALLOW_METHODS("Access-Control-Allow-Methods"),
+    ACCESS_CONTROL_ALLOW_HEADERS("Access-Control-Allow-Headers");
```
dmnjohns: To follow convention, should we move these headers into their own enumeration? It would live in the same package, so I'm not sure what it buys us apart from perhaps being more clear. It seems to be what we've done for Extended and PowerApi headers, though.

repose-aggregator/components/filters/cors/src/main/resources/META-INF/schema/config/cors-configuration.xsd
```diff
@@ -0,0 +1,117 @@
...
+        </xs:sequence>
+    </xs:complexType>
+
+    <xs:complexType name="Origins">
+        <xs:annotation>
+            <xs:documentation>
+                <html:p>
+                    List of valid cross-origins that are allowed to access resources on this host.  At a minimum, the
+                    current host should be included in this list for the clients that send the Origin header for
+                    same-origin requests (e.g. Chrome and Safari).  You can optionally mark a value as being a regex
```
dmnjohns: Can/should the current host be implicit in the origins list? Is there ever a time when you would not want a same-origin request to be allowed?

repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/http/CommonHttpHeader.java
Mario-Lopez: D'oh.  I was concerned about putting these in a "common" class, and I thought I looked around to see where else I could put them.  I'll move them to a CorsHttpHeader class.

repose-aggregator/components/filters/cors/src/main/resources/META-INF/schema/examples/cors.cfg.xml
```diff
@@ -0,0 +1,63 @@
...
+        <resource path="/.*">
+            <allowed-methods>
+                <method>POST</method>
+                <method>PUT</method>
+                <method>PATCH</method>
+                <method>DELETE</method>
+            </allowed-methods>
+        </resource>
+    </resources>
+</cross-origin-resource-sharing>
```
dmnjohns: Great example config.

repose-aggregator/components/filters/cors/src/main/resources/META-INF/schema/config/cors-configuration.xsd
Mario-Lopez: If it's possible, definitely.  Is there a way to get the exact host string that would be used to access the API?  It would have to include http/https, the full host name (e.g. subdomain.rackspace.com), and the port.  I thought this might be configured somewhere, but then I worried that there would be no definite way to get the exact string that would be used.

repose-aggregator/components/filters/cors/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
```diff
@@ -0,0 +1,180 @@
...
+    val xsdUrl: URL = getClass.getResource(SCHEMA_FILE_NAME)
+    configurationService.subscribeTo(filterConfig.getFilterName, configurationFile, xsdUrl, this, classOf[CorsConfig])
+
+    logger.trace("CORS filter initialized.")
+  }
+
+  override def doFilter(servletRequest: ServletRequest, servletResponse: ServletResponse, filterChain: FilterChain): Unit = {
+    val httpServletRequest = servletRequest.asInstanceOf[HttpServletRequest]
+    val httpServletResponse = servletResponse.asInstanceOf[HttpServletResponse]
+    val isOptions = httpServletRequest.getMethod == "OPTIONS"
```
dmnjohns: If it's available, I would use the constant "javax.wx.rs.HttpMethod.OPTIONS" here. I don't remember the dependency hierarchy off the top of my head. It probably isn't worth adding a dependency for, though.

repose-aggregator/components/filters/cors/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
```diff
@@ -0,0 +1,180 @@
...
+    initialized = true
+  }
+
+  override def isInitialized: Boolean = initialized
+
+  def isOriginAllowed(requestOrigin: String): Boolean = allowedOrigins.exists(_.findFirstIn(requestOrigin).isDefined)
+
+  def getValidMethodsForResource(path: String): Iterable[String] = {
+    val resource = resources.find(_.path.findFirstIn(path).isDefined)
+    logger.debug(s"Matched path $path with configured resource: $resource")
```
dmnjohns: This log message seems a little misleading. The path didn't necessarily match a configured resource, matching was merely attempted.

repose-aggregator/components/filters/cors/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
```diff
@@ -0,0 +1,180 @@
...
+          case (true, false) => MethodNotAllowed
+        }
+    }
+
+    validationResult match {
+      case Pass =>
+        requestType match {
+          case NonCorsRequest => filterChain.doFilter(httpServletRequest, httpServletResponse)
+          case PreflightRequest =>
+            httpServletResponse.setHeader(CommonHttpHeader.ACCESS_CONTROL_ALLOW_CREDENTIALS, "true")
```
dmnjohns: It doesn't matter in this case, but I'll note that true.toString would work as a replacement to the string literal "true". Just a "huh" type of thing.

repose-aggregator/components/filters/cors/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
Mario-Lopez: Awesome.  It looks like it's available, so I updated it.

repose-aggregator/components/filters/cors/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
Mario-Lopez: I updated it to match on the output of the .find(...) and used the existing log message for success and a new message for a failure.  It looks cleaner now.  Good stuff.

repose-aggregator/components/filters/cors/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
Mario-Lopez: 10/10 info.  I updated it since I was in that code anyway.

repose-aggregator/components/filters/cors/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
```diff
@@ -0,0 +1,180 @@
...
+            httpServletResponse.setHeader(CommonHttpHeader.ACCESS_CONTROL_ALLOW_ORIGIN, origin)
+            Option(httpServletRequest.getHeader(CommonHttpHeader.ACCESS_CONTROL_REQUEST_HEADERS)).foreach {
+              httpServletResponse.setHeader(CommonHttpHeader.ACCESS_CONTROL_ALLOW_HEADERS, _)}
+            getValidMethodsForResource(httpServletRequest.getRequestURI).foreach {
+              httpServletResponse.addHeader(CommonHttpHeader.ACCESS_CONTROL_ALLOW_METHODS, _)}
```
dmnjohns: Why is the validMethods variable not being used here? I don't think there's a need to call the getValidMethodsForResource method twice while processing a single request.

Mario-Lopez: Short story:  I updated it.

Long story: I initially only had to call getValidMethodsForResource once.  Later on when I needed to call it twice, I made the validMethods variable and used it for the second usage.  However, I guess I never went back to update this line.  Thanks for catching this!

repose-aggregator/components/filters/cors/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
```diff
@@ -0,0 +1,180 @@
...
+            httpServletResponse.setStatus(HttpServletResponse.SC_OK)
+          case ActualRequest =>
+            filterChain.doFilter(httpServletRequest, httpServletResponse)
+            httpServletResponse.setHeader(CommonHttpHeader.ACCESS_CONTROL_ALLOW_CREDENTIALS, "true")
+            httpServletResponse.setHeader(CommonHttpHeader.ACCESS_CONTROL_ALLOW_ORIGIN, origin)
+
+            // clone the list of header names so we can add headers while we iterate through it
+            (List() ++ httpServletResponse.getHeaderNames.asScala).foreach {
```
dmnjohns: In place of List(), you could also use List.empty or Nil. Nil is probably a bit unclear, and is generally used for matching. As for List() vs List.empty, it just comes down to preference. AFAIK, the Scala style guide does not have a suggestion for this case. Again, just a "huh" type of thing, but worth mentioning I guess.

dmnjohns: I'm not sure I understand the reason for cloning the header names here. Why not just iterate over the header names collection and drop the "List() ++" part?

Mario-Lopez: I rather dislike what I have there, but when I just iterated over the header names, I got a concurrent modification exception on the next line when I add a new header.

repose-aggregator/components/filters/cors/src/main/resources/META-INF/schema/config/cors-configuration.xsd
dmnjohns: That's a good point. The information should be available in the system model as a destination, but putting it in a form that matches the header may prove to be a challenge. That does raise the concern, however, that a user could very easily misconfigure an origin.

repose-aggregator/components/filters/cors/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
dmnjohns: Derp, right, that makes sense. Writing to a collection as you're iterating over it is bad, mmkay.

Whelp, this makes sense then.

repose-aggregator/components/filters/cors/src/main/resources/META-INF/schema/config/cors-configuration.xsd
Mario-Lopez: Yeah.  I put a big "Security Warning" box at the top of the documentation because of this.  You can configure ".*" as the origin to get things working quickly, but you're gonna have a bad time.  I also worried that they wouldn't understand why you'd configure your own origin, so I added the note about Chrome and Safari both in the XSD and in the sample config (just in case someone who doesn't understand CORS is configuring this CORS filter).

Can you think of anything else I should pro-actively do just in case?

dmnjohns: Not really. All of your documentation is descriptive and explicit. It's just something that is always present even though, as a user, I don't really care. It would be nice to find a way around that requirement, but it's not a big deal. So I'm fine with leaving it as-is. We can always iterate on it later, if need be.

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/filters/cors/CorsFilterBasicTest.groovy
```diff
@@ -0,0 +1,333 @@
...
+        mc.receivedResponse.code == '200'
+        mc.getHandlings().size() == 1  // OPTIONS request without origin will make to it origin service
+        !mc.receivedResponse.headers.getFirstValue(CommonHttpHeader.ACCESS_CONTROL_ALLOW_ORIGIN.toString())
+        !mc.receivedResponse.headers.findAll(CommonHttpHeader.ACCESS_CONTROL_ALLOW_METHODS.toString())
+        !mc.receivedResponse.headers.getFirstValue(CommonHttpHeader.ACCESS_CONTROL_ALLOW_CREDENTIALS.toString())
+        mc.receivedResponse.headers.contains("Vary")
+        mc.receivedResponse.headers.findAll('Vary') == ['origin', 'access-control-request-headers', 'access-control-request-method']
+
+        where:
+        [method, path, origin] << [['GET', 'HEAD'], ['/', '/status'], ['http://openrepose.com:80', 'http://test.repose.site:80']].combinations()
```
dmnjohns: This is a neat way to parameterize this test. I'll have to remember it.

dmnjohns:
This looks good to me. :+1: 

I'll merge once you push up the changes that you made and the tests have another chance to run.

dmnjohns:
Looks like a few references to CommonHttpHeader were missed when you moved the headers into their own enumeration file.

Mario-Lopez:
Groovy!!!  *shakes fist*

I took a quick look in IntelliJ for red lines and didn't see any which is how I missed updating the functional tests. :-1:

PR #1393, opened 2015-10-14
Update tests

jennyvo:
add keystonev2 for some test suites using auth filter as part of test config.


jennyvo:
these'll be covered in REP-3283

PR #1394, opened 2015-10-16
REP 2880

dkowis:
Un-ignored the ignored tests, updated their acceptance criteria. Added a couple more tests for the other response codes.

Updated the mess of the mock identity service, now it's even more of a mess :( Tried to make the least mess.

Did the implementation. When 400 is received from identity, I log the response that came from identity. This could expose credentials to the logs, but it would only be what identity already is approved to expose. Just goes into our logs instead of back to the client. It's logged at the warn level, because a bad request response from identity probably isn't a good thing.

### DIFF
--- a/repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/filters/identitybasicauth/BasicAuthTest.groovy
+++ b/repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/filters/identitybasicauth/BasicAuthTest.groovy
@@ -26,8 +26,6 @@ import org.apache.commons.lang.RandomStringUtils
 import org.rackspace.deproxy.Deproxy
 import org.rackspace.deproxy.MessageChain
 import org.rackspace.deproxy.Response
-import spock.lang.Ignore
-import spock.lang.IgnoreIf
 import spock.lang.Unroll
 
 import javax.servlet.http.HttpServletResponse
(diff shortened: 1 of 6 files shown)


dkowis:
Doh, I didn't update the expectations for these to delegate the right message. Should be a trivial fix. I don't mind if someone else does it to get it done sooner.

dkowis:
Thanks for the reformat fix. My bad.

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/filters/identitybasicauth/BasicAuthTest.groovy
```diff
@@ -167,30 +164,58 @@ class BasicAuthTest extends ReposeValveTest {
...
-        then: "response with 400 bad request"
-        mc.receivedResponse.code == HttpServletResponse.SC_BAD_REQUEST.toString()
+        then: "response with 401 Unauthorized"
+        mc.receivedResponse.code == HttpServletResponse.SC_UNAUTHORIZED.toString()
         mc.handlings.size() == 0
         mc.receivedResponse.getHeaders().findAll(HttpHeaders.WWW_AUTHENTICATE).contains("Basic realm=\"RAX-KEY\"")
     }
 
-    @Ignore
+    def "When identity returns a 403 Unauthorized, repose will also return a 403 unauthorized"() {
```
wdschei: It doesn't affect the test, but it should be `403 Forbidden` instead of `403 Unauthorized`.

dkowis: Words are hard. Will fix.

wdschei: Fixed. Merging.

PR #1395, opened 2015-10-16
Rep 2886

dkowis:
Just making the `description` an `Optional[String]` works.

Of note, this wouldn't be a problem if we refactor IdentityV3 to use play-json and json pathing. We'd only extract the things we care about, and we'd be more resiliant to contract oddities.


dkowis:
Updated search message is a great solution to the test!

PR #1399, opened 2015-10-20
DO NOT MERGE til v8.0. Fix Debian package name.

wdschei:
This was a crazy syntax thing that made it into the v7.0.0.1 when we were updating the package information and normalizing the artifact names/ID's.

We probably want to wait to merge this until v8.0.0.0 since it is a breaking change for anyone running Debian WAR. Although, since the filter bundles are currently requiring the wrong package name, repose-war instead of repose-deb-war, I not sure how we didn't catch this sooner.


wdschei:
These commits were cherry picked into the `repose-8-staging` branch.

PR #1400, opened 2015-10-21
REP-2320 Create an Atom Feed Consumption Service

dmnjohns:


dmnjohns:
retest this please

wdschei:
Retest this please.

dmnjohns:
retest this please

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/actors/FeedReader.scala
```diff
@@ -0,0 +1,114 @@
...
+  def props(feedUri: String,
+            authenticatorMaker: ActorRefFactory => ActorRef,
+            notifierMaker: ActorRefFactory => ActorRef,
+            order: EntryOrderType): Props =
+    Props(new FeedReader(feedUri, authenticatorMaker, notifierMaker, order))
+}
+
+class FeedReader(feedUri: String,
+                 authenticatorMaker: ActorRefFactory => ActorRef,
+                 notifierMaker: ActorRefFactory => ActorRef,
```
dmnjohns: Can we default these functions?

repose-aggregator/services/atom-feed/impl/src/main/resources/META-INF/schema/config/atom-feed-service.xsd
```diff
@@ -0,0 +1,183 @@
...
+                      default="org.openrepose.nodeservice.atomfeed.impl.auth.NoopAuthenticatedRequestFactory">
+            <xs:annotation>
+                <xs:documentation>
+                    <html:p>Defines the fully qualified class name of a class which can authenticate requests.</html:p>
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+    </xs:complexType>
+
+    <xs:complexType name="OpenStackIdentityFQCNType">
```
wdschei: It might be nice to go ahead and split this out into it's own XSD that is imported in the examples/tests so that it more explicitly shows how this great capability can be extended.

repose-aggregator/services/atom-feed/impl/src/main/resources/META-INF/schema/examples/atom-feed-service.cfg.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.1" encoding="UTF-8"?>
+
+<atom-feed-service xmlns="http://docs.openrepose.org/repose/atom-feed-service/v1.0"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
```
wdschei: Then include the new XSD here.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/actors/FeedReader.scala
dmnjohns: Nope, never mind.

repose-aggregator/services/atom-feed/impl/src/main/resources/META-INF/schema/config/atom-feed-service.xsd
dmnjohns: Thinking about this now, it seems this feat is actually quite a bit more challenging for the user than expected. Let us posit that the user does not have access to this schema (since it is hidden away in an artifact somewhere). Then that user cannot extend a type from this schema. The parser inside of Repose will always attempt to use this schema to validate a configuration file, and the user is unable to change this schema.

So I've come up with the following for now: allow any element or attribute to be attached to the AuthenticationType. The result is that the AuthenticationType can pass any data in any XML-supported format to the code. Unfortunately, the Repose parser will not be able to validate any of the elements or attributes which are not explicitly defined in this schema. This means that data can be passed around, but not validated like the rest of the configuration.

I believe there may be a way to have the XML configuration file tell the parser what to validate and where the validation rules live, but I haven't been able to figure that part out yet. To be more specific, the user would probably define a type which extends the AuthenticationType in their own XSD which sits alongside their configuration XML. Of course, doing so would require that their XSD be aware of the AuthenticationType, which probably relies on the parser to bridge that gap. Once the XSD is written, the user's configuration file would have to declare the type on their AuthenticationType element as I've done with the OpenStack Identity type. Doing so requires that the parser know where that type lives, which is in the user-defined XSD. If the parser is smart enough to do all of that, type-safe extensibility from the user perspective may be possible.

dmnjohns: Actually, I'll use phone-a-friend.

@jorgew Is there a way to allow a user to extend a XML schema defined by Repose and validate a configuration against the user's schema? Or, more broadly, is there a way for a user to inject their own type (which extends a type defined by Repose) for the parser in Repose to use when validating a configuration file?

RackerWilliams:
@dmnjohns Yes, there are several ways of having a user extend a schema and to extend types of an existing schema.  Don't have a lot of context.  If you give me an example of what you're trying to do I'll point you in the right direction.

dmnjohns:
retest this please

repose-aggregator/core/core-lib/pom.xml
```diff
@@ -137,6 +138,18 @@
 
         <dependency>
             <groupId>org.openrepose</groupId>
+            <artifactId>atom-feed-service-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
```
adrianjgeorge: why both? the only thing i can think of is that there is a reference to something in api

dmnjohns: Well, the same thing was done for the datastore service.

My guess is that including both as dependencies gets the classes/interfaces from both into the core-lib JAR. core-lib itself probably doesn't need access to the service.

repose-aggregator/services/atom-feed/impl/pom.xml
```diff
@@ -0,0 +1,181 @@
...
+
+        <!-- Update listener and configuration service -->
+        <dependency>
+            <groupId>org.openrepose</groupId>
+            <artifactId>core-service-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <!-- Guava Optional to satisfy transitive dependency check -->
+        <dependency>
```
adrianjgeorge: why?

dmnjohns: The system model interrogator returns a Guava Optional. The transitive dependency check was complaining when this dependency was missing.

repose-aggregator/services/atom-feed/impl/src/main/resources/META-INF/schema/config/atom-feed-service.xsd
```diff
@@ -0,0 +1,189 @@
...
+
+        <xs:attribute name="polling-frequency" type="PositiveIntType" use="optional" default="5">
+            <xs:annotation>
+                <xs:documentation>
+                    <html:p>Defines how often, in seconds, the Atom Feed service checks this feed for updates.</html:p>
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+
+        <xs:attribute name="entry-order" type="EntryOrderType" use="optional" default="update">
```
adrianjgeorge: update doesn't really tell me much as a value

repose-aggregator/services/atom-feed/impl/src/main/resources/META-INF/schema/config/atom-feed-service.xsd
```diff
@@ -0,0 +1,189 @@
...
+                        on this feed.
+                    </html:p>
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+    </xs:complexType>
+
+    <xs:simpleType name="EntryOrderType">
+        <xs:restriction base="xs:string">
+            <xs:enumeration value="random">
```
adrianjgeorge: what?

repose-aggregator/services/atom-feed/impl/src/main/resources/META-INF/schema/config/atom-feed-service.xsd
```diff
@@ -0,0 +1,189 @@
...
+                <xs:annotation>
+                    <xs:documentation>
+                        <html:p>
+                            With this ordering, a feed listener will receive events with no ordering guarantees.
+                        </html:p>
+                    </xs:documentation>
+                </xs:annotation>
+            </xs:enumeration>
+
+            <xs:enumeration value="update">
```
adrianjgeorge: This seems like it would be the publish order. where is the order we talked about being the default which was the feed read order?

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
```diff
@@ -0,0 +1,228 @@
...
+      }
+    }
+  }
+
+  private def ifRunning[T](f: => T): T = {
+    if (running) f
+    else throw new IllegalStateException("Service is not running")
+  }
+
+  private def getListenersForFeed(feedId: String): Map[String, AtomFeedListener] = {
```
adrianjgeorge: this doesn't seem like it's providing a lot of value. It really only makes sense to me if you are calling it a lot

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
```diff
@@ -0,0 +1,228 @@
...
+          logger.debug("Listener not registered: " + listenerId)
+      }
+    }
+  }
+
+  override def isRunning: Boolean = running
+
+  private def startService(): Unit = {
+    synchronized {
+      if (isInitialized && isServiceEnabled && !running) {
```
adrianjgeorge: suddenly don't like cases?

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
```diff
@@ -0,0 +1,228 @@
...
+
+  private def getListenersForFeed(feedId: String): Map[String, AtomFeedListener] = {
+    feedListeners.getOrElse(feedId, Map.empty)
+  }
+
+  def isInitialized: Boolean = {
+    SystemModelConfigurationListener.isInitialized && AtomFeedServiceConfigurationListener.isInitialized
+  }
+
+  object AtomFeedServiceConfigurationListener extends UpdateListener[AtomFeedServiceConfigType] {
```
adrianjgeorge: this is broken. you've restarted the actor system, but all the current listeners are not hooked into

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
dmnjohns: I love matching, but I don't use it when it doesn't buy me much. I don't actually have a good differentiation between when I do and don't use them right now, but for boolean expressions, I generally use if/thens. They are often less verbose, and they certainly generate less byte-code than matches. There's probably an argument to be made for performance, but I imagine any performance differences are incredibly marginal.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
dmnjohns: Fffffffuuuuuuuuuuuu

repose-aggregator/services/atom-feed/impl/src/main/resources/META-INF/schema/config/atom-feed-service.xsd
dmnjohns: For whatever reason, that seems to be absent. My mistake.

So looking into this a bit today, here's some information I found in the Atom Spec (RFC 4287):
- This specification assigns no significance to the order of atom:entry elements within the feed.
- If multiple atom:entry elements with the same atom:id value appear in
   an Atom Feed Document, they represent the same entry.  Their
   atom:updated timestamps SHOULD be different.  If an Atom Feed
   Document contains multiple entries with the same atom:id, Atom
   Processors MAY choose to display all of them or some subset of them.
(From the above two points, it seems as though the order of Atom entries (defined by their id) may change over the span of multiple reads. If that is true, the id alone is not a sufficient field for the high water mark. Could/must links be used?)
- atom:entry elements MUST contain exactly one atom:updated element. 
- atom:entry elements MUST NOT contain more than one atom:published element.
(From the above two points, if we're looking at time, we can only guarantee ordering using the updated field)

repose-aggregator/services/atom-feed/impl/src/main/resources/META-INF/schema/config/atom-feed-service.xsd
dmnjohns: It seems this option is not necessary or useful. I'll remove it.

repose-aggregator/core/core-lib/pom.xml
adrianjgeorge: if i recall it was done to datastore because of classpathing and deserialization related issues

repose-aggregator/services/atom-feed/impl/pom.xml
adrianjgeorge: ugh, we should note this api for when we move to java 8, to switch to java optionals

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
adrianjgeorge: Interesting, this is a case where to me the cases would add clarity, as to what conditions are being looked for

dmnjohns: Fair enough. So would you find the following to be more clear?:

```
(isInitialized, isServiceEnabled, running) match {
    case (true, true, false) => blah
    case (_, _, true) => blah
    case (_, false, _) => blah
    case _ => blah
}
```

Looking at that, I notice the detachment between the variable name and value. Like, I have to keep looking above to see that the true/false values map to.

repose-aggregator/core/core-lib/pom.xml
dmnjohns: It's also done for the httpclient service. How would the api and implementation classes be available to use at runtime if they weren't added to some JAR? For anything that will use this service, only the api will be needed, but here we just want to make the service usable for any installation of Repose.

repose-aggregator/services/atom-feed/impl/pom.xml
dmnjohns: Agreed.

repose-aggregator/services/atom-feed/impl/src/main/resources/META-INF/schema/config/atom-feed-service.xsd
dmnjohns: Oh? Does "publish" or "read"?

This seems like a trade-off between verbosity and clarity. I'm pretty okay with "update", "publish", and "feed" orders. It's worth noting that the atom:entry fields which would be used are named "updated" and "published".

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
dmnjohns: So this is kind of a hacky thing I did. Part of letting the service itself maintain the listener mapping is providing actors a way to access that mapping. We can't pass the mapping in when we construct an actor because it changes over time. To get around that, I pass a function with a reference to the mapping that the actors can use. So this does actually get called a bunch.

Originally, I just had the service sending messages to actors whenever a change to listeners came in. That approach led to a bunch of cookie-cutter code to forward a call to registerListener into the FeedReader actor into the Notifier actor (where the mapping was finally stored/updated). Admittedly, that may be the right way to do it, but I figured I could save some overhead by passing around a function instead.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
adrianjgeorge: Yeah, i have to look at the names, but it gives me a better idea overall of what situations are being covered.

adrianjgeorge:
non mergeable and failing tests

adrianjgeorge:
looks like tests that shouldn't be live till bills story, and time bombs

repose-aggregator/services/atom-feed/api/src/main/java/org/openrepose/nodeservice/atomfeed/AuthenticatedRequestFactory.java
```diff
@@ -0,0 +1,42 @@
...
+public interface AuthenticatedRequestFactory {
+
+    /**
+     * Mutates a request to adhere to some authentication scheme compatible with the Atom service.
+     *
+     * @param atomFeedUrlConnection The unauthenticated feed {@link URLConnection} with the Atom service.
+     * @param requestId             A unique identifier associated with all of the requests for this feed reading interaction.
+     * @return The {@link URLConnection} with authentication mutations applied.
+     */
+    URLConnection authenticateRequest(URLConnection atomFeedUrlConnection, String requestId, String reposeVersion);
```
adrianjgeorge: I hate these being on this method call. They are here for the admin call though aren't they, because we could put them on the main call ourselves. Instead of a pair of strings do we want to instead pas in a context object of some sort, that happens to have these strings? Then if we come up with other things we need later we can make additions to that class instead of having to change the method signature which is potentially backwards in compatible.

repose-aggregator/services/atom-feed/api/src/main/java/org/openrepose/nodeservice/atomfeed/AuthenticatedRequestFactory.java
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Invalidates any cached authentication data.
+     */
+    void invalidateCache();
```
adrianjgeorge: I don't like this name. You are dictating usage. I'd rather it be something like authFailed let the implementer decide what there behavior is when authentication fails.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomEntryStreamBuilder.scala
```diff
@@ -0,0 +1,105 @@
...
+    * @return a Scala [[scala.collection.immutable.Stream]] which will yield Atom entries in an Atom feed located at the
+    *         provided baseFeedUrl. The entries will be yielded in the order in which they are read from the feed XML,
+    *         from top to bottom.
+    * @throws AuthenticationException when the authenticator fails to authenticate a request
+    */
+  def build(baseFeedUrl: URL,
+            reposeVersion: String,
+            authenticator: Option[AuthenticatedRequestFactory] = None,
+            authenticationTimeLimit: Duration = 1 second): Stream[Entry] = {
+    val requestId = java.util.UUID.randomUUID().toString
```
adrianjgeorge: this probably shouldn't happen here, but instead in whatever is calling this, so it's logs match up.

repose-aggregator/services/atom-feed/api/src/main/java/org/openrepose/nodeservice/atomfeed/AuthenticatedRequestFactory.java
dmnjohns: Right, there are here to generate the trace header for the token call. I'm not exactly sure what you mean by "put them on the main call ourselves". A context object might be a good idea, though. Would these parameters just become fields in the context object? That allows us to add things to the context, but removing things is still backwards incompatible. I guess another option would be having a loose mapping where you provide a key and get back either some value or null. But that sounds terrible.

repose-aggregator/services/atom-feed/api/src/main/java/org/openrepose/nodeservice/atomfeed/AuthenticatedRequestFactory.java
dmnjohns: Totally legit. I'll change it.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomEntryStreamBuilder.scala
```diff
@@ -0,0 +1,105 @@
...
+
+    val authenticatedConnection = authenticator match {
+      case Some(arf) =>
+        val connectionFuture = Future(arf.authenticateRequest(baseFeedConnection, requestId, reposeVersion))
+        Option(Await.result(connectionFuture, authenticationTimeLimit))
+      case None =>
+        Some(baseFeedConnection)
+    }
+
+    authenticatedConnection match {
```
adrianjgeorge: so i know this will complicate a ton of things, but why isn't this using either of the service clients (akka/http)?

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomEntryStreamBuilder.scala
dmnjohns: Fair enough. I think I was thinking that this is where the request begins, but there may be logging outside of here. Or maybe I just didn't want to pass the requestId through a bunch of functions. At the moment, setting it here actually will set the MDC key for all log entries since this is the first thing that is called, but that could change in the future.

I'll move it one level up into the partial function for handling ReadFeed messages. I think that's the highest it can go and still be per-request.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomEntryStreamBuilder.scala
```diff
@@ -0,0 +1,105 @@
...
+        feedInputStream.close()
+
+        feed.getLinks.find(link => link.getRel.equals("next")) match {
+          case Some(nextPageLink) =>
+            feed.getEntries.toStream #::: buildR(nextPageLink.getResolvedHref.toURL, reposeVersion, authenticator, authenticationTimeLimit, requestId)
+          case None =>
+            feed.getEntries.toStream
+        }
+      case None =>
+        throw AuthenticationException
```
adrianjgeorge: Not handling the attempt to retry auth here probably breaks things. If the auth goes bad mid stream how does that effect high water marking, and or duplicate events? I suspect (haven't looked further yet) that you end up in a bad state, where as if you handle it here the stream can continue as it's intended.

dmnjohns: This is the case where the AuthenticatedRequestFactory returns a null (to signify that something went wrong). If we wanted to retry, that could absolutely be done in the AuthenticatedRequestFactory itself.

In this case, the page failed to be read due to failed authentication. The high water mark is not updated, but events are potentially sent. When the schedule kicks again, authentication will be attempted and flow will continue as normal. So there is a risk of duplicate events...

I think an easy fix would be to set the high water mark earlier. Rather than waiting until the end of the stream, do it for each page of the stream. Since we don't have page granularity, it would actually have to be done after each individual entry. That should be fine though, we'll just set a variable a bit more often.

dmnjohns: Also, good catch.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomEntryStreamBuilder.scala
dmnjohns: There may have been a better reason, but writing this was so long ago I can only guess at this point.

My guess is that the Abdera parser takes an InputStream, Reader, or a couple of other things. I know we can get that out of our service clients, but I might not have made the connection. Additionally, the HttpConnectionPoolService returns an Apache HttpClient while the AkkaServiceClient takes a URI and headers as explicit parameters. In the former case, that requires the AtomFeedService to either force an Apache dependency on anyone writing an AuthenticatedRequestFactory or to translate between a URLConnection and the Apache class, while in the latter case, I suspect we don't have the ability to manipulate the request arbitrarily since the AkkaServiceClient uses Apache's client underneath while only providing access to two parameters for one method call.

URL connection manipulation provides the most freedom, but we lose some of the utility provided by our other service clients. I /think/ I can get the Akka service to work, but doing so might compel me to change the AuthenticatedRequestFactory interface. I'd like to talk about this more in person so I can scattershot some thoughts, ideas, and concerns.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomEntryStreamBuilder.scala
adrianjgeorge: I totally forgot it was the return from the authenticator. The situation i was thinking of and was concerned with is when we get an authorized response from an authenticated stream, how is that being handled?

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
```diff
@@ -0,0 +1,230 @@
...
+
+    val notifierManager = feedActors.get(feedId) match {
+      case Some(FeedActorPair(manager, _)) =>
+        manager
+      case None =>
+        val manager = actorSystem.actorOf(Props[NotifierManager], name = feedId + NOTIFIER_MANAGER_TAG)
+        feedActors = feedActors + (feedId -> FeedActorPair(manager, None))
+
+        if (isServiceEnabled) {
+          manager ! ServiceEnabled
```
adrianjgeorge: I'm not sure what this syntax does?

dmnjohns: This is the "send and forget" syntax for sending Akka messages.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
```diff
@@ -0,0 +1,230 @@
...
+
+    override def configurationUpdated(configurationObject: AtomFeedServiceConfigType): Unit = {
+      synchronized {
+        logger.trace("Service configuration updated")
+        initialized = true
+
+        configurationObject.getFeed foreach { feedConfig =>
+          val notifierManager = feedActors.get(feedConfig.getId) match {
+            case Some(actorPair) =>
+              // todo: only replace the current feed reader if the configuration has changed
```
adrianjgeorge: might want to look into this?

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
adrianjgeorge: From the different contexts I see, it seems to send a message.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/actors/FeedReader.scala
```diff
@@ -0,0 +1,143 @@
...
+          newEntryStream foreach { entry =>
+            val entryString = new StringWriter()
+            entry.writeTo(entryString)
+            // todo: wrap atom entry in type-safe object
+            notifierManager ! Notify(entryString.toString)
+          }
+        }
+
+        firstReadDone = true
+        highWaterMark = entryStream.headOption.map(_.getId)
```
adrianjgeorge: streams remember their values right? or are they recalculated?

repose-aggregator/services/atom-feed/api/src/main/java/org/openrepose/nodeservice/atomfeed/AuthenticatedRequestFactory.java
adrianjgeorge: Yeah, i get they are for the call to get the token. I'm not sure we'd ever remove something, or if we do we'd flag it deprecated first. Yes, they'd just become fields of the context object. And the other solution you put forward is just a map, and i agree that's too loose.

adrianjgeorge:
A couple different todo's hanging out that you might want to address.

repose-aggregator/services/atom-feed/api/src/main/java/org/openrepose/nodeservice/atomfeed/AuthenticatedRequestFactory.java
dmnjohns: Alright. Will implement.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomEntryStreamBuilder.scala
dmnjohns: Right, I started going into the case in the comment above. Basically...

If we get an unauthorized response from the Atom Feed service AFTER the first page has already been read, then we will send duplicate events. All other flows should work.

The reason is that the high water mark is only being set after all new entries have been read and sent to listeners. If we fail at any point, the high water mark is not reset. However, we send new entries as soon as possible after reading a page. So we could send some entries but never update the high water mark. That happens when a page after the first page cannot be retrieved.

So the only solution I can think of that works is buffering events until we've finished reading all of the new entries, then shooting the buffered events off to the listeners. That way, if we fail at any point, we will send no events and not update the high water mark.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
dmnjohns: So I guess that's the question. I believe everything works doing it this way, but we may perform some unnecessary processing. When I was writing this code, it was looking like this would not be a trivial thing to do. However, I think we may be able to do it in a future story?

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomEntryStreamBuilder.scala
adrianjgeorge: so we definitely don't want to buffer. It seems to me that you could rewrite the stream building method to do a single retry. Add another parameter `firstAttempt`, catch the exception coming out of abdera, tell the authenticator that auth failed , and call the method again with the flag flipped.

adrianjgeorge: In psuedo code:
```
catch AbderaException {
    if(firstAttempt) {
        autheticator.authFailed
        buildR(..., firstAttempt= false)
    } else {
        rethrow AbderaException
    }
}
```

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
adrianjgeorge: If i'm not mistaken won't this make any change to the config reset all our high water marks?

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/actors/FeedReader.scala
dmnjohns: Yes, streams are memoizing.

Proof!

```
scala> def numStream(seed: Int): Stream[Int] = {
     | println("numStream")
     | seed #:: numStream(seed + 1)
     | }
numStream: (seed: Int)Stream[Int]

scala> val ns = numStream(0)
numStream
ns: Stream[Int] = Stream(0, ?)

scala> ns.head
res2: Int = 0

scala> ns.take(5).foreach(num => println(num))
0
numStream
1
numStream
2
numStream
3
numStream
4

scala> ns.head
res4: Int = 0

scala> ns.tail.head
res5: Int = 1
```

dmnjohns:
I believe the remaining TODOs are things that we could address in the future (optimizations or new features). I left the TODOs because I either wasn't sure what we wanted, or the implementation was not trivial and I assumed that we could complete it at some point in the future. There is one FIXME remaining that I just wasn't sure about. I'll comment on it to get your thoughts. Apart from that, let me know if there is anything that we actually need to make it into this PR.

dmnjohns:
Crap, accidentally hit the close and comment button rather than just comment...

repose-aggregator/services/atom-feed/api/src/main/java/org/openrepose/nodeservice/atomfeed/AuthenticatedRequestFactory.java
dmnjohns: Done.

repose-aggregator/services/atom-feed/api/src/main/java/org/openrepose/nodeservice/atomfeed/AuthenticatedRequestFactory.java
dmnjohns: Done.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomEntryStreamBuilder.scala
dmnjohns: Done.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomEntryStreamBuilder.scala
dmnjohns: Will leave for a future story.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
dmnjohns: Done.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/actors/FeedReader.scala
```diff
@@ -0,0 +1,149 @@
...
+      } catch {
+        // fixme: Is there any case where this feed reader actor should be killed? Or restarted after some time?
```
dmnjohns: This is the fixme I referred to in a previous comment. Currently the feed reader actor will just keep on chugging if an exception occurs. One example of where we might not want this is when user credentials are bad, and the authentication component keeps failing (throwing an exception). Every x seconds, this actor will log the error, but it won't turn itself off or anything.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
adrianjgeorge: Comment in wrong spot?

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/actors/FeedReader.scala
adrianjgeorge: I think i prefer it to keep going, because the issue might not be our side. if we poison our self when there is a temporary outage on feeds everyone has to do a repose restart. Just let it keep trying and error messaging.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
dmnjohns: Nope, I was just trying to mark that there was left to do for this comment stream.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
dmnjohns: Oh, fair point. I want to say I thought about that, because it seems pretty obvious... let me look at the code again.

Ah, okay. So I think this was my thinking... If we send the PoisonPill message, then the current FeedReader will finish any ReadFeed messages that came in prior to the PoisonPill. If that's the case, then the old reader should read all of the entries in the Feed, and when the new FeedReader starts, it will mark the most recent entry as the high water mark and start reading from there.

I guess there should be some concern over the sychronization... that's hard since we don't know exactly when the old FeedReader will stop. The old FeedReader still reading could lead to duplicate events, or even missed events if the old FeedReader dies immediately and a new entry is put on the feed before the new FeedReader reads for the first time.

If we hard stop the old FeedReader rather than using a PoisonPill, we know exactly when it stops, but not on which entry it stops.

So I guess the solution is to only replace a FeedReader if its configuration has changed. That solves the problem with replacing the FeedReader. How would we check the configuration delta? We have to store the old configuration in some form. We could store the configuration for a FeedReader within the FeedReader itself, but then we have to communicate with the ActorSystem from outside of it, which isn't very clean. Actually, I think the easiest way would be to store the entire old configuration object (or, at least, the portion pertaining to feeds). Then just check every field against itself in the new feed configuration. Does that sound reasonable?

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomEntryStreamBuilder.scala
dmnjohns: Oh, right, we could absolutely do a single retry. I think that's a different issue, though? Take the following example:

The FeedReader is kicked by the Akka scheduler. It reads the first page of a feed no problem, so all of those entries are send to associated listeners. Now it goes to read the second page, but the network is terribly noisy, so the request times out. Exception. Well the highWaterMark hasn't been updated because an Exception occurred, but we *did* send out some new entries. Of course, we can't set the high water mark to any of the entries we sent, because then we potentially be skipping older entries. If we don't set the high water mark, though, we are at risk of sending duplicate entries the next time the FeedReader attempts to read the feed. So what do we do?

If we were reading from the bottom of the feed to the top, we could simply set the high water mark after each entry we read. That doesn't work if we read from top to bottom (the natural order), though. The problem is, we don't know if a page will fail until we've read that page. The only solution I can think of, then, is to read all of the pages, buffer them, then send out the entries on a success. That slows the read top-to-bottom case down to the speed of the read bottom-to-top case, though. Both cases become read everything, then send entries (in whatever order).

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
adrianjgeorge: Yeah, my main concern was dropped entries considering our primary use case involves auth. As for accomplishing deltas. From what i recall of the configurable fields the only time we'd want a high water mark reset is on a url change for a feed or maybe auth change? A timing change would ideally maintain high water mark. I think we can live without that for now though, because it's the more complicated use case. You have the old config and the new config during update right? I think jaxb gives us a reasonable equals method, so you should be able to get the differences between the two lists both ways, and take appropriate action. Meaning old - new = deletions, new-old = additions. Am I making sense?

dmnjohns: Kind of? I think?

Right now we don't keep track of the old config, so I don't have both when the config changes, but that's easy enough to change.

If we're only looking at a couple of fields, I think we can maintain the high water mark on a timing change.

Let me try to implement this and you can tell me what you think. It might be much easier than it sounds.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomEntryStreamBuilder.scala
adrianjgeorge: yeah, that one is harder to solve, because we don't want to buffer needlessly. I'd say we just document it for now as possible duplication if we have catastrophic failure while reading in this direction.

repose-aggregator/services/atom-feed/impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
adrianjgeorge: Don't worry about the timing changes for now, we can do that as a future enhancement. Just do a drop and re add on any change.

dmnjohns: Will do.

dmnjohns:
After

PR #1402, opened 2015-10-23
REP-2793 IP User filter

dkowis:
IP Classification filter. Woo!

There's a couple TODOs regarding questions about functionality. PLZKTHX


wdschei:
Retest this please.

dkowis:
The test on this will still fail. The written tests don't follow the instructions in the sample code. The catch-all match is first, and is going to always match.

wdschei:
Cool. I was just trying to get context for the Monday morning. Figuring out what could be code reviewed and what I could jump in and help with.

jennyvo:
the test is written that way, it expected to hit all-match first.  I guess the repose didn't update reconfig quick enough.

repose-aggregator/components/filters/ip-classification/src/main/resources/META-INF/schema/examples/ip-classification.cfg.xml
```diff
@@ -0,0 +1,44 @@
...
+    <classifications>
+        <!-- Classifications follow the CIDR rules: https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing
+        You can specify ipv4-cidr and/or ipv6-cidr. If either matches, the label is applied to the specified header
+        and matching stops. If neither matches, nothing happens.
+        labels must be unique.
+        You may specify multiple cidr notations space separated in either IPv4 or IPv6.
+        -->
+        <classification label="sample-group" ipv4-cidr="192.168.1.0/24 192.168.0.1/32"/>
+        <classification label="sample-ipv6-group" ipv6-cidr="2001:db8::/48"/>
+        <classification label="bolth-group" ipv4-cidr="10.10.220.0/24" ipv6-cidr="2001:1938:80:bc::1/64"/>
```
dmnjohns: bolth-group? That may be a typo, heh.

repose-aggregator/components/filters/ip-classification/src/main/scala/org/openrepose/filters/ipclassification/IPClassificationFilter.scala
```diff
@@ -0,0 +1,153 @@
...
+import org.openrepose.commons.utils.servlet.http.ReadableHttpServletResponse
+import org.openrepose.core.filter.FilterConfigHelper
+import org.openrepose.core.filter.logic.{FilterAction, FilterDirector}
+import org.openrepose.core.filter.logic.common.AbstractFilterLogicHandler
+import org.openrepose.core.filter.logic.impl.{FilterDirectorImpl, FilterLogicHandlerDelegate}
+import org.openrepose.core.services.config.ConfigurationService
+import org.openrepose.filters.ipclassification.config.IpClassificationConfig
+
+@Named
+class IPClassificationFilter @Inject()(configurationService: ConfigurationService) extends AbstractFilterLogicHandler
```
dmnjohns: Ew, AbstractFilterLogicHandler.

Can we cut that pattern out of this filter? We seem to be moving away from it in favor of self-contained filters which define their own flow.

repose-aggregator/components/filters/ip-classification/src/main/scala/org/openrepose/filters/ipclassification/IPClassificationFilter.scala
```diff
@@ -0,0 +1,153 @@
...
+  def getClassificationLabel(ipAddress: String): Option[String] = {
+    import scala.collection.JavaConversions._
+    cidrList.toList.find { case (_, cidrUtil) =>
+      cidrUtil.isInRange(ipAddress)
+    } map { case (label, _) =>
+      label
+    }
+  }
+
+  override def configurationUpdated(config: IpClassificationConfig): Unit = {
```
dmnjohns: This parameter overrides the global "config" var. One of those names should probably be changed.

repose-aggregator/components/filters/ip-classification/src/main/resources/META-INF/schema/examples/ip-classification.cfg.xml
dkowis: It's just a label :P It could be anything. And "both" sometimes has "L"s in it. I'm trying to be inclusive.

repose-aggregator/components/filters/ip-classification/src/main/scala/org/openrepose/filters/ipclassification/IPClassificationFilter.scala
```diff
@@ -0,0 +1,153 @@
...
+class IPClassificationFilter @Inject()(configurationService: ConfigurationService) extends AbstractFilterLogicHandler
+with Filter
+with LazyLogging
+with UpdateListener[IpClassificationConfig] {
+
+  private final val DEFAULT_CONFIG = "ip-classification.cfg.xml"
+  private var initialized = false
+  private var config: String = _
+
+  type CIDRTuple = (String, CIDRUtils)
```
dmnjohns: Would this be clearer as a case class? I personally hate working with tuples, even if they are type aliased. Plus a case class gives names to parameters. This is a simple type, but even so...

repose-aggregator/components/filters/ip-classification/src/main/scala/org/openrepose/filters/ipclassification/IPClassificationFilter.scala
dkowis: Oh. I lifted this from the basic auth filter, and partially keystone v3. Which is a better one to use as a template?

repose-aggregator/components/filters/ip-classification/src/main/scala/org/openrepose/filters/ipclassification/IPClassificationFilter.scala
```diff
@@ -0,0 +1,153 @@
...
+      xsdURL,
+      this,
+      classOf[IpClassificationConfig]
+    )
+  }
+
+  override def destroy(): Unit = configurationService.unsubscribeFrom(config, this.asInstanceOf[UpdateListener[_]])
+
+  override def doFilter(servletRequest: ServletRequest, servletResponse: ServletResponse, filterChain: FilterChain): Unit = {
+    new FilterLogicHandlerDelegate(servletRequest, servletResponse, filterChain).doFilter(this)
```
dmnjohns: Again, this pattern :cry:

repose-aggregator/components/filters/ip-classification/src/main/scala/org/openrepose/filters/ipclassification/IPClassificationFilter.scala
```diff
@@ -0,0 +1,153 @@
...
+  override def handleRequest(httpServletRequest: HttpServletRequest, httpServletResponse: ReadableHttpServletResponse): FilterDirector = {
+    logger.debug("Handling HTTP request!")
+    val filterDirector = new FilterDirectorImpl()
+
+    getClassificationLabel(httpServletRequest.getRemoteAddr).foreach { label =>
+      filterDirector.requestHeaderManager.appendHeader(groupHeaderName, label, groupHeaderQuality)
+    }
+
+    //Always set the user header name to the current IP address
+    //TODO: apparently we append users as well?
```
dmnjohns: Seems a little odd to have multiple users on one request, but I guess we take the constructive approach over the destructive one in general. It's easy for a user to get the behavior they want if we append. If they want us to append, done. If they don't, just add the header-normalization filter to strip the x-pp-user header before it reaches this filter. If we set, though, the previous values for that header are gone. You could do something hacky, like move them for this filter then move them back, but we shouldn't make the a requirement. After all, who knows what people will use this filter for? It could conceivably be paired with the keystone filter for fine-grained rate limiting.

Which is all to say, appending seems right.

repose-aggregator/components/filters/ip-classification/src/main/scala/org/openrepose/filters/ipclassification/IPClassificationFilter.scala
dkowis: Yeah that's my bad.

repose-aggregator/components/filters/ip-classification/src/main/scala/org/openrepose/filters/ipclassification/IPClassificationFilter.scala
dkowis: Yeah, it probably is clearer as a case class, I can do that. It is certainly more clear as a case class proper.

repose-aggregator/components/filters/ip-classification/src/main/scala/org/openrepose/filters/ipclassification/IPClassificationFilter.scala
dkowis: The old ip-identity filter did appending, because there's a quality associated with the user field. I suppose multiple filters could assign "users" based on whatever. It didn't make a lot of sense for me either, but I wasn't sure that I wanted to re-define that behavior.

repose-aggregator/components/filters/ip-classification/src/main/scala/org/openrepose/filters/ipclassification/IPClassificationFilter.scala
```diff
@@ -0,0 +1,153 @@
...
+          List.empty[String]
+        }
+      }
+
+      splitCIDR(classification.getIpv4Cidr).map { cidr =>
+        items.add((label, new CIDRUtils(cidr)))
+      }
+      splitCIDR(classification.getIpv6Cidr).map { cidr =>
+        items.add((label, new CIDRUtils(cidr)))
+      }
```
dmnjohns: This seems redundant. Why not just take the set union of the CIDRs and map once? Something something DRY something something.

repose-aggregator/components/filters/ip-classification/src/main/scala/org/openrepose/filters/ipclassification/IPClassificationFilter.scala
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    //We have loaded all the config, clear it and add all the new ones
+    cidrList.clear()
+    cidrList.addAll(items)
```
dmnjohns: I guess this is why we're using the ConcurrentLinkedQueue. So since multiple threads can be in this method at once, are these two lines guaranteed to be correct? It seems like maybe not. The more recent update could .addAll, after which the previous update (that is still processing) could .clear giving us consistent, but old, configuration.

That's assuming addAll is atomic. I don't actually know the contract for the ConcurrentLinkedQueue. If not, though, multiple threads could call addAll and jumble the list with the union of those items.

repose-aggregator/components/filters/ip-classification/src/main/resources/META-INF/schema/config/ip-classification.xsd
```diff
@@ -0,0 +1,120 @@
...
+
+        <xs:all>
+            <xs:element name="user-header-name" type="HeaderType" minOccurs="0" maxOccurs="1" default="x-pp-user"/>
+            <xs:element name="group-header-name" type="HeaderType" minOccurs="0" maxOccurs="1" default="x-pp-groups"/>
+            <xs:element name="classifications" type="ClassificationsType" minOccurs="1" maxOccurs="1"/>
+        </xs:all>
+    </xs:complexType>
+
+    <!-- because this becaomse a complex type now, I cannot use defaults. Have to put them in the scala :( -->
+    <!-- well, the quality could be optional, but not the name, and if the name isn't there, the quality isn't there either -->
```
dmnjohns: I suppose you could make the 4 configuration items into their own attributes instead... a little messy, because you lose the well-defined type, but it would give us the same information and allow us to keep the defaults in the XSD.

Or you could make the name an attribute on the complex type and default it in child types. That would require that you extend the type so that a type exists for both the user header name and group header name. Also gross.

It's super annoying that defaults for elements don't work.

repose-aggregator/components/filters/ip-classification/src/main/scala/edazdarevic/commons/net/CIDRUtils.java
```diff
@@ -0,0 +1,142 @@
+/*
```
dmnjohns: ...

I'm just going to assume this is correct.

repose-aggregator/components/filters/ip-classification/src/main/scala/org/openrepose/filters/ipclassification/IPClassificationFilter.scala
dkowis: Because we specify both ipv4 and ipv6. As well there's a TODO listed about how an IPv6 catchall will match an IPv4 address. So, if that needs to be changed, I left the logic paths separate. Otherwise, we could probably do some fancyness to smush them together, but they're originally treated as two separate entities.

repose-aggregator/components/filters/ip-classification/src/main/scala/org/openrepose/filters/ipclassification/IPClassificationFilter.scala
dkowis: This is true. The other way to do this would be to use a read-only list, and put it in an [AtomicReference](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicReference.html). This exists in java 7 as well. That's probably the right way to do it. Then it's an atomic swap out.

repose-aggregator/components/filters/ip-classification/src/main/resources/META-INF/schema/config/ip-classification.xsd
dkowis: Yeah. I've covered it in code, but it's annoying. I prefer to go on the side of clearer configuration, and we'll deal with the complexity in the code, because we're putting this in front of users.

What do you think is best for the user?

repose-aggregator/components/filters/ip-classification/src/main/scala/edazdarevic/commons/net/CIDRUtils.java
dkowis: Well, I have some unit tests that verify that some addresses match :D, and I found this guy like 3 times on stack overflow

repose-aggregator/components/filters/ip-classification/src/main/scala/org/openrepose/filters/ipclassification/IPClassificationFilter.scala
dkowis: Honestly, I think I wrote maybe one filter during my time on the team. Most of my time was spent punching `core` in the face.

repose-aggregator/components/filters/ip-classification/src/main/scala/org/openrepose/filters/ipclassification/IPClassificationFilter.scala
dmnjohns: Keystone v2v2 =P

Or HERP. Or Valkyrie.

repose-aggregator/components/filters/ip-classification/src/main/scala/org/openrepose/filters/ipclassification/IPClassificationFilter.scala
dmnjohns: It was a much needed beating. :+1:

repose-aggregator/components/filters/ip-classification/src/main/scala/org/openrepose/filters/ipclassification/IPClassificationFilter.scala
dmnjohns: Fair enough.

repose-aggregator/components/filters/ip-classification/src/main/scala/org/openrepose/filters/ipclassification/IPClassificationFilter.scala
dmnjohns: Legit.

repose-aggregator/components/filters/ip-classification/src/main/resources/META-INF/schema/config/ip-classification.xsd
dmnjohns: I think either what you have now or the type hierarchy would be be best for the user. The type hierarchy approach keeps the defaults in the XSD, which is nice, but it makes the XSD types a bit more complex. Your approach keeps the XSD simple, but moves the defaults into Scala code.

Best solution: get JAXB to not be stupid about default values for elements.

dmnjohns:
retest this please

dmnjohns:
Looks good to me once tests pass. The only thing I'm not sure about is having IPv6 addresses match IPv4 addresses... but it's not a big enough concern that I wouldn't merge this.

repose-aggregator/components/filters/filter-bundle/pom.xml
```diff
@@ -204,6 +204,11 @@
             <artifactId>cors</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.openrepose</groupId>
+            <artifactId>ip-classification</artifactId>
```
adrianjgeorge: we had a different name paradigm we were going to start using for these filters

repose-aggregator/components/filters/ip-classification/src/main/scala/edazdarevic/commons/net/CIDRUtils.java
adrianjgeorge: you have a java class in the scala source folder, also why did we copy pasta this?

repose-aggregator/components/filters/filter-bundle/pom.xml
dkowis: I asked about it when building it, and nobody said anything. Dunno what it is.

repose-aggregator/components/filters/ip-classification/src/main/resources/META-INF/schema/examples/ip-classification.cfg.xml
```diff
@@ -0,0 +1,44 @@
...
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+  =_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_=_
+  -->
+
+<ip-classification xmlns="http://docs.openrepose.org/repose/ip-classification/v1.0">
```
adrianjgeorge: I super don't like this configuration. As soon as you have more than a couple IPs it becomes completely illegible. Also it appears that v4 and v6 gets treated the same so why have a distinction?
How about something more like:
```xml
<ip-classification>
    <group name="butts">
        <ip>some ipv4 ip</ip>
        <ip>some other ipv4 ip</ip>
        <ip>some ipv6 ip</ip>
    </group>
</ip-classifciation>
```

repose-aggregator/components/filters/ip-classification/src/main/resources/META-INF/schema/config/ip-classification.xsd
```diff
@@ -0,0 +1,120 @@
...
+            <xs:extension base="xs:string">
+                <xs:annotation>
+                    <xs:documentation>
+                        <html:p>The header name to put the classification label in</html:p>
+                    </xs:documentation>
+                </xs:annotation>
+                <xs:attribute name="quality" type="doubleBetweenZeroandOne" use="optional" default="0.4">
+                    <xs:annotation>
+                        <xs:documentation>
+                            <html:p>The IPv6 CIDR notation to match against.</html:p>
```
adrianjgeorge: pretty sure this is wrong

repose-aggregator/components/filters/ip-classification/src/main/resources/META-INF/schema/examples/ip-classification.cfg.xml
```diff
@@ -0,0 +1,44 @@
...
+    <!--
+     specify what header to put the group name in. This is the default. It will always add an additional header, not
+     clobbering the existing value.
+    -->
+    <group-header-name quality="0.4"/>
```
adrianjgeorge: given usage this is a terrible name. also i think this just changed it from the default to blank.
just go with group header and have the name be an attribute of the element. do the same for the other header.

repose-aggregator/components/filters/ip-classification/src/main/resources/META-INF/schema/examples/ip-classification.cfg.xml
dkowis: The ipv4 vs ipv6 thing was always an open question for me. If that's okay behavior, I don't have any problems with thiis. It feels a bit more verbose, as I don't expect people to have more than a couple IP addresses, especially given that they can assign entire subnets with CIDR notation. But maybe I misjudged the use case.

repose-aggregator/components/filters/ip-classification/src/main/resources/META-INF/schema/examples/ip-classification.cfg.xml
dkowis: This wouldn't have changed it from the default to blank, because of how JAXB works. Using attributes does make the defaults more sane in the JAXB world. No objections here.

repose-aggregator/components/filters/ip-classification/src/main/scala/edazdarevic/commons/net/CIDRUtils.java
dkowis: It doesn't matter if there's a java source in the scala source folder, scala can compile it. I didn't feel like separating the things since it was the only one. It's copy pasta, because I have no reason to re-type it? Where is it copy-pasta'd from? It handles all the use cases for CIDR notations, so there's no reason to re-implement it.

repose-aggregator/components/filters/ip-classification/src/main/resources/META-INF/schema/examples/ip-classification.cfg.xml
adrianjgeorge: i've seen configs with a dozen or more

dkowis: Because they couldn't use CIDR? That is new in this filter. (Granted, at this point I'm just asking, I'm not arguing against doing the `<ip>` style.)

repose-aggregator/components/filters/ip-classification/src/main/scala/edazdarevic/commons/net/CIDRUtils.java
adrianjgeorge: yeah, it needs to move to java sources. let me change the question. why was it copy pastad instead of a dependency brought in?

repose-aggregator/components/filters/ip-classification/src/main/resources/META-INF/schema/examples/ip-classification.cfg.xml
adrianjgeorge: the old one supported that or at least our investigation suggested it did, they just had enough different ip's that doing so wasn't pracitical.

repose-aggregator/components/filters/ip-classification/src/main/scala/edazdarevic/commons/net/CIDRUtils.java
dkowis: Because there isn't one.

dmnjohns:
The old ip-identity filter should be deprecated, and it should be noted that it will be removed in Repose 8.

wdschei:
Retest this please.

repose-aggregator/components/filters/filter-bundle/pom.xml
adrianjgeorge: I looked it up it was whatever-user filter. The only one currently in that paradigm is rackspace-auth-user for identity.

wdschei: So do we want to change this one?
If so, then to what?
Just append user to it `ip-classification-user`?
Or do we leave it alone?

repose-aggregator/components/filters/ip-classification/src/main/resources/META-INF/schema/config/ip-classification.xsd
```diff
@@ -0,0 +1,128 @@
...
+    <xs:complexType name="HeaderType">
+        <!-- quality attribute follows the default from ip-identity -->
+        <xs:simpleContent>
+            <xs:extension base="xs:string">
+                <xs:annotation>
+                    <xs:documentation>
+                        <html:p>The header name to put the classification label in</html:p>
+                    </xs:documentation>
+                </xs:annotation>
+                <xs:attribute name="quality" type="doubleBetweenZeroandOne" use="optional" default="0.4">
```
adrianjgeorge: name should be here as well, just not defaulted

repose-aggregator/components/filters/ip-classification/src/main/resources/META-INF/schema/config/ip-classification.xsd
```diff
@@ -0,0 +1,128 @@
...
+            <xs:element name="group" type="GroupType" minOccurs="1" maxOccurs="unbounded"/>
+        </xs:sequence>
+    </xs:complexType>
+
+    <!-- because this became a complex type now, I cannot use defaults. Have to put them in the scala :( -->
+    <!-- well, the quality could be optional, but not the name, and if the name isn't there, the quality isn't there either -->
+    <xs:complexType name="HeaderType">
+        <!-- quality attribute follows the default from ip-identity -->
+        <xs:simpleContent>
+            <xs:extension base="xs:string">
```
adrianjgeorge: probably should extend string, because we are not expecting content in the element.

repose-aggregator/components/filters/ip-classification/src/main/scala/edazdarevic/commons/net/CIDRUtils.java
adrianjgeorge: still needs to move to java src

repose-aggregator/components/filters/ip-classification/src/main/scala/org/openrepose/filters/ipclassification/IPClassificationFilter.scala
```diff
@@ -0,0 +1,142 @@
...
+      val label = group.getName
+      group.getCidrIp.map { cidr =>
+        LabeledCIDR(label, new CIDRUtils(cidr))
+      }
+    }
+
+    //Replace our object! Atomically
+    cidrList.set(replacementCidrList)
+
+    //Blergh, no useful defaults in XSD when I add complex types :(
```
adrianjgeorge: i would think this wouldn't be necessary now that they all have defaults in the xsd.

repose-aggregator/components/filters/ip-classification/src/test/scala/org/openrepose/filters/ipclassification/ConfigValidationTest.scala
```diff
@@ -0,0 +1,102 @@
...
+      resultant shouldNot be(null)
+      resultant.getClass should equal(classOf[IpClassificationConfig])
+
+      //Verify the group settings and such are not the defaults
+      resultant.getUserHeader.getName should equal("user-header")
+      resultant.getUserHeader.getQuality should equal(0.5D)
+      resultant.getGroupHeader.getName should equal("group-header")
+      resultant.getGroupHeader.getQuality should equal(0.5D)
+    }
+    //TODO: IF we want default true values, THEN we have to re-work where the defaults are specified; because XSD
```
adrianjgeorge: i think these todo's are ready to be reexamined.

repose-aggregator/components/filters/ip-classification/src/main/scala/org/openrepose/filters/ipclassification/IPClassificationFilter.scala
wdschei: If the element is not present, then it doesn't provide defaults; it returns `null`.

repose-aggregator/components/filters/ip-classification/src/test/scala/org/openrepose/filters/ipclassification/ConfigValidationTest.scala
wdschei: These are referring to the way XML handles (or doesn't) if an element is not present. That is what the other monkey business is about.

repose-aggregator/components/filters/ip-classification/src/main/resources/META-INF/schema/config/ip-classification.xsd
wdschei: Updated.

repose-aggregator/components/filters/ip-classification/src/main/resources/META-INF/schema/config/ip-classification.xsd
wdschei: That makes sense. Fixed.

repose-aggregator/components/filters/filter-bundle/pom.xml
adrianjgeorge: that is the question. i like the idea of these types of filters having a common name, but i'm not 100% sold on -user. I understand why we went with it, but i'm open to better names.

repose-aggregator/components/filters/ip-classification/src/test/scala/org/openrepose/filters/ipclassification/ConfigValidationTest.scala
adrianjgeorge: ohh, i missed that they wouldn't be there. legit

repose-aggregator/components/filters/ip-classification/src/main/resources/META-INF/schema/config/ip-classification.xsd
wdschei: This was updated and then had to be backed out because it didn't work. After discussing this with @jorgew, the decision was to not use either the deprecated `redefine` or the newer `override` capabilites of XML Schemas due to their complexities involved, but to instead leave the `name` attribute with a defined default in both of the extending types.

wdschei:
This was rebased and force pushed after all of the updates based on the code review feedback.

adrianjgeorge:
Ugh, now it's not mergeable for some reason

wdschei:
I'll take a look.

wdschei:
There was a conflict with the FilterBundle's pom.xml and web-fragment.xml since there were two filters added at the same time. Rebased, fIxed, and force pushed.

repose-aggregator/components/filters/ip-user/pom.xml
```diff
@@ -0,0 +1,123 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.openrepose</groupId>
+        <artifactId>filters-support</artifactId>
+        <version>7.2.2.0-SNAPSHOT</version>
```
adrianjgeorge: this needs an update now

wdschei: Yep.

PR #1404, opened 2015-10-23
REP-2941 Use the Java environment variable rather than checking the path

dmnjohns:
Manually tested, and this seems to work just fine.

### DIFF
--- a/repose-aggregator/installation/configs/etc/init.d/deb/repose-valve
+++ b/repose-aggregator/installation/configs/etc/init.d/deb/repose-valve
@@ -34,12 +34,12 @@ CLEAN=/usr/bin/clean-repose-deploy
 . /etc/sysconfig/repose
 
 # First validate that we have a java executable on the path
-command -v java >/dev/null 2>&1 || { echo "Repose requires a java executable on the path! Exiting!"; exit 1; }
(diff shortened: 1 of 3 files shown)


dmnjohns:
The comments need to be updated.

dkowis:
My bad, I thought we didn't want to use environment vars :|

dmnjohns:
How come? It seems reasonable since that's what we're using to launch Repose, but maybe there's something I don't know here.

dkowis:
No this is fine. I didn't even know we had an `/etc/[sysconfig|default]/repose` that we were using to set the java location explicitly.

dmnjohns:
retest this please

dmnjohns:
retest this please

adrianjgeorge:
retest this please

dmnjohns:
So there was a comment about using JAVA_HOME, although that would be a change potentially requiring users to update their environment after upgrading Repose.

Otherwise, is this ready to be pulled in?

adrianjgeorge:
It mostly ready, i'm just waiting on confirmation from @jennyvo that she's manually tested it.

jennyvo:
QE manually tested.

PR #1405, opened 2015-10-26
AkkaServiceClient - Connection pool ID

Mario-Lopez:
Added ability to set a connection pool ID in the AkkaServiceClient.

If none is specified, the old behavior is used, so this update is backwards compatible.

### DIFF
--- a/repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
+++ b/repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
@@ -37,7 +37,7 @@ import org.openrepose.commons.utils.servlet.http.MutableHttpServletRequest
 import org.openrepose.core.filter.FilterConfigHelper
 import org.openrepose.core.services.config.ConfigurationService
 import org.openrepose.core.services.datastore.{Datastore, DatastoreService}
-import org.openrepose.core.services.serviceclient.akka.{AkkaServiceClient, AkkaServiceClientException}
+import org.openrepose.core.services.serviceclient.akka.{AkkaServiceClientFactory, AkkaServiceClient, AkkaServiceClientException}
 import org.openrepose.core.systemmodel.SystemModel
 import org.openrepose.filters.keystonev2.KeystoneRequestHandler._
 import org.openrepose.filters.keystonev2.config._
@@ -48,7 +48,7 @@ import scala.util.{Failure, Random, Success, Try}
 
 @Named
 class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
-                                 akkaServiceClient: AkkaServiceClient,
+                                 akkaServiceClientFactory: AkkaServiceClientFactory,
                                  datastoreService: DatastoreService)
   extends Filter
   with HttpDelegationManager
@@ -62,6 +62,7 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
   private val datastore: Datastore = datastoreService.getDefaultDatastore //Which happens to be the local datastore
 
   var keystoneV2Config: KeystoneV2Config = _
+  var akkaServiceClient: AkkaServiceClient = _
 
   override def init(filterConfig: FilterConfig): Unit = {
     configurationFile = new FilterConfigHelper(filterConfig).getFilterConfig(DEFAULT_CONFIG)
@@ -83,6 +84,7 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
   }
 
   override def destroy(): Unit = {
+    Option(akkaServiceClient).foreach(_.destroy())
     configurationService.unsubscribeFrom(configurationFile, KeystoneV2ConfigListener)
     configurationService.unsubscribeFrom(SYSTEM_MODEL_CONFIG, SystemModelConfigListener)
   }
@@ -626,6 +628,10 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
       // Removes an extra slash at the end of the URI if applicable
       val serviceUri = keystoneV2Config.getIdentityService.getUri
       keystoneV2Config.getIdentityService.setUri(serviceUri.stripSuffix("/"))
+
+      val akkaServiceClientOld = Option(akkaServiceClient)
+      akkaServiceClient = akkaServiceClientFactory.newAkkaServiceClient(keystoneV2Config.getIdentityService.getConnectionPoolId)
+      akkaServiceClientOld.foreach(_.destroy())
 
       initialized = true
     }
--- a/repose-aggregator/components/filters/openstack-identity-v3/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3HandlerFactory.scala
+++ b/repose-aggregator/components/filters/openstack-identity-v3/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3HandlerFactory.scala
@@ -24,14 +24,15 @@ import java.util
 import org.openrepose.commons.config.manager.UpdateListener
 import org.openrepose.core.filter.logic.AbstractConfiguredFilterHandlerFactory
 import org.openrepose.core.services.datastore.DatastoreService
-import org.openrepose.core.services.serviceclient.akka.AkkaServiceClient
+import org.openrepose.core.services.serviceclient.akka.{AkkaServiceClientFactory, AkkaServiceClient}
 import org.openrepose.filters.openstackidentityv3.config.OpenstackIdentityV3Config
 import org.openrepose.filters.openstackidentityv3.utilities.OpenStackIdentityV3API
 
-class OpenStackIdentityV3HandlerFactory(akkaServiceClient: AkkaServiceClient, datastoreService: DatastoreService)
+class OpenStackIdentityV3HandlerFactory(akkaServiceClientFactory: AkkaServiceClientFactory, datastoreService: DatastoreService)
   extends AbstractConfiguredFilterHandlerFactory[OpenStackIdentityV3Handler] {
 
   private var keystoneHandler: OpenStackIdentityV3Handler = _
+  private var akkaServiceClient: AkkaServiceClient = _
 
   override def buildHandler: OpenStackIdentityV3Handler = {
     if (isInitialized) keystoneHandler
@@ -46,10 +47,18 @@ class OpenStackIdentityV3HandlerFactory(akkaServiceClient: AkkaServiceClient, da
     listenerMap
   }
 
+  def destroy(): Unit = {
+    akkaServiceClient.destroy()
+  }
+
   private class KeystoneV3ConfigurationListener extends UpdateListener[OpenstackIdentityV3Config] {
     private var initialized = false
 
     def configurationUpdated(config: OpenstackIdentityV3Config) {
+      val akkaServiceClientOld = Option(akkaServiceClient)
+      akkaServiceClient = akkaServiceClientFactory.newAkkaServiceClient(config.getConnectionPoolId)
+      akkaServiceClientOld.foreach(_.destroy())
+
       val identityAPI = new OpenStackIdentityV3API(config, datastoreService.getDefaultDatastore, akkaServiceClient)
       keystoneHandler = new OpenStackIdentityV3Handler(config, identityAPI)
       initialized = true
--- a/repose-aggregator/components/filters/openstack-identity-v3/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3HandlerFactory.scala
+++ b/repose-aggregator/components/filters/openstack-identity-v3/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3HandlerFactory.scala
@@ -48,7 +48,7 @@ class OpenStackIdentityV3HandlerFactory(akkaServiceClientFactory: AkkaServiceCli
   }
 
   def destroy(): Unit = {
-    akkaServiceClient.destroy()
+    Option(akkaServiceClient).foreach(_.destroy())
   }
 
   private class KeystoneV3ConfigurationListener extends UpdateListener[OpenstackIdentityV3Config] {
--- a/repose-aggregator/services/service-client/api/src/main/java/org/openrepose/core/services/serviceclient/akka/AkkaServiceClient.java
+++ b/repose-aggregator/services/service-client/api/src/main/java/org/openrepose/core/services/serviceclient/akka/AkkaServiceClient.java
@@ -32,4 +32,6 @@ public interface AkkaServiceClient {
     ServiceClientResponse get(String token, String uri, Map<String, String> headers) throws AkkaServiceClientException;
 
     ServiceClientResponse post(String requestKey, String uri, Map<String, String> headers, String payload, MediaType contentMediaType) throws AkkaServiceClientException;
+
+    void destroy();
 }
--- a/repose-aggregator/services/service-client/impl/src/main/java/org/openrepose/core/services/serviceclient/akka/impl/AkkaServiceClientImpl.java
+++ b/repose-aggregator/services/service-client/impl/src/main/java/org/openrepose/core/services/serviceclient/akka/impl/AkkaServiceClientImpl.java
@@ -35,9 +35,6 @@ import org.slf4j.Logger;
 import scala.concurrent.Await;
 import scala.concurrent.Future;
 
-import javax.annotation.PreDestroy;
-import javax.inject.Inject;
-import javax.inject.Named;
 import javax.ws.rs.core.MediaType;
 import java.util.Map;
 import java.util.concurrent.Callable;
@@ -47,7 +44,6 @@ import java.util.concurrent.TimeUnit;
 import static akka.pattern.Patterns.ask;
 import static akka.routing.ConsistentHashingRouter.ConsistentHashable;
 
-@Named
 public class AkkaServiceClientImpl implements AkkaServiceClient {
 
     private static final Logger LOG = org.slf4j.LoggerFactory.getLogger(AkkaServiceClientImpl.class);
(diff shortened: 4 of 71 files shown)


repose-aggregator/services/service-client/impl/src/main/java/org/openrepose/core/services/serviceclient/akka/impl/AkkaServiceClientImpl.java
```diff
@@ -71,26 +74,49 @@ public AkkaServiceClientImpl(HttpClientService httpClientService) {
...
 
-                return new AuthTokenFutureActor(serviceClient);
-            }
-        }).withRouter(new RoundRobinRouter(numberOfActors)), "authRequestRouter");
+        if (oldTokenActorRef != null) {
+            actorSystem.stop(oldTokenActorRef);
+        }
     }
 
+    private synchronized void createTokenActor(ForceCreate forceCreate) {
```
dkowis: Instead of trying to handle threads, you should just send messages to actors telling them to change their connection Pool ID. The actors will then change their own state, and no threading needs to happen. The actors can internally see if they are using the right connection pool, and if not change it. If they get multiple messages, it doesn't matter, the actor will know enough to throw it away.

Trying to manage it entirely outside of the actors seems like an error prone way to do this. Much better idea is to tell the actors that you want them to change state, and do it via a message. Then there aren't any synchronization issues, no attempting to hopefully create things in the proper thread.

Mario-Lopez: I like that idea.

Before I put in a lot of effort to dig deeper into Akka and refactor this, do you know if these can be handled:

1. I want to lazily initialization the ServiceClient instance being passed to the AuthTokenFutureActor constructor.
2. I want to be able to update the ServiceClient instance in case the configuration is updated to use a different connection pool id and update the number of actors (based on the pool size).

I may not need to do #1 if it's cheap to always use the default connection pool id and then just update the actors to use the new connection pool id once it's set for the first time.

I may not need to do #2 if we don't want to support configuration updates to the connection pool id.

I need to support at least one of those unless I can get Spring to inject an instance of AkkaServiceClient with the configured connection pool ID (which I don't see how with the current setup).

dkowis: Just always use the default connection pool ID, unless another is specified. 

Did the number of actors correspond to the connection pool size ever? What was the behavior before? Create an actor on each request? (Actors are super cheap, and you can create and throw them away.) If it's using a fixed actor pool size, the number of actors needs to at least be the size of the pool, so you can consume the entire pool. 

I'd recommend creating then a hierarchy of actors, so that an actor can control state below it, you can send a message to that actor to tell it to change it's connection pool. It would then send a kill message to the existing actors, and create a new set. The existing actors would finish up their work and then exit properly and any new work would go to the new actors with the new configuration.

OR: this might be a good time to refactor this. We really don't need actors here at all. All we need are futures, and an executor service.

https://github.com/rackerlabs/repose/blob/master/repose-aggregator/services/service-client/impl/src/main/java/org/openrepose/core/services/serviceclient/akka/impl/AuthTokenFutureActor.java

All this does is abstract a call so that's asynchronous. This can be done with the java Future constructs. It might make things a whole lot simpler, and it might be less work than trying to create a hierarchy of actors (which is how you're supposed to use akka to do what you're trying to do). It might also be a lot more work, I haven't wrapped my head around all the context.

Mario-Lopez: Yes, previously it did:
.withRouter(new RoundRobinRouter(numberOfActors)), "authRequestRouter");
... where the number of actors was the connection pool size.

I agree this class could stand to be refactored.  I don't think that should happen in this story, though.  I can update the class to remove the lazy initialization and just start off with a null connection pool id.  That would simplify the multi-threading logic.  Does that sound fine?

dkowis: What would a null connection pool ID represent?

Mario-Lopez: It's what it does now (without my changes).  It creates a new pool using the default pool configuration.

dkowis: That's probably a pretty good option, lets see what that looks like?

Mario-Lopez: I will show you what I got.

dkowis: show me what you got

repose-aggregator/services/service-client/impl/src/main/java/org/openrepose/core/services/serviceclient/akka/impl/AkkaServiceClientImpl.java
```diff
@@ -72,27 +74,42 @@ public AkkaServiceClientImpl(HttpClientService httpClientService) {
...
     @PreDestroy
     public void destroy() {
-        //Tie this actor system to our bean lifecycle
+        // tie this actor system to our bean lifecycle
         actorSystem.shutdown();
     }
 
     @Override
+    public void setConnectionPoolId(String connectionPoolId) {
+        if (StringUtils.equals(this.connectionPoolId, connectionPoolId)) {
```
dkowis: This state should live in the actor, not in here. You should let the actors decide if they need to change their connection pool ID or not.

Getting sent multiple messages asking them to change their connection pool ID shouldn't matter.

repose-aggregator/services/service-client/impl/src/main/java/org/openrepose/core/services/serviceclient/akka/impl/AkkaServiceClientImpl.java
```diff
@@ -72,27 +74,42 @@ public AkkaServiceClientImpl(HttpClientService httpClientService) {
...
+            return;
+        }
+
+        ActorRef oldTokenActorRef = tokenActorRef;
```
dkowis: You don't want to recreate the actor, you want to change the state in the actor, and then change the size of the pool of actors used for routing.

repose-aggregator/services/service-client/impl/src/main/java/org/openrepose/core/services/serviceclient/akka/impl/AkkaServiceClientImpl.java
```diff
@@ -72,27 +74,42 @@ public AkkaServiceClientImpl(HttpClientService httpClientService) {
...
+        this.connectionPoolId = connectionPoolId;
+        createTokenActor();
+
+        if (oldTokenActorRef != null) {
+            actorSystem.stop(oldTokenActorRef);
+        }
+    }
+
+    private void createTokenActor() {
+        ServiceClient serviceClient = new ServiceClient(connectionPoolId, httpClientService);
```
dkowis: The actor can be created giving a reference to the HttpClient Service, and a default ConnectionPoolId.

Then you can just issue it messages and have it create it's own service client depending on whatever connection ID it needs.

repose-aggregator/services/service-client/impl/src/main/java/org/openrepose/core/services/serviceclient/akka/impl/AkkaServiceClientImpl.java
```diff
@@ -72,27 +74,42 @@ public AkkaServiceClientImpl(HttpClientService httpClientService) {
...
+        this.timeout = new Timeout(serviceClient.getSocketTimeout() + CONNECTION_TIMEOUT_BUFFER_MILLIS, TimeUnit.MILLISECONDS);
+        this.tokenActorRef = actorSystem.actorOf(Props.create(AuthTokenFutureActor.class, serviceClient)
```
dkowis: I think we'll have to level up our abstraction here. We're probably going to have to have a parent actor that relays messages to the children so that we can recreate the children with a new connection PoolID. 

Or, create another actor that you can use to have all these children talk to to get that ServiceClient. Then you can poisonpill the router, and create a new one, after updating the connection pool provider router. 

There isn't a really clean way to do this, without adding more actors, and it's a bit difficult to describe without having a whiteboard :(

repose-aggregator/services/service-client/impl/src/main/java/org/openrepose/core/services/serviceclient/akka/impl/AkkaServiceClientImpl.java
```diff
@@ -54,15 +55,16 @@
     private static final long FUTURE_CACHE_TTL = 500;
     private static final TimeUnit FUTURE_CACHE_UNIT = TimeUnit.MILLISECONDS;
     private static final int CONNECTION_TIMEOUT_BUFFER_MILLIS = 1000;
-    private final ServiceClient serviceClient;
     private final Cache<Object, Future> quickFutureCache;
+    private final HttpClientService httpClientService;
     private ActorSystem actorSystem;
-    private ActorRef tokenActorRef;
+    private volatile Timeout timeout;
+    private volatile ActorRef tokenActorRef;
```
dkowis: Because this doesn't post in chronological order. I want to avoid any `volatile` anything. If we end up there, we're doing concurrency wrong. We shouldn't need any `synchronized` either, we just need to find the right way to use the actor system in a way that makes sense. It might require more work than we thought, but if we don't and fall back to concurrency primitives we'll probably get it wrong. And it's a huge pain to get people to upgrade to newer versions of Repose, so we really need to do as much as possible to get it right the first time.

repose-aggregator/services/service-client/impl/src/main/java/org/openrepose/core/services/serviceclient/akka/impl/AkkaServiceClientImpl.java
dkowis: Hrm, this doesn't solve the number of actors problem. Blerg.

repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -626,6 +626,7 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
       // Removes an extra slash at the end of the URI if applicable
       val serviceUri = keystoneV2Config.getIdentityService.getUri
       keystoneV2Config.getIdentityService.setUri(serviceUri.stripSuffix("/"))
+      akkaServiceClient.setConnectionPoolId(keystoneV2Config.getIdentityService.getConnectionPoolId)
```
adrianjgeorge: this isn't safe. the client is used by multiple filters and you just changed them all to using this one's configured pool

Mario-Lopez: Are you saying that AkkaServiceClient is a singleton that is shared among multiple filters?  I thought that was the case, but when I tested to confirm that wasn't the case.  Or are you saying something else?

repose-aggregator/services/service-client/api/src/main/java/org/openrepose/core/services/serviceclient/akka/AkkaServiceClient.java
```diff
@@ -29,6 +29,8 @@
  */
 public interface AkkaServiceClient {
 
+    void setConnectionPoolId(String connectionPoolId);
```
adrianjgeorge: This guy is a big problem, there are very few reasons for a setter on a shared resource

repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
adrianjgeorge: That is what i'm saying, and everything points to it being a singleton. And I can see no way that it is not

repose-aggregator/services/service-client/api/src/main/java/org/openrepose/core/services/serviceclient/akka/AkkaServiceClient.java
Mario-Lopez: Yeah, I don't like it either, but I'm not sure how else to pass in the connection pool id without either manually creating the object (i.e. not injecting it) or doing some multi-threaded funkiness to ensure it's only set one time upfront (which could also make it a major version change to boot).  In theory (famous last words), this shouldn't cause an issues as the filters that are setting it are doing it before their configs are set to initialized, and the only time they'd ever set it after that would be if the config changes, and I believe a config change of this magnitude would result in Repose just restarting.

repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
adrianjgeorge: What did your test look like? Unless there are some specific hoops setup all of our spring beans are singletons, at least within their context.

Mario-Lopez: What's making it a singleton?

adrianjgeorge: it's a singleton by default, because that's how spring works.

Mario-Lopez: I believe I added some logging to the constructor to see if it was created once or multiple times.  I saw it created multiple times.  Although I'm doubting myself since this was 3 weeks ago.  This was a point of contention in my head, so I know I could be wrong.

repose-aggregator/services/service-client/api/src/main/java/org/openrepose/core/services/serviceclient/akka/AkkaServiceClient.java
adrianjgeorge: So i was thinking about this. I think we revert all the changes to the akka service client itself, and stop making it a bean. Instead we make a factory bean. We wire the factory bean into the places that currently use the client. They ask the factory for a client for a given pool. then we can just add another arguement to the current client constructor and it works almost as is.

repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
adrianjgeorge: i think this client might be one of the per node services, so there could be a couple with the right config. or the update service could be randomly forcing multiple reloads of the main context

repose-aggregator/services/service-client/api/src/main/java/org/openrepose/core/services/serviceclient/akka/AkkaServiceClient.java
adrianjgeorge: We can give the factory a destroy method too, if we want to make sure we get clean teardowns of the actor system

adrianjgeorge: nevermind that second, the client already has a destroy

adrianjgeorge: also, also wik. This is internal api changes so it's only minor version number. major is reserved when we are forcing people to change their configs or operating environment (java 8).

Mario-Lopez: It would break MaaS' custom filter.  How would we communicate that adequately?

https://github.com/rackerlabs/cloud-monitoring-filter-bundle/blob/master/extract-device-id/src/main/java/org/openrepose/filters/custom/extractdeviceid/ExtractDeviceIdFilter.java#L88

Mario-Lopez: Never mind.  I just checked the wiki on this.  Minor it would be. :+1:

adrianjgeorge: yeah, we'd probably send them an email since we know it would be breaking for them. our version numbers are a little weird because all our internal changes causing major version numbers were freaking some customers out, mostly because they had heavy weight processes in place for those kind of version changes.

repose-aggregator/components/filters/openstack-identity-v3/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3HandlerFactory.scala
```diff
@@ -46,10 +47,18 @@ class OpenStackIdentityV3HandlerFactory(akkaServiceClient: AkkaServiceClient, da
     listenerMap
   }
 
+  def destroy(): Unit = {
+    akkaServiceClient.destroy()
```
adrianjgeorge: did you want to option this one too?

Mario-Lopez: Thanks for catching.  Fix pushed.

PR #1408, opened 2015-10-30
REP-2972 Valkyrie account_admin enhancement.

wdschei:
This enhancement provided a backwards compatible way to make the Valkyrie filter's config able to support the new expected default behavior. Part of this behavior required adding a second call to Valkyrie in order to obtain the Account Device Inventory when the original Direct Permissions by Contact call returns a role of 'account_admin'.

### DIFF
--- a/repose-aggregator/components/filters/valkyrie-authorization/src/main/scala/org/openrepose/filters/valkyrieauthorization/ValkyrieAuthorizationFilter.scala
+++ b/repose-aggregator/components/filters/valkyrie-authorization/src/main/scala/org/openrepose/filters/valkyrieauthorization/ValkyrieAuthorizationFilter.scala
@@ -139,17 +139,17 @@ class ValkyrieAuthorizationFilter @Inject()(configurationService: ConfigurationS
 
     def getInventory(headerResult: ValkyrieResult): ValkyrieResult = {
       def parseInventory(inputStream: InputStream): Try[UserPermissions] = {
-        def parseJson(values: Array[JsValue]): UserPermissions = {
+
+        @tailrec
+        def parseJson(deviceToPermissions: List[DeviceToPermission], values: List[JsValue]): UserPermissions = {
           if (values.isEmpty) {
-            UserPermissions(Vector.empty[String], Vector.empty[DeviceToPermission])
+            UserPermissions(List(ACCOUNT_ADMIN).toVector, deviceToPermissions.toVector)
           } else {
-            val permissions: UserPermissions = parseJson(values.tail)
             val currentItem: JsValue = values.head
             (currentItem \ "id").as[Int] match {
               case id if id > 0 =>
-                UserPermissions(permissions.roles,
-                  DeviceToPermission(id, ACCOUNT_ADMIN) +: permissions.devices)
-              case _ => permissions
+                parseJson(DeviceToPermission(id, ACCOUNT_ADMIN) +: deviceToPermissions, values.tail)
+              case _ => parseJson(deviceToPermissions, values.tail)
             }
           }
         }
@@ -157,8 +157,8 @@ class ValkyrieAuthorizationFilter @Inject()(configurationService: ConfigurationS
         val input: String = Source.fromInputStream(inputStream).getLines() mkString ""
         try {
           val json = Json.parse(input)
-          val inventory: Array[JsValue] = (json \ "inventory").as[Array[JsValue]]
-          Success(parseJson(inventory))
+          val inventory: List[JsValue] = (json \ "inventory").as[List[JsValue]]
+          Success(parseJson(List.empty[DeviceToPermission], inventory))
         } catch {
           case e: Exception =>
             logger.error(s"Invalid Json response from Valkyrie: $input", e)
@@ -186,6 +186,7 @@ class ValkyrieAuthorizationFilter @Inject()(configurationService: ConfigurationS
             case "view_product" if List("GET", "HEAD").contains(method) => permissionsWithDevicePermissions
             case "edit_product" => permissionsWithDevicePermissions
             case "admin_product" => permissionsWithDevicePermissions
+            case ACCOUNT_ADMIN => permissionsWithDevicePermissions
             case _ => ResponseResult(403, "Not Authorized")
           }
         } getOrElse {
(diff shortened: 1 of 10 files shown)


repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/framework/mocks/MockValkyrie.groovy
```diff
@@ -76,10 +76,12 @@ class MockValkyrie {
     String account_perm = "also_butts"
     String contact_id = ""
     String tenant_id = ""
+    Integer fixnumber = 10
+    Integer multiplier = 1
```
wdschei: If this is going to be split into two separate values, then the names should probably reflect what they are tied to (e.g. device, inventory). For example deviceCount/invnentoryCount, devMulti/invMulti, device_multiplier/inventory_multiplier.

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/framework/mocks/MockValkyrie.groovy
```diff
@@ -311,7 +306,105 @@ class MockValkyrie {
...
+                        ]
+                    },""").make(params))
+        }
+        lotsOJson.deleteCharAt(lotsOJson.size() - 1) //Delete the pesky trailing comma
+
         //Now glue all the things together
         originalTemplate + lotsOJson + "]}"
     }
+
+
```
wdschei: What's up with the extra lines?

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/framework/mocks/MockValkyrie.groovy
jennyvo: yes, that why I don't want to slow other tests just set fixnumber as 10 and use multiplier as flexible number and use only as inventory call is made

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/framework/mocks/MockValkyrie.groovy
jennyvo: don't know it added after run code format

repose-aggregator/components/filters/valkyrie-authorization/src/main/scala/org/openrepose/filters/valkyrieauthorization/ValkyrieAuthorizationFilter.scala
```diff
@@ -135,6 +137,47 @@ class ValkyrieAuthorizationFilter @Inject()(configurationService: ConfigurationS
...
+        }
+      }
+
+      headerResult match {
+        case UserInfo(tenant, contact) =>
+          //  authorize device            || cull list                  || translate account permissions
+          if (requestedDeviceId.isDefined || matchingResources.nonEmpty || translateAccountPermissions.isDefined) {
+            datastoreValue(tenant, contact, ACCOUNT_ADMIN, configuration.getValkyrieServer, _.asInstanceOf[UserPermissions], parseInventory, tracingHeader)
+          } else {
+            ResponseResult(200)
```
Mario-Lopez: I think this line is unreachable.  The if check is already done in the getPermissions method, and if it fails there then the returned ResponseResult(200) would prevent this method from ever being called.

wdschei: Good catch. That was a copy/paste of the original that didn't get completely cleaned up. ;-)

repose-aggregator/components/filters/valkyrie-authorization/src/main/scala/org/openrepose/filters/valkyrieauthorization/ValkyrieAuthorizationFilter.scala
```diff
@@ -135,6 +137,42 @@ class ValkyrieAuthorizationFilter @Inject()(configurationService: ConfigurationS
...
+      def parseInventory(inputStream: InputStream): Try[UserPermissions] = {
+
+        @tailrec
+        def parseJson(deviceToPermissions: List[DeviceToPermission], values: List[JsValue]): UserPermissions = {
+          if (values.isEmpty) {
+            UserPermissions(List(ACCOUNT_ADMIN).toVector, deviceToPermissions.toVector)
+          } else {
+            val currentItem: JsValue = values.head
+            (currentItem \ "id").as[Int] match {
+              case id if id > 0 =>
```
Mario-Lopez: Why don't we want negative numbers?

repose-aggregator/components/filters/valkyrie-authorization/src/main/scala/org/openrepose/filters/valkyrieauthorization/ValkyrieAuthorizationFilter.scala
```diff
@@ -135,6 +137,42 @@ class ValkyrieAuthorizationFilter @Inject()(configurationService: ConfigurationS
       }
     }
 
+    def getInventory(headerResult: ValkyrieResult): ValkyrieResult = {
+      def parseInventory(inputStream: InputStream): Try[UserPermissions] = {
+
+        @tailrec
+        def parseJson(deviceToPermissions: List[DeviceToPermission], values: List[JsValue]): UserPermissions = {
+          if (values.isEmpty) {
+            UserPermissions(List(ACCOUNT_ADMIN).toVector, deviceToPermissions.toVector)
```
Mario-Lopez: Later on we add these to the existing UserPermissions which means the "account_admin" role will show up twice now.  This should be Vector.empty.  Or, better yet, have the parseJson method just return a Vector of DeviceToPermissions instead of UserPermissions.  Up to you.

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/framework/mocks/MockValkyrie.groovy
Mario-Lopez: I'm confused.  To me, fixnumber and multiplier imply that at some point you will multiply the fixed number "fixnumber" by "multiplier" which means you could only ever have some number that's a multiple of the fixed number (10 in this case).  That's not what I'm seeing.  Like Bill mentioned, it seems like "fixnumber" really means "number of devices in the effective call" and "multiplier" really means "number of devices in the inventory call".

Mario-Lopez:
Maybe because it's past my bedtime, but did I miss a functional test for having enable-bypass-account-admin set to true?

repose-aggregator/components/filters/valkyrie-authorization/src/main/scala/org/openrepose/filters/valkyrieauthorization/ValkyrieAuthorizationFilter.scala
wdschei: From my understanding, no.

repose-aggregator/components/filters/valkyrie-authorization/src/main/scala/org/openrepose/filters/valkyrieauthorization/ValkyrieAuthorizationFilter.scala
wdschei: I made it an empty `Vector`, but for now I left it returning `UserPermissions` so that it remains compatible with a `ValkyrieResult`.

wdschei:
There was not a functional test, but that was the only case tested in the entirety of the original unit test suite.

PR #1415, opened 2015-11-12
Updated to API-Checker v1.1.3 and in turn WADL-Tools v1.0.31.

wdschei:


adrianjgeorge:
looks good, but lets add a functional test that proves it solves the bug

repose-aggregator/extensions/pom.xml
```diff
@@ -18,9 +18,33 @@
         No other filters should be developed here.
     </description>
 
+    <properties>
+        <api-checker.version>1.1.3</api-checker.version>
```
adrianjgeorge: this is weirdly inconsistent, you have dashes and dots

repose-aggregator/extensions/pom.xml
```diff
@@ -18,9 +18,33 @@
...
         <module>simple-rbac</module>
         <module>extensions-filter-bundle</module>
     </modules>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.openrepose</groupId>
+                <artifactId>core-service-api</artifactId>
+                <version>${project.version}</version>
```
adrianjgeorge: i'd rather not dependency manage in project dependencies

adrianjgeorge: It makes things difficult down the road if we try to break things up

repose-aggregator/extensions/pom.xml
wdschei: I was following the format that API-Checker uses:
* https://github.com/rackerlabs/api-checker/blob/master/pom.xml#L53

What would you rather it be?
All dashes or all dots?

repose-aggregator/extensions/pom.xml
wdschei: So push the `org.openrepose` dependencies back into the individual module POM's, but leave the API-Checker one up here?

repose-aggregator/extensions/pom.xml
adrianjgeorge: all dots, also feel free to fix that other one, because that is terrible

repose-aggregator/extensions/pom.xml
adrianjgeorge: makes sense to me

repose-aggregator/extensions/pom.xml
wdschei: Updated.

repose-aggregator/extensions/pom.xml
wdschei: Done.

wdschei:
Retest this please.

adrianjgeorge:
commits already pulled in elsewhere

PR #1417, opened 2015-11-18
REP-2974 Repose Lint and Tenantless Traffic Check

dmnjohns:


repose-aggregator/components/repose-lint/src/main/scala/org/openrepose/lint/commands/VerifyTryItNowCommand.scala
```diff
@@ -0,0 +1,466 @@
...
+      )
+    }
+    val clustersArray = clusterJsonObjects.foldLeft(Json.arr())((arr, obj) => arr :+ obj)
+
+    println(Json.prettyPrint(Json.obj("clusters" -> clustersArray)))
+  }
+
+  object FoyerStatus extends Enumeration {
+    type FoyerStatus = Value
+    val Ready, NotReady, Unknown = Value
```
dmnjohns: Rather than Ready, NotReady, and Unknown, use Allowed, NotAllowed, and Unknown.

dmnjohns:
retest this please

repose-aggregator/components/repose-lint/pom.xml
```diff
@@ -0,0 +1,121 @@
...
+        A command-line utility which flags suspicious usage of Repose.
+    </description>
+
+    <properties>
+        <sonar.jacoco.itReportPath>${project.basedir}/../../target/jacoco-it.exec</sonar.jacoco.itReportPath>
+    </properties>
+
+    <dependencies>
+        <!-- Scala language support -->
+        <dependency>
```
adrianjgeorge: I thought we started doing this at a higher level, so it didn't have to be done everywhere?

repose-aggregator/components/repose-lint/pom.xml
```diff
@@ -0,0 +1,121 @@
...
+        <!-- Command-line argument parser -->
+        <dependency>
+            <groupId>com.github.scopt</groupId>
+            <artifactId>scopt_2.10</artifactId>
+            <version>3.2.0</version>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <finalName>repose-lint</finalName>
```
adrianjgeorge: why? this is already the name you'll get

repose-aggregator/components/repose-lint/pom.xml
```diff
@@ -0,0 +1,121 @@
...
+
+        <resources>
+            <resource>
+                <directory>src/main/resources</directory>
+                <filtering>true</filtering>
+                <includes>
```
adrianjgeorge: pretty sure we get everything by default, was there something you didn't want?

repose-aggregator/components/repose-lint/pom.xml
```diff
@@ -0,0 +1,121 @@
...
+                    <include>lint.conf</include>
+                </includes>
+            </resource>
+        </resources>
+
+        <testResources>
+            <testResource>
+                <directory>src/test/resources</directory>
```
adrianjgeorge: pretty sure this is the default

repose-aggregator/components/repose-lint/src/main/scala/org/openrepose/lint/CommandExecutor.scala
```diff
@@ -0,0 +1,93 @@
...
+    // Attach the actual console output streams to the passed in streams
+    Console.setOut(out)
+    Console.setIn(in)
+    Console.setErr(err)
+
+    val lintVer = config.getString("version")
+
+    val parser = new OptionParser[LintConfig]("repose-lint") {
+      head("repose-lint", lintVer)
+      opt[File]('c', "config-dir") valueName "<dir>" action { (x, c) =>
```
adrianjgeorge: maybe put in a line break between each option, and a comment saying what it is? then even though i can't read this gobblygook i at least know what they are

repose-aggregator/components/repose-lint/src/main/scala/org/openrepose/lint/CommandExecutor.scala
```diff
@@ -0,0 +1,93 @@
...
+              0
+            } catch {
+              case t: Throwable =>
+                // The command has failed
+                Console.err.println(s"${command.getCommandToken} command failed")
+                Console.err.println(s"Cause: ${t.getMessage}")
+                1
+            }
+          case None =>
+            // Failed to lookup the command (this should never happen since the parser should catch it first)
```
adrianjgeorge: maybe a message on this and the one below, in case the impossible happens we can at least know what impossible thing happened.

repose-aggregator/components/repose-lint/src/main/scala/org/openrepose/lint/commands/VerifyTryItNowCommand.scala
```diff
@@ -0,0 +1,491 @@
...
+
+  override def perform(lintConfig: LintConfig): Unit = {
+    val reposeVersion = lintConfig.reposeVersion
+
+    def versionGreaterThanOrEqualTo(otherVersion: String): Boolean = {
+      val reposeVersionSplit = reposeVersion.split('.').map(_.toInt)
+      val otherVersionSplit = otherVersion.split('.').map(_.toInt)
+
+      val isReposeVersionShorter = reposeVersionSplit.length < otherVersionSplit.length
+      val comparisonIdx = if (isReposeVersionShorter) {
```
adrianjgeorge: what is this trying to do?

adrianjgeorge: nevermind, i finally figured it out. this is super not intuitive

dmnjohns: I know. I really hate this code, but couldn't figure out a better way to do it.

repose-aggregator/components/repose-lint/src/main/scala/org/openrepose/lint/CommandExecutor.scala
dmnjohns: Will do.

repose-aggregator/components/repose-lint/src/main/scala/org/openrepose/lint/CommandExecutor.scala
dmnjohns: Will do.

adrianjgeorge:
It occured to me last night that part of our problem was we were trying to get down to two states (not counting the indeterminate). I think there might be three, will not allow, will allow securely, and will allow insecurely. Would that help or just compound matters further?

repose-aggregator/components/repose-lint/pom.xml
dmnjohns: It looks like the dependency is managed, but not declared as a hard dependency at the "profile-support" level.

repose-aggregator/components/repose-lint/pom.xml
dmnjohns: This was a copy and paste from the cli-utils pom. However, I think there is a reason it's being done. From the Maven pom reference:

> finalName: This is the name of the bundled project when it is finally built (sans the file extension, for example: my-project-1.0.jar). It defaults to ${artifactId}-${version}. The term "finalName" is kind of a misnomer, however, as plugins that build the bundled project have every right to ignore/modify this name (but they usually do not). For example, if the maven-jar-plugin is configured to give a jar a classifier of test, then the actual jar defined above will be built as my-project-1.0-test.jar.

So it looks like finalName is being used to drop the version. I accepted that at some point in the past, but I don't remember why at this exact moment.

repose-aggregator/components/repose-lint/pom.xml
dmnjohns: Nope, I just wanted to make sure that the lint.conf file that Maven sticks the Repose version is makes it into the JAR.

I'll make sure we get it by default and drop this section if so.

dmnjohns: Looks like it gets added by default. Will remove this section.

repose-aggregator/components/repose-lint/pom.xml
dmnjohns: It looks like it. The resources and testResources configuration in this file was actually copied and pasted from the Valve pom, so we could perhaps clean that up a smidgen.

Will remove this section.

repose-aggregator/components/repose-lint/src/main/scala/org/openrepose/lint/CommandExecutor.scala
dmnjohns: In the second (bottom) case scopt should have already printed an error message. I can print another if we don't mind the redundant output, but I'll hold off on doing that for the moment.

dmnjohns:
That's a fair question. It seems to me that, ideally, we would be able to declare whether or not Repose can handle foyer user traffic appropriately for a given service. In some cases we can, but in the cases where we can't, why not? I've come up with the following:
* At the moment, we believe foyer users will only be granted access to the Containers service endpoints. We do not have a list of these endpoints to compare to users' configuration. Additionally, in the future, accessible endpoints could change, invaliding the verify-try-it-now command for this version of Repose Lint (which may not actually matter that much).
1. Most services want to block foyer user traffic. The exception, at the moment is the Containers service. There could be more exceptions in the future. I'll note that this is the reason we are using Allowed/Disallowed rather than Ready/NotReady.
2. There are a ton of variables. I'll note that this is the reason we created truth tables for the variables we identified as being pertinent.
3. Runtime data. We don't know the URL of the request made by a foyer user until that request is actually made. Thus, we have no way of knowing whether or not a filter with a uri-regex set will be applied to all relevant requests. This is also true of user endpoints, although we have been given some information on that front.

So back to your suggestion. It might help clear up some of the unknown states (I'm looking at AuthZ specifically, which falls victim to point 1 above). Rather than saying "Unknown" when foyer is not on the ignore-tenant-roles list, we could say "AllowedIfAuthorized" or "AllowedIfEndpointMatches". To generalize, we could add states for conditions which are determined by runtime data.

dmnjohns:
Forgot to update the functional tests...

Fixing now.

PR #1418, opened 2015-11-18
REP-3043 Updated the Valkyrie filter to only cull 2xx responses.

wdschei:

PR #1419, opened 2015-11-19
REP-3025 Updated Valkyrie filter's culling to be HTTP method aware.

wdschei:
This has "REP-3043 Updated the Valkyrie filter to only cull 2xx responses (#1418)" cherry-picked in to ensure no conflicts on merge.


repose-aggregator/components/filters/valkyrie-authorization/src/main/scala/org/openrepose/filters/valkyrieauthorization/ValkyrieAuthorizationFilter.scala
```diff
@@ -75,7 +76,15 @@ class ValkyrieAuthorizationFilter @Inject()(configurationService: ConfigurationS
     val tracingHeader = nullOrWhitespace(Option(mutableHttpRequest.getHeader(CommonHttpHeader.TRACE_GUID.toString)))
     val urlPath: String = new URL(mutableHttpRequest.getRequestURL.toString).getPath
     val matchingResources: Seq[Resource] = Option(configuration.getCollectionResources)
-      .map(_.getResource.asScala.filter(_.getPathRegex.r.pattern.matcher(urlPath).matches())).getOrElse(Seq.empty[Resource])
+      .map(_.getResource.asScala.filter(resource => {
+        val pathRegex = resource.getPathRegex
+        pathRegex.getValue.r.pattern.matcher(urlPath).matches() && {
+          val httpMethods = pathRegex.getHttpMethods
+          httpMethods.isEmpty ||
```
adrianjgeorge: this should never be possible, and if it is some how possible this would be the opposite of the behavior we'd want.

wdschei: During testing we saw the default, ALL, was not always present. Since the default is ALL, if it is empty, then it would be the equivalent of ALL. How is this opposite behavior?

adrianjgeorge: because if they have set it to empty, that would imply that they want it to happen for no http methods. I'm not sure how you were getting empty in system tests, that would make sense in unit tests though where a config isn't getting marshaled.

wdschei: It appears the reason the default was missing is actually a JAXB/XJC bug.
* http://stackoverflow.com/questions/27528698/is-there-any-way-i-can-specify-defaultvalue-for-collections-with-jaxb

PR #1420, opened 2015-11-20
Rep 2128 url extraction to header filter

Mario-Lopez:
Added new filter that let's you extract part of the URL and put it in a header.  The header name is configurable, the header value is the first group in the configured regex, and a default value can be specified (or left out if the header shouldn't added) when the regex doesn't match the URL.

### DIFF
--- a/repose-aggregator/components/filters/url-extractor-to-header/src/main/scala/org/openrepose/filters/urlextractortoheader/UrlExtractorToHeaderFilter.scala
+++ b/repose-aggregator/components/filters/url-extractor-to-header/src/main/scala/org/openrepose/filters/urlextractortoheader/UrlExtractorToHeaderFilter.scala
@@ -1,0 +1,30 @@
+/*
+ * _=_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_=
+ * Repose
+ * _-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-
+ * Copyright (C) 2010 - 2015 Rackspace US, Inc.
+ * _-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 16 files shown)


repose-aggregator/components/filters/url-extractor-to-header/src/main/scala/org/openrepose/filters/urlextractortoheader/UrlExtractorToHeaderFilter.scala
```diff
@@ -0,0 +1,93 @@
...
+class UrlExtractorToHeaderFilter @Inject()(configurationService: ConfigurationService)
+  extends Filter with UpdateListener[UrlExtractorToHeaderConfig] with LazyLogging {
+
+  import UrlExtractorToHeaderFilter._
+
+  private var configurationFile: String = DEFAULT_CONFIG
+  private var initialized = false
+  private var extractions: Iterable[Extraction] = _
+
+  override def init(filterConfig: FilterConfig): Unit = {
```
adrianjgeorge: We should extract this to a base class, because it looks pretty much exactly the same for every filter we write. This is not work for you right now, i'm just making a note.

repose-aggregator/components/filters/url-extractor-to-header/src/main/scala/org/openrepose/filters/urlextractortoheader/UrlExtractorToHeaderFilter.scala
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  override def doFilter(servletRequest: ServletRequest, servletResponse: ServletResponse, filterChain: FilterChain): Unit = {
+    val mutableHttpRequest = MutableHttpServletRequest.wrap(servletRequest.asInstanceOf[HttpServletRequest])
+
+    extractions.foreach { extraction =>
+      (extraction.urlRegex.findFirstIn(mutableHttpRequest.getRequestURI), extraction.defaultValue) match {
+        case (Some(extraction.urlRegex(headerValue)), _) => mutableHttpRequest.addHeader(extraction.headerName, headerValue)
+        case (None, Some(defaultValue)) => mutableHttpRequest.addHeader(extraction.headerName, defaultValue)
+        case (None, None) => // don't add a header
```
adrianjgeorge: I'm torn between leaving our third case explicit, and future proofing by using case (_, _).

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/filters/urlextractortoheader/UrlExtractorToHeaderWValkyrieTest.groovy
```diff
@@ -0,0 +1,140 @@
+package features.filters.urlextractortoheader
```
adrianjgeorge: license

PR #1430, opened 2016-01-06
Adding rackspace basic auth password filter

the2hill:
Really just a CP of the Rackspace basic auth filter. Could probably have been done in one, but this allows a user to submit username and password to retrieve a token. 

This may not be specific to rackspace and could potentially be renamed. 

The use case for this is:

I have a user that only has a username and password, usually the admin. The admin needs to be validated against the auth service in order to use a service behind repose, but must retrieve an auth token first. This filter allows the user to submit an Authorization header in the basic auth format to securely authenticate with only a username and a password. .

### DIFF
--- a/repose-aggregator/components/filters/filter-bundle/pom.xml
+++ b/repose-aggregator/components/filters/filter-bundle/pom.xml
@@ -123,6 +123,11 @@
         </dependency>
         <dependency>
             <groupId>org.openrepose.rackspace</groupId>
+            <artifactId>rackspace-identity-basic-auth-password</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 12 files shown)


adrianjgeorge:
ok to test

adrianjgeorge:
While this works, I'm not a big fan of the large amount of duplication it brings. I'd much rather see a version that is a configuration item on the original filter, that determines if it tries to use it as an api key or password.

dmnjohns:
Just to expand on something that was said, I believe the reason that the filter as it currently stands is Rackspace specific, since username/api-key authentication is an Identity extension.

If we merge this functionality into the existing filter, we don't have to keep the Rackspace specific name, but it should be noted (perhaps in documentation) that api-key authentication is an extension.

adrianjgeorge:
Superseded by #1437

PR #1432, opened 2016-01-08
REP 3175 Add Header Filter config update

Mario-Lopez:
Previously the Add Header filter required you to configure headers for both the request and the response.  I updated the XSD to allow a user to configure just one of them if they wanted to.  The code already handles having just one of them configured.

There's an existing functional test with both configured.  I added two additional tests with just the request and just the response to ensure it worked.

### DIFF
--- a/repose-aggregator/components/filters/add-header/src/main/resources/META-INF/schema/config/add-header.xsd
+++ b/repose-aggregator/components/filters/add-header/src/main/resources/META-INF/schema/config/add-header.xsd
@@ -30,8 +30,8 @@
 
     <xs:complexType name="AddHeadersConfig">
         <xs:all>
-            <xs:element name="request" type="HttpMessage"/>
-            <xs:element name="response" type="HttpMessage"/>
+            <xs:element name="request" type="HttpMessage" minOccurs="0"/>
+            <xs:element name="response" type="HttpMessage" minOccurs="0"/>
         </xs:all>
     </xs:complexType>
 
--- a/repose-aggregator/components/filters/add-header/src/main/resources/META-INF/schema/config/add-header.xsd
+++ b/repose-aggregator/components/filters/add-header/src/main/resources/META-INF/schema/config/add-header.xsd
@@ -22,6 +22,9 @@
 
 <xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
            xmlns:html="http://www.w3.org/1999/xhtml"
+           xmlns:vc="http://www.w3.org/2007/XMLSchema-versioning"
+           xmlns:xerces="http://xerces.apache.org"
+           xmlns:saxon="http://saxon.sf.net/"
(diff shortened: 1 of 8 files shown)


repose-aggregator/components/filters/add-header/src/main/resources/META-INF/schema/config/add-header.xsd
```diff
@@ -30,8 +30,8 @@
 
     <xs:complexType name="AddHeadersConfig">
         <xs:all>
-            <xs:element name="request" type="HttpMessage"/>
-            <xs:element name="response" type="HttpMessage"/>
+            <xs:element name="request" type="HttpMessage" minOccurs="0"/>
+            <xs:element name="response" type="HttpMessage" minOccurs="0"/>
```
wdschei: This should probably have the logic to ensure that there is at least one of them.

Mario-Lopez: I totally agree.  Do you know of an easy way to do that?  I only found two unideal ways of doing it on the web.  The less insane option was to do an assertion.  I vaguely remember seeing some other filter's doc mention having such a case of not configuring anything and the filter just being a "pass through", so I figured there was precedent.  Thoughts?

repose-aggregator/components/filters/add-header/src/main/resources/META-INF/schema/config/add-header.xsd
```diff
@@ -30,8 +30,8 @@
...
         </xs:all>
```
wdschei: Something like:
```
           test="count(*)>=1"
           xerces:message="At least one header must be defined."
           saxon:message="At least one header must be defined."/>
```

Mario-Lopez: Sounds good.  I'll add that in.  Thanks!

repose-aggregator/components/filters/add-header/src/main/resources/META-INF/schema/config/add-header.xsd
```diff
@@ -30,9 +33,13 @@
...
+        <xs:assert vc:minVersion="1.1"
+                   test="count(*)>=1"
+                   xerces:message="At least one header must be defined."
+                   saxon:message="At least one header must be defined."/>
```
wdschei: Was there a test to make sure this worked?

jennyvo: will check and add if need

wdschei: It looks like this is the only thing missing before I can sign off on it.

PR #1433, opened 2016-01-13
REP-3251 Add Header filter wrapper

Mario-Lopez:
I refactored out the old delegation stuff putting the little bit of business logic all in the AddHeaderFilter class.  I rewrote the unit tests since the old ones were dependent on the filter director.  I also updated the filter to use the new wrappers instead of the old.


repose-aggregator/components/filters/add-header/src/main/scala/org/openrepose/filters/addheader/AddHeaderFilter.scala
```diff
@@ -21,42 +21,79 @@ package org.openrepose.filters.addheader
...
+
+    addHeaders(requestWrapper, config.getRequest)
+
+    chain.doFilter(requestWrapper, responseWrapper)
+
+    addHeaders(responseWrapper, config.getResponse)
+    responseWrapper.commitToResponse()
+  }
+
+  def addHeaders(wrapper: HeaderInteractor, configuredHeaders: HttpMessage): Unit = {
```
adrianjgeorge: Boom, interface already proved it's worth.

repose-aggregator/components/filters/add-header/src/test/scala/org/openrepose/filters/addheader/AddHeaderFilterTest.scala
```diff
@@ -0,0 +1,222 @@
...
+    }
+  }
+
+  def getPostFilterRequest: HttpServletRequestWrapper = {
+    val requestCaptor = ArgumentCaptor.forClass(classOf[HttpServletRequestWrapper])
+    verify(filterChain).doFilter(requestCaptor.capture(), any(classOf[ServletResponse]))
+    requestCaptor.getValue
+  }
+
+  def addResponseHeaders(headers: Iterable[SimpleHeader]): Unit = {
```
adrianjgeorge: this works, but wouldn't it have been simpler to just examine the response you send into the filter call to begin with?

adrianjgeorge: Hmmm, maybe not because you need some magic to happen in the middle. I forgot about that.

adrianjgeorge:
Man i wish we could get it to report how many failures there were.

PR #1434, opened 2016-01-14
REP-3252 Compression filter wrapper updates

Mario-Lopez:
In CompressingFilter.java, I moved the code that removes the Content-Encoding and Accept-Encoding headers since it really only needs to happen if the request doesn't get wrapped in the CompressedHttpServletRequest wrapper (which hides those headers).  I also removed the response buffer commit as that should be the job of the previous filter to do such a commit.

In CompressionFilter.java, I removed the whole thing and converted it to Scala.  JVMs love Scala.  I moved over the CompressionHandlerFactory logic (to build up the config) and the exception handling logic in CompressionHandler to CompressionFilter.scala.

I rewrote CompressionConfigWrapper.java in Scala as well, and I created CompressingFilterFactory since the CompressingFilter can be created multiple times with different config (I'm not absolutely sure this is necessary, but I kept it this way for now since that's how it was before; REP-3310 will fix this).

I converted the old unit tests (that were dependent on FilterDirector) to Scala Test and added additional tests around config loading.

These updates add 3 more failing functional tests to the REP-2188 branch for a total of 27 failed functional tests.  The three tests are due to the response headers that Deproxy is sending back to Repose not making it to the Compression filter.  Bill says this is the same issue with the original 24 failing tests, and we expect them to be fixed when we work on the Filter Director.

I didn't carry over some of the try/catches that were allowing initialization to silently fail.


Mario-Lopez:
27 functional tests failed.  24 were pre-existing, and the 3 new failures are due to the same reason.  These will be fixed in REP-3309.

repose-aggregator/components/filters/compression/src/main/scala/org/openrepose/filters/compression/CompressionFilter.scala
```diff
@@ -0,0 +1,149 @@
...
+
+    logger.info(s"Initializing Compression filter using config $configurationFile")
+    val xsdUrl: URL = getClass.getResource(SCHEMA_FILE_NAME)
+    configurationService.subscribeTo(filterConfig.getFilterName, configurationFile, xsdUrl, this, classOf[ContentCompressionConfig])
+
+    logger.trace("Compression filter initialized.")
+  }
+
+  override def doFilter(request: ServletRequest, response: ServletResponse, filterChain: FilterChain): Unit = {
+    val responseWrapper: HttpServletResponseWrapper = new HttpServletResponseWrapper(
```
adrianjgeorge: I don't see why you are wrapping it? The only method call i see is `setStatus` which is a standard part of the `HttpsServletResponse` contract

repose-aggregator/components/filters/compression/src/main/scala/org/openrepose/filters/compression/CompressingFilterFactory.scala
```diff
@@ -0,0 +1,37 @@
...
+
+package org.openrepose.filters.compression
+
+import javax.inject.Named
+import javax.servlet.FilterConfig
+
+import org.openrepose.external.pjlcompression.CompressingFilter
+
+@Named
+class CompressingFilterFactory {
```
adrianjgeorge: This is pointless. I know you kept it because it existed before, but just inline this code and kill a pointless class.

adrianjgeorge: Helps with testing I guess. I'd still rather murder it in it's sleep.

repose-aggregator/components/filters/compression/src/main/scala/org/openrepose/filters/compression/CompressionFilter.scala
Mario-Lopez: I should check the HttpServletResponse interface more often.  I'll update it.

repose-aggregator/components/filters/compression/src/main/scala/org/openrepose/filters/compression/CompressingFilterFactory.scala
Mario-Lopez: Yeah, that's the only reason why it's there (for testing).  Should I leave it then?

adrianjgeorge: I'd prefer it gone, but if the testing is a huge pain without it, i guess i can live with it for now.

Mario-Lopez: Cool.

PR #1435, opened 2016-01-14
REP-3204 - Fix Keystone v2^2 Get Groups call to Keystone/Identity

wdschei:
The `mockAkkaServiceClient.validate()` call in `KeystoneV2FilterTest` was moved from the `after()` method into each test. This was done so that all the tests would run and not just abort on the first failure with no indication of the why it had stopped or even which test had caused the issue.


adrianjgeorge:
why isn't this running in the pr builder?

adrianjgeorge:
oh now, it shows up

adrianjgeorge:
looks good, if it passes we'll merge it

PR #1437, opened 2016-01-18
REP-3304 Add Username/Password Authentication to Rackspace Basic Auth Filter

dmnjohns:


dmnjohns:
Functional tests pending

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/framework/mocks/MockIdentityV2Service.groovy
```diff
@@ -469,8 +485,10 @@ class MockIdentityV2Service {
             // THEN return the Admin token response.
             if (request.body.contains("username") &&
                     request.body.contains(client_username) &&
-                    request.body.contains("apiKey") &&
-                    request.body.contains(client_apikey)) {
+                    ((request.body.contains("apiKey") &&
```
adrianjgeorge: spacing is weird here

adrianjgeorge:
I assume the url for the request is the same?

dmnjohns:
Yes, the URL is the same. A POST to "/v2.0/tokens" works for both username/api-key and username/password authentication. However, the URL we make the request against is just whatever is in the configuration file, so it's really up to the user to provide the correct endpoint.

jennyvo:
retest this please

dmnjohns:
To follow up on the URL being the same bit...

It appears that sometime in the past, our tests were expecting the path to be "/tokens" rather than "/v2.0/tokens". That is why you see some changes that Jenny made to the mock Identity service. "/v2.0/tokens" is the correct path defined by the Identity API. In the future, we should make sure that the mocks expect a path of "/v2.0/tokens", but whether or not we do that as part of this PR is up to @adrianjgeorge .

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/framework/mocks/MockIdentityV2Service.groovy
dmnjohns: Interestingly, IntelliJ's auto-formatter keeps this formatting. I think the reason is that the logic is kind of complex, so parenthesis are being used to group predicates so that the connectives fit together in the way we want them to.

*tldr;* Parenthesis for excepting operator precedence lead to odd auto-formatting.

adrianjgeorge:
I don't see a need to change it right now especially since we want to get this in now.

adrianjgeorge:
I think we should add a story for v9 that makes sure our identity urls in configs are consistent in how much of the endpoint they are expecting.

adrianjgeorge:
Actually, we could try to drop it into v8

adrianjgeorge:
I made the story

PR #1438, opened 2016-01-18
Updated the API-Checker version to 1.1.4.

wdschei:


adrianjgeorge:
We also wanted to turn the new option on  by default

wdschei:
Yes, we did. Updated.

adrianjgeorge:
Of course now it needs a test to prove that the feature is on and working.

wdschei:
We aren't going to rely on the API-Checker tests for this?

adrianjgeorge:
We need a test to prove that we are defaulting it to on. I'm fine with a unit test.

adrianjgeorge:
You made a prod code change that had no test requiring it.

wdschei:
Unfortunately The API-Validator filter uses the old pattern which precludes adding easy unit tests. Also since there were other functional tests dealing with required headers, I was able to easily add a new set of tests around them.

adrianjgeorge:
That works, hopefully it goes green and i'll merge it.

PR #1442, opened 2016-01-19
REP-2196 Keystone v2^2 Cache Invalidation via Atom Feed service.

wdschei:


repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -299,6 +304,21 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
...
+      if (CacheInvalidationFeedListener.invalidationEnabled()) {
+        //Have to use Vector, because List isn't serializeable in 2.10
+        val oldTokens = Option(datastore.get(s"$USER_ID_KEY_PREFIX$userId").asInstanceOf[Vector[String]]).getOrElse(Vector.empty[String])
+        if (!oldTokens.contains(authToken)) {
+          // @TODO: Do we need to clean up the oldTokens which may have irrelevant values.
+          // @TODO: Nothing is wrong with the data, but it may have extra values that would try to be removed later.
+          val newTokens = oldTokens ++ Vector(authToken)
+          // Updated for later cache invalidation via Atom Feed User events.
+          datastore.remove(s"$USER_ID_KEY_PREFIX$userId")
+          datastore.put(s"$USER_ID_KEY_PREFIX$userId", newTokens, ttl, TimeUnit.SECONDS)
```
adrianjgeorge: you've created a potential threading issue

adrianjgeorge: i was going to say just do a put, but there's a bigger overall issue, if we get multiple requests concurrently for the same user but different tokens they will potentially stomp each other and info will get lost. I think the right choice is to do a patch here, then you have no need to do the get either. Make it a set of some sort instead of a vector and just blindly fire off a patch.

repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -645,6 +669,58 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
...
+      registeredFeeds.synchronized {
+        registeredFeeds.foreach(feed =>
+          atomFeedService.unregisterListener(feed)
+        )
+      }
+    }
+
+    def registerFeeds(feeds: List[AtomFeedType]): Unit = {
+      registeredFeeds.synchronized {
+        unRegisterFeeds()
```
adrianjgeorge: We shouldn't just drop and re add here, your are going to potentially lose events which means people aren't necessarily going to get evicted like they should and that's a problem.

repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -299,6 +304,21 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
       }
     }
 
+    def doUserTokensUpdate(userId: String, authToken: String, ttl: Int): Unit = {
+      if (CacheInvalidationFeedListener.invalidationEnabled()) {
```
adrianjgeorge: I don't feel like this particularly needs to be conditionally added, is they turn it on after the fact it'd be nice for it to start working for the items already in the cache.

repose-aggregator/components/filters/keystone-v2/src/test/scala/org/openrepose/filters/keystonev2/KeystoneV2FilterCacheInvalidationTest.scala
```diff
@@ -0,0 +1,180 @@
...
+  Mockito.when(mockDatastoreService.getDefaultDatastore).thenReturn(mockDatastore)
+
+  before {
+    reset(mockDatastore)
+    reset(mockConfigurationService)
+    mockAkkaServiceClient.reset()
+  }
+
+  after {
+    mockAkkaServiceClient.validate()
```
adrianjgeorge: didn't you just fix this on another test?

repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
adrianjgeorge: Patch is not as little work as i'd like it to be...

repose-aggregator/components/filters/keystone-v2/src/test/scala/org/openrepose/filters/keystonev2/KeystoneV2FilterCacheInvalidationTest.scala
wdschei: Yep. I missed this one since it wasn't causing any issues yet. Updated.

repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
wdschei: I made it conditional the more the actual update code grew. It is probably premature optimization, but the other code was so ugly.

repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
wdschei: So do we wait until all or just the first of the new feeds are registered before unregistering the old feeds?

adrianjgeorge: You're probably going to have to delta, and just ditch the ones that are gone and add the ones that are new.

repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
wdschei: I agree that this is the standard threading data concurrency issue. Should I move away from storing a serialized Scala Vector in favor of a tokenized String that can be more easily Patched?

adrianjgeorge: The problem is in how patch support works in the datastore and the dependency pains we have to go through to make it work the way we want. I can't come up with any other way that is fully thread safe, which is a serious concern with ae tokens. So i'd lean towards extending a scala set adding `with Patchable`

repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
wdschei: This was complicated by the unregistering not wanting the **Feed ID**, rather the **Unique ID** returned from the `registerListener()` call.

wdschei: This is now only removing what is no longer wanted and adding what was not already there. There is also a bunch of tests around this now too just to make sure that new logic was working as expected.

repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -299,6 +304,20 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
       }
     }
 
+    def doUserTokensUpdate(userId: String, authToken: String, ttl: Int): Unit = {
+      // Retrieve the current delimited string value.
+      // IF the token is NOT already there,
+      // THEN append it using a patch.
+      // NOTE: This is at least thread safe, but does not prevent duplication from other threads.
+      val oldTokens = Option(datastore.get(s"$USER_ID_KEY_PREFIX$userId").asInstanceOf[String])
```
adrianjgeorge: This should be `StringValue`

wdschei: Yes, it should. Fixed.

PR #1443, opened 2016-01-19
REP-3259 Flush Output filter wrapper updates

Mario-Lopez:
I removed the old wrapper usage in the filter, and I refactored it away from the old pattern.  I also marked the filter class deprecated.


Mario-Lopez:
The expected 24 tests failed.

PR #1447, opened 2016-01-21
REP-3261 - Header ID Mapping filter

Mario-Lopez:
Updated filter:
1. away from the old filter pattern
2. away from Java
3. away from using the old and busted wrappers

I wrote a bunch more unit tests to verify the behavior is the same as what the old filter was doing.


Mario-Lopez:
retest this please

Mario-Lopez:
The errors are weird.  I ran a few tests locally, and they all passed.  I'm re-running the tests.

Mario-Lopez:
Got the expected 27 failed tests.

repose-aggregator/components/filters/header-id-mapping/src/main/scala/org/openrepose/filters/headeridmapping/HeaderIdMappingFilter.scala
```diff
@@ -0,0 +1,97 @@
...
+
+    filterChain.doFilter(wrappedRequest, servletResponse)
+  }
+
+  private def extractUserGroup(request: HttpServletRequestWrapper, configuredHeader: HttpHeader): Option[UserGroupValues] =
+    getFirstHeaderValue(request.getHeader(configuredHeader.getUserHeader))
+      .map(UserGroupValues(_, getFirstHeaderValue(request.getHeader(configuredHeader.getGroupHeader)), configuredHeader.getQuality))
+
+  // carry-over behavior from the old filter - be sure we only get the first value of the header
+  def getFirstHeaderValue(header: String): Option[String] = Option(header).map(_.split(",", 2)(0).trim).filterNot(_.isEmpty)
```
adrianjgeorge: This is unneccessary. The wrapper already will do splitting, and it maintains ordering. Call `getSplittableHeaders` or `getSplittableHeaderScala`.

Mario-Lopez: Excellent.  I'll do that.

repose-aggregator/components/filters/header-id-mapping/src/main/scala/org/openrepose/filters/headeridmapping/HeaderIdMappingFilter.scala
```diff
@@ -0,0 +1,97 @@
...
+
+    logger.info(s"Initializing Header Identity Mapping filter using config $configurationFile")
+    val xsdUrl: URL = getClass.getResource(SCHEMA_FILE_NAME)
+    configurationService.subscribeTo(filterConfig.getFilterName, configurationFile, xsdUrl, this, classOf[HeaderIdMappingConfig])
+
+    logger.trace("Header Identity Mapping filter initialized.")
+  }
+
+  override def doFilter(servletRequest: ServletRequest, servletResponse: ServletResponse, filterChain: FilterChain): Unit = {
+    val wrappedRequest = new HttpServletRequestWrapper(servletRequest.asInstanceOf[HttpServletRequest])
```
dmnjohns: So with the old pattern, the HandlerFactory would return a null if the isInitialized call on the configuration listeners returned false. The result was the FilterLogicHandlerDelegate returning a 503 indicating that the filter was not yet ready to service requests.

To maintain that behavior, should a check be added to the doFilter method to check if the configuration has been read? And return a 503 if not?

Mario-Lopez: Since it's so common across the filters I thought it would be handled else where.  I thought it was already, but now I see the code that was doing that will be removed:
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/core/core-lib/src/main/java/org/openrepose/core/filter/logic/AbstractConfiguredFilterHandlerFactory.java#L79

So...  I think we should consider putting this in some central place.  Or would the behavior differ between filters?

dmnjohns: That sounds like a good idea. Even before it could vary based on the implementation of the HandlerFactory, which wasn't great.

Design wise, only the UpdateListener will know when it has been passed a configuration. Without changing any interfaces, that means that we probably don't know if a filter has read the configuration until we try sending it a request. Which means that logic being invoked by every filter, since we can't change the Filter interface.

One alternative would be, perhaps, to amend the ConfigurationService to intelligently handle configuration processing. Then we could have some centralized code (say in the PowerFilter) that asks the ConfigurationService is all configurations have been processed. If not, reject with a 503, if so, send the request through the chain.

In the latter case, I would suggest putting that work in a new story. Should we still, then, add the logic to each filter for the time being?

Paging @adrianjgeorge.

Mario-Lopez: Before I continue digging into this, is this a common concern?  The only instance where config isn't eagerly loaded is here:
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/components/filters/translation/src/main/java/org/openrepose/filters/translation/XslUpdateListener.java#L58

Otherwise, I expect the init method of each filter to eagerly load configuration.  The init method being called here:
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/filtercontext/FilterContextFactory.java#L119

Mario-Lopez: Damn it.  I eagerly pressed enter.  Anyway, will requests have access to filters that haven't had "init" called on them?

repose-aggregator/components/filters/header-id-mapping/src/main/scala/org/openrepose/filters/headeridmapping/HeaderIdMappingFilter.scala
```diff
@@ -0,0 +1,98 @@
...
+
+    // find the first configured "user-header" that exists in the request and use that config section to populate the user/group headers
+    configuredHeaders.toStream.flatMap(extractUserGroup(wrappedRequest, _)).headOption.foreach { userGroupValues =>
```
dmnjohns: I like this line. The behavior is fairly complex -- lazily take the first header which can have a user and group extracted, then do something if such a header exists -- but it reads so nicely.

repose-aggregator/components/filters/header-id-mapping/src/main/scala/org/openrepose/filters/headeridmapping/HeaderIdMappingFilter.scala
dmnjohns: Oh, the init method is absolutely called eagerly, as required by the Filter specification. My concern is over the actual reading of the configuration file. For filters that require configuration, processing (probably) can't occur until that configuration has been presented. That configuration may not be presented when the filter is constructed, however. We subscribe to the ConfigurationService which will notify the filter of configuration changes (well, not deletions, but otherwise). Then the filter sets the "initialized" flag on its listener when the filter has finished processing the configuration. Generally, that is when the filter can actually start processing requests. Prior to that, behavior is undefined.

Requests will not have access to filters that haven't had "init" called on them, as far as I know.

dmnjohns:
Looks good to me, potentially pending what we do with the configuration loading stuff.

repose-aggregator/components/filters/header-id-mapping/src/main/scala/org/openrepose/filters/headeridmapping/HeaderIdMappingFilter.scala
```diff
@@ -0,0 +1,104 @@
...
+
+@Named
+class HeaderIdMappingFilter @Inject()(configurationService: ConfigurationService)
+  extends Filter with UpdateListener[HeaderIdMappingConfig] with LazyLogging {
+
+  import HeaderIdMappingFilter._
+
+  private var configurationFile: String = DEFAULT_CONFIG
+  private var initialized = false
+  private var configuredHeaders: Iterable[HttpHeader] = _
```
dmnjohns: Oops. I merged this since it is otherwise good to go, but did you want to update these to be AtomicThings?

dmnjohns: Additional thoughts: do we have a reason to use AtomicThings rather than volatile (or the @volatile annotation in Scala)? I suppose we could also lock with sychronized. If we're only using .get() and .set(...) on AtomicThings, though, those three options should be equivalent, I think?

Also, wouldn't making "initialized" volatile suffice? Assuming we set initialized at the end of the configurationUpdated(...) method. Although it's probably clearer to mark all of the fields which may be mutated and read by multiple threads as volatile. Does doing so hamper performance? I wouldn't think so since it's the same block of code being locked, as it were.

Mario-Lopez: I'm not concerned about adding it to this filter right now.  I added an item to tomorrow's dev meeting to discuss.

repose-aggregator/components/filters/header-id-mapping/src/test/scala/org/openrepose/filters/headeridmapping/HeaderIdMappingFilterTest.scala
```diff
@@ -0,0 +1,268 @@
...
+
+  var filter: HeaderIdMappingFilter = _
+  var servletRequest: MockHttpServletRequest = _
+  var servletResponse: ServletResponse = _
+  var filterChain: FilterChain = _
+
+  before {
+    servletRequest = new MockHttpServletRequest
+    servletResponse = mock(classOf[ServletResponse])
+    filterChain = mock(classOf[FilterChain])
```
dmnjohns: I totally missed it, but you could clean this up a tiny bit by mixing in the MockitoSugar trait.

Mario-Lopez: I keep forgetting about MockitoSugar, and I want to say I've had naming conflicts before with it which is why I don't always use it.  I'll give it a try on the next filter.

PR #1449, opened 2016-01-26
REP-3262 Header Normalization filter wrapper

Mario-Lopez:
Refactored and replaced wrappers.  Also added some fantastic unit tests.


repose-aggregator/components/filters/header-normalization/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
```diff
@@ -0,0 +1,133 @@
...
+  private var configurationFile: String = DEFAULT_CONFIG
+  private var initialized = false
+  private var config: Seq[Target] = _
+  private var metricsMeter: Option[MeterByCategorySum] = _
+
+  override def init(filterConfig: FilterConfig): Unit = {
+    logger.trace("Header Normalization filter initializing...")
+    configurationFile = new FilterConfigHelper(filterConfig).getFilterConfig(DEFAULT_CONFIG)
+
+    logger.info(s"Initializing Header Normalization filter using config $configurationFile")
```
dmnjohns: So while it doesn't matter realistically, I'll just leave this here...

http://stackoverflow.com/questions/20442244/string-interpolation-vs-concatenation

I used to use string interpolation all the time, but after seeing that, I decided it may be better to use it more judiciously. Just another silly thing I spend cycles on.

repose-aggregator/components/filters/header-normalization/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
```diff
@@ -0,0 +1,133 @@
...
+    val xsdUrl: URL = getClass.getResource(SCHEMA_FILE_NAME)
+    configurationService.subscribeTo(filterConfig.getFilterName, configurationFile, xsdUrl, this, classOf[HeaderNormalizationConfig])
+
+    metricsMeter = Option(metricsService) match {
```
dmnjohns: This is curious. Is the metricsService ever null? Isn't it a managed Spring bean?

Mario-Lopez: I know.  It's totally weird.  That's how it used to be, though.  Do you think I should always expect it?

dmnjohns: I wonder if it was that way before because we used to manually build services... prior to the great spring refactor of 2015, it's possible that we didn't build a service unless it was used. As opposed to now, where the beans are always built, even if they are never used.

I think it's probably safe to always expect it, but I'll defer to @adrianjgeorge again.

repose-aggregator/components/filters/header-normalization/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
```diff
@@ -0,0 +1,133 @@
...
+
+    logger.trace("Header Normalization filter initialized.")
+  }
+
+  override def doFilter(servletRequest: ServletRequest, servletResponse: ServletResponse, filterChain: FilterChain): Unit = {
+    val wrappedRequest = new HttpServletRequestWrapper(servletRequest.asInstanceOf[HttpServletRequest])
+
+    config.find { target =>
+      // find the first "target" config element that matches this request (if any)
+      target.url.findFirstIn(wrappedRequest.getRequestURL).isDefined &&
```
dmnjohns: It looks like the filter used to do:
```
target.getPattern().matcher(uri).matches()
```

The semantics of .findFirstIn for a Java regex are not the same as the semantics of Java's Pattern .matches(). The former will check if any substring of the input matches the regex and return the first such match, while the latter will require that the full input matches the input.

Example:
```
scala> "hello".r.findFirstIn("hello world!")
res0: Option[String] = Some(hello)

scala> "hello".r.pattern.matcher("hello world!").matches()
res1: Boolean = false
```

I'd recommend changing this line to:
```
target.url.pattern.matcher(wrappedRequest.getRequestURL).matches() &&
```

repose-aggregator/components/filters/header-normalization/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
```diff
@@ -0,0 +1,133 @@
...
+        None
+    }
+
+    logger.trace("Header Normalization filter initialized.")
+  }
+
+  override def doFilter(servletRequest: ServletRequest, servletResponse: ServletResponse, filterChain: FilterChain): Unit = {
+    val wrappedRequest = new HttpServletRequestWrapper(servletRequest.asInstanceOf[HttpServletRequest])
+
+    config.find { target =>
```
dmnjohns: I'd probably replace the ".find" with " find" here. That is, replace the period with a space.

See the Arity-1 section of:
http://docs.scala-lang.org/style/method-invocation.html

repose-aggregator/components/filters/header-normalization/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
```diff
@@ -0,0 +1,133 @@
...
+      // find the first "target" config element that matches this request (if any)
+      target.url.findFirstIn(wrappedRequest.getRequestURL).isDefined &&
+        (target.methods.contains(wrappedRequest.getMethod) || target.methods.contains(AllHttpMethods))
+    }.foreach { target =>
+      // figure out which headers to remove, and remove them
+      (target.access match {
+        case WhiteList => wrappedRequest.getHeaderNamesScala.diff(target.headers)
+          // written like this to maintain the case-insensitive string comparisons - do not swap
+        case BlackList => target.headers.intersect(wrappedRequest.getHeaderNamesScala)
```
dmnjohns: Ah. So this is where that question yesterday came from.

I don't love that we switch the order going from .diff to .intersect. Stupid Scala collections...

repose-aggregator/components/filters/header-normalization/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
adrianjgeorge: For simple uses i find interpolation way easier to read than concatenation.

repose-aggregator/components/filters/header-normalization/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
```diff
@@ -0,0 +1,133 @@
...
+      case None =>
+        logger.warn("Metrics meter not started for Header Normalization filter due to missing metrics service")
+        None
+    }
+
+    logger.trace("Header Normalization filter initialized.")
+  }
+
+  override def doFilter(servletRequest: ServletRequest, servletResponse: ServletResponse, filterChain: FilterChain): Unit = {
+    val wrappedRequest = new HttpServletRequestWrapper(servletRequest.asInstanceOf[HttpServletRequest])
```
dmnjohns: I guess I'll point out the if (!isInitialized) thing again, just for the sake of completeness.

repose-aggregator/components/filters/header-normalization/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
adrianjgeorge: Yeah, if it's an autowired spring bean, it's never not there, because we won't be able to start the spring context.

repose-aggregator/components/filters/header-normalization/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
Mario-Lopez: Interesting.  I'm hoping it's a safe bet to use interpolation as the defacto method of concatenating strings that should continuously see improvements in terms of speed.  Just like how concatenating strings have been changed in the background to be faster in later versions of the JVM.

adrianjgeorge: the problem is that java doesn't yet have a good string interpolation solution so they are unlikely to target that for optimization.

Mario-Lopez: I'm hoping the Scala devs compile it down to something awesome, not the Java folks.

adrianjgeorge: also the second guy saw conflicting results
It's a pretty big non concern for me.

dmnjohns: I'm hoping Java eventually natively supports interpolation. At any rate, it's interesting to see what Java code is generated for interpolation.

dmnjohns: Looks like I was late to the party.

Yeah, I really don't have a concern, I just wanted to point that out. It's useful information if you ever need to do a bunch of string templating stuff.

repose-aggregator/components/filters/header-normalization/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
```diff
@@ -0,0 +1,133 @@
...
+    logger.trace("Header Normalization filter destroyed.")
+  }
+
+  override def configurationUpdated(config: HeaderNormalizationConfig): Unit = {
+    this.config = config.getHeaderFilters.getTarget.asScala.map { target =>
+      val access = if (target.getBlacklist.isEmpty) WhiteList else BlackList
+      val headers = (access match {
+        case WhiteList => target.getWhitelist
+        case BlackList => target.getBlacklist
+      }).get(0).getHeader.asScala.map(_.getId).toSet
```
dmnjohns: This code makes me think our XSD is a mess. I know you didn't touch it as part of this story, but...

```
<xs:choice>
            <xs:element name="whitelist" type="hnorm:HttpHeaderList" minOccurs="0" maxOccurs="unbounded"/>
            <xs:element name="blacklist" type="hnorm:HttpHeaderList" minOccurs="0" maxOccurs="unbounded"/>
        </xs:choice>
```
This is weird. You have to choose a white list or a black list, but then you can have multiple lists? Why? I would think that if you define a target, you should also have to defined exactly one whitelist OR blacklist. Which is what the code here assumes.

dmnjohns: Side note: JavaConversions provides implicit Java <-> Scala conversions much like JavaConverters does explicitly. It saves you having to type .asScala all the time. This is another thing that doesn't particularly bother me, but knowing about both can be useful. As I've begun to trust implicits more, I've started using JavaConversions myself.

Mario-Lopez: OMG Yes.  When I first saw this jank I was like WTF!?

I can only guess they mistakenly thought the minOccurs and maxOccurs applied to its children (which I've seen someone suggest as much on StackOverflow).  I was seriously tempted to fix this, but meh.

dmnjohns: Well, if you happen to fix it, I might happen to pull it in... =P

Mario-Lopez: ain't nobody got time for that

repose-aggregator/components/filters/header-normalization/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
Mario-Lopez: I almost forgot about this.  I pushed the change now.

repose-aggregator/components/filters/header-normalization/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
Mario-Lopez: It's quite annoying.  That's why I put the note about not swapping the order because I know someone down the line will want to.

PR #1450, opened 2016-01-26
REP-3265 IP User Filter Wrapper

dmnjohns:


repose-aggregator/components/filters/ip-user/src/main/scala/org/openrepose/filters/ipuser/IpUserFilter.scala
```diff
@@ -28,23 +28,22 @@ import javax.servlet.http.{HttpServletRequest, HttpServletResponse}
...
+  with LazyLogging
+  with UpdateListener[IpUserConfig] {
 
   private final val DEFAULT_CONFIG = "ip-user.cfg.xml"
-  private var initialized = false
-  private var configName: String = _
-
-  case class LabeledCIDR(label: String, cidr: CIDRUtils)
 
   private val cidrList: AtomicReference[List[LabeledCIDR]] = new AtomicReference[List[LabeledCIDR]]()
```
Mario-Lopez: Unrelated to your changes...  I've been wondering if we should be doing this in the other filters since we're updating a shared object (cidrList in this case) without worrying about the object being half built when another thread tries to use it.

dmnjohns: I also just noticed that @dkowis did this in a number of places. It seems like a good practice to adopt.

repose-aggregator/components/filters/ip-user/src/main/scala/org/openrepose/filters/ipuser/IpUserFilter.scala
```diff
@@ -72,7 +71,7 @@ with UpdateListener[IpUserConfig] {
       servletResponse.asInstanceOf[HttpServletResponse].sendError(500)
     } else {
       logger.trace("IP Classification filter handling request...")
```
Mario-Lopez: Also unrelated to your changes, but at some point we should fix the log messages to use the new filter name.

repose-aggregator/components/filters/ip-user/src/main/scala/org/openrepose/filters/ipuser/IpUserFilter.scala
```diff
@@ -72,7 +71,7 @@ with UpdateListener[IpUserConfig] {
...
-      val request = MutableHttpServletRequest.wrap(servletRequest.asInstanceOf[HttpServletRequest])
+      val request = new HttpServletRequestWrapper(servletRequest.asInstanceOf[HttpServletRequest])
 
       getClassificationLabel(servletRequest.getRemoteAddr).foreach { label =>
         request.addHeader(groupHeaderName, s"$label;q=$groupHeaderQuality")
```
Mario-Lopez: I'd rather us use the addHeader method where you pass along the quality instead of using string interpolation to do it in this class.

repose-aggregator/components/filters/ip-user/src/main/scala/org/openrepose/filters/ipuser/IpUserFilter.scala
dmnjohns: I'll go ahead and change those lines now, because why not?

repose-aggregator/components/filters/ip-user/src/main/scala/org/openrepose/filters/ipuser/IpUserFilter.scala
dmnjohns: Man, I totally missed that. I'll update it, of course.

Mario-Lopez:
Looks good; waiting on build to finish.

Mario-Lopez:
retest this please

PR #1453, opened 2016-01-28
REP-3270 Rackspace Auth User filter wrapper updates

Mario-Lopez:
Updated filter:
1. away from the old filter pattern
2. away from using the old and busted wrappers

I kept the handler class and updated its interface away from the old pattern since it had a lot of logic and more importantly unit tests.  I moved a bit of the logic into the Filter class since I thought it made more sense there.

The design structure isn't the way I would have written it from scratch, but I was able to make relatively few changes to accomplish our goals.


wdschei:
The code changes look good, but it seems to have failed 20+ functional tests. We'll see what @jennyvo has to say during her review. Maybe they are just needed updates to the tests, but I don't think they are.

jennyvo:
will take a look those tests failed

jennyvo:
wdschei, Mario-Lopez, I checked test failed for real, the filter is used to extract username and group to x-pp-user and x-pp-groups from auth user payload to rate limiting purposes

wdschei:
That is what I was afraid of at first glance.

jennyvo:
debugged through seeing that mutableHttpRequest added header but username as key and group name as value

repose-aggregator/components/filters/rackspace-auth-user/src/main/scala/org/openrepose/filters/rackspaceauthuser/RackspaceAuthUserFilter.scala
```diff
@@ -19,40 +19,74 @@
...
+    }
+
+    val httpServletRequest = servletRequest.asInstanceOf[HttpServletRequest]
+    if (httpServletRequest.getMethod != "POST") {
+      // this filter only operates on POST requests that have a body to parse
+      filterChain.doFilter(servletRequest, servletResponse)
+    } else {
+      val wrappedRequest = new HttpServletRequestWrapper(httpServletRequest)
+      handler.get.parseUserGroupFromInputStream(wrappedRequest.getInputStream, wrappedRequest.getContentType) foreach { rackspaceAuthUserGroup =>
+        wrappedRequest.addHeader(rackspaceAuthUserGroup.user, rackspaceAuthUserGroup.group, rackspaceAuthUserGroup.quality)
```
jennyvo: added header but username as name and group name as value?

Mario-Lopez: D'oh!  Thanks for finding this!  I pushed a fix.

Mario-Lopez:
Correct number of functional test failures now.

wdschei:
Yea. Only 27 tests failed!!! Something just doesn't seem right about saying that.

PR #1454, opened 2016-02-01
REP-3269 Identity v3 filter refactored and replaced wrappers

dmnjohns:


repose-aggregator/components/filters/openstack-identity-v3/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3Handler.scala
```diff
@@ -26,225 +26,149 @@ import com.rackspace.httpdelegation.HttpDelegationManager
...
+      }
+      token.get.rax_impersonator.foreach { impersonator =>
+        impersonator.id.foreach(wrappedHttpRequest.replaceHeader(OpenStackIdentityV3Headers.X_IMPERSONATOR_ID.toString, _))
+        impersonator.name.foreach(wrappedHttpRequest.replaceHeader(OpenStackIdentityV3Headers.X_IMPERSONATOR_NAME.toString, _))
+      }
+      if (filterConfig.fullConfig.isForwardCatalog) {
+        token.get.catalog.foreach(catalog => wrappedHttpRequest.replaceHeader(PowerApiHeader.X_CATALOG.toString, base64Encode(catalog.toJson.compactPrint)))
+      }
+      if (filterConfig.fullConfig.isForwardGroups) {
+        userGroups.foreach(group => wrappedHttpRequest.addHeader(PowerApiHeader.GROUPS.toString, group + ";q=1.0"))
```
Mario-Lopez: Quality being set through the header value.

repose-aggregator/components/filters/openstack-identity-v3/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3Handler.scala
```diff
@@ -26,225 +26,149 @@ import com.rackspace.httpdelegation.HttpDelegationManager
...
+    // Attempt to validate the request token with the Identity service
+    val token = authenticate(wrappedHttpRequest, tracingHeader) match {
+      case Success(tokenObject) =>
+        Some(tokenObject)
+      case Failure(e: InvalidSubjectTokenException) =>
+        filterFailure = Some(FilterFailure(HttpServletResponse.SC_UNAUTHORIZED, e.getMessage))
+        None
+      case Failure(e: IdentityServiceOverLimitException) =>
+        filterFailure = Some(FilterFailure(HttpServletResponse.SC_SERVICE_UNAVAILABLE, e.getMessage))
+        authServiceOverLimit(e, httpResponse)
```
Mario-Lopez: Will this impact delegation?  Previously this was only called if delegation was turned off.

repose-aggregator/components/filters/openstack-identity-v3/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3Handler.scala
```diff
@@ -26,225 +26,149 @@ import com.rackspace.httpdelegation.HttpDelegationManager
...
-        retry = new HttpDate(retryCalendar.getTime).toRFC1123
+    // Attempt to fetch groups if configured to do so
+    val userGroups = if (filterFailure.isEmpty && filterConfig.fullConfig.isForwardGroups) {
+      token.get.user.id map { userId =>
+        identityApi.getGroups(userId, tracingHeader) match {
+          case Success(groupsList) =>
+            groupsList.map(_.name)
+          case Failure(e: IdentityServiceOverLimitException) =>
+            filterFailure = Some(FilterFailure(HttpServletResponse.SC_SERVICE_UNAVAILABLE, e.getMessage))
+            authServiceOverLimit(e, httpResponse)
```
Mario-Lopez: Same here about previously only running if delegation was off.

repose-aggregator/components/filters/openstack-identity-v3/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3Handler.scala
```diff
@@ -26,225 +26,149 @@ import com.rackspace.httpdelegation.HttpDelegationManager
...
+            List.empty[String]
+          case Failure(e) =>
+            filterFailure = Some(FilterFailure(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, e.getMessage))
+            logger.error(e.getMessage)
```
Mario-Lopez: I've always regretted not passing the logger the exception so it could print the stack trace, especially since we're not rethrowing the exception for some other code to deal with.

dmnjohns:
I think I've come up with a better way to handle this filter. My brain just wasn't playing well with the mutability before, which is both good and bad. Anyway, that put me in a weird spot.

I'll probably rework this all tonight if you want to hold off on reviewing it until tomorrow. :smile:

repose-aggregator/components/filters/openstack-identity-v3/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3Handler.scala
dmnjohns: In some cases where Exceptions are expected (i.e., they are used as to control the flow) I understand only printing the messages and not the full stack trace.

Since this is a catch-all, however, I agree that printing the full stack trace is probably the right call.

Mario-Lopez:
Sounds good.  I'll put this on hold for tomorrow.

dmnjohns:
Alright, it's ready for another go!

Mario-Lopez:
This was much easier to follow.  Looks legit.  Waiting on build to succeed.

dmnjohns:
Yeah, I was making things way more difficult for myself by trying to come up with a way to remove some but not all of the mutability. It was a misguided first attempt, and the result was not satisfactory.

I blame it on a case of the Mondays. Now I'm going to forget it ever happened and move on with my life.

repose-aggregator/components/filters/openstack-identity-v3/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3Handler.scala
```diff
@@ -59,43 +57,43 @@ class OpenStackIdentityV3Handler(identityConfig: OpenstackIdentityV3Config, iden
...
     if (isUriWhitelisted(request.getRequestURI, identityConfig.getWhiteList)) {
       logger.debug("Request URI matches a configured whitelist pattern! Allowing request to pass through.")
-      filterDirector.setFilterAction(FilterAction.PASS)
+      filterAction = FilterAction.PASS
     } else {
-      val requestHeaderManager = filterDirector.requestHeaderManager
-
       // Set the default behavior for this filter
-      filterDirector.setFilterAction(FilterAction.RETURN)
-      filterDirector.setResponseStatusCode(HttpServletResponse.SC_INTERNAL_SERVER_ERROR)
```
dmnjohns: Found another pattern that will break things. Setting these "default" values without overriding them. Specifically, if the response code is >= 500 when it reaches the end of the filter chain, the code that actually routes the request will simply return it instead.

So here's what changed. The FilterLogicHandlerDelegate would not apply the change to the response status until after calling the handleResponse method. In other words, it would delay applying the status to the underlying response until the last moment. With the new wrappers, that change is applied immediately.

dmnjohns: As far as a solution goes, we can:
1. Leave the default there (to save LOC, perhaps at the cost of clarity) and then reset the status to a 200 if no failures occur
2. We could stop setting the default behavior, and explicitly set a 500 and return on any failure that warrants doing so.
3. We could pull the logic in the PowerFilter which prevents a request from being routed if the response is >= 500.

dmnjohns: For reference, see:
https://github.com/rackerlabs/repose/blob/REP-2188_SpikeToUpdateToNewWrappers/repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/PowerFilterChain.java#L259

Mario-Lopez:
Build failure was spectacular.  Probably needs to be rebased onto the 2188 branch.

PR #1455, opened 2016-02-02
REP-3271 Rackspace Basic Auth Filter Refactored

dmnjohns:


repose-aggregator/components/filters/rackspace-identity-basic-auth/src/main/scala/org/openrepose/filters/rackspaceidentitybasicauth/RackspaceIdentityBasicAuthFilter.scala
```diff
@@ -231,47 +244,43 @@ class RackspaceIdentityBasicAuthFilter @Inject()(configurationService: Configura
...
         case failure: TokenCreationInfo =>
           processFailedToken(failure, encodedCredentials)
       }
     }
 
     // We need to process the Response unless a couple of specific conditions occur.
-    filterDirector.setFilterAction(FilterAction.PROCESS_RESPONSE)
-    if (!httpServletRequest.getHeaderNames.asScala.toList.contains(X_AUTH_TOKEN)) {
-      withEncodedCredentials(httpServletRequest) { encodedCredentials =>
+    if (!httpServletRequestWrapper.getHeaderNames.asScala.toList.contains(X_AUTH_TOKEN)) {
```
Mario-Lopez: Probably worth using getHeaderNamesScala to avoid converting back and forth.

repose-aggregator/components/filters/rackspace-identity-basic-auth/src/main/scala/org/openrepose/filters/rackspaceidentitybasicauth/RackspaceIdentityBasicAuthFilter.scala
```diff
@@ -100,19 +98,34 @@ class RackspaceIdentityBasicAuthFilter @Inject()(configurationService: Configura
   }
 
   override def doFilter(servletRequest: ServletRequest, servletResponse: ServletResponse, filterChain: FilterChain) {
-    new FilterLogicHandlerDelegate(servletRequest, servletResponse, filterChain).doFilter(this)
+    val wrappedHttpRequest = new HttpServletRequestWrapper(servletRequest.asInstanceOf[HttpServletRequest])
+    val httpResponse = servletResponse.asInstanceOf[HttpServletResponse]
+
+    val filterAction = handleRequest(wrappedHttpRequest, httpResponse)
+    filterAction match {
```
Mario-Lopez: FilterAction.PASS and .NOT_SET are never used.

dmnjohns: That's true. I added all of them for the sake of completeness, and to maintain the behavior using the FilterLogicHandlerDelegate would have provided. Well, except for NOT_SET, which is no longer applicable. I'm actually not sure if the Scala compiler requires all values of an enum to be represented in a match as it does with sealed traits.

repose-aggregator/components/filters/rackspace-identity-basic-auth/src/main/scala/org/openrepose/filters/rackspaceidentitybasicauth/RackspaceIdentityBasicAuthFilter.scala
dmnjohns: Legit.

repose-aggregator/components/filters/rackspace-identity-basic-auth/src/main/scala/org/openrepose/filters/rackspaceidentitybasicauth/RackspaceIdentityBasicAuthFilter.scala
Mario-Lopez: That's fine.  I guess I expected something _like_ the FilterFailure case class used in the v3 filter to be returned from handleRequest.  That way there's either a failure or it's good to go as opposed to 4 possible states, 2 of which aren't possible.  Up to you.

dmnjohns: So nope, the Scala compiler does not require an exhaustive match.

If we leave those actions off, and they are used in the future without this code being update, the error the user would see would be something like:
```
scala.MatchError: NOT_SET (of class org.openrepose.commons.utils.servlet.filter.FilterAction)
	at org.openrepose.filters.openstackidentityv3.A$A8$A$A8.get$$instance$$res0(Test.sc0.tmp:5)
	at #worksheet#.#worksheet#(Test.sc0.tmp:18)
```

If we include them here, they won't see an error for PASS, and for NOT_SET they will see "Unexpected internal filter state".

In both cases, a 500 response will be sent for the associated request.

Thoughts/opinion?

Mario-Lopez: When I first saw the pattern matching I wondered in what situations you would PASS and when it would ever be NOT_SET, so I think there's some confusion there.  It also feels wrong to leave them out and risk a possible run time error in the event someone tries to use PASS and it's not handled.  The third option is to use something other than FilterAction, but then that's additional effort for something that's not broken.

dmnjohns: In a general sense, PASS is the most common action. Of course, it isn't used in this filter.

NOT_SET would have been the default on the FilterDirector, which was the return type from these methods in the old pattern. I guess the idea was that you had to be explicit about the FilterAction. Since that code is now in the filters themselves, it will always be explicit, so we probably don't need NOT_SET anymore.

It's true that FilterAction is an enumeration that lends itself to the old pattern. In fact, that's the reason I've kept using it -- it prevents me from having to refactor all of the handleRequest(...) and handleResponse(...) methods. Even so, if we're not getting rid of mutability this pass, I don't have any real desire to get rid of FilterAction either.

I'll also note that I used @wdschei FilterAction in commons utils rather than the one in core.

So here are a couple of options going forward:
1. Do nothing.
2. Leave code comments indicating that PASS and NOT_SET are not expected to be used. Note that these comments will have to be maintained with the code, which sometimes doesn't happen since the compile doesn't check comments.
3. Drop NOT_SET from FilterAction. I don't know if this is actually possible for the moment, since other code might rely on it. Also, we still don't use PASS in this filter, so that may need to be addressed separately.
4. Only react to expected FilterActions by using an if/else if/else block. This is basically just a match, though, so why?
5. Combine PASS and NOT_SET into a catch-all in the match (i.e., case _ => ...). This seems to indicate that only the other FilterActions are expected, but if the unexpected happens, do something. I'm not sure what the something that we should be. Log and fail with a 500 (or just throw an exception)?

Mario-Lopez: I vote for the catch-all.  I think that would make it clear what we expect to handle while still handling other cases with a 500 just in case.

dmnjohns: Fair enough. I'll push it momentarily.

wdschei: The original `FilterAction` in `org.openrepose.core.filter.logic` will be going away with the `FilterDirector`. The new one I added to `org.openrepose.commons.utils.servlet.filter` allows us to do this transition fairly cleanly. The original behavior in `FilterDirector` for the `NOT_SET` action was to throw a 500 if no filter had changed the value. I left it in the new one strictly for ease of conversion, but I agree that it should probably be tossed. The remaining three options:

* `PASS`
* `PROCESS_RESPONSE`
* `RETURN`

provide a way to route the logic internal to the filter if needed. I'm have no strong opinion either way on using the new `FilterAction` moving forward or just using local definitions other than to say it helps make them consistent if we use a common convention. What that convention is, makes no difference to me. I would just like to be able to look at a new to me filter and be able to easily follow the flow. To me, that was the one nice thing the `FilterAction` provided and that is why I brought it along.

dmnjohns: I figured that was the case. I am not opposed to using that FilterAction and dropping NOT_SET. Perhaps a future story to do that should be created?

wdschei: Story added: https://repose.atlassian.net/browse/REP-3455

Mario-Lopez:
Looks legit.  Waiting on build to succeed.

PR #1457, opened 2016-02-02
REP-3283 Deleted the deprecated Auth-N and Auth-Z filters.

wdschei:


wdschei:
That last round of test updates brings us back to the baseline 27 failures.

PR #1458, opened 2016-02-03
REP-3278 Valkyrie filter wrapper replacement

dmnjohns:


Mario-Lopez:
Build failure was spectacular.  Probably needs to be rebased onto the 2188 branch.

dmnjohns:
A bit of investigation on the failing test where the response body of a HEAD request is non-empty. It's not actually an issue with the new wrappers, Repose core, or the Valkyrie filter updates:

So here is the overall flow in the Mutable wrapper:
.getOutputStream() (called by the new wrapper)
->
.popOutputStream() (called by the PowerFilterChain)
->
.commitBufferToServletOutputStream() (called by the PowerFilter)
The result is the empty internal buffer created by the first call being written as the body of the response in the last call. Even though there is no real data to write, hence the 1024 -1s.

This will be fixed once we drop the old wrappers entirely. I expect the issue with intrafilter logging will also be fixed for the same reason.

PR #1459, opened 2016-02-08
REP-3269 Identity v3 filter refactored and replaced wrappers

dmnjohns:
Tracking repose-8-staging.

For old comments, see:
#1454


Mario-Lopez:
I didn't expect to see this failure:
`features.filters.addheader.AddHeaderTest.Verify miss config add header - no header`

Is that expected now?

PR #1460, opened 2016-02-08
REP-3278 Valkyrie filter wrapper replacement

dmnjohns:
Tracking repose-8-staging.

For old comments, see:
#1458


Mario-Lopez:
I didn't expect to see this failure:
`features.filters.addheader.AddHeaderTest.Verify miss config add header - no header`

Is that expected now?

dmnjohns:
Yes. That is what @jennyvo was checking late last week. It's the 500 uninitialized status code we're getting now rather than the 503 we used to get using the FilterDirector. Before the main branch is merged, we'll have to fix that.

PR #1462, opened 2016-02-09
REP-3272 Rate Limiting wrappers updates

Mario-Lopez:
I updated the Rate Limiting filter to move away from using the old pattern.  I moved the HandlerFactory code to the Filter class, and I created an internal class in the Filter in order to atomically hold the current config state like AbstractConfiguredFilterHandlerFactory used to essentially do for us.  I updated the Handler class to return a FilterAction to the Filter class which minimized the number of changes required to move away from the old pattern.

I made a few more changes to move away from the old wrappers and get rid of all of the many FilterDirector side-effects.  Stuff was crazy.

There were a couple of exceptions (CacheException and DatastoreOperationException) in the Handler that didn't seem to be properly handled and could potentially result in a request making it to the origin service despite the error state, so I fixed that.


Mario-Lopez:
Tons of unrelated failures.  Rebased and restarted tests.

repose-aggregator/commons/commons-utilities/src/main/java/org/openrepose/commons/utils/http/media/MimeType.java
```diff
@@ -66,11 +66,13 @@ public static MimeType guessMediaTypeFromString(String mimeType) {
         if (StringUtilities.isNotBlank(mimeType)) {
 
             for (MimeType ct : values()) {
+                // worst case scenario this will match on UNKNOWN because everything contains "" (an empty string)
                 if (mimeType.contains(ct.getType()) && mimeType.contains(ct.getSubType())) {
                     return ct;
                 }
             }
 
+            // this is unreachable code, and at this point I'm too afraid to fix it
```
dmnjohns: lol

repose-aggregator/components/filters/rate-limiting-filter/src/main/java/org/openrepose/filters/ratelimiting/RateLimitingFilter.java
```diff
@@ -56,22 +80,127 @@ public RateLimitingFilter(
...
+                LOG.info("Using requested datastore {}", requestedDatastore);
+                return datastoreService.getDefaultDatastore();
+            }
+
+            Datastore datastore = datastoreService.getDatastore(requestedDatastore);
+            if (datastore != null) {
+                LOG.info("Using requested datastore {}", requestedDatastore);
+                return datastore;
+            } else {
+                LOG.warn("Requested datastore not found");
```
dmnjohns: Is this what the behavior was before? If the request datastore does not exist, use the distributed datastore? I would have thought that the default datastore would be used instead.

Mario-Lopez: Yeah, this method is copy/pasta as I didn't want to mess with it.

https://github.com/rackerlabs/repose/blob/master/repose-aggregator/components/filters/rate-limiting/src/main/java/org/openrepose/filters/ratelimiting/RateLimitingHandlerFactory.java#L70

dmnjohns: Weird, but alrighty then.

repose-aggregator/components/filters/rate-limiting-filter/src/test/java/org/openrepose/filters/ratelimiting/RateLimitingHandlerTest.java
```diff
@@ -242,11 +194,12 @@ public void beforeAny() throws Exception {
...
+            mockedResponse = mock(HttpServletResponseWrapper.class);
+        }
 
-            mockedRequest = mock(HttpServletRequest.class);
-            mockedResponse = mock(ReadableHttpServletResponse.class);
+        public RateLimitingHandler newHandler() {
+            return RateLimitingTestSupport.createHandler(defaultRateLimitingConfiguration(), eventService, datastore);
         }
 
     }
```
dmnjohns: All of the test cleanup here is wonderful.

repose-aggregator/components/filters/rate-limiting-filter/pom.xml
```diff
@@ -130,6 +130,20 @@
...
+            <artifactId>jcl-over-slf4j</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <!-- MockHttpServletRequest/Response -->
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-test</artifactId>
+            <scope>test</scope>
+        </dependency>
```
dmnjohns: Do these need to be added to a Gradle build file?

dmnjohns:
Looks good to me. I will merge once testing is completed.

repose-aggregator/components/filters/rate-limiting-filter/pom.xml
Mario-Lopez: Oh, yeah.  The gradle build wasn't a thing when I initially worked this story.  Since the build isn't working at the moment, I'm not sure if I should add the lines I expect would work without testing it or hold off and let the gradle work take care of it.

PR #1463, opened 2016-02-09
REP-3273 SLF4JHttpLogging filter conversion and refactor

dmnjohns:


dmnjohns:
I have to say, converting this filter to Scala really cleaned things up. At least, in my opinion.

dmnjohns:
Those failures. Looking into it.

repose-aggregator/components/filters/slf4j-http-logging/src/main/scala/org/openrepose/filters/slf4jlogging/Slf4jLoggerWrapper.scala
```diff
@@ -17,32 +17,11 @@
...
-
-    public String getFormatString() {
-        return formatString;
-    }
-
-    public Logger getLogger() {
-        return logger;
-    }
-}
+case class Slf4jLoggerWrapper(logger: Logger, formatString: String) {
```
adrianjgeorge: Doesn't seem like it's worth it to keep this as a separate class at this point.

dmnjohns: You mean as it's own file? I can totally move it.

PR #1464, opened 2016-02-09
REP-3212 Handling 404 from groups call for Rackers

dmnjohns:
There are some decisions I'm not sure about, but the best way to get feedback is probably to open a pull request, so here we go!

### DIFF
--- a/repose-aggregator/components/filters/keystone-v2/src/main/resources/META-INF/schema/config/keystone-v2.xsd
+++ b/repose-aggregator/components/filters/keystone-v2/src/main/resources/META-INF/schema/config/keystone-v2.xsd
@@ -108,6 +108,13 @@
                 </xs:documentation>
             </xs:annotation>
         </xs:attribute>
+        <xs:attribute name="set-rackers-groups" type="xs:string" use="optional">
+            <xs:annotation>
+                <xs:documentation>
+                    <html:p>For users with a "racker" role, set the x-pp-groups header to some value</html:p>
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
         <xs:attribute name="set-catalog-in-header" type="xs:boolean" use="optional" default="false">
             <xs:annotation>
                 <xs:documentation>
--- a/repose-aggregator/components/filters/keystone-v2/src/main/resources/META-INF/schema/config/keystone-v2.xsd
+++ b/repose-aggregator/components/filters/keystone-v2/src/main/resources/META-INF/schema/config/keystone-v2.xsd
@@ -63,10 +63,6 @@
     </xs:complexType>
 
     <xs:complexType name="IdentityServiceType">
-        <xs:sequence>
-            <xs:element name="allow-racker-group-fail" type="AllowRackerGroupFailType" minOccurs="0" maxOccurs="1"/>
-        </xs:sequence>
-
         <xs:attribute name="username" type="xs:string" use="optional">
             <xs:annotation>
                 <xs:documentation>
@@ -124,31 +120,6 @@
                    test="if (@username or @password) then (@username and @password) else true()"
                    xerces:message="Must provide both a username and a password"
                    saxon:message="Must provide both a username and a password"/>
-    </xs:complexType>
-
-    <xs:complexType name="AllowRackerGroupFailType">
-        <xs:annotation>
-            <xs:documentation>
-                <html:p>
-                    If included, users with the "Racker" role will continue past the groups call.
-                    <html:br/>
-                    If excluded, the groups call will fail for users with the "Racker" role.
-                </html:p>
-            </xs:documentation>
-        </xs:annotation>
-
-        <xs:attribute name="set-group" type="xs:string" use="optional">
-            <xs:annotation>
-                <xs:documentation>
-                    <html:p>
-                        In included, the X-PP-Groups header will be populated with the value of this
-                        attribute.
-                        <html:br/>
-                        If excluded, the X-PP-Groups header will not be populated.
-                    </html:p>
-                </xs:documentation>
-            </xs:annotation>
-        </xs:attribute>
     </xs:complexType>
 
     <xs:complexType name="TenantHandlingType">
--- a/repose-aggregator/components/filters/keystone-v2/src/main/resources/META-INF/schema/config/keystone-v2.xsd
+++ b/repose-aggregator/components/filters/keystone-v2/src/main/resources/META-INF/schema/config/keystone-v2.xsd
@@ -112,13 +112,6 @@
                 </xs:documentation>
             </xs:annotation>
         </xs:attribute>
-        <xs:attribute name="set-rackers-groups" type="xs:string" use="optional">
-            <xs:annotation>
-                <xs:documentation>
-                    <html:p>For users with a "racker" role, set the x-pp-groups header to some value</html:p>
-                </xs:documentation>
-            </xs:annotation>
-        </xs:attribute>
         <xs:attribute name="set-catalog-in-header" type="xs:boolean" use="optional" default="false">
             <xs:annotation>
                 <xs:documentation>
--- a/repose-aggregator/components/filters/keystone-v2/src/main/resources/META-INF/schema/config/keystone-v2.xsd
+++ b/repose-aggregator/components/filters/keystone-v2/src/main/resources/META-INF/schema/config/keystone-v2.xsd
@@ -63,6 +63,10 @@
     </xs:complexType>
 
     <xs:complexType name="IdentityServiceType">
+        <xs:sequence>
+            <xs:element name="allow-racker-group-fail" type="AllowRackerGroupFailType" minOccurs="0" maxOccurs="1"/>
+        </xs:sequence>
+
         <xs:attribute name="username" type="xs:string" use="optional">
             <xs:annotation>
                 <xs:documentation>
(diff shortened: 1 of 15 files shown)


repose-aggregator/components/filters/keystone-v2/src/main/resources/META-INF/schema/config/keystone-v2.xsd
```diff
@@ -108,6 +108,13 @@
                 </xs:documentation>
             </xs:annotation>
         </xs:attribute>
+        <xs:attribute name="set-rackers-groups" type="xs:string" use="optional">
```
dmnjohns: I don't love this being another attribute side-by-side with the "set-groups-in-header" attribute. Also, I'm not sure about the name.

dmnjohns: Also, there's still the question about what we want to do with regards to translating roles to groups. If we were to do that translation, there may not be a need for a configuration group value here.

repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -409,6 +409,8 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
                   case Success(newAdminToken) => requestHandler.getGroups(newAdminToken, validToken.userId)
                   case Failure(x) => Failure(IdentityAdminTokenException("Unable to reacquire admin token", x))
                 }
+              case _: InvalidTokenException if Option(keystoneV2Config.getIdentityService.getSetRackersGroups).isDefined && validToken.roles.exists(_.equalsIgnoreCase("racker")) =>
```
dmnjohns: I made this a case-insensitive check, which seems reasonable, but worth pointing out. Actually, I could see it either way.

To be clear, roles are case sensitive, and "Racker" seems to be the role for Rackers in Rackspace Identity.

repose-aggregator/components/filters/keystone-v2/src/main/resources/META-INF/schema/config/keystone-v2.xsd
adrianjgeorge: you've tied two things together here, which i'm not certain should be. i assume this attribute is what determines whether to let the groups call fail for rackers, but you are also using it for a value to put in. These are separate items. Only sort of though, it's an option on an option, which suggests it should be an element with an attribute. also your right the name is bad.

repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
adrianjgeorge: Are we 100% on it being 'Racker'? If so i agree checking against that specifically is reasonable, but I have concerns over the consistency of identity's data.

adrianjgeorge:
Random test fail?

dmnjohns:
Single test failure. Interestingly, it's in a Valkyrie test that apparently also uses keystone-v2. Still, it looks like the test shouldn't be affected by my changes. I'll look into it when I get a chance, but I'll roll the dice and retest it for now.

dmnjohns:
Retest this please.

repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
dmnjohns: If you have concerns, we may as well leave it case-insensitive. It's configurable behavior specific to Rackspace, and I assume Rackspace will never create a "racker" role AND a "Racker" role that are distinct. If so, oh man, Identity and I are going to get into a fight.

repose-aggregator/components/filters/keystone-v2/src/main/resources/META-INF/schema/config/keystone-v2.xsd
dmnjohns: I don't name things well. It is known.

Yes, this attribute enables/disables the feature, and, in this case, forces the user to provide a value for x-pp-groups.

I can split this into enable/disable the feature and set a value/don't set a value. Easy enough. I wasn't sure we wanted to, though. Then we have three loosely-coupled attributes which determine the desired behavior.

adrianjgeorge: you can add an element and keep backwards compatibility no problem, just set it to minimum uses of 0

repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
adrianjgeorge: lol. it's entirely possible that it's not actually dynamic data, but statically added to anyone it has to do an ldap lookup to populate. In the end we are pretty safe just doing an insensitive check in this one case.

repose-aggregator/components/filters/keystone-v2/src/main/resources/META-INF/schema/config/keystone-v2.xsd
dmnjohns: Yes, well I would want to pull the set-groups attribute into that new element. Which means either moving it, which isn't backwards compatible, or having multiple ways to configure it, which is ugly and a pain.

jamesyli:
It looks like the initial implementation was to add a `set-rackers-groups` config option, but somehow it got removed in the newer commits. Some review comments seem to suggest using a translation of `x-roles`. However I have a couple of opinions on this:

(1) `x-roles` header is usually a *huge* list, especially for rackers. User probably does not want to translate all of them into `x-pp-groups`, which ends up with writing/changing `header-transform.xsl`, and it is a pain; By contrast `set-rackers-groups` is much simpler to use and less error prone;

(2) Translating `x-roles` to `x-pp-groups` looks to me conceptually incorrect. Roles defines who has the permission but Groups defines the service level. Doing any translation between the two makes people (at least me)  confused.

My 2 cents.

jennyvo:
I looked into 11 tests failed, they were all related to the way we handled getGroups response if we are moving forward with this change, I need to update these tests to accept current responses.

adrianjgeorge:
From my expereince the only rackers that have large numbers of groups are QEs who have been given many so that they can do integration type testing. We had problems settling on a configuration, and we finally realized it was because we were trying to do something that we really shouldn't. That solution was adding a large of amount of very specific business logic to a filter where it didn't belong. The solution as we've done it now doesn't involve that fiddly logic and gives us better integration into generic keystone not the rackspace specific implementation. We still try to make the rackspace extension call, but we aren't depend on it's success. The solution of translating roles to groups came from the architect of both identity and repose, this suggests that there isn't as big of a conceptual gap between the two as you might think. As has been mentioned if you don't want all the roles, you can use the translation filter. It does require a bit of XML knowledge to use though. In the future we plan on building a scripting filter, that allows a number of scripting languages to be used to write tiny filters without going through the whole process of writing a custom filter. The work related to that feature will be tracked (here)[https://repose.atlassian.net/browse/REP-2226]. If that interests you contact our product manager @karidavis and let her know so that it's priority can be potentially adjusted.

repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -409,6 +409,8 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
                   case Success(newAdminToken) => requestHandler.getGroups(newAdminToken, validToken.userId)
                   case Failure(x) => Failure(IdentityAdminTokenException("Unable to reacquire admin token", x))
                 }
+              case _: InvalidTokenException =>
```
adrianjgeorge: Might want to chnge the exception you are throwing, because as we figured out in this case a 404 is not telling us that the token is invalid, just that they lack groups.

dmnjohns: Yeah, I thought about that when I saw it, but apparently forgot to actually change it.

Problematically, that means that the generic function which was being used to handle the responses may not work for this particular call.

dmnjohns: Actually, I can just change it to "NotFoundException" and change some wording to be a little more generic.

repose-aggregator/components/filters/keystone-v2/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -409,6 +409,8 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
                   case Success(newAdminToken) => requestHandler.getGroups(newAdminToken, validToken.userId)
                   case Failure(x) => Failure(IdentityAdminTokenException("Unable to reacquire admin token", x))
                 }
+              case _: he newxception =>
```
jennyvo: look like a typo

jamesyli:
Glad to see this one got merged, thanks guys!

PR #1465, opened 2016-02-12
REP-3504 Rate Limiting default HTTP Methods fix.

wdschei:
I also brought in the minor consistency update to the Valkyrie Authorization filter so it would behave exactly like the Rate Limiting filter in the future.


dmnjohns:
Looks good :+1:

PR #1466, opened 2016-02-15
REP-3274 - URI User wrapper updates

Mario-Lopez:
Rewrote filter in Scala using the new wrappers.  Added some nice unit tests.

Will also update the name from "URI Identity" to "URI User" in a future commit.


repose-aggregator/components/filters/uri-user-filter/build.gradle
```diff
@@ -1,12 +1,15 @@
...
+    compile "org.scala-lang:scala-library"
+    compile "com.typesafe.scala-logging:scala-logging-slf4j_2.10"
 
     provided "javax.servlet:javax.servlet-api"
     provided "javax.inject:javax.inject"
 
     testCompile "junit:junit"
+    testCompile "org.scalatest:scalatest_2.10"
     testCompile "org.mockito:mockito-all"
+    testCompile "org.springframework:spring-test"
```
dmnjohns: Huh, does Gradle prevent the need for the following artifact somehow?

<artifactId>jcl-over-slf4j</artifactId>

repose-aggregator/components/filters/uri-user-filter/src/main/scala/org/openrepose/filters/uriuser/UriUserFilter.scala
```diff
@@ -0,0 +1,105 @@
...
+  override def isInitialized: Boolean = initialized
+}
+
+object UriUserFilter {
+  private final val DefaultConfigFileName = "uri-user.cfg.xml"
+  private final val SchemaFileName = "/META-INF/schema/config/uri-user-configuration.xsd"
+
+  // while the XSD specifies these defaults, the generated getters don't provide these values for some reason
+  private final val DefaultQuality = 0.5d
+  private final val DefaultGroup = "User_Standard"
```
dmnjohns: I keep forgetting that the Scala convention is to capitalize the first letter of each word for constants. If you happen to review my code in the future and I've capitalized every letter, please correct me! =)

repose-aggregator/components/filters/uri-user-filter/build.gradle
Mario-Lopez: No idea.  It's copy/pasta from another file that seemed to work.  I had not realized it didn't have the jcl-over-slf4j in there.

@adrianjgeorge ?

repose-aggregator/components/filters/uri-user-filter/src/test/scala/org/openrepose/filters/uriuser/UriUserFilterTest.scala
```diff
@@ -0,0 +1,207 @@
...
+
+  var filter: UriUserFilter = _
+  var request: MockHttpServletRequest = _
+  var response: ServletResponse = _
+  var filterChain: FilterChain = _
+
+  before {
+    request = new MockHttpServletRequest
+    response = mock[ServletResponse]
+    filterChain = mock[FilterChain]
```
dmnjohns: Would we want to use Spring's MockHttpServletResponse and MockFilterChain here instead? It's more consistent, and you don't have to use an ArgumentCaptor. On the other hand, it may be considered more "heavy-weight" than is necessary.

I don't really mind either way, but thought I would call attention to it nonetheless.

dmnjohns:
Apart from the Gradle dependency, this looks good to me. I'll merge after that is settled and testing is completed.

repose-aggregator/components/filters/uri-user-filter/build.gradle
dmnjohns: Any update on this?

dmnjohns:
Looks like we're at 32 failures right now. Are the 2 new failures expected, or are we still waiting on updates?

repose-aggregator/components/filters/uri-user-filter/build.gradle
adrianjgeorge: It's only needed if you are looking to parse the logs that you would get from jcl.

adrianjgeorge: Or if you want the further output for debugging, but we have no hard dependency on it

dmnjohns: Right. It exists in the Maven build currently, so I assumed we actually needed it for something. I suppose if the tests run in Gradle, we should be good.

adrianjgeorge: it hangs out several places, that aren't strictly necessary. I suspect it was copy pasta in the maven builds. it was not needed in the gradle build at the time I converted it, but things might have changed since then. Or somebody was having a problem and decided to include it to get better output from their tests for debugging porpoises

Mario-Lopez:
40 test failures.  Not sure if that's the expected amount these days.

dmnjohns:
Kind of. Not really. We've seen those across other branches, but something got into repose-8-staging that broke tests which were previously passing.

I may look at addressing those this morning.

PR #1467, opened 2016-02-16
Wrapper changes

dmnjohns:
- URI/URL manipulation support
- Query parameter manipulation support

### DIFF
--- a/repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
+++ b/repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
@@ -46,7 +46,8 @@ class HttpServletRequestWrapper(originalRequest: HttpServletRequest, inputStream
 
   private var requestUri: String = originalRequest.getRequestURI
   private var requestUrl: StringBuffer = originalRequest.getRequestURL
-  private var queryParameterMap: Option[ListMap[String, Array[String]]] = None
+  private var queryString: String = originalRequest.getQueryString
+  private var parameterMap: Option[ListMap[String, Array[String]]] = None
   private var headerMap: Map[String, List[String]] = new TreeMap[String, List[String]]()(caseInsensitiveOrdering)
   private var removedHeaders: Set[String] = new TreeSet[String]()(caseInsensitiveOrdering)
 
@@ -169,48 +170,48 @@ class HttpServletRequestWrapper(originalRequest: HttpServletRequest, inputStream
   /**
     * @return a string representation of the query parameters for this request
     */
-  override def getQueryString: String = {
-    queryParameterMap match {
-      case Some(parameterMap) if parameterMap.isEmpty =>
-        null
-      case Some(parameterMap) =>
-        parameterMap map { case (key, values) =>
-          values.map(value => key + "=" + value).mkString("&")
-        } mkString "&"
-      case None =>
-        super.getQueryString
-    }
+  override def getQueryString: String = queryString
+
+  /**
+    * @param string the desired query string for this request
+    */
+  def setQueryString(string: String): Unit = {
+    // todo: update parameter map
+    // parse the current query string and remove from parameter map
+    // parse the new query string and add to parameter map
+    queryString = string
   }
 
   /**
-    * @param key a query parameter key
-    * @return the first query parameter value associated with the provided key for this request, or null if no value exists
+    * @param key a parameter key
+    * @return the first parameter value associated with the provided key for this request, or null if no value exists
     */
   override def getParameter(key: String): String =
     Option(getParameterValues(key)).map(_.head).orNull
+
   /**
-    * @param key a query parameter key
-    * @return all query parameter values associated with the provided key for this request
+    * @param key a parameter key
+    * @return all parameter values associated with the provided key for this request
     */
   override def getParameterValues(key: String): Array[String] =
-    queryParameterMap.map(_.get(key).orNull).getOrElse(super.getParameterValues(key))
+    parameterMap.map(_.get(key).orNull).getOrElse(super.getParameterValues(key))
 
   /**
-    * @return all query parameter names for this request
+    * @return all parameter names for this request
     */
   override def getParameterNames: util.Enumeration[String] =
-    queryParameterMap.map(_.keysIterator.asJavaEnumeration).getOrElse(super.getParameterNames)
+    parameterMap.map(_.keysIterator.asJavaEnumeration).getOrElse(super.getParameterNames)
 
   /**
-    * @return the query parameter map for this request
+    * @return the parameter map for this request
     */
   override def getParameterMap: util.Map[String, Array[String]] =
-    queryParameterMap.map(_.asJava).getOrElse(super.getParameterMap)
+    parameterMap.map(_.asJava).getOrElse(super.getParameterMap)
 
-  /** Sets the query parameter map for this request.
+  /** Sets the parameter map for this request.
     *
     * The provided map parameter will have its contents copied into an immutable map.
-    * As a result, modifications to the map parameter after calling this method will have no effect on the query
+    * As a result, modifications to the map parameter after calling this method will have no effect on the
     * parameters of this request.
     * The iteration order of the provided map will be maintained.
     *
@@ -224,6 +225,6 @@ class HttpServletRequestWrapper(originalRequest: HttpServletRequest, inputStream
       val arrayCopy = util.Arrays.copyOf(entry.getValue, entry.getValue.length)
       mapCopy += (entry.getKey -> arrayCopy)
     }
-    queryParameterMap = Option(ListMap(mapCopy.toSeq: _*))
+    parameterMap = Option(ListMap(mapCopy.toSeq: _*))
   }
 }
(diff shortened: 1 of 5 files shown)


dmnjohns:
We should probably verify that the behavior for setting the URI/URL is correct. Should they be handled separately? Or should they be tied together?

dmnjohns:
I may have made a bad assumption on the query parameter work.

It looks like getParameterMap returns not only query parameters, but also body parameters. If that's the case, this implementation doesn't allow for separating the two.

So I propose we allow for getting and setting the query string. When the query string is set, the parameter map must also be updated. We may or may not want to retain the ability to set the parameter map. Since changes to the parameter map cannot be assumed to be changes to the query string, we could leave modification of the request to input stream manipulation (maybe).

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
```diff
@@ -145,4 +150,81 @@ class HttpServletRequestWrapper(originalRequest: HttpServletRequest, inputStream
...
+    */
+  override def getQueryString: String = queryString
+
+  /**
+    * @param string the desired query string for this request
+    */
+  def setQueryString(string: String): Unit = {
+    // todo: update parameter map
+    // parse the current query string and remove from parameter map
+    // parse the new query string and add to parameter map
```
dmnjohns: Is this something we actually want to do? Since the parameter map can be set independently, keeping it in sync with the query string map not be possible.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
```diff
@@ -145,4 +150,81 @@ class HttpServletRequestWrapper(originalRequest: HttpServletRequest, inputStream
...
+    * @return all parameter names for this request
+    */
+  override def getParameterNames: util.Enumeration[String] =
+    parameterMap.map(_.keysIterator.asJavaEnumeration).getOrElse(super.getParameterNames)
+
+  /**
+    * @return the parameter map for this request
+    */
+  override def getParameterMap: util.Map[String, Array[String]] =
+    parameterMap.map(_.asJava).getOrElse(super.getParameterMap)
```
dmnjohns: This is probably broken. If setParameterMap(...) has not been called, the underlying mutable parameter map is return. It can be mutated, and those mutations will persist. This can be done until setParameterMap(...) is called and the map is made immutable.

One option would be to copy over the mutable map into the immutable map upon construction. That could be costly, and honestly, mutability is kind of baked into the servlet spec.

So instead, we could just allow the parameter map to be mutable. Originally, I thought that the parameter map and query string had the same contents, but that turns out not to be the case. If it were the case, a mutable parameter map would enable de-synchronization between the parameter map and the query string. If we wanted to allow the parameter map to be mutable, I don't believe we would have to override any of the parameter methods (since, again, the underlying map is mutable).

If we take that mutable parameter map approach, what about the query string? Modifications to the parameter map should not be reflected in the query string, since the parameter map is comprised of more than just query parameters. To change the query string, a user should have to use the setQueryString(...) method. However, since the parameter map does include the query string parameters, the setQueryString(...) method should update the parameter map.

dmnjohns: Apparently I missed that getParameterMap(...) returns an unmodifiable map, as I discovered last week.

dmnjohns: Just a reference:

http://docs.oracle.com/javaee/7/api/javax/servlet/ServletRequest.html#getParameterMap--

adrianjgeorge:
Man, i knew about the form posts, and completely forgot

adrianjgeorge:
I'm curious what the current system does with a form post, especially one where the url in question also had query parameters.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
```diff
@@ -145,4 +151,71 @@ class HttpServletRequestWrapper(originalRequest: HttpServletRequest, inputStream
...
+    if (Option(getRequestURI).exists(_.nonEmpty)) {
+      url.append(getRequestURI)
+    }
+
+    url
+  }
+
+  override def getRequestURI: String = requestUri
+
+  def setRequestURI(uri: String): Unit = {
```
dmnjohns: I am going to assume, for now, that the following methods are smart enough to call getRequestURI/getRequestURL and reflect the changes:
getContextPath
getPathInfo
getPathTranslated
getServletPath

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
```diff
@@ -145,4 +151,71 @@ class HttpServletRequestWrapper(originalRequest: HttpServletRequest, inputStream
   def getSplittableHeaderScala(headerName: String): List[String] = getHeadersScala(headerName).foldLeft(List.empty[String])((list, s) => list ++ s.split(","))
 
   override def getSplittableHeaders(headerName: String): util.List[String] = getSplittableHeaderScala(headerName).asJava
+
+  /**
+    * @return a [[StringBuffer]] containing the reconstructed URL for this request (note that
+    *         mutation of this [[StringBuffer]] will have no effect on the request URL)
```
dmnjohns: Thought I'd point out this comment. The JavaDoc states:
"Because this method returns a StringBuffer, not a string, you can modify the URL easily, for example, to append query parameters."

However, that seems altogether like a bad idea. I'm not sure why the contract specifies a StringBuffer, to be honest.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
dmnjohns: Hm, Jetty and Glassfish both build the URL using the getRequestUri() method. So we can probably get away with not implementing that.

dmnjohns:
Test failures are caused by the MockRunner mock request being silly. Calling setRequestURL(...) on that object is not reflect in getRequestURI(). The wrapper changes I've made here rely on getRequesrURI to build the request URL, and so, breakage.

dmnjohns:
The current failing tests seem unrelated to these changes. 

The AuthHerpDerp tests appear to be failing because a reponse-messaging configuration is being provided but probably shouldn't be.

The StartupErrorLog test is failing because the Repose version was bumped to 8.

The Repose Lint tests appear to be failing because the artifact is the updated code while the test is not. It looks like all of the commits made it onto this branch, though, so that may have something to do with moving the project recently?

The rest of the failures appear to be the expected failures.

Too many parts moving on the repose-8-branch.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
adrianjgeorge: so you are explicitly breaking the contract here? we can't do that. people down stream don't know that they are getting one of our wrapper, and not a request that is written to spec.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
```diff
@@ -145,4 +144,120 @@ class HttpServletRequestWrapper(originalRequest: HttpServletRequest, inputStream
...
+      parameterMap.toMap
+    }
+
+    val updatedParameterMap = mutable.Map.empty[String, Array[String]]
+    val curQueryMap = Option(getQueryString).map(parseQueryString).getOrElse(Map.empty[String, Array[String]])
+    val newQueryMap = Option(newQueryString).map(parseQueryString).getOrElse(Map.empty[String, Array[String]])
+
+    // Remove all current query parameters from the parameter map
+    getParameterMap.asScala foreach { case (key, values) =>
+      val formValues = values.filterNot(value => curQueryMap.get(key).exists(_.contains(value)))
```
adrianjgeorge: this is overly aggressive. it'll remove the form value as well as the query value if they have the same value.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
adrianjgeorge: wait it doesn't return them? that's super weird. forget it then. mark it as a known oddity

adrianjgeorge:
looks fine, we'll let the tests finish out then pull it in

dmnjohns:
retest this please

Mario-Lopez:
Build only had 30 failures, so I'm going to merge this since Adrian said it looks fine.

PR #1470, opened 2016-02-18
REP-3281 - Simple RBAC wrapper updates

Mario-Lopez:
I removed the old wrapper usage in Simple RBAC.  I don't see why they were wrapped in the first place, so I didn't replace them with the new wrappers.  I also removed the core-lib dependency and had to add the jcl-over-slf4j dependency for the tests.

The gradle build worked without any updates.


Mario-Lopez:
Got the expected 30 test failures.

dmnjohns:
Nice! This one was super clean. I love it.

Will merge after QE is happy with it.

wdschei:
That's because I wrote it right the first time. ;-)

PR #1473, opened 2016-02-19
do not merge: Deletes are fun

adrianjgeorge:
this is so i can get a full functional test run with my deleteions


adrianjgeorge:
le sigh, apparently we use most of this stuff

PR #1474, opened 2016-02-19
REP-3167 - Tracing Header updates

Mario-Lopez:
I updated the tracing header config in the system model from being a pair of attributes to an element with those two attributes.  I also updated code to use the new generated source.

Previously, the tracing header attribute was set to true by default (i.e. if it wasn't explicitly configured, it would be turned on).  With these updates, if the element is not specified, it will default to being turned on.  If the element is specified, but the attribute is not, it will still default to true.  As before, the only way to turn it off is to specify the attribute with a value of false.  I was going to make the attribute required (because if you already have the element then why not?), but I kept it optional in case all you wanted to do was change the rewrite behavior.

The rewrite attribute pretty much acts the same way except it's in this new element now.

### DIFF
--- a/repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
+++ b/repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
@@ -604,7 +604,7 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
     private var initialized = false
 
     override def configurationUpdated(configurationObject: SystemModel): Unit = {
-      sendTraceHeader = configurationObject.getTracingHeader == null || configurationObject.getTracingHeader.isEnabled
+      sendTraceHeader = Option(configurationObject.getTracingHeader).isEmpty || configurationObject.getTracingHeader.isEnabled
       initialized = true
     }
 
--- a/repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
+++ b/repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
@@ -604,7 +604,7 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
     private var initialized = false
 
     override def configurationUpdated(configurationObject: SystemModel): Unit = {
-      sendTraceHeader = configurationObject.isTracingHeader
+      sendTraceHeader = configurationObject.getTracingHeader == null || configurationObject.getTracingHeader.isEnabled
       initialized = true
     }
 
--- a/repose-aggregator/functional-tests/spock-functional-test/src/test/configs/features/core/tracing/rewritetransid/system-model.cfg.xml
+++ b/repose-aggregator/functional-tests/spock-functional-test/src/test/configs/features/core/tracing/rewritetransid/system-model.cfg.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
-<system-model xmlns="http://docs.openrepose.org/repose/system-model/v2.0" rewrite-tracing-header="true">
+<system-model xmlns="http://docs.openrepose.org/repose/system-model/v2.0">
     <repose-cluster id="repose-cluster">
         <nodes>
             <node id="node" hostname="localhost" http-port="${reposePort}"/>
@@ -13,4 +13,5 @@
                       default="true"/>
         </destinations>
     </repose-cluster>
+    <tracing-header rewrite-header="true"/>
 </system-model>
--- a/repose-aggregator/components/filters/keystone-v2-filter/src/test/scala/org/openrepose/filters/keystonev2/KeystoneV2FilterPrepTest.scala
+++ b/repose-aggregator/components/filters/keystone-v2-filter/src/test/scala/org/openrepose/filters/keystonev2/KeystoneV2FilterPrepTest.scala
@@ -32,7 +32,7 @@ import org.mockito.{ArgumentCaptor, Matchers => MockitoMatcher, Mockito}
 import org.openrepose.core.services.config.ConfigurationService
 import org.openrepose.core.services.datastore.{Datastore, DatastoreService}
 import org.openrepose.core.services.serviceclient.akka.{AkkaServiceClient, AkkaServiceClientFactory}
-import org.openrepose.core.systemmodel.SystemModel
+import org.openrepose.core.systemmodel.{TracingHeaderConfig, SystemModel}
 import org.openrepose.filters.keystonev2.config.KeystoneV2Config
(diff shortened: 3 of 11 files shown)


repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -604,7 +604,7 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
     private var initialized = false
 
     override def configurationUpdated(configurationObject: SystemModel): Unit = {
-      sendTraceHeader = configurationObject.isTracingHeader
+      sendTraceHeader = configurationObject.getTracingHeader == null || configurationObject.getTracingHeader.isEnabled
```
dmnjohns: I know it's irrational, but just seeing "null" makes me angry. I propose using `Option(getTracingHeader).isEmpty`

Mario-Lopez: Sure, why not.  I'll make that change.

repose-aggregator/functional-tests/spock-functional-test/src/test/configs/features/core/tracing/rewritetransid/system-model.cfg.xml
```diff
@@ -13,4 +13,5 @@
                       default="true"/>
         </destinations>
     </repose-cluster>
+    <tracing-header rewrite-header="true"/>
```
dmnjohns: I know this is contrary to what we decided for the story, but I may as well voice my concern.

I worry that the new behavior around the `tracing-header` element will confuse users. The line highlighted above shows why -- the `enabled` attribute is not shown be default. So if I were trying to turn the tracing heading off, my first thought might be that I can do so by removing the element. A reasonable thought, since that is the behavior in other configs.

Now that's all well and good, but what can be done about it? Well, I don't have a solution, but I do have an idea. Why not let the presence dictate whether or not to add the `tracing-header`? If we want the `tracing-header` to be added by default, that's fine, we ship a config which includes the element. The user still has to explicitly do *something* to disable the `tracing-header`, so in practical terms, we've set a default value. We also remove an attribute and keep the behavior for the element consistent with other elements in other configs.

Oh, I'd also consider renaming the element to `add-tracing-header` or `set-tracing-header` if the `enabled` attribute were to be removed.

Just my :moneybag: (there is two cents in there)
:two: :hankey: (two scents?)
:two: :email: (two sent)
:v: :nose: (two scents again?)
:trollface: (that's apparently a thing)

Mario-Lopez: I initially had it require the "enabled" attribute to make it clearer but took it out at the last minute.  If I made that required, then the options would be:

`--nothing--` : add tracing header, no rewrite
`<tracing-header enabled="true" />` : add tracing header, no rewrite
`<tracing-header enabled="false" />` : don't add tracing header
`<tracing-header enabled="true" rewrite-header="true" />` : add tracing header, rewrite if needed

Which would be a bit better than being in the situation where "`<tracing-header enabled="true" />`", "`<tracing-header />`" and nothing mean the same thing.

The other option is we also just make tracing-header a required element.  Explicit is better, right?  I'm not sure how I feel about making that a necessary change when updating to Repose 8, but I suppose this would be the time to do it.

I think putting a verb in the element name would be inconsistent.

dmnjohns: Even requiring the `enabled` attribute doesn't help when the element is not present.

Making the `tracing-header` element required seems like a good option, although it might require a change for some users as you point out.

As for the verb, do you mean inconsistent in that we currently use an attribute named `tracing-header`? If so, that's true, but since it's a major version anyway I'd prioritize making this configuration as clear as possible going forward.

Mario-Lopez: Inconsistent with the other elements in `system-model` which are all nouns.  Not that that's a big deal at all, only my reasoning for going with `tracing-header` as the name of the element.  I'm fine either way. :sunglasses: 

I totally see your reasoning, though.  The three scenarios we want are:
- No tracing header
- Tracing header with no rewrite (default)
- Tracing header with rewrite

Ideally, I would expect them configured like this respectively:
- (nothing)
- `<tracing-header />`
- `<tracing-header rewrite="true" />`

The only problem is that default behavior, tho.  Something they were getting for nothing previously would now require an extra element.  I wonder if anyone would not even realize the connection between the header and that configuration item.  Even worse, downstream Repose instances could be impacted.  I'm just guessing as to why we might want to go with our original plan in the story.

@adrianjgeorge can we get a verdict, plz? :hammer: (why is there no gavel emoti?)

dmnjohns: Gotcha.

Yeah, that's why I would also be totally okay with making the new element required and keeping the enabled attribute. That way there is no confusion.

I still don't love the original solution, though.

Anyway, that's @adrianjgeorge call to make. :stuck_out_tongue:

adrianjgeorge: Making it mandatory doesn't fix the problem, because it's a change in a config that we don't replace on upgrade.

Our usage breaks down like this:
97% Want the header on
3% Want rewrite
0% Want it turned off (I haven't heard of anybody not wanting the header, but we give the option, because there is always one weirdo out there)

I don't want to make people have configuration for something that they want all the damn time, it's the opposite of convention over configuration which is where the industry is moving. We are still giving them the option of being explicit if they want, but why specify something that is always going to be the same. It's only if you need to change it's default behavior that you need to specify it and interact with it.

make sense?

adrianjgeorge: So along with this we have to be clear in our documentation that this behavior is on by default, and that the only way to turn it off is to set it so explicitly.

Mario-Lopez: Sounds good to me.  I'll update the null check @dmnjohns pointed out and call it a day.

dmnjohns: I guess that's the thing. If we aim for convention over configuration, then our documentation needs to be very clear.

Your explanation makes sense, but it feels like we're fighting the configuration system. In this case, I suppose it doesn't matter too much since, as you pointed out, virtually no user is disabling this feature. For a feature that is configured more often, though, would you make the same decision?

Anyway, I'm not going to lose sleep at night because this configuration doesn't give me warm fuzzies, so leaving it as written sounds like what we want to do.

dmnjohns:
retest this please

PR #1475, opened 2016-02-22
Removing dependencies on mocks-util module

Mario-Lopez:
I updated three spots that were using the mocks-util module.  Two were easy to replace.  The third involved using XmlSlurper to do some unmarshalling.  There's still a dependency in mocks-servlet on mocks-util, but that's going away I hear.  Of note, EmbeddedTomcatProxyTest uses the WAR generated by mocks-servlet, so additional updates will be needed in EmbeddedTomcatProxyTest before mocks-servlet and thus mocks-util can be removed.


Mario-Lopez:
Work was done under sub-task REP-3568.

PR #1482, opened 2016-03-01
Adding a Gradle wrapper

dmnjohns:
The version is easy enough to update in the future. I think the `gradle wrapper` task handles everything, actually (and you can specify a version, or it will use the version the task was run with). Or you can modify the `gradle-wrapper.properties` file manually, but if the Gradle wrapper was changed between versions that won't be reflected.

Another option would be to add a wrapper task to our root build file. Then we could specify the version there, like:
```
task wrapper(type: Wrapper) {
    gradleVersion = '2.0'
}
```


adrianjgeorge:
i already did this on my branch, and you've selected the wrong version

PR #1485, opened 2016-03-02
REP-3317 - Defaulted TLS Renegotiation to false

Mario-Lopez:
I updated the XSD to default to false (which is just for show), and I updated the code to specifically set it to disabled (false) if the config element is null (since behind the covers the default is true).

### DIFF
--- a/repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
+++ b/repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
@@ -128,7 +128,7 @@ class ReposeJettyServer(val clusterId: String,
           }
         }
 
-        cf.setRenegotiationAllowed(Option(ssl.isTlsRenegotiationAllowed).getOrElse[java.lang.Boolean](java.lang.Boolean.FALSE))
+        cf.setRenegotiationAllowed(Option(ssl.isTlsRenegotiationAllowed).getOrElse(false).asInstanceOf[Boolean])
 
         sslConnector
       } getOrElse {
--- a/repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
+++ b/repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
@@ -128,9 +128,7 @@ class ReposeJettyServer(val clusterId: String,
           }
         }
 
-        Option(ssl.isTlsRenegotiationAllowed).foreach { tls =>
-          cf.setRenegotiationAllowed(tls)
-        }
+        cf.setRenegotiationAllowed(Option(ssl.isTlsRenegotiationAllowed).getOrElse[java.lang.Boolean](false))
 
         sslConnector
       } getOrElse {
--- a/repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
+++ b/repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
@@ -128,7 +128,7 @@ class ReposeJettyServer(val clusterId: String,
           }
         }
 
-        cf.setRenegotiationAllowed(Option(ssl.isTlsRenegotiationAllowed).getOrElse[java.lang.Boolean](false))
+        cf.setRenegotiationAllowed(Option(ssl.isTlsRenegotiationAllowed).getOrElse[java.lang.Boolean](java.lang.Boolean.FALSE))
 
         sslConnector
       } getOrElse {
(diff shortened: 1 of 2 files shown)


repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
```diff
@@ -128,9 +128,7 @@ class ReposeJettyServer(val clusterId: String,
           }
         }
 
-        Option(ssl.isTlsRenegotiationAllowed).foreach { tls =>
-          cf.setRenegotiationAllowed(tls)
-        }
+        cf.setRenegotiationAllowed(Option(ssl.isTlsRenegotiationAllowed).getOrElse[java.lang.Boolean](false))
```
dmnjohns: Ew, switching from boxing to a Scala Boolean to boxing to a Java Boolean. The type system differences are super apparent in this line. Could you use the Java `Boolean.FALSE` constant instead (https://docs.oracle.com/javase/7/docs/api/java/lang/Boolean.html#FALSE)?

I never realized that XJC boxes booleans.

dmnjohns:
Looks good apart from any prior comments!

repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
Mario-Lopez: Sure.  I thought the line looked ugly that way, but why not.

Of note, I tried:
```
Option(ssl.isTlsRenegotiationAllowed).getOrElse(java.lang.Boolean.FALSE)
```
but the build failed with:
```
[ERROR]  found   : Any
[ERROR]  required: Boolean
[ERROR]         cf.setRenegotiationAllowed(Option(ssl.isTlsRenegotiationAllowed).getOrElse(java.lang.Boolean.FALSE))
[ERROR]                                                                                   ^
[ERROR] one error found
```
which doesn't make a lot of sense to me.

I considered for a moment importing java.lang.Boolean to avoid writing out the whole thing twice, but I didn't want to risk messing something else up that depended on the scala boolean.

repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
```diff
@@ -128,9 +128,7 @@ class ReposeJettyServer(val clusterId: String,
           }
         }
 
-        Option(ssl.isTlsRenegotiationAllowed).foreach { tls =>
-          cf.setRenegotiationAllowed(tls)
-        }
+        cf.setRenegotiationAllowed(Option(ssl.isTlsRenegotiationAllowed).getOrElse[java.lang.Boolean](java.lang.Boolean.FALSE))
```
dmnjohns: Using java.lang.Boolean.FALSE, you should be able to drop the explicit type on `getOrElse`.

Mario-Lopez: That's what I thought, too, but the maven build fails with that.

Mario-Lopez: I tried a few things, and this is what I found:

```
Works:
cf.setRenegotiationAllowed(Option(ssl.isTlsRenegotiationAllowed).getOrElse(java.lang.Boolean.FALSE).asInstanceOf[Boolean])
cf.setRenegotiationAllowed(Option(ssl.isTlsRenegotiationAllowed).getOrElse[java.lang.Boolean](java.lang.Boolean.FALSE))
cf.setRenegotiationAllowed(Option(ssl.isTlsRenegotiationAllowed).getOrElse[java.lang.Boolean](false))
cf.setRenegotiationAllowed(Option(ssl.isTlsRenegotiationAllowed).map(Boolean.unbox).getOrElse(false))
cf.setRenegotiationAllowed(Option(ssl.isTlsRenegotiationAllowed).exists(Boolean.unbox))

Does not work:
cf.setRenegotiationAllowed(Option(ssl.isTlsRenegotiationAllowed).getOrElse(java.lang.Boolean.FALSE))
cf.setRenegotiationAllowed(Option(ssl.isTlsRenegotiationAllowed).getOrElse(false))
cf.setRenegotiationAllowed(Option(ssl.isTlsRenegotiationAllowed).getOrElse[scala.Boolean](false))
```

dmnjohns: Weird.

After seeing what you tried, I tried something myself. It worked.

```
cf.setRenegotiationAllowed(Option(ssl.isTlsRenegotiationAllowed).getOrElse(java.lang.Boolean.FALSE).booleanValue())
```

So the Option(java.lang.Boolean).getOrElse(java.lang.Boolean) works and returns a java.lang.Boolean. But `setRenegotiationAllowed` takes a `boolean`, the Java primitive. So Scala doesn't know how to convert `java.lang.Boolean` to boolean, I guess. Scala must do some magic to make a `java.lang.Boolean` into a Scala `Boolean` which provides the `booleanValue` method.

So yeah, pick your favorite. I really don't care about much, but this was an interesting discovery!

dmnjohns:
The test failures are all known, and should now be fixed. I don't want to wait for another test cycle to complete, so I'm pulling this in.

PR #1486, opened 2016-03-03
Updated to the latest version of the gradle-jaxb-plugin which supports parallel project execution.

wdschei:
There were a few tests that appear to not work with the parallel builds that were unrelated to the JAXB issues, so I temporarily `ignored` them for the time being.


wdschei:
This is no longer needed and will be kept in the history for the sake of identifying these tests.

PR #1487, opened 2016-03-03
REP-3577 - Configure request headers in HTTP connection pools

Mario-Lopez:
I updated the HTTP Connection Pool configuration to allow a list of headers to be configured for a connection pool.  I also updated the HttpConnectionPoolProvider to set the configured headers as default headers in the HttpClient.

I added a functional test (with the v2v2 filter configured to use a connection pool with a configured header) to verify that the call to identity contained the configured header and that the origin service (which used the default connection pool) did not receive said header.

### DIFF
--- a/build-tools/pom.xml
+++ b/build-tools/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.openrepose</groupId>
         <artifactId>repose</artifactId>
-        <version>8.0.0.0-SNAPSHOT</version>
+        <version>7.3.3.0-SNAPSHOT</version>
     </parent>
 
     <name>Repose - Build Tools</name>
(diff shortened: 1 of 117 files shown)


wdschei:
Looks good. Tests pass. I will wait and see what @jennyvo adds and take another look.

adrianjgeorge:
I like the functional test, but why no unit test?

Mario-Lopez:
I wasn't sure HttpClient was going to do what I wanted, so I wrote the functional test to verify stuff.  At that point it felt weird to write a unit test to confirm that I populated a parameter, but I totally can if you want.

adrianjgeorge:
If it's realtively easy do it, otherwise don't worry about it.

Mario-Lopez:
No problem at all.  Unit tests added.

PR #1488, opened 2016-03-04
REP-3276 - URI Stripper wrapper updates

Mario-Lopez:
I rewrote the URI Stripper filter in Scala moving it away from the old handler pattern and old wrappers.  I also spruced up the unit tests a bit.


dmnjohns:
This code really demonstrates some of the power of Options.

Anyway, everything looks good to me. Merging. :+1:

PR #1494, opened 2016-03-10
Reviving the header-identity filter

dmnjohns:
Maybe.

This might work.


dmnjohns:
A better way of doing this would probably have been to revert the previous commits, and then do an interactive rebase on those.

Mario-Lopez:
I'm taking care of this in this pull request:  https://github.com/rackerlabs/repose/pull/1498

PR #1495, opened 2016-03-11
Java 8

Mario-Lopez:
Updated pom.xml to target Java 8.


wdschei:
This is already part of [REP-3166](https://github.com/rackerlabs/repose/blob/REP-3166_UpgradeScala2-10to2-11/pom.xml#L82-L92).

PR #1496, opened 2016-03-15
Fix URI Normalization functional test

Mario-Lopez:
This functional test is inconsistently failing:
UriNormalizationFilterTest.When no media-variants config in the filter should not throw NPE

The test is specifically looking for this string in the logs:
UriNormalizationConfig@(.*)[mediaVariants=<null>, uriFilters=<null>]

The output of the config class's toString method was being logged resulting in the string above.  Recently, the fluent-api plugin was added to the URI Normalization build, and this change caused the JAXB generated toString method to not be generated.  (Why the test ever passed is a mystery.)

To fix, I added the toString plugin to the build to ensure the method got generated.  I ran the functional test locally in IntelliJ to confirm this fixed the problem.  I did this for the Maven and Gradle build.


repose-aggregator/components/filters/uri-normalization-filter/pom.xml
```diff
@@ -92,6 +92,7 @@
 
                     <args>
                         <arg>-Xfluent-api</arg>
+                        <arg>-XtoString</arg>
                     </args>
```
adrianjgeorge: these args are different than the gradle build. which is correct?

versions.properties
```diff
@@ -79,6 +79,8 @@ org.jvnet.jaxb2_commons:jaxb2-basics-ant=$jaxbBasicsVersion
 org.jvnet.jaxb2_commons:jaxb2-basics-runtime=$jaxbBasicsVersion
 org.jvnet.jaxb2_commons:jaxb2-default-value=1.1
 org.jvnet.jaxb2_commons:jaxb2-fluent-api=3.0
+org.jvnet.jaxb2_commons:jaxb2-basics=0.6.0
+org.jvnet.jaxb2_commons:jaxb2-basics-runtime=0.6.0
```
adrianjgeorge: this is already present 4 lines up. also the basics should probably use the predefine version so that they stay hooked together. and finally these are alphabetized, so basics should be above the others.

repose-aggregator/components/filters/uri-normalization-filter/pom.xml
Mario-Lopez: I'm not sure why someone added the default-value arg to the gradle build.  I cleaned it up.

versions.properties
Mario-Lopez: d'oh.  Fixed.

PR #1497, opened 2016-03-16
REP-3257 Convert destination router filter

dmnjohns:


dmnjohns:
I'm not sure why I thought this was ready for a PR. The "tmp" commit at the end clearly indicates that it is not.

I'll reopen this PR when it has a chance of being accepted. Oops!

PR #1498, opened 2016-03-16
REP-3640 - Header User wrapper updates

Mario-Lopez:
The Header Identity filter was revived, refactored, updated to use the new wrappers, and renamed to Header User filter.


repose-aggregator/commons/commons-configuration/src/test/resources/unmarshallerValidator/xsd/header-user-configuration.xsd
```diff
@@ -20,51 +20,47 @@
   -->
 
 
-<xs:schema xmlns="http://docs.openrepose.org/repose/header-id-mapping/v1.0"
-           xmlns:html="http://www.w3.org/1999/xhtml"
-           xmlns:xs="http://www.w3.org/2001/XMLSchema"
+<xs:schema xmlns:header-user="http://docs.openrepose.org/repose/header-user/v1.0" xmlns:html="http://www.w3.org/1999/xhtml"
```
dmnjohns: If you make this the default namespace (drop the :header-user) you won't have to include the namespace when referring to your own types (e.g., header-user:HeaderUserConfig).

repose-aggregator/components/filters/header-user-filter/src/main/scala/org/openrepose/filters/headeruser/HeaderUserFilter.scala
```diff
@@ -0,0 +1,104 @@
...
+    if (!isInitialized) {
+      logger.warn("Header User filter has not yet initialized.")
+      servletResponse.asInstanceOf[HttpServletResponse].sendError(HttpServletResponse.SC_SERVICE_UNAVAILABLE)
+      return
+    }
+
+    val wrappedRequest = new HttpServletRequestWrapper(servletRequest.asInstanceOf[HttpServletRequest])
+
+    configuredHeaders.get flatMap { configuredHeader =>
+      wrappedRequest.getSplittableHeaderScala(configuredHeader.name).headOption
```
dmnjohns: Would we want to handle splittable vs non-splittable headers here, as well? I guess that question loosely translates to "are usernames allowed to contain a comma?"

Did we decide to continue only moving the first header value? I think that's fine, I just want to verify.

repose-aggregator/commons/commons-configuration/src/test/resources/unmarshallerValidator/xsd/header-user-configuration.xsd
Mario-Lopez: Sounds good.  I had made the change to the real XSD and forgot to update this XSD to be the same.  Update pushed.

repose-aggregator/components/filters/header-user-filter/src/test/scala/org/openrepose/filters/headeruser/HeaderUserFilterTest.scala
```diff
@@ -0,0 +1,170 @@
...
+      servletRequest.addHeader("some-header", "why-not-one-more")
+
+      filter.doFilter(servletRequest, servletResponse, filterChain)
+
+      val wrappedRequest = getPostFilterRequest
+      wrappedRequest.getHeader(User) shouldBe "some-value" + DefaultQualty
+    }
+  }
+
+  private def getPostFilterRequest = filterChain.getRequest.asInstanceOf[HttpServletRequestWrapper]
```
dmnjohns: This looks so much nicer than the argument captor. :smile:

repose-aggregator/components/filters/header-user-filter/src/main/scala/org/openrepose/filters/headeruser/HeaderUserFilter.scala
Mario-Lopez: Good question.  I kept the old behavior for now.  Would we also want to add a "splittable" config attribute here?  This might be a good story to add to the backlog if so.

repose-aggregator/components/filters/header-user-filter/src/test/scala/org/openrepose/filters/headeruser/HeaderUserFilterTest.scala
Mario-Lopez: Definitely.  I didn't have to verify anything in the response, so I went this route here.

repose-aggregator/components/filters/header-user-filter/src/main/scala/org/openrepose/filters/headeruser/HeaderUserFilter.scala
dmnjohns: Fair enough.

Yeah, I expect we'd want to add that config attribute since we could be translating from any header with any set of semantics around ','. We should probably write the story and talk about it at a story grooming.

dmnjohns:
Looks like we're seeing the expected 30 test failures.

Everything looks good. I'll let the build finish again for super extra certainty, then I'll merge it. :+1:

repose-aggregator/components/filters/header-user-filter/src/main/scala/org/openrepose/filters/headeruser/HeaderUserFilter.scala
Mario-Lopez: Wrote https://repose.atlassian.net/browse/REP-3726

dmnjohns:
Looks like 31 tests failed that time. Is it anything to be concerned over?

Mario-Lopez:
31 failures in the Gradle build, 30 in the Maven build.  Last I heard Gradle was supposed to have 33 failures.

dmnjohns:
Ah, right, all the builds.

Alright. MERGINATING!

PR #1501, opened 2016-03-21
RPE-3257 Destination Router Filter Conversion and Core Changes

dmnjohns:
I also fixed a few random bugs in this PR that were either in the way, or just annoying me.


repose-aggregator/components/filters/destination-router-filter/src/main/scala/org/openrepose/filters/destinationrouter/DestinationRouterFilter.scala
```diff
@@ -0,0 +1,116 @@
...
+import org.openrepose.filters.routing.servlet.config.DestinationRouterConfiguration
+
+@Named
+class DestinationRouterFilter @Inject()(configurationService: ConfigurationService, metricsService: MetricsService)
+  extends Filter with UpdateListener[DestinationRouterConfiguration] with LazyLogging {
+
+  private final val DefaultConfigFileName = "destination-router.cfg.xml"
+  private final val SchemaFilePath = "/META-INF/schema/config/destination-router-configuration.xsd"
+
+  private val mbcsRoutedReponse: Option[MeterByCategorySum] = Option(metricsService) flatMap { ms =>
```
Mario-Lopez: I thought we decided we didn't have to worry about the metrics service being null these days.

Mario-Lopez: In fact, you're the one that told me. :P
https://github.com/rackerlabs/repose/pull/1449#discussion_r50881077

dmnjohns: You can't expect me to remember all the junk I say... :stuck_out_tongue_closed_eyes: 

I'll update this.

repose-aggregator/components/filters/destination-router-filter/src/main/scala/org/openrepose/filters/destinationrouter/DestinationRouterFilter.scala
```diff
@@ -0,0 +1,114 @@
...
+    val httpResponse = response.asInstanceOf[HttpServletResponse]
+
+    if (!isInitialized) {
+      logger.error("Destination Router Filter has not yet initialized")
+      httpResponse.sendError(HttpServletResponse.SC_SERVICE_UNAVAILABLE)
+    } else {
+      val target = configuration.getTarget
+
+      // Set the default quality since XJC won't
+      if (!target.isSetQuality) target.setQuality(0.5)
```
Mario-Lopez: Why not do this once in the configurationUpdated method?

repose-aggregator/core/core-service-api/src/main/resources/META-INF/schema/system-model/system-model.xsd
```diff
@@ -333,7 +333,7 @@
                    saxon:message="Must have at least one destination defined.">
         </xs:assert>
         <xs:assert vc:minVersion="1.1"
-                   test="((count(mod:endpoint)+count(mod:target)) eq 1) or (count(mod:*[xs:boolean(@default) = true()]) eq 1)"
+                   test="count(mod:*[xs:boolean(@default) = true()]) eq 1"
```
Mario-Lopez: Out of curiosity, why are we no longer assuming a single item is the default?

dmnjohns: Actually, while that was the assumption in the schema, it was not an assumption we made in code. So that mismatch in expectation led to a bug.

Now, rather than this change, we could have updated the code to assume that a single item is the default. I think in the ended I opted to go this route for the sake of clarity and consistency. This approach provides the same experience regardless of whether there is one item in the sequence or many. Also, this change was one line versus the multi-line code change the other approach would have required. So that minimizes complexity in some small way.

repose-aggregator/components/filters/destination-router-filter/src/main/scala/org/openrepose/filters/destinationrouter/DestinationRouterFilter.scala
dmnjohns: I guess I was derping when I wrote that. I'll move it.

repose-aggregator/core/core-service-api/src/main/resources/META-INF/schema/system-model/system-model.xsd
Mario-Lopez: Legit.  Thanks.

Mario-Lopez:
Man, I was ready to merge this, but there's a merge conflict.  Could I get a rebase, please?

dmnjohns:
Rebased!

PR #1502, opened 2016-03-21
REP-3603 Updated the Saxon-EE library dependency from v9.4.0.9 to v9.5.1-8.

wdschei:
The API-Checker was already updated from v1.1.4 to v2.0.0-SNAPSHOT and the latest snapshot is also using Saxon-EE v9.5.1-8.


adrianjgeorge:
If this passes tests, and the saxon ee features still work when needed, ship it.
I'm assuming you tried to run everything in a situation where the license wasn't present?

PR #1503, opened 2016-03-21
REP-3629 - Header Translation - configure quality

Mario-Lopez:
I added "quality" and "splittable" attributes to the header config item for the Header Translation filter.  The filter now supports adding a quality parameter to the headers it creates (while not carrying over any quality parameter that was on the original header value).  If no "quality" attribute is specified, none is added (i.e. the quality parameter on the original header will be used if it exists).

In case the header is splittable, a config item was added to have the filter split the header values before applying a new quality when creating the new headers.

### DIFF
--- a/repose-aggregator/components/filters/header-translation-filter/src/main/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilter.scala
+++ b/repose-aggregator/components/filters/header-translation-filter/src/main/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilter.scala
@@ -65,10 +65,11 @@ class HeaderTranslationFilter @Inject()(configurationService: ConfigurationServi
 
       sourceHeaders foreach { sourceHeader =>
         val originalHeaderName = sourceHeader.getOriginalName
-        val originalHeaderValues = if (sourceHeader.isSplittable)
-          httpRequest.getSplittableHeaderScala(originalHeaderName) else
-          httpRequest.getHeadersScala(originalHeaderName)
         val quality = Option(sourceHeader.getQuality)
+        val originalHeaderValues = if (sourceHeader.isSplittable)
+          httpRequest.getSplittableHeaderScala(originalHeaderName)
+        else
+          httpRequest.getHeadersScala(originalHeaderName)
 
         if (originalHeaderValues.nonEmpty) {
           sourceHeader.getNewName foreach { newHeaderName =>
@@ -112,5 +113,5 @@ class HeaderTranslationFilter @Inject()(configurationService: ConfigurationServi
 object HeaderTranslationFilter {
   private final val DefaultConfig = "header-translation.cfg.xml"
   private final val SchemaFileName = "/META-INF/schema/config/header-translation.xsd"
-  private final val QualityReqex = ";q=\\d[^;]*".r
+  private final val QualityReqex = """;q=[^;]*""".r
 }
--- a/repose-aggregator/components/filters/header-translation-filter/src/main/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilter.scala
+++ b/repose-aggregator/components/filters/header-translation-filter/src/main/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilter.scala
@@ -36,17 +36,18 @@ import scala.collection.JavaConversions._
 class HeaderTranslationFilter @Inject()(configurationService: ConfigurationService)
   extends Filter with UpdateListener[HeaderTranslationType] with LazyLogging {
 
-  private final val DEFAULT_CONFIG = "header-translation.cfg.xml"
+  private final val DefaultConfig = "header-translation.cfg.xml"
+  private final val SchemaFileName = "/META-INF/schema/config/header-translation.xsd"
 
   private var configFilename: String = _
   private var initialized: Boolean = false
   private var sourceHeaders: List[Header] = List.empty
 
   override def init(filterConfig: FilterConfig): Unit = {
-    configFilename = new FilterConfigHelper(filterConfig).getFilterConfig(DEFAULT_CONFIG)
+    configFilename = new FilterConfigHelper(filterConfig).getFilterConfig(DefaultConfig)
     logger.info("Initializing filter using config " + configFilename)
 
-    val xsdURL = getClass.getResource("/META-INF/schema/config/header-translation.xsd")
+    val xsdURL = getClass.getResource(SchemaFileName)
     configurationService.subscribeTo(
       filterConfig.getFilterName,
       configFilename,
@@ -72,15 +73,15 @@ class HeaderTranslationFilter @Inject()(configurationService: ConfigurationServi
             originalHeaderValues foreach { originalHeaderValue =>
               httpRequest.addHeader(newHeaderName, originalHeaderValue)
             }
-            logger.trace("Header added: " + newHeaderName)
+            logger.trace("Header added: {}", newHeaderName)
           }
 
           if (sourceHeader.isRemoveOriginal) {
             httpRequest.removeHeader(originalHeaderName)
-            logger.trace("Header removed: " + originalHeaderName)
+            logger.trace("Header removed: {}", originalHeaderName)
           }
         } else {
-          logger.trace("Header for translation not found: " + originalHeaderName)
+          logger.trace("Header for translation not found: {}", originalHeaderName)
         }
       }
 
@@ -88,9 +89,7 @@ class HeaderTranslationFilter @Inject()(configurationService: ConfigurationServi
     }
   }
 
-  override def isInitialized: Boolean = {
-    initialized
-  }
+  override def isInitialized: Boolean = initialized
 
   override def destroy(): Unit = {
     configurationService.unsubscribeFrom(configFilename, this)
--- a/repose-aggregator/components/filters/header-translation-filter/src/main/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilter.scala
+++ b/repose-aggregator/components/filters/header-translation-filter/src/main/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilter.scala
@@ -36,8 +36,7 @@ import scala.collection.JavaConversions._
 class HeaderTranslationFilter @Inject()(configurationService: ConfigurationService)
   extends Filter with UpdateListener[HeaderTranslationType] with LazyLogging {
 
-  private final val DefaultConfig = "header-translation.cfg.xml"
-  private final val SchemaFileName = "/META-INF/schema/config/header-translation.xsd"
+  import HeaderTranslationFilter._
 
   private var configFilename: String = _
   private var initialized: Boolean = false
@@ -66,12 +65,19 @@ class HeaderTranslationFilter @Inject()(configurationService: ConfigurationServi
 
       sourceHeaders foreach { sourceHeader =>
         val originalHeaderName = sourceHeader.getOriginalName
-        val originalHeaderValues = httpRequest.getHeadersScala(originalHeaderName)
+        val originalHeaderValues = if (sourceHeader.isSplittable)
+          httpRequest.getSplittableHeaderScala(originalHeaderName) else
+          httpRequest.getHeadersScala(originalHeaderName)
+        val quality = Option(sourceHeader.getQuality)
 
         if (originalHeaderValues.nonEmpty) {
           sourceHeader.getNewName foreach { newHeaderName =>
             originalHeaderValues foreach { originalHeaderValue =>
-              httpRequest.addHeader(newHeaderName, originalHeaderValue)
+              if (quality.isEmpty) {
+                httpRequest.addHeader(newHeaderName, originalHeaderValue)
+              } else {
+                httpRequest.addHeader(newHeaderName, withoutQuality(originalHeaderValue), quality.get)
+              }
             }
             logger.trace("Header added: {}", newHeaderName)
           }
@@ -89,6 +95,8 @@ class HeaderTranslationFilter @Inject()(configurationService: ConfigurationServi
     }
   }
 
+  private def withoutQuality(headerValue: String): String = QualityReqex.replaceFirstIn(headerValue, "")
+
   override def isInitialized: Boolean = initialized
 
   override def destroy(): Unit = {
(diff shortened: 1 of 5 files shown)


repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/filters/headertranslation/HeaderTranslationWQualityTest.groovy
```diff
@@ -53,6 +52,42 @@ class HeaderTranslationWQualityTest extends ReposeValveTest {
...
+                headers: ["x-pp-user"    : "test, repose",
+                          "x-tenant-name": "b",
+                          "x-roles"      : "test"]
+        )
+        def reposehandling = mc.getHandlings()[0]
+
+        then:
+        reposehandling.request.getHeaders().contains("x-rax-username")
+        reposehandling.request.getHeaders().findAll("x-rax-username").contains("test;q=0.5")
+        reposehandling.request.getHeaders().findAll("x-rax-username").contains("repose;q=0.5")
```
Mario-Lopez: Jenny, this is actually testing the opposite of what the configuration says is supposed to happen.  To get this behavior, the config should have said splittable="true".  By default, this value is false.  Here's the schema:  https://github.com/rackerlabs/repose/blob/REP-3629-config-quality-header-translation/repose-aggregator/components/filters/header-translation-filter/src/main/resources/META-INF/schema/config/header-translation.xsd#L98

The reason why the test worked anyway is that Repose always splits certain headers.  X-PP-User is one of them, so it shouldn't be used for this test.  For a list of automatically split headers that should be avoided in this test, check out:  https://repose.atlassian.net/wiki/display/REPOSE/Merge+Header+filter#MergeHeaderfilter-Headers

I would also test the non-splittable case as well.

For the splittable case, it should look like what you have there:
test;q=0.5
repose;q=0.5

For the non-splittable case, it should have a single value like this:
test, repose;q=0.5

repose-aggregator/components/filters/header-translation-filter/src/main/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilter.scala
```diff
@@ -101,3 +108,9 @@ class HeaderTranslationFilter @Inject()(configurationService: ConfigurationServi
     initialized = true
   }
 }
+
+object HeaderTranslationFilter {
+  private final val DefaultConfig = "header-translation.cfg.xml"
+  private final val SchemaFileName = "/META-INF/schema/config/header-translation.xsd"
+  private final val QualityReqex = ";q=\\d[^;]*".r
+}
```
dmnjohns: No newline at EOF?! Blasphemy!

repose-aggregator/components/filters/header-translation-filter/src/main/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilter.scala
```diff
@@ -36,17 +36,17 @@ import scala.collection.JavaConversions._
 class HeaderTranslationFilter @Inject()(configurationService: ConfigurationService)
   extends Filter with UpdateListener[HeaderTranslationType] with LazyLogging {
 
-  private final val DEFAULT_CONFIG = "header-translation.cfg.xml"
+  import HeaderTranslationFilter._
```
dmnjohns: I know we've done this in quite a few places, but I wonder if it's worth breaking constants out into an Object. Specifically for filters, since they are singleton Spring beans anyway. It may be easier to see the constants inline, and there is likely a very, _very_ small overhead that comes with defining a class and an object.

It doesn't bother me that much, but I figured I'd write out my musings anyway.

Mario-Lopez: I've been wondering, too.  I'll do some research.  My guess is that the only difference is going to be where in memory the value is stored.

repose-aggregator/components/filters/header-translation-filter/src/test/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilterTest.scala
```diff
@@ -109,40 +141,25 @@ class HeaderTranslationFilterTest extends FunSpec with BeforeAndAfter with Match
...
 
     it("should add a new header with a single value") {
       filter.doFilter(mockRequest, null, mockFilterChain)
 
-      val requestCaptor = ArgumentCaptor.forClass(classOf[HttpServletRequest])
-      verify(mockFilterChain).doFilter(requestCaptor.capture(), any[ServletResponse])
-
-      val capturedRequest = requestCaptor.getValue
-      capturedRequest.getHeaders("X-New-One-One").toSeq should contain theSameElementsAs Seq("valueOne")
+      getCapturedRequest.getHeaders("X-New-One-One").toSeq should contain theSameElementsAs Seq("valueOne")
```
dmnjohns: You could probably drop the ".toSeq" and drop "theSameElementsAs" in this case. There exists a ScalaTest Matcher to check that a Java collection contains a single value. I assume it works for Enumerations...

Not that it overly matters.

repose-aggregator/components/filters/header-translation-filter/src/test/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilterTest.scala
```diff
@@ -174,22 +188,98 @@ class HeaderTranslationFilterTest extends FunSpec with BeforeAndAfter with Match
...
+      getCapturedRequest.getHeaders("X-New-Six-One").toSeq should contain theSameElementsAs Seq("mustard,ketchup,mayo;q=0.9,pickles;q=0.1")
+    }
+
+    it("should split the headers when configured") {
+      val mockRequest = new MockHttpServletRequest()
+      mockRequest.addHeader("X-Five", "mustard,ketchup,mayo;q=0.9,pickles;q=0.1")
+
+      filter.doFilter(mockRequest, null, mockFilterChain)
+
+      getCapturedRequest.getHeaders("X-New-Five-One").toSeq should contain theSameElementsInOrderAs Seq("mustard", "ketchup", "mayo;q=0.9", "pickles;q=0.1")
```
dmnjohns: You may be able to use the "inOrder" matcher rather than "theSameElementsInOrderAs". That would let you drop the Seq constructor. Or you could go a step further with the "inOrderOnly" matcher.

Again, this is of little consequence.

repose-aggregator/components/filters/header-translation-filter/src/main/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilter.scala
```diff
@@ -65,22 +65,29 @@ class HeaderTranslationFilter @Inject()(configurationService: ConfigurationServi
 
       sourceHeaders foreach { sourceHeader =>
         val originalHeaderName = sourceHeader.getOriginalName
-        val originalHeaderValues = httpRequest.getHeadersScala(originalHeaderName)
+        val originalHeaderValues = if (sourceHeader.isSplittable)
+          httpRequest.getSplittableHeaderScala(originalHeaderName) else
+          httpRequest.getHeadersScala(originalHeaderName)
```
dmnjohns: What is going on with the formatting here?

dmnjohns: Oh, was it one line that got split by the IntelliJ formatter? That would make sense.

One line is correct, but due to the length you may prefer:
```
if (sourceHeader.isSplittable)
    httpRequest.getSplittableHeaderScala(originalHeaderName)
else
    httpRequest.getHeadersScala(originalHeaderName)
```

Mario-Lopez: I could be wrong, but I thought a style-check would complain about not having braces in that case making what should be a one-liner be like 5 lines.  Still, I was iffy on it to begin with, and if it wasn't obvious to you what that meant on first glance, it's worth changing.  I'll change it.

repose-aggregator/components/filters/header-translation-filter/src/main/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilter.scala
```diff
@@ -101,3 +108,9 @@ class HeaderTranslationFilter @Inject()(configurationService: ConfigurationServi
     initialized = true
   }
 }
+
+object HeaderTranslationFilter {
+  private final val DefaultConfig = "header-translation.cfg.xml"
+  private final val SchemaFileName = "/META-INF/schema/config/header-translation.xsd"
+  private final val QualityReqex = ";q=\\d[^;]*".r
```
dmnjohns: I like how this regex is like, "the first character of the parameter value must be a digit, but after that... f*** it, anything goes!"

dmnjohns: Well, except a ';'.

Mario-Lopez: Yeah, after reading the spec, I figured this would be Safe Enough (TM).

dmnjohns: Side note: I've always found that using single-quotes for regexes may lead to confusion, especially when you have to escape (or double escape, as it were) things. You might consider using triple quotes instead.

`""";q=\\d[^;]*""".r`

dmnjohns: Oh, it's totally Safe Enough. But why even enforce the first digit? It seems like a quirky mismatch of caring, but not that much.

repose-aggregator/components/filters/header-translation-filter/src/main/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilter.scala
dmnjohns: Let me know what you find out.

repose-aggregator/components/filters/header-translation-filter/src/main/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilter.scala
dmnjohns: Well, the behavior was obvious, but the formatting wasn't. I think it's worth changing, but unless you can find where the Scala style guide describes this situation, you can decide what to change it too. Mostly, if on multiple lines, I would like and expect to see "if" and "else" starting on the same column.

dmnjohns:
I like this PR. Super straight forward. A+.

I assume test changes are still inbound, so I'll wait for those and give them a chance to run before I merge.

repose-aggregator/components/filters/header-translation-filter/src/main/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilter.scala
Mario-Lopez: I wanted to enforce as much as possible, but once you get past the first digit, it gets a little nuts.  After the first digit, there might be a period, and if there's a period, there might be up to three digits, of which must be zeros if the first digit was a "1".  This seemed like the most I could easily check without going crazy.  Still, I see your point.  Actually, I really should have made the \d be [01] or not even care and just go for the [^;].  Hmmm.....

dmnjohns: Whatever makes you happy, as long as it cuts out the ;q=BLAH bit.

repose-aggregator/components/filters/header-translation-filter/src/main/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilter.scala
Mario-Lopez: I gave that EOF a newline.  EOFs love newlines.

repose-aggregator/components/filters/header-translation-filter/src/test/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilterTest.scala
Mario-Lopez: I went with the inOrderOnly option.

repose-aggregator/components/filters/header-translation-filter/src/test/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilterTest.scala
Mario-Lopez: I was able to get rid of the "theSameElementsAs" and Seq.apply, but not the .toSeq.

repose-aggregator/components/filters/header-translation-filter/src/main/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilter.scala
Mario-Lopez: I found the idiomatic Scala way and went with that.

repose-aggregator/components/filters/header-translation-filter/src/test/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilterTest.scala
dmnjohns: Fair enough. Thinking about it, a Java Enumeration is **NOT** a Java Collection. So that's probably why.

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/filters/headertranslation/HeaderTranslationWQualityTest.groovy
dmnjohns: Once this test is corrected, I'll merge this PR.

Mario-Lopez: She updated the name of that test and added the tests I described in her latest commits.

dmnjohns: Ah, I see. Fair enough. Merging!

PR #1506, opened 2016-03-29
REP-3393 Make our identity URLs in config consistent

dmnjohns:
Only the rackspace-identity-basic-auth filter needed to be updated. All of the others expected just scheme, host, and port -- not the path.

Before this gets merged, the task to talk to users to make sure we aren't breaking their proxies should be completed.


wdschei:
This one looks good to me. I like an easy one every once in a while. When @jennyvo clears it, then I'll merge it.

jennyvo:
When build is completed and passed from jenkins, it can be merged.

PR #1510, opened 2016-04-05
REP-3573 - Intra-filter logging wrapper updates

Mario-Lopez:
Instead of wrapping the request and response with the old mutable wrappers in the PowerFilterChain.doReposeFilter() method, I use what's passed in when intrafilter logging is turned off and wrap it using the new wrappers when it's set to trace.

I also check if the InputStream in the request supports mark/reset, and if it doesn't, I copy the data over to an InputStream that does (and use that when wrapping the request).

For the response, I wrap it using READONLY mode on the body so I can read the OutputStream.

I'm not a huge fan of the maybeWrapped stuff, but I expect the PowerFilterChain to be updated soon anyway.

I added a new functional test to verify:
- the origin service receives the request body sent from the client
- the client receives the response body sent from the origin service
- each intrafilter log line contains the full request and/or response body

### DIFF
--- a/repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/intrafilterLogging/RequestLog.java
+++ b/repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/intrafilterLogging/RequestLog.java
@@ -53,7 +53,7 @@ public class RequestLog {
             requestBody = IOUtils.toString(inputStream); //http://stackoverflow.com/a/309448
             inputStream.reset();
         } else {
-            throw new RuntimeException("Servlet input stream does not support mark/reset: " + inputStream);
+            throw new IllegalArgumentException("Servlet input stream does not support mark/reset: " + inputStream);
         }
     }
 
--- a/repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/intrafilterLogging/RequestLog.java
+++ b/repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/intrafilterLogging/RequestLog.java
@@ -53,7 +53,7 @@ public class RequestLog {
             requestBody = IOUtils.toString(inputStream); //http://stackoverflow.com/a/309448
             inputStream.reset();
         } else {
-            throw new RuntimeException("Servlet input stream does not support mark/reset.");
+            throw new RuntimeException("Servlet input stream does not support mark/reset: " + inputStream);
         }
     }
 
--- a/repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/intrafilterLogging/RequestLog.java
+++ b/repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/intrafilterLogging/RequestLog.java
@@ -22,11 +22,10 @@ package org.openrepose.powerfilter.intrafilterLogging;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.joda.time.DateTime;
-import org.openrepose.commons.utils.servlet.http.MutableHttpServletRequest;
 import org.openrepose.core.systemmodel.Filter;
-import org.openrepose.powerfilter.filtercontext.FilterContext;
 
 import javax.servlet.ServletInputStream;
+import javax.servlet.http.HttpServletRequest;
 import java.io.IOException;
 import java.util.*;
 
@@ -38,41 +37,34 @@ public class RequestLog {
     String httpMethod;
     String requestURI;
     String requestBody;
-    HashMap<String, String> headers;
+    Map<String, String> headers;
 
-    public RequestLog(MutableHttpServletRequest mutableHttpServletRequest,
-                      FilterContext filterContext) throws IOException {
-
-        Filter filter = filterContext.getFilterConfig();
-
+    public RequestLog(HttpServletRequest httpServletRequest, Filter filter) throws IOException {
         preamble = "Intrafilter Request Log";
         timestamp = new DateTime().toString();
         currentFilter = StringUtils.isEmpty(filter.getId()) ? filter.getName() : filter.getId() + "-" + filter.getName();
-        httpMethod = mutableHttpServletRequest.getMethod();
-        requestURI = mutableHttpServletRequest.getRequestURI();
-        headers = convertRequestHeadersToMap(mutableHttpServletRequest);
+        httpMethod = httpServletRequest.getMethod();
+        requestURI = httpServletRequest.getRequestURI();
+        headers = convertRequestHeadersToMap(httpServletRequest);
 
-        //Have to wrap the input stream in something that can be buffered, as well as reset.
-        ServletInputStream bin = mutableHttpServletRequest.getInputStream();
-        bin.mark(Integer.MAX_VALUE); //Something doesn't support mark reset
-        requestBody = IOUtils.toString(bin); //http://stackoverflow.com/a/309448
-        bin.reset();
+        ServletInputStream inputStream = httpServletRequest.getInputStream();
+        if (inputStream.markSupported()) {
+            inputStream.mark(Integer.MAX_VALUE);
+            requestBody = IOUtils.toString(inputStream); //http://stackoverflow.com/a/309448
+            inputStream.reset();
+        } else {
+            throw new RuntimeException("Servlet input stream does not support mark/reset.");
+        }
     }
 
-    private HashMap<String, String> convertRequestHeadersToMap(
-            MutableHttpServletRequest mutableHttpServletRequest) {
-
-        HashMap<String, String> headerMap = new LinkedHashMap<>();
-        List<String> headerNames = Collections.list(mutableHttpServletRequest.getHeaderNames());
+    private Map<String, String> convertRequestHeadersToMap(HttpServletRequest httpServletRequest) {
+        Map<String, String> headerMap = new LinkedHashMap<>();
+        List<String> headerNames = Collections.list(httpServletRequest.getHeaderNames());
 
         for (String headerName : headerNames) {
-            StringBuilder allHeaderValues = new StringBuilder();
-            for (String value : Collections.list(mutableHttpServletRequest.getHeaders(headerName))) {
-                allHeaderValues.append(value).append(",");
-            }
-            //Clobber the last character
-            allHeaderValues.deleteCharAt(allHeaderValues.length() - 1);
-            headerMap.put(headerName, allHeaderValues.toString());
+            StringJoiner stringJoiner = new StringJoiner(",");
+            Collections.list(httpServletRequest.getHeaders(headerName)).forEach(stringJoiner::add);
+            headerMap.put(headerName, stringJoiner.toString());
         }
 
         return headerMap;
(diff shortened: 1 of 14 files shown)


repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/intrafilterLogging/RequestLog.java
```diff
@@ -38,41 +37,34 @@
...
-            StringBuilder allHeaderValues = new StringBuilder();
-            for (String value : Collections.list(mutableHttpServletRequest.getHeaders(headerName))) {
-                allHeaderValues.append(value).append(",");
-            }
-            //Clobber the last character
-            allHeaderValues.deleteCharAt(allHeaderValues.length() - 1);
-            headerMap.put(headerName, allHeaderValues.toString());
+            StringJoiner stringJoiner = new StringJoiner(",");
+            Collections.list(httpServletRequest.getHeaders(headerName)).forEach(stringJoiner::add);
+            headerMap.put(headerName, stringJoiner.toString());
```
dmnjohns: I didn't know the StringJoiner class existed. Looks like it came in with Java 8. 

Also, if you wanted, you could do this whole function using Java 8's new functional interface. Something like:
```
return Collections.list(mutableHttpServletRequest.getHeaderNames()).stream()
                .collect(Collectors.toMap(Function.identity(), 
                        headerName -> Collections.list(mutableHttpServletRequest.getHeaders(headerName)).stream()
                                .collect(Collectors.joining(","))));
```

That formatting is kind of insane, but yeah.

dmnjohns: Side note: Enumerations make this uglier than it ought to be.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/intrafilterLogging/ResponseLog.java
```diff
@@ -39,37 +35,26 @@
...
-            }
-            //Clobber the last character
-            allHeaderValues.deleteCharAt(allHeaderValues.length() - 1);
-            headerMap.put(headerName, allHeaderValues.toString());
+            StringJoiner stringJoiner = new StringJoiner(",");
+            wrappedServletResponse.getHeaders(headerName).forEach(stringJoiner::add);
+            headerMap.put(headerName, stringJoiner.toString());
         }
 
         return headerMap;
```
dmnjohns: Same thing here for the stream processing, if you want.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/intrafilterLogging/RequestLog.java
Mario-Lopez: I seriously thought about it, but damn does that not look pretty.  Maybe if it was Groovy... ;-)

`httpServletRequest.getHeaderNames().toList().collectEntries {
	[(it): mutableHttpServletRequest.getHeaders(it).toList().join(",")]
}`

I think it's more understandable with the old school enhanced for loop, but I'm open to changing it if you want.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/intrafilterLogging/RequestLog.java
```diff
@@ -38,41 +37,34 @@
...
-        bin.mark(Integer.MAX_VALUE); //Something doesn't support mark reset
-        requestBody = IOUtils.toString(bin); //http://stackoverflow.com/a/309448
-        bin.reset();
+        ServletInputStream inputStream = httpServletRequest.getInputStream();
+        if (inputStream.markSupported()) {
+            inputStream.mark(Integer.MAX_VALUE);
+            requestBody = IOUtils.toString(inputStream); //http://stackoverflow.com/a/309448
+            inputStream.reset();
+        } else {
+            throw new RuntimeException("Servlet input stream does not support mark/reset: " + inputStream);
```
dmnjohns: I just want to make sure that we want to throw an exception here rather than, saying, logging a warning. The input stream _should_ always support mark since we pass the request in from our code. If at some point in the future we pass a request with an input stream that does not support mark, throwing an exception should make our tests bomb immediately so that we know what we did. Of course, we likely have tests verifying that the body gets logged by intra-filter logging, so that may be redundant.

Also, RuntimeException is pretty generic. Is there a better exception to use from the standard library? Maybe IllegalArgumentException? It doesn't matter too much, just a thought.

Really, I'd just like to hear your thoughts on the matter.

repose-aggregator/functional-tests/spock-functional-test/src/test/configs/features/core/intrafilterlogging/header-normalization.cfg.xml
```diff
@@ -22,14 +22,6 @@
...
         </target>
 
-        <!-- applies to all POST and PUT requests -->
-        <target http-methods="POST PUT">
-            <whitelist id="modification">
-                <header id="X-THIRD-Filter"/>
-                <header id="x-auth-token"/>
-            </whitelist>
-        </target>
-
```
dmnjohns: Is it just me, or is it odd that the intra-filter logging tests use header-normalization to being with? Perhaps other filters as well. Not that it's bad to test the behavior of multiple filters working together, we should do that, but hopefully we have tests that minimize the scope (to only the filter under test) so that we can assert something about the behavior of the filter and not just one specific composition of filters.

dmnjohns: In this case, we aren't testing filters, so... s/filter/component/.

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/core/intrafilterlogging/IntraFilterLoggingResponseBodyTest.groovy
```diff
@@ -0,0 +1,105 @@
...
+        repose.start()
+
+        deproxy.addEndpoint(properties.targetPort, 'origin service')
+
+        repose.waitForNon500FromUrl(reposeEndpoint)
+    }
+
+    def cleanupSpec() {
+        deproxy?.shutdown()
+        repose?.stop()
```
dmnjohns: I always wondered why we used if statements elsewhere. Lovely Groovy syntax.

Although, I think you inherit a cleanupSpec from the `ReposeValveTest` class. So this may not be necessary at all. Of course, the inherited method could probably do with using this syntax.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/intrafilterLogging/RequestLog.java
Mario-Lopez: I was on the fence about throwing that exception (or throwing one at all).  IllegalArgumentException would definitely be better (and it's unchecked!).  I could also go the logging route and just not include the body in those cases.  And, like you said, we could also just not even check since we should be the only ones calling that code.

Because the class is pretty useless to outsiders (no getters and only package-visible fields), I'm okay with removing the check all together (which is how it used to be).  Thoughts?

dmnjohns:
I like it. Good work :+1:

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/intrafilterLogging/RequestLog.java
dmnjohns: Personally, I wouldn't be able to resist the new fanciness :stuck_out_tongue: 

That's legit, though. If you find it to be more understandable as-is, then leave it as-is.

repose-aggregator/functional-tests/spock-functional-test/src/test/configs/features/core/intrafilterlogging/header-normalization.cfg.xml
Mario-Lopez: The original intra-filter logging functional test includes all of the filters, and for the new functional test, I removed some of the filters that would have required additional effort to get the request to successfully complete.  I actually would have preferred testing as many filters as possible to ensure my updates worked with how every filter handles the new wrappers.  During development, I ran into a problem with how the rate limiting filter messes with the streams which fed my paranoia.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/intrafilterLogging/RequestLog.java
dmnjohns: I think I like having some sort of catch here in case we break it and don't have a test in place. I hope that is never the case, but better safe than sorry. If it's a hard assert that we expect an input stream which supports mark, then an exception seems like the right call. It's a simpler path than modifying the output. So I think my preference would be to throw an IllegalArgumentException.

I'll ask @adrianjgeorge to chime in with his thoughts on the matter.

repose-aggregator/functional-tests/spock-functional-test/src/test/configs/features/core/intrafilterlogging/header-normalization.cfg.xml
dmnjohns: Interesting. My worry would just be that the way two filters work together prevent us from catching an issue with the component under test, or that we end up testing everything with multiple filters and when one has a bug we have to fix it across a huge number of tests.

I think ideally, we would test every component in isolation and in a production-type environment. Both approaches have their pros and cons. In the end, this works, so I have no real complaints.

Mario-Lopez: Completely agree with that concern.  I guess this is more of an integration test than a functional test. :-/

repose-aggregator/functional-tests/spock-functional-test/src/test/groovy/features/core/intrafilterlogging/IntraFilterLoggingResponseBodyTest.groovy
Mario-Lopez: Interesting.  I hadn't noticed the method was in ReposeValveTest already.  I removed the one in my test.  I also renamed the test since I test both the request and the response body.  When I had originally named the class, I was only planning on testing the response body.

Also, fun fact, I noticed that both the test and the ReposeValveTest cleanupSpec methods were getting called.  I guess Spock goes through the effort of going through each parent to find a cleanupSpec method.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/intrafilterLogging/RequestLog.java
adrianjgeorge: Seems legit to throw an exception to me, but not directly `RuntimeException` some inheritor.

repose-aggregator/functional-tests/spock-functional-test/src/test/configs/features/core/intrafilterlogging/header-normalization.cfg.xml
adrianjgeorge: So personally i think functional tests weise there should be multiple tests, a set where it's in as much isolation as we can manage, and a set that more closely resembles it's live environment amongst a bunch of filters.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/intrafilterLogging/RequestLog.java
Mario-Lopez: Cool.  I changed it to IllegalArgumentException.

Mario-Lopez:
Down to 11 test failures. 😄

PR #1511, opened 2016-04-05
REP-3404 - Rename Basic Auth filter

Mario-Lopez:
Renamed the filter from "Rackspace Identity Basic Auth" to "Keystone v2 Basic Auth".


wdschei:
That was a very good exercise in how to use the Find & Replace features of the IDE. ;-)

PR #1514, opened 2016-04-12
REP-1618 Removing Mutable Wrappers and Old Pattern

dmnjohns:
Compilation and unit tests are passing now, but I imagine some things are still borked. I'll let the Jenkins job tell me what's up.


repose-aggregator/commons/commons-utilities/src/main/java/org/openrepose/commons/utils/http/normal/ExtendedStatusCodes.java
```diff
@@ -17,15 +17,11 @@
...
  */
-public enum FilterAction {
-
-    PASS,
-    RETURN,
-    PROCESS_RESPONSE,
-    NOT_SET
+public class ExtendedStatusCodes {
+    public static final int SC_TOO_MANY_REQUESTS = 429;
 }
```
adrianjgeorge: why did it think this was a rename?

dmnjohns: No idea...

repose-aggregator/commons/commons-utilities/src/main/java/org/openrepose/commons/utils/logging/apache/format/stock/ResponseBytesClfHandler.java
```diff
@@ -19,19 +19,34 @@
  */
 package org.openrepose.commons.utils.logging.apache.format.stock;
 
+import org.openrepose.commons.utils.StringUtilities;
```
adrianjgeorge: why would you use our string utils instead of one of the standard ones? we are trying to get rid of ours.

repose-aggregator/commons/commons-utilities/src/main/java/org/openrepose/commons/utils/logging/apache/format/stock/ResponseBytesHandler.java
```diff
@@ -19,17 +19,32 @@
  */
 package org.openrepose.commons.utils.logging.apache.format.stock;
 
+import org.openrepose.commons.utils.StringUtilities;
```
adrianjgeorge: oh god, you did it here too

dmnjohns: Well you've got to be consistent about things like this!

dmnjohns: I'm going to replace these both. I wasn't sure if we were moved towards or away from this class.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -99,6 +102,48 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
   def this(originalResponse: HttpServletResponse, headerMode: ResponseMode, bodyMode: ResponseMode) =
     this(originalResponse, headerMode, bodyMode, originalResponse.getOutputStream)
 
+  override def setStatus(i: Int, s: String): Unit = {
+    message = s
+    super.setStatus(i, s)
```
adrianjgeorge: Not sure calling this here is the right call. Seems like something that should happen on commit. Hmm, would it be based on the headers or the body mutability though. Shouldn't this fail if the response is committed?

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -99,6 +102,48 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
...
+  }
+
+  // todo: how does this affect the container processing invoked by sendError? should we track that this method was
+  // called? should commitToResponse take a parameter indicating if an error should bubble up?
+  /** Sets a status code and message on the first line of the HTTP response. The container may do additional processing
+    * to, for example, generate an error page as the response entity.
+    *
+    * Note that in any mutable mode, this method will call through to the [[sendError(i)]] method of the wrapped
```
adrianjgeorge: This seems like the opposite of the behavior we want?

adrianjgeorge: Looks like the code does the right thing, but the doc is backwards.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -99,6 +102,48 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
...
+
+  /** See [[sendError(i)]].
+    *
+    * @param i the status code to set
+    */
+  override def sendError(i: Int, s: String): Unit = {
+    committed = true
+    message = s
+    if (headerMode == ResponseMode.MUTABLE || bodyMode == ResponseMode.MUTABLE) {
+      super.setStatus(i)
```
adrianjgeorge: aren't you supposed to write to the body as well?

repose-aggregator/commons/commons-utilities/src/main/java/org/openrepose/commons/utils/logging/apache/format/stock/ResponseBytesClfHandler.java
wdschei: We should make a story to replace all uses of `org.openrepose.commons.utils.StringUtilities` with `org.apache.commons.lang3.StringUtils`.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: I call it here just to have a place to store the status. The status is 100% mutable, though. We can always call setStatus later to replace it. What does only setting the status on commit buy us?

As for failing if the response is committed, it probably should. The Javadocs didn't say anything about failing if the response has already been committed, even though it does call it out on other methods.

http://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpServletResponse.html#setStatus-int-

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: Will fix.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -316,6 +361,18 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
 
     // Track that the user intended to commit the response.
     flushedBuffer = true
+    committed = true
+  }
+
+  /**
+    * @throws IllegalStateException when the wrapped response has already been committed
+    */
+  def uncommit(): Unit = {
```
adrianjgeorge: Is there any case where the underlying won't be commited, but someone tries to call this when in non-mutable mode? If so we should check that as well.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -333,13 +390,20 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
       throw new IllegalStateException("Cannot call resetBuffer after the response has been committed")
     } else {
       super.resetBuffer()
+      // try to remove the content-type and content-length headers (best-effort)
+      Try {
+        removeHeader(CommonHttpHeader.CONTENT_TYPE.toString)
+        removeHeader(CommonHttpHeader.CONTENT_LENGTH.toString)
+      }
+      Try {
```
adrianjgeorge: won't this add them, even if we succeeded in removing them?

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -333,13 +390,20 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
...
+        setHeader(CommonHttpHeader.CONTENT_TYPE.toString, null)
+        setHeader(CommonHttpHeader.CONTENT_LENGTH.toString, null)
+      }
+      responseBodyType = ResponseBodyType.Available
       bodyOutputStream.resetBuffer()
     }
   }
 
-  /**
-    * @throws IllegalStateException when neither headerMode nor bodyMode is ResponseMode.MUTABLE
```
adrianjgeorge: Thought we weren't getting rid of this?

repose-aggregator/commons/commons-utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapperTest.scala
```diff
@@ -2003,5 +2013,137 @@ class HttpServletResponseWrapperTest extends FunSpec with BeforeAndAfter with Ma
...
+
+    it("should mark the wrapper as committed when given two arguments") {
+      val wrappedResponse = new HttpServletResponseWrapper(originalResponse, ResponseMode.PASSTHROUGH, ResponseMode.PASSTHROUGH)
+
+      wrappedResponse.sendError(418, "TEAPOT")
+
+      wrappedResponse.isCommitted shouldBe true
+    }
+
+    it("should not commit the underlying response if not in a mutable header mode") {
```
adrianjgeorge: what?

repose-aggregator/commons/commons-utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapperTest.scala
```diff
@@ -2003,5 +2013,137 @@ class HttpServletResponseWrapperTest extends FunSpec with BeforeAndAfter with Ma
...
+
+    it("should not commit the underlying response if in a mutable header mode") {
+      val wrappedResponse = new HttpServletResponseWrapper(originalResponse, ResponseMode.MUTABLE, ResponseMode.PASSTHROUGH)
+
+      wrappedResponse.sendError(418, "TEAPOT")
+
+      originalResponse.wasErrorSent() shouldBe false
+    }
+
+    it("should not commit the underlying response if not in a mutable body mode") {
```
adrianjgeorge: same problem

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: Is there a reason to keep it?

repose-aggregator/commons/commons-utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapperTest.scala
```diff
@@ -2003,5 +2013,137 @@ class HttpServletResponseWrapperTest extends FunSpec with BeforeAndAfter with Ma
...
+    it("should not commit the underlying response if in a mutable body mode") {
+      val wrappedResponse = new HttpServletResponseWrapper(originalResponse, ResponseMode.PASSTHROUGH, ResponseMode.MUTABLE)
+
+      wrappedResponse.sendError(418, "TEAPOT")
+
+      originalResponse.wasErrorSent() shouldBe false
+    }
+  }
+
+  describe("isCommitted") {
```
adrianjgeorge: should probably also have tests that confirm that it reports being committed in the right instances

repose-aggregator/core/repose-core/src/main/java/org/openrepose/core/services/rms/ResponseMessageServiceImpl.java
```diff
@@ -93,10 +96,19 @@ public boolean isInitialized() {
     }
 
     @Override
-    public void handle(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    public void handle(HttpServletRequest request, HttpServletResponseWrapper response) throws IOException {
         final StatusCodeMatcher matchedCode = getMatchingStatusCode(String.valueOf(response.getStatus()));
-        final MutableHttpServletRequest mutableRequest = MutableHttpServletRequest.wrap(request);
-        MediaRangeProcessor processor = new MediaRangeProcessor(mutableRequest.getPreferredHeaders("Accept", DEFAULT_TYPE));
+        final HttpServletRequestWrapper wrappedRequest = new HttpServletRequestWrapper(request);
```
adrianjgeorge: Since you are already passing in the wrapped response, why not go ahead and pass in the wrapped request, instead of immediately turning around and wrapping it?

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
```diff
@@ -350,72 +352,69 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
         final long startTime = System.currentTimeMillis();
         long streamLimit = containerConfigurationService.getContentBodyReadLimit();
 
-        final MutableHttpServletRequest mutableHttpRequest = MutableHttpServletRequest.wrap((HttpServletRequest) request, streamLimit);
-        final MutableHttpServletResponse mutableHttpResponse = MutableHttpServletResponse.wrap(mutableHttpRequest, (HttpServletResponse) response);
+        final HttpServletRequestWrapper wrappedRequest = new HttpServletRequestWrapper((HttpServletRequest) request,
+                new BufferedServletInputStream(new LimitedReadInputStream(streamLimit, request.getInputStream())));
+        final HttpServletResponseWrapper wrappedResponse = new HttpServletResponseWrapper((HttpServletResponse) response,
+                ResponseMode.PASSTHROUGH,
```
adrianjgeorge: this seems wrong, don't we need to change the content related headers potentially in rms? I guess we could leave it like this and be sure to call the set methods in rms. Either way it's imperative that the response is uncommited before we hand off to rms.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
```diff
@@ -350,72 +352,69 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
-            mutableHttpResponse.setLastException(use);
+            LOG.debug("{}:{} -- Invalid URI requested: {}", clusterId, nodeId, wrappedRequest.getRequestURI(), use);
+            wrappedResponse.sendError(HttpServletResponse.SC_BAD_REQUEST, "Error processing request");
         } catch (Exception ex) {
             LOG.error("{}:{} -- Exception encountered while processing filter chain.", clusterId, nodeId, ex);
-            mutableHttpResponse.sendError(HttpServletResponse.SC_BAD_GATEWAY, "Error processing request");
-            mutableHttpResponse.setLastException(ex);
+            wrappedResponse.sendError(HttpServletResponse.SC_BAD_GATEWAY, "Error processing request");
         } finally {
+            wrappedResponse.uncommit();
```
adrianjgeorge: add a comment saying why this needs to be here, because it does need to be.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilterChain.java
```diff
@@ -261,33 +255,30 @@ private String convertPojoToJsonString(Object object) throws IOException {
...
         );
 
         try {
-            if (isResponseOk(httpServletResponseWrapper)) {
-                containerFilterChain.doFilter(mutableHttpRequest, httpServletResponseWrapper);
-            }
-            if (isResponseOk(httpServletResponseWrapper)) {
-                // todo: rip out the mutable wrapper
-                router.route(new HttpServletRequestWrapper(mutableHttpRequest), httpServletResponseWrapper);
+            if (isResponseOk(wrappedResponse)) {
```
adrianjgeorge: undo this combining

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilterChain.java
```diff
@@ -261,33 +255,30 @@ private String convertPojoToJsonString(Object object) throws IOException {
...
+                containerFilterChain.doFilter(httpRequest, wrappedResponse);
+                router.route(new HttpServletRequestWrapper(httpRequest), wrappedResponse);
```
adrianjgeorge: does this still need to be rewrapped?

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: Yes. In fact, we never need to chain through to sendError since we wrap the request in the PowerFilter with a mode set to mutable.

Also, we can probably flush here as well.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: Sure, that could totally happen.

Do we care, though?

repose-aggregator/core/repose-core/src/main/java/org/openrepose/core/services/rms/ResponseMessageServiceImpl.java
dmnjohns: I was trying to scope where we wrap the request and response. I want it to be clear where we actually need the wrapper, like in this case, where we're only using it to gain access to the `getPreferredHeaders(...)` method. If we call this method conditionally, why should we wrap the request outside of this method?

On the whole, scoping things in this way give us less overhead, and I like the encapsulation.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
dmnjohns: Will do.

repose-aggregator/commons/commons-utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapperTest.scala
dmnjohns: Copy pasta'd the name. Apparently I changed the negation on the mode, but not on the commit. Will fix.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
adrianjgeorge: Stupid vague docs. it's clearly gone if they flushed though.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
adrianjgeorge: It gives a safety net if we later add other mutating methods and forget to make them throw an exception when they are operating when non-mutable.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/core/services/rms/ResponseMessageServiceImpl.java
adrianjgeorge: fair enough. I assumed as much, but i thought i'd bring it up.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: Fair enough. I'm trying to remember why I wanted to remove it, but I super can't. I want to say it would have been helpful if the Scripting filter conditionally chose a `ResponseMode` since it would prevent us from having to conditionally call this method.

Will re-add.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: Alright, I'll add in the Exception for things that don't specifically call out that they do nothing.

dmnjohns: According to the servlet 3.1 specification, "Headers set after the response is committed will be ignored by the servlet container."

This seems to imply that headers _can_ be set after the response is committed, they just won't be sent in the response. That is how the code is written now, so I'll leave it that way for the time being.

Also, "The sendError method will set the appropriate headers and content body for an
error message to return to the client. An optional String argument can be provided
to the sendError method which can be used in the content body of the error." That's basically all that the Servlet 3.1 specification says about `sendError(...)`. I still wonder if all containers set the HTTP status message, or if that's just something Jetty does. I'm not certain if we should set the HTTP status message or just the content body. For the moment, I've chained `sendError(i, s)` through to `setStatus(i, s)` so that we'll get the behavior of the container (although the `s` parameter of those methods does not necessarily represent the same thing).

wdschei: Internally the way the `setStatus(i, s)` variant is used seems to be wrong in several places. The `String` parameter is what the `ResponseMessageService` is keying off of. This looks like a bastardization of a deprecated call to get the desired behavior. When I ran across it a year or so ago, we didn't have time to fix it and we still haven't got back around to clean it up yet.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
dmnjohns: We do call the set methods, and I'm pretty sure this works. However, there isn't much of a reason not to make it mutable, and that gives us a little more power if we need it. So I'll change it.

Also the response is always uncommitted.

repose-aggregator/commons/commons-utilities/src/test/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapperTest.scala
dmnjohns: I'll add those.

jennyvo:
there're still test failed from common-utilities folder
shouldThrowExceptionForInvalidAddress

dmnjohns:
What test is that? I can't seem to find it.

As a side note, I am able to perform `gradle build` locally without any issues. Also, Jenkins seems to have made it past the unit testing phase, making me think all unit tests passed there as well. Maybe it's something in your environment?

jennyvo:
sorry it is a test case in IpAddressRangeTest

Mario-Lopez:
I wonder if that's the test that fails if you're not physically connected to the Rackspace network or VPN'd in.  @jennyvo, could you try connecting to the VPN and running the test again?

dmnjohns:
Ah. Knowing the test, what @Mario-Lopez said is very likely the case.

IpAddressRangeTest and ReposeJettySSLTest both fail on me locally at home. The latter because of the network issue, and the former because of the difference between the openssl command on modern Linux OSes versus Mac OS X.

jennyvo:
trying...

repose-aggregator/commons/commons-utilities/src/main/java/org/openrepose/commons/utils/logging/apache/LogFormatArgument.java
```diff
@@ -43,13 +43,13 @@
...
     STRING("STRING"),
-    ERROR_MESSAGE("M"),
+    RESPONSE_REASON("M"),
     RESPONSE_TIME_MICROSECONDS("D"),
     RESPONSE_TIME_SECONDS("T");
 
     private final String argument;
 
-    private LogFormatArgument(String argument) {
+    LogFormatArgument(String argument) {
```
adrianjgeorge: there a reason you changed this scope?

dmnjohns: I don't believe this is actually a change in scope, just dropping a redundant scope declaration. This is one of the things IntelliJ highlighted.

From https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html:
```
Note: The constructor for an enum type must be package-private or private access. It automatically creates the constants that are defined at the beginning of the enum body. You cannot invoke an enum constructor yourself.
```

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -99,6 +102,64 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
...
+    *
+    * @param i the status code to set
+    */
+  override def sendError(i: Int): Unit = {
+    // Set that status.
+    setStatus(i)
+
+    // Writes the response data and marks the response as committed.
+    flushBuffer()
+  }
```
adrianjgeorge: why not reset on this one too?

dmnjohns: It should 100% reset. I fix.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -261,6 +323,12 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
   def getOutputStreamAsInputStream: InputStream = bodyOutputStream.getOutputStreamAsInputStream
 
   /**
+    * @throws IllegalStateException when bodyMode is ResponseMode.PASSTHROUGH
+    */
+  def getOutputStreamAsString: String =
```
adrianjgeorge: why add this?

dmnjohns: It's just a convenience method. I was often getting the output stream as an input stream to then convert it to a string. It came up a lot in testing, but I can see it being useful in production code as well. I don't know that it has been used too much yet since I added it fairly late.

repose-aggregator/tests/functional-tests/src/test/groovy/features/core/proxy/RequestQueryParamTest.groovy
```diff
@@ -87,8 +87,11 @@ class RequestQueryParamTest extends ReposeValveTest {
 
         where:
         uriSuffixGiven                             | uriSuffixExpected                          | method
-        "/path/to/resource?key=value@%"            | "/path/to/resource?key=value%40%25"        | "GET"
-        "/path/to/resource?key=va%lu@e"            | "/path/to/resource?key=va%25lu%40e"        | "GET"
+        // todo: The '%' character breaks Jetty -- getQueryString returns the raw character while all of the
+        // getParameter methods return garbage. What do we want to do about this?
```
adrianjgeorge: they are illegal input from a url/query parameter perspective right? What happens in the system right now with them?

dmnjohns: Yes, they seem to be illegal characters unless encoded. However, it appears that these tests are passing now. I would bet that is because of the change to the way we HTTP Client. Where we used to call getParameterMap and use the values from that to add query parameters, we now use getQueryString and set that on the outgoing request. HTTP Client expects an unencoded string, which is what Jetty gives us, so HTTP Client performs the encoding for us.

From https://tools.ietf.org/html/rfc3986#section-2.4:
>  Because the percent ("%") character serves as the indicator for
>  percent-encoded octets, it must be percent-encoded as "%25" for that
>  octet to be used as data within a URI.

In other words, I think we answered what we want to do about this while we were doing other work. I'll uncomment these tests.

Mario-Lopez:
image

PR #1516, opened 2016-04-18
Use entities for illegal angle brackets in RPM %pre scripts

mikenerone:
This fixes broken sed commands in the RPM %pre scripts, which have been causing installation failures in cases where /etc/repose is already populated with config files at install time.


dmnjohns:
Note to team: make sure these changes are reflected on the repose-8-staging branch.

adrianjgeorge:
I wonder if these changes are necessary in the new branch. We aren't embedded inside of xml there, so it may be working as intended already there.

Mario-Lopez:
Confirmed this issue is resolved in Repose 8.

PR #1517, opened 2016-04-19
REP-3667 - URI Normalization config ordering

Mario-Lopez:
I added a unit test that consistently triggered the ordering issue when loading the URI Norm config, and I changed the scala HashMap to a scala LinkedHashMap in the filter which preserves the ordering of the config.


Mario-Lopez:
testing comments

PR #1518, opened 2016-04-20
REP-3765 Learn SystemD

wdschei:
Since the spike went so well and we got such great feedback from @mikenerone and @gregswift in a timely manner, this is now complete and fully functional.

This PR has a few extraneous things in it that are not specifically SystemD related, but they were in the context of keeping the old SysInit and new SystemD ways similar and being able to create fully testable artifacts.


repose-aggregator/artifacts/src/config/scripts/repose-valve-deb
```diff
@@ -33,18 +33,24 @@ CLEAN=/usr/bin/clean-repose-deploy
 # Can override the defaults in /etc/sysconfig
 . /etc/sysconfig/repose
```
gregswift: isn't this ```/etc/defaults/repose``` on debian?

wdschei: It should be by Debian convention, but it never has been.
The Redhat way has just always been forced on the Debian boxes.

repose-aggregator/artifacts/src/config/scripts/repose-valve-systemd
```diff
@@ -0,0 +1,65 @@
...
+fi
+
+VERSION=$("$JAVA_CMD" -version 2>&1 | grep ' version' | awk '{ print substr($3, 2, length($3)-2); }')
+JAVA_MINOR=$(echo $VERSION | tr "." " " | cut -d " " -f2)
+if [ "$JAVA_MINOR" -lt "8" ]; then
+    echo 'Repose requires a Java version of at least 8 to function.'
+    echo 'Please install a JRE 1.8 or greater.'
+    exit 2
+else
+    echo 'Java is at JRE 1.8 or greater.'
```
Mario-Lopez: I'm curious why you print something out for the success case here.  It doesn't hurt anything, just curious.

wdschei: Good question. This is the only one that prints a positive message. Removing now.

Mario-Lopez:
Everything else looks good.  Waiting on the build to merge.

PR #1521, opened 2016-04-21
Updated the Gradle build files to add the same info to the EAR and WAR manifests that were previously added to the Jar's.

wdschei:


Mario-Lopez:
I like it.

PR #1522, opened 2016-04-22
REP-3838 - IP User filter should check X-Forwarded-For header first

Mario-Lopez:
The IP User filter now grabs the first splittable value of the X-Forwarded-For header to populate the x-pp-user header if it's available.  Otherwise, it'll use the remote IP address in the request (like it previously did).

There were some unit tests for this filter already, but none of them verified that the IP address was put into the X-PP-User header.  I added some unit tests to verify this and the behavior for when the X-Forwarded-For header is in the request.

### DIFF
--- a/repose-aggregator/components/filters/ip-user-filter/src/test/scala/org/openrepose/filters/ipuser/IpUserFilterTest.scala
+++ b/repose-aggregator/components/filters/ip-user-filter/src/test/scala/org/openrepose/filters/ipuser/IpUserFilterTest.scala
@@ -42,11 +42,12 @@ class IpUserFilterTest extends FunSpec with BeforeAndAfter with Matchers {
     servletRequest = new MockHttpServletRequest
     servletResponse = new MockHttpServletResponse
     filterChain = new MockFilterChain
+
+    ipUserFilter.configurationUpdated(createConfig())
   }
 
   describe("setting the request x-pp-user header") {
     it("should use the value of the remote IP address of the request when no X-Forwarded-For header is included") {
-      ipUserFilter.configurationUpdated(createConfig())
       servletRequest.setRemoteAddr("10.1.2.3")
 
       ipUserFilter.doFilter(servletRequest, servletResponse, filterChain)
@@ -55,7 +56,6 @@ class IpUserFilterTest extends FunSpec with BeforeAndAfter with Matchers {
     }
 
     it("should use the value of the X-Forwarded-For header when it is included") {
-      ipUserFilter.configurationUpdated(createConfig())
       servletRequest.setRemoteAddr("10.1.2.3")
       servletRequest.addHeader("X-Forwarded-For", "10.55.66.77")
 
@@ -65,7 +65,6 @@ class IpUserFilterTest extends FunSpec with BeforeAndAfter with Matchers {
     }
 
     it("should use the first value of the X-Forwarded-For header when there are multiple non-split values") {
-      ipUserFilter.configurationUpdated(createConfig())
       servletRequest.setRemoteAddr("10.1.2.3")
       servletRequest.addHeader("X-Forwarded-For", "10.123.123.123")
       servletRequest.addHeader("X-Forwarded-For", "10.55.66.77")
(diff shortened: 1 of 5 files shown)


Mario-Lopez:
Has the expected 13 failures, but QE review hasn't been started.

repose-aggregator/components/filters/ip-user-filter/src/test/scala/org/openrepose/filters/ipuser/IpUserFilterTest.scala
```diff
@@ -0,0 +1,100 @@
...
+    ipUserFilter = new IpUserFilter(null)
+    servletRequest = new MockHttpServletRequest
+    servletResponse = new MockHttpServletResponse
+    filterChain = new MockFilterChain
+  }
+
+  describe("setting the request x-pp-user header") {
+    it("should use the value of the remote IP address of the request when no X-Forwarded-For header is included") {
+      ipUserFilter.configurationUpdated(createConfig())
+      servletRequest.setRemoteAddr("10.1.2.3")
```
wdschei: Why are these two lines that are in the top of each test not in the `before`?
If it does need to be accomplished before each test, then why not also use the `BeforeAndAfterEach` trait which provides `beforeEach`?

Mario-Lopez: I wrote it like that when I thought I needed to write more tests around different configurations.  I'll move it into the before block.

wdschei: After a second look at the `BeforeAndAfter` trait, it's `before` method is run before each test. The nuance between the two is that a failure in the `before` method will abort the rest of the suite marking the remainder as skipped and a failure in the `beforeEach` method will continue to fail each test of the suite.

Maybe we should be using only the `BeforeAndAfterAll` and `BeforeAndAfterEach` traits instead of the `BeforeAndAfter` trait across the entire codebase. There was also a weird deprecation and then reuse of the `BeforeAndAfter` trait between ScalaTest 1.0 and 2.0. That lends even more to not using it, if only for the explicitness of the other two.

Thoughts?

Mario-Lopez: I'm not sure if we'd want to remain consistent with past behavior or forge on ahead with the more explicit names.  Perhaps a dev meeting topic?

wdschei: Agreed. For consistency, we should probably leave it the same. I also concur that we should probably discuss this further as a team.

wdschei: So just moving those two lines out of the individual tests and into the `before` method should do it.

repose-aggregator/tests/functional-tests/src/test/groovy/features/filters/ipuser/IpUserTest.groovy
```diff
@@ -117,4 +117,65 @@ class IpUserTest extends ReposeValveTest {
...
+        MessageChain mc = deproxy.makeRequest(url: reposeEndpoint, method: 'get', headers: headers)
+        def sentRequest = mc.handlings[0]
+
+        then: "Repose will send x-pp-group with the configured value"
+        mc.handlings.size() == 1
+
+        and: "Repose will x-pp-user base on x-forward-for header plus default quality"
+        sentRequest.request.headers.findAll("x-pp-user").size() == 1
+        sentRequest.request.headers.getFirstValue("x-pp-user") == '192.25.25.15;q=0.4'
+    }
```
wdschei: These new tests and the `def "classifying a request by its IP"()` test don't need to restart Repose in between tests do they?

jennyvo: failed when not restart Repose. I don't know why

jennyvo: will try again

Mario-Lopez:
Got the expected 12 test failures.

PR #1524, opened 2016-04-26
REP-3311 - Fix RPM installer

Mario-Lopez:
The install script for the RPM installer was encoded to be in an XML file which it isn't anymore, so I fixed that.

Also, the filter RPMs wouldn't install because they were trying to create directories that already existed (from the valve RPM).  I updated the Gradle build to have the filter RPMs not try to create the directories.

I was able to install the valve RPM and the three filter RPMs in a Vagrant CentOS 7 VM.


Mario-Lopez:
Got the expected 12 test failures.

PR #1525, opened 2016-04-29
Rep 3842 Munging Filter

adrianjgeorge:
So far just the config, but the rest will come later


repose-aggregator/components/filters/munging-filter/src/main/resources/META-INF/schema/config/munging.xsd
```diff
@@ -0,0 +1,176 @@
...
+        <xs:annotation>
+            <xs:documentation>
+                <html:p>
+                    The set of patches to be applied for a given url path.
+                </html:p>
+            </xs:documentation>
+        </xs:annotation>
+
+        <xs:sequence>
+            <xs:element name="header-filter" type="HeaderFilter" minOccurs="0" maxOccurs="1"/>
```
Mario-Lopez: Why limit to 1?

adrianjgeorge: Because they only need one, and i don't want to deal with the crappy way jaxb handles lists more than i have to

repose-aggregator/components/filters/munging-filter/src/main/resources/META-INF/schema/examples/munging.cfg.xml
```diff
@@ -0,0 +1,65 @@
...
+            <!--<xml>-->
+                <!--<![CDATA[-->
+                <!--<diff>-->
+                    <!--<replace sel="example/greeting/text()">Hello XML patch</replace>-->
+                <!--</diff>-->
+                <!--]]>-->
+            <!--</xml>-->
+        </response>
+    </change>
+    <change path="/foo/bar/*">
```
Mario-Lopez: Should be .* I assume.  Other than that, this works.

adrianjgeorge: I have no idea what you are talking about

repose-aggregator/components/filters/munging-filter/src/main/resources/META-INF/schema/config/munging.xsd
```diff
@@ -0,0 +1,176 @@
...
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+
+        <xs:assert vc:minVersion="1.1"
+                   xpathDefaultNamespace="##targetNamespace"
+                   test="request or response"
+                   xerces:message="You must specify request and/or response behavior"
+                   saxon:message="You must specify request and/or response behavior">
+        </xs:assert>
```
dmnjohns: I think there's a way to get the same behavior without relying on an assert. Something like...

```
<xs:sequence>
    <xs:choice>
        <xs:sequence>
            <xs:element name="request" type="Patch" minOccurs="1" maxOccurs="1"/>
            <xs:element name="response" type="Patch" minOccurs="0" maxOccurs="1"/>
        </xs:sequence>
        <xs:element name="response" type="Patch" minOccurs="1" maxOccurs="1"/>
    </xs:choice>
</xs:sequence>
```

I made the minOccurs and maxOccurs explicit, but in some places, they don't need to be. I didn't want to pull up the docs to figure out what places, though.

adrianjgeorge: That works, but is much harder to read.

adrianjgeorge:
le sigh, another one that i pointed at master

PR #1526, opened 2016-05-02
REP-3843 Body Extractor to Header filter

wdschei:


repose-aggregator/components/filters/body-extractor-to-header/src/main/resources/META-INF/schema/config/body-extractor-to-header.xsd
```diff
@@ -0,0 +1,115 @@
...
+        <xs:attribute name="jsonpath" type="xs:string" use="required">
+            <xs:annotation>
+                <xs:documentation>
+                    <html:p>
+                        The JPath to be applied to the Body in order to extract the desired value.
+                    </html:p>
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+        <xs:attribute name="default" type="xs:string" use="optional">
```
dmnjohns: I don't love naming this attribute "default". I worry that name is overloaded with meaning. Also, while it behaves much like "nullValue", there isn't a clear link between the two, which may be helpful.

"onNull" and "onMiss" were proposed last week. I would also throw "ifNull" and "ifAbsent" into the mix.

I'll let @adrianjgeorge give his opinion on this so that it doesn't come back later.

repose-aggregator/components/filters/body-extractor-to-header/src/main/resources/META-INF/schema/config/body-extractor-to-header.xsd
```diff
@@ -0,0 +1,115 @@
...
+    <xs:element name="body-extractor-to-header" type="BodyExtractorToHeaderConfig"/>
+
+    <xs:complexType name="BodyExtractorToHeaderConfig">
+        <xs:annotation>
+            <xs:documentation>
+                <html:p>The root config type for the Body Extractor to Header filter configuration file.</html:p>
+            </xs:documentation>
+        </xs:annotation>
+        <xs:sequence>
+            <xs:element name="extraction" type="Extractor" minOccurs="1" maxOccurs="unbounded"/>
```
dmnjohns: I would also consider either changing this name to "extractor" or changing the root element to "body-extraction-to-header". Especially since the type is "Extractor". Not necessary, but I like the consistency.

dmnjohns:
Apart from my prior comments, the configuration looks good to me. 👍

repose-aggregator/components/filters/body-extractor-to-header/src/main/resources/META-INF/schema/config/body-extractor-to-header.xsd
Mario-Lopez: I had the same concern, but I thought it was fine since our other "X extract to header" filters use the same attribute name "default" to mean that.  Either consistent between the two attributes or consistent with the other filters.

That being said, I like ifNull/ifAbsent.

repose-aggregator/components/filters/body-extractor-to-header/src/main/resources/META-INF/schema/config/body-extractor-to-header.xsd
Mario-Lopez: I actually like the idea that the extractor performs extractions.  Not that it matters.  Naming things, amirite?

dmnjohns: Fair enough.

Generally speaking, naming things doesn't matter from a functional perspective, but certainly does from a usability perspective. In this case, it probably doesn't matter.

repose-aggregator/components/filters/body-extractor-to-header/src/main/resources/META-INF/schema/config/body-extractor-to-header.xsd
adrianjgeorge: default has a clear and obvious meaning on those other filters, where here it's a little ambiguous. I prefer them to both be explicit. I'm fine with if or on. We could even split the difference and use `defaultIfMiss` and `defaultIfNull`.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
```diff
@@ -0,0 +1,152 @@
...
+  private final val DEFAULT_CONFIG = "body-extractor-to-header.cfg.xml"
+  private final val SCHEMA_FILE_NAME = "/META-INF/schema/config/body-extractor-to-header.xsd"
+
+  case class Extraction(headerName: String,
+                        jsonPath: String,
+                        defaultValue: Option[String],
+                        nullValue: Option[String],
+                        overwrite: Boolean,
+                        quality: Option[java.lang.Double]
+                       )
```
dmnjohns: I think then Scala style guide suggests terminating the parameter list on the same line as the last parameter. See: 
http://docs.scala-lang.org/style/declarations.html#classes
http://docs.scala-lang.org/style/indentation.html#methods-with-numerous-arguments

Not a huge deal, but a thing we can totally do consistently.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
```diff
@@ -0,0 +1,152 @@
...
+import org.openrepose.core.services.config.ConfigurationService
+import org.openrepose.filters.bodyextractortoheader.config.BodyExtractorToHeaderConfig
+
+import scala.collection.JavaConverters._
+import scala.io.Source
+import scala.util.{Failure, Success, Try}
+
+@Named
+class BodyExtractorToHeaderFilter @Inject()(configurationService: ConfigurationService)
+  extends Filter with UpdateListener[BodyExtractorToHeaderConfig] with LazyLogging {
```
dmnjohns: You could use @adrianjgeorge's new abstract filter to cut down on some of the code in here:
https://github.com/rackerlabs/repose/blob/REP-3842/repose-aggregator/components/filters/munging-filter/src/main/scala/org/openrepose/filters/munging/AbstractConfiguredFilter.scala

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  override def destroy(): Unit = {
+    logger.trace("Body Extractor to Header filter destroying...")
+    configurationService.unsubscribeFrom(configurationFile, this.asInstanceOf[UpdateListener[_]])
+    logger.trace("Body Extractor to Header filter destroyed.")
+  }
+
+  override def configurationUpdated(config: BodyExtractorToHeaderConfig): Unit = {
+    extractions = config.getExtraction.asScala.map { extraction =>
```
dmnjohns: Another style guide thing. I'd drop the period before map. See:
http://docs.scala-lang.org/style/method-invocation.html#higher-order-functions

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
```diff
@@ -0,0 +1,152 @@
...
+      Extraction(extraction.getHeader,
+        extraction.getJsonpath,
+        Option(extraction.getDefaultIfMiss),
+        Option(extraction.getDefaultIfNull),
+        extraction.isOverwrite,
+        Option(extraction.getQuality)
+      )
```
dmnjohns: Moar style.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
```diff
@@ -0,0 +1,152 @@
...
+@Named
+class BodyExtractorToHeaderFilter @Inject()(configurationService: ConfigurationService)
+  extends Filter with UpdateListener[BodyExtractorToHeaderConfig] with LazyLogging {
+
+  import BodyExtractorToHeaderFilter._
+
+  private var configurationFile: String = DEFAULT_CONFIG
+  private var initialized = false
+  private var extractions: Iterable[Extraction] = _
+  val jsonPathConfiguration = JsonConfiguration.defaultConfiguration()
```
dmnjohns: I'm not sure why this is the only variable which isn't private?

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
```diff
@@ -0,0 +1,152 @@
...
+
+    def addHeader(name: String, value: String, quality: Option[java.lang.Double], overwrite: Boolean): Unit = {
+      if (overwrite) {
+        mutableHttpRequest.removeHeader(name)
+      }
+      val hdrValue = quality match {
+        case Some(qual) => s"$value;q=$qual"
+        case None => value
+      }
+      mutableHttpRequest.addHeader(name, hdrValue)
```
dmnjohns: Rather than appending the quality by hand, just call the three-argument `addHeader` method.

```
quality match {
  case Some(q) => mutableHttpRequest.addHeader(name, value, q)
  case None => mutableHttpRequest.addHeader(name, value)
}
```

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
```diff
@@ -0,0 +1,152 @@
...
+          } else {
+            new Failure(new Exception)
+          }
+        case _ =>
+          logger.debug("Content-Type was not defined.")
+          new Failure(new Exception)
+      }
+    }
+
+    extractions.foreach { extraction =>
```
dmnjohns: Another style thing.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
```diff
@@ -0,0 +1,152 @@
...
+            val jsonString = Source.fromInputStream(is).mkString
+            if (is.markSupported()) is.reset()
+            logger.trace(s"is.available = ${is.available()}")
+            Try(JsonPath.using(jsonPathConfiguration).parse(jsonString))
+          } else {
+            new Failure(new Exception)
+          }
+        case _ =>
+          logger.debug("Content-Type was not defined.")
+          new Failure(new Exception)
```
dmnjohns: You shouldn't need the `new` keyword before `Failure`. `Failure` should have a defined `apply` method that handles it for you.

That being said, I would suggest making `jsonDoc` an `Option` of a `Try`. After all, that is what comes out of this naturally. 
```
val jsonDoc: Option[Try[DocumentContext]] = Option(mutableHttpRequest.getContentType) filter { contentType =>
  contentType.toLowerCase.contains("json")
} map { contentType =>
  val is = mutableHttpRequest.getInputStream
  if (is.markSupported()) is.mark(0)
  val jsonString = Source.fromInputStream(is).mkString
  if (is.markSupported()) is.reset()
  logger.trace(s"is.available = ${is.available()}")
  Try(JsonPath.using(jsonPathConfiguration).parse(jsonString))
}
```

Then below, you'll probably end up checking `jsonDoc.isDefined && jsonDoc.get.isSuccess`.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
```diff
@@ -0,0 +1,152 @@
...
+      }
+      mutableHttpRequest.addHeader(name, hdrValue)
+    }
+
+    val jsonDoc: Try[DocumentContext] = {
+      Option(mutableHttpRequest.getContentType) match {
+        case Some(contentType) =>
+          if (contentType.toLowerCase.contains("json")) {
+            val is = mutableHttpRequest.getInputStream
+            if (is.markSupported()) is.mark(0)
```
dmnjohns: Oddly, I can't find the behavior of `mark` with an argument of 0 in the JavaDocs. I assume this means there is no read limit?

Mario-Lopez: _(Unrelated to this code review)_
If so, maybe we should also be doing that here:
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/core/core-lib/src/main/java/org/openrepose/powerfilter/intrafilterLogging/RequestLog.java#L57

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
```diff
@@ -0,0 +1,152 @@
...
+            val jsonString = Source.fromInputStream(is).mkString
+            if (is.markSupported()) is.reset()
```
dmnjohns: What if `mark` isn't supported? We shouldn't consume the body in this filter. You might have to read the input stream into your own input stream that does support mark, then pass that to the `HttpServletRequestWrapper` that will be passed down the chain.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
```diff
@@ -0,0 +1,152 @@
...
+        case _ =>
+          logger.debug("Content-Type was not defined.")
+          new Failure(new Exception)
+      }
+    }
+
+    extractions.foreach { extraction =>
+      if (jsonDoc.isSuccess) {
+        val extracted = jsonDoc match {
+          case Success(doc) => Try(doc.read[Any](extraction.jsonPath))
```
dmnjohns: Why do the `jsonDoc.isSuccess` check if you're going to match here anyway? Just stick everything in a match. Of course, that means you'll have to handle the other case, but it can just do nothing as it does now.

Also, match will probably be cleared if you switch `jsonDoc` to an `Option[Try[DocumentContext]`.

```
jsonDoc match {
  case Some(Success(doc)) => 
    val extracted = Try(doc.read[Any](extraction.jsonPath))
    ...
  case _ => 
    // do nothing
}
```

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
```diff
@@ -0,0 +1,152 @@
...
+          case (Success(headerValue), _, _) if Option(headerValue).isDefined =>
+            addHeader(extraction.headerName, headerValue.toString, extraction.quality, extraction.overwrite)
+          // JSONPath value was extracted AND is Null AND NullValue is defined
+          case (Success(headerValue), _, Some(nullValue)) =>
+            addHeader(extraction.headerName, nullValue, extraction.quality, extraction.overwrite)
+          // JSONPath value was NOT extracted AND DefaultValue is defined
+          case (Failure(e), Some(defaultValue), _) =>
+            addHeader(extraction.headerName, defaultValue, extraction.quality, extraction.overwrite)
+          case (_, _, _) => // don't add a header
+        }
```
dmnjohns: I like this block of code. It's a good use of tuples, matching, and guards.

repose-aggregator/components/filters/body-extractor-to-header/src/test/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilterTest.scala
```diff
@@ -0,0 +1,452 @@
...
+  val extractedHeader = "X-Extracted-Id"
+
+  before {
+    reset(mockConfigurationService)
+    config = new BodyExtractorToHeaderConfig
+    filter = new BodyExtractorToHeaderFilter(mockConfigurationService)
+    requestCaptor = ArgumentCaptor.forClass(classOf[MutableHttpServletRequest])
+    servletRequest = new MockHttpServletRequest
+    servletResponse = new MockHttpServletResponse
+    filterChain = mock(classOf[FilterChain])
```
dmnjohns: I haven't looked at how this is used too much yet, but you may consider using Spring's `MockFilterChain` in place of a Mockito mock and the `ArgumentCaptor`.

repose-aggregator/components/filters/body-extractor-to-header/src/test/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilterTest.scala
```diff
@@ -0,0 +1,452 @@
...
+import org.springframework.mock.web.{MockHttpServletRequest, MockHttpServletResponse}
+
+import scala.collection.JavaConverters._
+
+@RunWith(classOf[JUnitRunner])
+class BodyExtractorToHeaderFilterTest extends FunSpec with BeforeAndAfter with Matchers {
+
+  val mockConfigurationService = mock(classOf[ConfigurationService])
+  val mockFilterConfig = new MockFilterConfig
+  var config: BodyExtractorToHeaderConfig = _
```
dmnjohns: Mixing val and var declarations isn't my favorite. I generally do a block of vals followed by a block of vars to keep things straight. That might be a personal preference thing.

repose-aggregator/components/filters/body-extractor-to-header/src/test/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilterTest.scala
```diff
@@ -0,0 +1,452 @@
...
+        |      ],
+        |      "true": true,
+        |      "false": false,
+        |      "null": null
+        |    },
+        |    "regex": {
+        |      "group": "hybrid:123456"
+        |    }
+        |  }
+        |}""".stripMargin.getBytes(StandardCharsets.UTF_8)
```
dmnjohns: Either these bytes should be an ISO-8859-1, not UTF-8, representation, or when the Content-Type gets set, it should declare the charset to be UTF-8. By default, HTTP uses ISO-8859-1 encoding. See:
https://www.w3.org/International/articles/http-charset/index

dmnjohns:
Of course, the Gradle PR builder fails for this branch, and the Maven PR builder is disabled. After you've made changes and are happy with it, I'd suggest manually kicking a Maven PR build and linking the results as a comment.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
wdschei: Updated.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
wdschei: I copy/pasted it into this project with a `TODO` to remove it when these are brought into `repose-8-staging`.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
wdschei: Updated.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
wdschei: Updated.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
wdschei: It originally wasn't `private` because it is a `val` which is `final` by definition, so there wasn't a chance of something modifying it. But it doesn't need to have expanded scope, so it it now `private` too.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
wdschei: This should totally be done when we bring it into `repose-8-staging` and have access to the new wrappers, but for now that is not an option. I added a `TODO` to this stating as much. ;-)

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
wdschei: Updated.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
wdschei: D'oh. Yep the `new` is totally not needed. That is my Java sneaking in. ;-)
I'm not sure the `Option[Try[DocumentContext]]` is any better of a solution, but it isn't any worse.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
wdschei: I couldn't find it documented anywhere about `InputStream.mark(0)` being unlimited either, but it has always worked. That being said, I updated it to `Integer.MAX_VALUE` which is actually correct.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
wdschei: This too should totally be done when we bring it into `repose-8-staging` and have access to the new wrappers, but for now this is not an option. I added a `TODO` to this stating as much. ;-)

https://github.com/rackerlabs/repose/blob/REP-3843_BodyExtractorToHeader/repose-aggregator/commons/utilities/src/main/java/org/openrepose/commons/utils/servlet/http/MutableHttpServletRequest.java#L124-L125

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
wdschei: Good call on the extra `isSuccess` check, but by changing the `jsonDoc` to `Option[Try[DocumentContext]` then I end up doing a `get.get` and that looks kind of odd.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
wdschei: This was totally ripped off from @Mario-Lopez's `UrlExtractorToHeaderFilter`.

repose-aggregator/components/filters/body-extractor-to-header/src/test/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilterTest.scala
wdschei: This was also totally ripped off from @Mario-Lopez's `UrlExtractorToHeaderFilterTest`. That said I'm not for or against using `ArgumentCaptor`'s so I updated this to use `MockFilterChain`.

repose-aggregator/components/filters/body-extractor-to-header/src/test/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilterTest.scala
wdschei: All of the `var`'s were moved to a block below the `val`'s.

repose-aggregator/components/filters/body-extractor-to-header/src/test/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilterTest.scala
wdschei: Updated.

wdschei:
Retest this please.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
dmnjohns: Derp. My bad. But yay for the TODO.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
```diff
@@ -0,0 +1,121 @@
...
+      val hdrValue = quality match {
+        case Some(qual) => s"$value;q=$qual"
+        case None => value
+      }
+      mutableHttpRequest.addHeader(name, hdrValue)
+      //val hdrValue = quality match {
+      //  case Some(qual) => mutableHttpRequest.addHeader(name, hdrValue, qual)
+      //  case None => mutableHttpRequest.addHeader(name, hdrValue)
+      //}
+      //////////
```
dmnjohns: You probably don't need to set the `hdrValue` val either since nothing will be done with it.

wdschei: That is true.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
```diff
@@ -0,0 +1,121 @@
...
+import com.jayway.jsonpath.{DocumentContext, JsonPath, Configuration => JsonConfiguration, Option => JsonOption}
+import org.openrepose.commons.utils.servlet.http.MutableHttpServletRequest
+import org.openrepose.core.services.config.ConfigurationService
+import org.openrepose.filters.bodyextractortoheader.config.BodyExtractorToHeaderConfig
+
+import scala.collection.JavaConverters._
+import scala.io.Source
+import scala.util.{Failure, Success, Try}
+
+class BodyExtractorToHeaderFilter(configurationService: ConfigurationService)
```
dmnjohns: Why did the dependency management annotations get dropped? Am I missing something here?

wdschei: It is in the parent `AbstractConfiguredFilter`.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
dmnjohns: Yeah, I'm not a fan of the `get.get`. I'd still suggest just going with the match.

```
jsonDoc match {
  case Some(Success(doc)) => 
    val extracted = Try(doc.read[Any](extraction.jsonPath))
    
    (extracted, extraction.defaultValue, extraction.nullValue) match {
        // JSONPath value was extracted AND is NOT Null
        case (Success(headerValue), _, _) if Option(headerValue).isDefined =>
            addHeader(extraction.headerName, headerValue.toString, extraction.quality, extraction.overwrite)
        // JSONPath value was extracted AND is Null AND NullValue is defined
        case (Success(headerValue), _, Some(nullValue)) =>
            addHeader(extraction.headerName, nullValue, extraction.quality, extraction.overwrite)
        // JSONPath value was NOT extracted AND DefaultValue is defined
        case (Failure(e), Some(defaultValue), _) =>
            addHeader(extraction.headerName, defaultValue, extraction.quality, extraction.overwrite)
        case (_, _, _) => // don't add a header
    }
  case _ => 
    // do nothing
}
```

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
adrianjgeorge: i think he meant the dependency injection annotations? which you do still totally need

adrianjgeorge: and seemingly left off mine so far

wdschei: They are pulled in by the parent?

adrianjgeorge: unless the tests are working, in which case there's a little more spring magic in our other fix than i was expecting

wdschei: We'll see shortly, but it seemed to work locally.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
```diff
@@ -0,0 +1,121 @@
...
+  extends AbstractConfiguredFilter[BodyExtractorToHeaderConfig](configurationService) {
+  override val DEFAULT_CONFIG: String = "body-extractor-to-header.cfg.xml"
+  override val SCHEMA_LOCATION: String = "/META-INF/schema/config/body-extractor-to-header.xsd"
```
dmnjohns: The Scala style guide recommends a different format for constants:
http://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods

Although, since these are inherited from the abstract filter, they might be legit. I guess the question is, can we use the abstract filter in Java? If so, then this is fine. If not, I'd adopt the Scala style.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
```diff
@@ -0,0 +1,122 @@
...
+
+  override def configurationUpdated(configurationObject: BodyExtractorToHeaderConfig): Unit = {
+    extractions = configurationObject.getExtraction.asScala map {
+      extraction =>
+        Extraction(extraction.getHeader,
+          extraction.getJsonpath,
+          Option(extraction.getDefaultIfMiss),
+          Option(extraction.getDefaultIfNull),
+          extraction.isOverwrite,
+          Option(extraction.getQuality))
```
dmnjohns: Why'd the formatting change here? `extraction =>` got stuck on a new line.

wdschei:
Retest this please.

repose-aggregator/components/filters/body-extractor-to-header/src/main/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderFilter.scala
wdschei: IntelliJ magic?

dmnjohns:
Last tests pass, and the only change was text formatting.

I'm merging this now. 👍

PR #1527, opened 2016-05-02
REP-2470 - Intrafilter logging when request input stream is closed

Mario-Lopez:
Since the request's input stream is sometimes being closed by the API Validator filter (more likely in API Checker), turning on IntraFilter logging can be risky since the request will bomb out when we try to read the closed input stream.

I added a try/catch for IOException around the code that reads the input stream so we can log a warning when this happens and carry on with the request.  Since I was able to easily add in logging without impacting what gets printed out to the log, I also updated the mark/reset error to log a warning instead of throwing an exception.

Since this only happens in a very specific case (using the XBOX huge Feeds WADL), I'm not sure how to write a functional test around this.  I...  tested by hand to verify it works (in addition to a unit test). 😨


dmnjohns:
So record what we can, and log an error if we can't record everything. That makes sense to me. The code looks good. If you rebase off of repose-8-staging again, you should be able to eliminate that one test failure.

Will merge once QE signs off. 👍

Mario-Lopez:
retest this please

dmnjohns:
It appears something strange is happening with the PR builder for this branch.

> * What went wrong:
Cannot infer Groovy class path because no Groovy Jar was found on class path: file collection

Odd.

jennyvo:
moved test to right location but build still failed from my local with different issue

dmnjohns:
It looks like the issue is that some changes which made it into master did not make it into repose-8-staging.

Take, for example, the `ReposeLintWReposeVerTest`. Since it was a new file that appeared on master after `repose-8-staging` branched, it merged in cleanly when @wdschei merged master in `repose-8-staging`. `CommandExecutor`, on the other hand, existed when `repose-8-staging` branched off of master. It was changed further after that branching point, but those changes are not reflected in `repose-8-staging`. It would appear that Git handled the merge by dropping those changes for whatever reason. It's worth noting that the `CommandExecutor` file moved, which likely led to a case where the file was moved before it was changed, then it was changed, then the move attempted to be rebased on the change and Git ultimately just dropped the change.

Mario-Lopez:
So what you're saying is... it got the 3 expected failures. :P

dmnjohns:
I suppose I am saying that.

Okay then... merginating!

PR #1531, opened 2016-05-09
Abstract Configured Filter

adrianjgeorge:
The abstract configured filter in it's own pr so we can pull it in sooner


adrianjgeorge:
Whoops, accidentally forgot to mark it to not got to master

PR #1532, opened 2016-05-09
Abstract Configured Filter

adrianjgeorge:
The abstract configured filter in it's own pr so we can pull it in sooner

### DIFF
--- a/repose-aggregator/core/repose-core-api/src/main/scala/org/openrepose/core/filter/AbstractConfiguredFilter.scala
+++ b/repose-aggregator/core/repose-core-api/src/main/scala/org/openrepose/core/filter/AbstractConfiguredFilter.scala
@@ -30,18 +30,37 @@ import org.openrepose.core.services.config.ConfigurationService
 import scala.reflect.{ClassTag, _}
 
 /**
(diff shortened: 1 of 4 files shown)


repose-aggregator/core/repose-core-api/src/main/scala/org/openrepose/core/filter/AbstractConfiguredFilter.scala
```diff
@@ -0,0 +1,132 @@
...
+  /**
+    * Subscribes with the configuration service. If the filter config doesn't have a custom file name,
+    * it uses the name provided by DEFAULT_CONFIG. It tries to load and use the schema provided by SCHEMA_LOCATION.
+    * @param filterConfig
+    */
+  override def init(filterConfig: FilterConfig): Unit = {
+    logger.trace("{} initializing ...", this.getClass.getSimpleName)
+    configFile = new FilterConfigHelper(filterConfig).getFilterConfig(DEFAULT_CONFIG)
+
+    logger.info("Initializing filter using config " + configFile)
```
dmnjohns: String concatenation rather than interpolation.

I was told to put this comment over here.

PR #1533, opened 2016-05-09
REP-3842 Body Patcher Filter

adrianjgeorge:
It's done, maybe?


repose-aggregator/components/filters/body-patcher-filter/src/main/resources/META-INF/schema/config/body-patcher.xsd
```diff
@@ -0,0 +1,142 @@
...
+        <xs:annotation>
+            <xs:documentation>
+                <html:p>
+                    The root config type for the Munging filter configuration file.
+                </html:p>
+            </xs:documentation>
+        </xs:annotation>
+
+        <xs:sequence>
+            <xs:element name="change" type="ChangeDetails" minOccurs="1" maxOccurs="unbounded"/>
```
dmnjohns: Was there a specific reason for naming this `change` rather than `patch`? Not that it matters, I'm just curious.

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
```diff
@@ -0,0 +1,150 @@
...
+          httpRequest
+        }
+      case Xml =>
+        //todo: implement xml support
+        httpRequest
+      case Other =>
+        httpRequest
+    }
+
+
```
dmnjohns: Two blank lines?! Something went down here.

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
```diff
@@ -0,0 +1,150 @@
...
+  */
+@Named
+class BodyPatcherFilter @Inject()(configurationService: ConfigurationService)
+  extends AbstractConfiguredFilter[BodyPatcherConfig](configurationService) with LazyLogging {
+  override val DEFAULT_CONFIG: String = "body-patcher.cfg.xml"
+  override val SCHEMA_LOCATION: String = "/META-INF/schema/config/body-patcher.xsd"
+
+  override def doWork(request: ServletRequest, response: ServletResponse, chain: FilterChain): Unit = {
+    val httpRequest: HttpServletRequest = request.asInstanceOf[HttpServletRequest]
+    val httpResponse: HttpServletResponse = response.asInstanceOf[HttpServletResponse]
```
dmnjohns: Local variables don't really need type annotations (usually). See:
http://docs.scala-lang.org/style/types.html

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
```diff
@@ -0,0 +1,150 @@
...
+          }
+        case Xml => //todo implement xml support
+        case Other =>
+      }
+      wrappedResponse.commitToResponse()
+    }
+  }
+
+  def filterChanges(request: HttpServletRequest): List[ChangeDetails] = {
+    val urlPath: String = new URL(request.getRequestURL.toString).getPath
```
dmnjohns: Why not use `request.getRequestURI` instead? It should just return the path string. See:
https://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpServletRequest.html#getRequestURI--

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
```diff
@@ -0,0 +1,150 @@
...
+    configuration.getChange.asScala.toList
```
dmnjohns: I wonder what `.asScala` returns here. I'm guessing it's an `Iterable` of some type. Why make it a `List` explicitly?

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
```diff
@@ -0,0 +1,150 @@
...
+        .filter(_.getPath.r.findFirstIn(urlPath).isDefined)
```
dmnjohns: I should note that this approach to regex matching only requires that the configured regex matches some portion of the actual string. It's an issue that has come up before.

For example, if I've configured a change to occur on `http://example.com/` and the request URI is `http://example.com/foo/bar` I might not expect the change to occur, but it will. Of course, that can be remedied by explicitly included start of string or end of string matchers, like so: `^http://example.com/$`. 
```
scala> "/".r.findFirstIn("/foo").isDefined
res1: Boolean = true

scala> "^/$".r.findFirstIn("/foo").isDefined
res3: Boolean = false
```

The alternative would be to turn the Scala regex into a Java pattern and call `matches` which attempts a full match of the regex against the string.
```
scala> "/".r.pattern.matcher("/foo").matches
res2: Boolean = false
```

Ultimately, it just depends what we want the user experience to be.

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
```diff
@@ -0,0 +1,150 @@
...
+
+  def filterResponseChanges(changes: List[ChangeDetails]): List[Patch] = {
+    changes.flatMap(change => Option(change.getResponse))
+  }
+
+  def determineContentType(contentType: String): ContentType = {
+    //magic code from stack overflow, i'm a terrible person, http://stackoverflow.com/questions/4636610/how-to-pattern-match-using-regular-expression-in-scala
+    implicit class Regex(sc: StringContext) {
+      def r = new util.matching.Regex(sc.parts.mkString, sc.parts.tail.map(_ => "x"): _*)
+    }
```
dmnjohns: I still don't love this. It's a bit more magic than I prefer.

Mario-Lopez: I thought Scala's regex got much better in 2.11.  Is this magic still necessary?

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
```diff
@@ -0,0 +1,150 @@
...
+
+  def filterJsonPatches(patches: List[Patch]): List[String] = {
+    patches.flatMap(patch => Option(patch.getJson))
+  }
+
+  def applyJsonPatches(body: JsValue, patches: List[String]): JsValue = {
+    patches.foldLeft(body)((content: JsValue, patch: String) => {
+      val jsPatch: JsonPatch = JsonPatch.parse(patch)
+      jsPatch(content)
+    })
```
dmnjohns: Huh. So we apply all matching JSON patches in order. Is that what we want? That could make for some seriously complicated behavior if the config has a bunch of patches all of which get applied. Of course, what is the alternative? Only process the first matching patch? Eh.

repose-aggregator/components/filters/body-patcher-filter/src/main/resources/META-INF/schema/config/body-patcher.xsd
adrianjgeorge: because it potentially contains up to 4 patches?

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
adrianjgeorge: I'm aware, but i prefer it, because it makes reading it clearer when i the reader know the types. Code is about more than making the compiler happy, we have to maintain it.

repose-aggregator/components/filters/body-patcher-filter/src/test/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilterTest.scala
```diff
@@ -0,0 +1,315 @@
...
+  extends FunSpec
+    with Matchers
+    with MockitoSugar
+    with BeforeAndAfter {
+
+  var filter: BodyPatcherFilter = _
+  var configurationService: ConfigurationService = _
+
+  before {
+    configurationService= mock[ConfigurationService]
```
dmnjohns: Weirdly no space before the `=`.

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
adrianjgeorge: That's not obvious from the name of the method. Apparently it's not decoded from that method (i'm not saying it is in my way). I totally copy pastad this from soemwhere else i had to solve this problem, where it was the only thing that gave me what i wanted. This does seem simpler though.

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
adrianjgeorge: It returns some funky type that no one wants to deal in, specifically i think it returns a mutable collection as i recall. I went with list because it's a nice standard type.

repose-aggregator/components/filters/scripting-filter/pom.xml
```diff
@@ -1,142 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
dmnjohns: Huh? Is this a file we missed that you're taking care of as part of this PR?

repose-aggregator/components/filters/valkyrie-authorization-filter/src/main/scala/org/openrepose/filters/valkyrieauthorization/ValkyrieAuthorizationFilter.scala
```diff
@@ -34,6 +34,7 @@ import com.josephpconley.jsonpath.JSONPath
 import com.rackspace.httpdelegation.HttpDelegationManager
```
dmnjohns: Another file.

I'm guessing these were changed on `repose-8-staging` since the time you branched off of it.

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
adrianjgeorge: I remembered some problem with these, but couldn't exactly remember the details. We had to go down to the java types before, and that's what i'll end up doing here.

repose-aggregator/components/filters/valkyrie-authorization-filter/src/main/scala/org/openrepose/filters/valkyrieauthorization/ValkyrieAuthorizationFilter.scala
dmnjohns: Actually, looking at the file, did you change the Parser?

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
adrianjgeorge: Not in any way i could determine

repose-aggregator/core/repose-core-api/src/main/scala/org/openrepose/core/filter/AbstractConfiguredFilter.scala
```diff
@@ -0,0 +1,132 @@
...
+  /**
+    * Subscribes with the configuration service. If the filter config doesn't have a custom file name,
+    * it uses the name provided by DEFAULT_CONFIG. It tries to load and use the schema provided by SCHEMA_LOCATION.
+    * @param filterConfig
+    */
+  override def init(filterConfig: FilterConfig): Unit = {
+    logger.trace("{} initializing ...", this.getClass.getSimpleName)
+    configFile = new FilterConfigHelper(filterConfig).getFilterConfig(DEFAULT_CONFIG)
+
+    logger.info("Initializing filter using config " + configFile)
```
dmnjohns: Using string concatenation rather than interpolation here.

repose-aggregator/core/repose-core-api/src/test/resources/stubbed.xsd
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dmnjohns: I have no idea what this file is for.

dmnjohns: For testing the `AbstractConfiguredFilter`, apparently.

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
adrianjgeorge: Yeah, applying them all makes composition way easier. otherwise you have to get very explicit and repeat yourself a bunch potentially.

repose-aggregator/components/filters/body-patcher-filter/src/test/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilterTest.scala
adrianjgeorge: that is weird

repose-aggregator/components/filters/scripting-filter/pom.xml
adrianjgeorge: I have no idea. I didn't delete it, but it showed up when i rebased to the 8 branch.

repose-aggregator/components/filters/valkyrie-authorization-filter/src/main/scala/org/openrepose/filters/valkyrieauthorization/ValkyrieAuthorizationFilter.scala
adrianjgeorge: I had to change a bunch of things because of the move to a newer play, which moved the other library

adrianjgeorge: It's actually the same parser, i'm just invoking it myself instead of through the other library.

repose-aggregator/core/repose-core-api/src/main/scala/org/openrepose/core/filter/AbstractConfiguredFilter.scala
adrianjgeorge: copy pasta ftw.

adrianjgeorge: Because of course the work i did to stop copy pastaing was itself copy pastad

repose-aggregator/components/filters/body-patcher-filter/src/main/resources/META-INF/schema/config/body-patcher.xsd
dmnjohns: But those patches aren't contained in an element named patch. I suppose there's the potential for confusion since patch is (more or less) synonymous with change, but it's also a single specific... thing.

Anyway, I was just curious.

repose-aggregator/core/repose-core-api/src/main/scala/org/openrepose/core/filter/AbstractConfiguredFilter.scala
adrianjgeorge: also this comment should be on the other pr

repose-aggregator/core/repose-core-api/src/test/resources/stubbed.xsd
adrianjgeorge: Yep

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
dmnjohns: Of course it is. Which is why I would still drop the explicit type annotations. In cases where the type is obvious, it seems extraneous. I suspect that is why the Scala style guide suggests letting the type inferencing system handle it. With Groovy, you can drop almost drop typing entirely. Now I love my types, but man does Java frustrate me with the pattern where you end up typing the type twice (i.e., `Object myObject = new Object();`).

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
dmnjohns: I didn't even think about decoding. Do we want the URI to be decoded before testing it against the regex? I guess I would think so.

Does the `URL` class' `getPath` method decode it?

This is probably something worth throwing a couple of tests around.

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
dmnjohns: Odd. I would have expected it to return a standard type higher up on the hierarchy.

I think technically it returns a wrapper around the Java collection, which costs next to nothing. `toList` may actually re-arrange the data if it's not in a list-like format, though.

Anyway, I'm not worried about it. It will almost certainly be a tiny list anyway.

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
dmnjohns: Fair enough.

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
adrianjgeorge: My issue is where it's not obvious to the reader, then you have the weirdness of only sometimes having it.

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
adrianjgeorge: I'm actually not that concerned, because they can regex it either way

repose-aggregator/components/filters/body-patcher-filter/src/test/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilterTest.scala
adrianjgeorge: i'm kind of surprised the style check didn't spot that one

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
dmnjohns: I suppose. I'll stick that one under personal preferences and move on.

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
dmnjohns: Can they? Isn't the regex different if a space is encoded versus if it isn't? Or, perhaps a better example, what if I want to match on the character `a`? I don't want to have to deal with the fact that the `a` could be encoded. I want `/a` to match `/a` and `/%61`.

repose-aggregator/components/filters/valkyrie-authorization-filter/src/main/scala/org/openrepose/filters/valkyrieauthorization/ValkyrieAuthorizationFilter.scala
dmnjohns: Ah. That makes sense.

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
adrianjgeorge: It mostly is

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
adrianjgeorge: Fair enough, i was only considering control characters, but i suppose some crazy person might encode some of the other characters

adrianjgeorge: Yeah, my tests show's that it doesn't give back just the path. of course that could just be a poor implementation on the spring mock

adrianjgeorge: actually if you read the docs the description doesn't match there table, hooray crappy documentation

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
adrianjgeorge: I'm changing this, but it's much less readable now

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
```diff
@@ -0,0 +1,144 @@
...
+  */
+@Named
+class BodyPatcherFilter @Inject()(configurationService: ConfigurationService)
+  extends AbstractConfiguredFilter[BodyPatcherConfig](configurationService) with LazyLogging {
+  override val DEFAULT_CONFIG: String = "body-patcher.cfg.xml"
+  override val SCHEMA_LOCATION: String = "/META-INF/schema/config/body-patcher.xsd"
+
+  override def doWork(request: ServletRequest, response: ServletResponse, chain: FilterChain): Unit = {
+    val httpRequest = request.asInstanceOf[HttpServletRequest]
+    val httpResponse = response.asInstanceOf[HttpServletResponse]
```
dmnjohns: I thought you weren't changing this?

adrianjgeorge: meh

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
```diff
@@ -120,14 +119,9 @@ class BodyPatcherFilter @Inject()(configurationService: ConfigurationService)
...
-    //magic code from stack overflow, i'm a terrible person, http://stackoverflow.com/questions/4636610/how-to-pattern-match-using-regular-expression-in-scala
-    implicit class Regex(sc: StringContext) {
-      def r = new util.matching.Regex(sc.parts.mkString, sc.parts.tail.map(_ => "x"): _*)
-    }
-
     Option(contentType).getOrElse("").toLowerCase match {
-      case r".*json.*" => Json
-      case r".*xml.*" => Xml
+      case Json.regex() => Json
+      case Xml.regex() => Xml
```
dmnjohns: Honestly, I don't think this looks any worse than it did before. I also appreciate not having to figure out the magics. I mean, it was obvious what the match was doing before, just not how.

dmnjohns:
Looks like this branch needs to be rebased.

adrianjgeorge:
I'm aware, I'm just waiting till it's otherwise in working order

adrianjgeorge:
retest this please

dmnjohns:
Merging since the failing tests appear to be unrelated.

PR #1537, opened 2016-05-24
REP-3951 Package Enhancements (maybe)

wdschei:
Updated the Body Extractor to Header filter's schema test to use the static-like variables in the new framework to prevent misnamed examples from getting/remaining in the code base. Thoughts?

### DIFF
--- a/repose-aggregator/components/filters/body-extractor-to-header-filter/src/test/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderSchemaTest.scala
+++ b/repose-aggregator/components/filters/body-extractor-to-header-filter/src/test/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderSchemaTest.scala
@@ -20,17 +20,20 @@
 package org.openrepose.filters.bodyextractortoheader
 
 import org.junit.runner.RunWith
+import org.mockito.Mockito._
 import org.openrepose.commons.test.ConfigValidator
+import org.openrepose.core.services.config.ConfigurationService
 import org.scalatest.junit.JUnitRunner
 import org.scalatest.{FunSpec, Matchers}
 
 @RunWith(classOf[JUnitRunner])
 class BodyExtractorToHeaderSchemaTest extends FunSpec with Matchers {
-  val validator = ConfigValidator("/META-INF/schema/config/body-extractor-to-header.xsd")
+  val filter: BodyExtractorToHeaderFilter = new BodyExtractorToHeaderFilter(mock(classOf[ConfigurationService]))
+  val validator = ConfigValidator(filter.SCHEMA_LOCATION)
 
   describe("schema validation") {
     it("should successfully validate the example config") {
-      validator.validateConfigFile("/META-INF/schema/examples/body-extractor-to-header.cfg.xml")
+      validator.validateConfigFile(s"/META-INF/schema/examples/${filter.DEFAULT_CONFIG}")
     }
   }
 }


repose-aggregator/components/filters/body-extractor-to-header-filter/src/test/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderSchemaTest.scala
```diff
@@ -20,17 +20,20 @@
...
+import org.mockito.Mockito._
 import org.openrepose.commons.test.ConfigValidator
+import org.openrepose.core.services.config.ConfigurationService
 import org.scalatest.junit.JUnitRunner
 import org.scalatest.{FunSpec, Matchers}
 
 @RunWith(classOf[JUnitRunner])
 class BodyExtractorToHeaderSchemaTest extends FunSpec with Matchers {
-  val validator = ConfigValidator("/META-INF/schema/config/body-extractor-to-header.xsd")
+  val filter: BodyExtractorToHeaderFilter = new BodyExtractorToHeaderFilter(mock(classOf[ConfigurationService]))
```
dmnjohns: I don't love that you have to instantiate the filter (and mock a configuration service) just to get the paths you want.

Can we make those variables static? In Scala, that would be moving them to a companion object. I'm not sure if you can override them in the companion object, since you're using the abstract filter, but if not, you should be able to define them in a companion object, and override them in the class by pointing the references to the companion object.

dmnjohns:
We should definitely be doing this in our tests to keep them in sync. I like.

dmnjohns:
I take it the main package enhancements branch isn't ready for a PR yet? You just wanted feedback on this way of testing? I went to check the story and was like, huh, none of that work is here. Then I noticed that this PR was one of your branches onto another of your branches.

repose-aggregator/components/filters/body-extractor-to-header-filter/src/test/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderSchemaTest.scala
wdschei: We could definitely move these into a companion object to provide a true static scoping, but then we loose the inherited functionality provide by overriding the abstract definitions in `AbstractConfiguredFilter`.

That is why I left this as a separate PR. This way didn't feel completely right, but it did provide a way to enforce that there is at least one example that is named the default configuration.

wdschei:
The main is ready too, I just hadn't opened the PR util we figured out how we wanted to do this. Or if we even wanted to do this.

repose-aggregator/components/filters/body-extractor-to-header-filter/src/test/scala/org/openrepose/filters/bodyextractortoheader/BodyExtractorToHeaderSchemaTest.scala
dmnjohns: Right, so it would be a bit redundant, but I would suggest doing something like this:
```
class MyFilter extends AbstractConfiguredFilter {
  override def SCHEMA_LOCATION = MyFilter.SchemaLocation
  override def DEFAULT_CONFIG = MyFilter.DefaultConfig
}

object MyFilter {
  final val SchemaLocation = "/path/to/schema.xsd"
  final val DefaultConfig = "/path/to/default/config.cfg.xml"
}
```

Then in the test, you can use the static definitions. Ideally, you would only have static definitions (at least, at public visibility) but I don't know how you would get that to work.

dmnjohns: There may even be a way to get rid of the redundancy by splitting out some of the configuration stuff from the `AbstractConfiguredFilter` into a trait. Scala objects can extend traits, so we get can define things statically, but then the `AbstractConfiguredFilter` would need to take that information on construction. Something like:
```
class MyFilter extends AbstractConfiguredFilter(MyFilter) {
  ...
}

object MyFilter extends ConfiguredFilterPaths {
  final val SchemaLocation = "/path/to/schema.xsd"
  final val DefaultConfig = "/path/to/default/config.cfg.xml"
}
```

wdschei: This is kind of why I made this a separate PR. Since it is getting into more of changing the abstract parent and/or making new stuff, that is kind of out of scope for this story. So probably the suggestion is to just close this PR, but don't delete the branch, and then reference it in a new story.

dmnjohns: Fair enough.

wdschei:
Since this PR was leaving the scope of the original story, a new story was created to accommodate it:

- https://repose.atlassian.net/browse/REP-3971

This PR will be closed, but the branch left intact for future reference.

PR #1538, opened 2016-05-24
Removed feature to allow blank bodies to work on the patcher

adrianjgeorge:
This resolves the failing tests.

### DIFF
--- a/repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
+++ b/repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
@@ -68,7 +68,7 @@ class BodyPatcherFilter @Inject()(configurationService: ConfigurationService)
           val originalValue: JsValue = Try(PJson.parse(httpRequest.getInputStream))
             .recover({
               case jpe: JsonParseException =>
-                logger.trace("Bad Json body")
+                logger.debug("Bad Json body", jpe)
                 throw new RequestBodyUnparseableException("Couldn't parse the body as json", jpe)
             }).get
           val patchedValue: JsValue = applyJsonPatches(originalValue, jsonPatches)
--- a/repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
+++ b/repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
@@ -65,19 +65,12 @@ class BodyPatcherFilter @Inject()(configurationService: ConfigurationService)
         val jsonPatches: List[String] = filterJsonPatches(requestPatches)
         if (jsonPatches.nonEmpty) {
           logger.debug("Applying json patches on the request")
-          val originalValue: JsValue = Try({
-            val inputStream: ServletInputStream = httpRequest.getInputStream
-            if (inputStream.available() > 0) {
-              PJson.parse(inputStream)
-            } else {
-              logger.trace("No request body found, creating empty json")
-              PJson.obj()
-            }
-          }).recover({
-            case jpe: JsonParseException =>
-              logger.trace("Bad Json body")
-              throw new RequestBodyUnparseableException("Couldn't parse the body as json", jpe)
-          }).get
+          val originalValue: JsValue = Try(PJson.parse(httpRequest.getInputStream))
+            .recover({
+              case jpe: JsonParseException =>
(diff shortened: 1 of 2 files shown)


repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
```diff
@@ -65,19 +65,12 @@ class BodyPatcherFilter @Inject()(configurationService: ConfigurationService)
...
-          }).recover({
-            case jpe: JsonParseException =>
-              logger.trace("Bad Json body")
-              throw new RequestBodyUnparseableException("Couldn't parse the body as json", jpe)
-          }).get
+          val originalValue: JsValue = Try(PJson.parse(httpRequest.getInputStream))
+            .recover({
+              case jpe: JsonParseException =>
+                logger.trace("Bad Json body")
+                throw new RequestBodyUnparseableException("Couldn't parse the body as json", jpe)
```
Mario-Lopez: Is the original exception going to make it back to the client?

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
```diff
@@ -65,19 +65,12 @@ class BodyPatcherFilter @Inject()(configurationService: ConfigurationService)
...
-            }
-          }).recover({
-            case jpe: JsonParseException =>
-              logger.trace("Bad Json body")
-              throw new RequestBodyUnparseableException("Couldn't parse the body as json", jpe)
-          }).get
+          val originalValue: JsValue = Try(PJson.parse(httpRequest.getInputStream))
+            .recover({
+              case jpe: JsonParseException =>
+                logger.trace("Bad Json body")
```
Mario-Lopez: I assume that the exception thrown in the next line will be logged at some point.  If that's the case, this line seems redundant.  If not, then we should also log the jpe exception here, perhaps at the debug level.

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
adrianjgeorge: Clarify? I've attached the exception to the new one, which will intentionally bubble out as a 500.

adrianjgeorge: wait, crap i think i wanted a 400 here. let me look again at it

Mario-Lopez: To be clear, I absolutely love including the original exception inside of a new exception.  I'm just wondering out loud how we convey the error to the client.

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
adrianjgeorge: It will be, there use to be a couple of paths here, so the logger was showing us which one was coming through, this is no longer the case.

repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
adrianjgeorge: The original exception is actually pretty useless when it comes to content, if i recall. The intent here, is that the xml will catch it's funky exceptions and also wrap to this same type, so the rest of the code only has to deal with the one thing.

adrianjgeorge: Also i confirmed that this exception on the request side is caught and made into a 400, on the response it's intentionally allowed out as a 500, because it means the origin service isn't returning bodies that can be patched as expected.

adrianjgeorge: Also the exception wasn't getting logged anywhere else, just caught and handled. I've added the exception to this log line now so we have the info if we need it.

adrianjgeorge:
retest this please

PR #1539, opened 2016-05-24
REP-3951 Package Enhancements

wdschei:


repose-aggregator/artifacts/filter-bundle/src/config/filters/rate-limiting.cfg.xml
```diff
@@ -1,23 +1,15 @@
...
+        <limit id="four" uri="/service/*" uri-regex="/service/([\d^/]*)/.*" http-methods="POST" unit="DAY" value="2"/>
+    </limit-group>
+    <limit-group id="admin-limits" groups="admin">
+        <limit id="five" uri="/service/admin/*" uri-regex="/service/admin/(.*)" http-methods="GET POST PUT DELETE"
+               unit="SECOND" value="1"/>
     </limit-group>
-
-    <!-- Limits for WhiteListed IPs -->
-    <limit-group id="unlimited" groups="unlimited" default="false"/>
-
```
dmnjohns: I feel like the config that already existed actually covered use-cases better. Really, a combination of the two would be best. I imagine that you just replaced the existing shipped configs with the example configs for this and the others, correct?

It's almost certainly worth looking at exactly what we ship. In some cases, it may have been better to replace the examples with what we ship.

repose-aggregator/artifacts/filter-bundle/src/config/filters/merge-header.cfg.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<merge-header xmlns='http://docs.openrepose.org/repose/merge-header/v1.0'>
+    <!-- both response/request configs are optional, if neither are specified, this filter is a no-op -->
+    <response>
+        <header>your-mother</header>
+        <header>your-father</header>
+        <header>your-sister</header>
+    </response>
```
dmnjohns: This is hilarious, but do we want to ship it?

dmnjohns:
Apart from some of the changes made to shipped configs, this looks good to me.

repose-aggregator/artifacts/filter-bundle/src/config/filters/rate-limiting.cfg.xml
wdschei: Yes. I defaulted to the working examples since those were all guaranteed to be at least syntactically correct since they were being tested. There were several that were out of sync between the shipping example and the current features.

repose-aggregator/artifacts/filter-bundle/src/config/filters/merge-header.cfg.xml
wdschei: And I honestly didn't look at these for content after I decided to use the already tested examples from the filter META-INF directories.

adrianjgeorge:
I'm not certain we want to ship a config for every filter. I'm not certain what the original intent of the shipped configs was. Was it commonly used configs only, or was it meant to be comprehensive, and nobody was doing it so it became way out of date? More importantly what do we want to do?

repose-aggregator/artifacts/filter-bundle/src/config/filters/rate-limiting.cfg.xml
wdschei: Updated with a combination of the two and put it in both the filter's example and the bundle's files.

dmnjohns:
One final question: do we want to a set of configurations to be put in `/etc/repose` which run Repose in a pass-through mode? Just something to let a user install and run Repose before needing to fiddle with configs. It would be a very minimal set: system-model, container, log4j2, and maybe some others.

Also, it appears that with the move to Gradle, we aren't currently shipping a system-model config, container config, log4j2 config, and maybe some others?

dmnjohns:
I lied. Those are in the `src` directory of the `artifacts` directory. So system-model and others are being shipped.

wdschei:
The `system-model.cfg.xml` that ships with both Valve and Root WAR was updated some time ago and it already runs out of the box.

dmnjohns:
Ah. The part I missed was that you only put filter bundle files into the new `examples` directory. WAR and Valve configs still go into `/etc/repose`. That's probably legit.

dmnjohns:
The build is failing because of the api-checker version thing, but I have no problem merging this now. Are there are objections?

adrianjgeorge:
did he move the configs for th other bundles as well?

adrianjgeorge:
looks like it

PR #1544, opened 2016-05-27
Valkyrie changes to support keystone based auth

adrianjgeorge:


wdschei:
The schema changes look good.

adrianjgeorge:
I've updated the code to actually deal with bad status codes. There are a million magic number, i'll fix them monday.

repose-aggregator/components/filters/valkyrie-authorization-filter/src/main/scala/org/openrepose/filters/valkyrieauthorization/ValkyrieAuthorizationFilter.scala
```diff
@@ -321,27 +325,59 @@ class ValkyrieAuthorizationFilter @Inject()(configurationService: ConfigurationS
...
+      (Option(valkyrieServer.getUsername), Option(valkyrieServer.getPassword)) match {
+        case (Some(username), Some(password)) =>
+          Try(akkaServiceClient.get(cacheKey(callType, transformedTenant, contactId),
+            valkyrieServer.getUri + uri,
+            Map("X-Auth-User" -> username, "X-Auth-Token" -> password) ++ requestTracingHeader)
+          )
+        case _ =>
+          Try(akkaServiceClient.get(cacheKey(callType, transformedTenant, contactId),
+            valkyrieServer.getUri + uri,
+            Map("X-Auth-Token" -> authToken.getOrElse(null)) ++ requestTracingHeader)
```
dmnjohns: What happens if an auth token isn't available here? How does the Akka Service Client handle a null value in a header map? I'll check the tests when I get down there.

dmnjohns: Apparently it's legit, but I'm not sure what that looks like in the actual HTTP request. Moving right along.

adrianjgeorge: Yeah, it works, but not looking at it has given me a better way to handle it. Update inbound.

dmnjohns:
So. Many. Magic. Number. Replacements.

repose-aggregator/components/filters/valkyrie-authorization-filter/src/main/scala/org/openrepose/filters/valkyrieauthorization/ValkyrieAuthorizationFilter.scala
```diff
@@ -321,27 +325,59 @@ class ValkyrieAuthorizationFilter @Inject()(configurationService: ConfigurationS
...
+      (Option(valkyrieServer.getUsername), Option(valkyrieServer.getPassword)) match {
+        case (Some(username), Some(password)) =>
+          Try(akkaServiceClient.get(cacheKey(callType, transformedTenant, contactId),
+            valkyrieServer.getUri + uri,
+            Map("X-Auth-User" -> username, "X-Auth-Token" -> password) ++ requestTracingHeader)
+          )
+        case _ =>
+          Try(akkaServiceClient.get(cacheKey(callType, transformedTenant, contactId),
+            valkyrieServer.getUri + uri,
+            authToken.map(token => Map("X-Auth-Token" -> token)).getOrElse(Map.empty) ++ requestTracingHeader)
```
adrianjgeorge: Boom, no more ugly null.

dmnjohns: I like it.

dmnjohns:
This all looks good to me. Tests were passing before that last commit, so I expect them to continue passing... if time is an issue, I would be fine merging now, otherwise I would wait until tests pass again.

👍

PR #1545, opened 2016-05-31
Adding a hard version dependency for virtual packages

dmnjohns:
This change allowed me to install packages on a local Ubuntu VM, whereas before dpkg was complaining about requiring a specific version of a virtual package.


adrianjgeorge:
Are the deb packages produced here fine when not running release or is package name illegal?

PR #1547, opened 2016-06-01
Ignored a test for the time being

adrianjgeorge:


dmnjohns:
Yup. That's definitely an ignored test.

Merging.

PR #1550, opened 2016-06-07
REP-999 - Identity v3 json pathing

Mario-Lopez:
Updated the v3 filter to use play instead of spray.  I was able to completely remove spray from our dependencies including versions.properties.


repose-aggregator/components/filters/openstack-identity-v3-filter/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3Handler.scala
```diff
@@ -87,6 +86,19 @@ class OpenStackIdentityV3Handler(identityConfig: OpenstackIdentityV3Config, iden
...
+        "endpoints" -> JsArray(catalog.endpoints.map(endpoint => JsObject(
+          Seq("id" -> JsString(endpoint.id)) ++
+            endpoint.name.map(name => Seq("name" -> JsString(name))).getOrElse(Seq()) ++
+            endpoint.interface.map(interface => Seq("interface" -> JsString(interface))).getOrElse(Seq()) ++
+            endpoint.region.map(region => Seq("region" -> JsString(region))).getOrElse(Seq()) ++
+            Seq("url" -> JsString(endpoint.url)) ++
+            endpoint.service_id.map(serviceId => Seq("service_id" -> JsString(serviceId))).getOrElse(Seq()))))) ++
+        catalog.id.map(id => Seq("id" -> JsString(id))).getOrElse(Seq()) ++
+        catalog.name.map(name => Seq("name" -> JsString(name))).getOrElse(Seq()))
+    }
```
dmnjohns: Not sure if this was one of the things you wanted to change, but you may be able to simplify this block using function construction. 

"Using class construction" on https://www.playframework.com/documentation/2.5.x/ScalaJson

Also, I always use `Collection.empty[Type]` over `Collection[Type]()`. It's a personal preference thing, but I find it to read more clearly and cut down on the sometimes obscene number of parenthesis you see nested in Scala.

dmnjohns: Also, is there not an explicit conversion from `String` to `JsString`? You may be able to drop the construction on those.

dmnjohns: It is also odd that line 92 and line 93 are indented differently. They are sibling fields in a JSON object, right?

dmnjohns: Ah, I see. Everything is pretty much being treated as a `Seq` to handle cases where something shouldn't be added, so explicitly constructing `JsValue`s is a simple way to handle that.

I wonder if that is how we want to handle that case? Would it be better to have a static JSON structure and place a JSON null value if we don't have anything to put? That may make parsing more reliable. Also, I believe in that case you can make use of Scala's `Option` -- I _think_ `None` will be treated as JSON null.

Mario-Lopez: Yup, this is one of the things I plan on changing.

1. It's using the fancy implicit writes thing to be able to do `Json.toJson` later in the class.  I think I'm going to move the JSON construction down so it'll be closer to where it's being used and avoid using an implicit completely.  https://github.com/rackerlabs/repose/blob/70c346711aaa39ebd7fdaf6576b1a0ba45125b7e/repose-aggregator/components/filters/openstack-identity-v3-filter/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3Handler.scala#L235

2. It is using class construction.  I'm not sure what you mean???

3. If you're talking about the `Seq()`, the alternative doesn't look that great:
`Seq.empty[(String, JsValue)]`

4. When using JsObject(Seq(...)), there is no automatic conversion from String to JsString.  It must be specified.

4a. I'm using JsObject(Seq(...)) over Json.obj(...) since (in order to maintain the old behavior which may be something we don't need to do) I need to be able to leave out fields if they're going to be null.  It's still possible to do it using Json.obj(...), but it looks messier.

5. Yeah, I planned on fixing indentation issues like that.

Mario-Lopez: I like how it replaced my "5" with a "1" because I put a "4a".  Silly Markdown.

dmnjohns: I should just wait for you to finish with your changes, but...

1. Sounds legit.

2. Right, right, so while it's in the same section in the docs, Play JSON supports class construction and function construction. So rather than writing `JsObject(stuff)` you could write `Json.obj(stuff)`. Mostly, this should allow you to replace the `Seq`s with calls to functions which take varargs.

3. Well, technically it would be `Seq[(String, JsValue)]` vs `Seq.empty[(String, JsValue)]`. In both cases, though, Scala's type system should be able to figure out what you want without being explicit about the types. So really, it would be `Seq()` vs `Seq.empty`.

4. So there is absolutely a conversion from `String` to `JsString`, but the type system has to know to do that conversion. In your case, you can coerce the type system by writing something like `Seq[(String, JsString)]("foo" -> "bar")`. That should invoke the implicit conversion. Of course, if you collapse some of the other stuff here, you shouldn't even need to be explicit about the types. The type system will handle a lot of it for you.

4a. Ah. Totally didn't realize the old behavior left stuff out. Man, that's kind of gross.

5. Cool, cool.

Mario-Lopez: I'll try to make some of these changes to see how they flow.

repose-aggregator/components/filters/openstack-identity-v3-filter/src/main/scala/org/openrepose/filters/openstackidentityv3/utilities/OpenStackIdentityV3API.scala
```diff
@@ -196,14 +234,15 @@ class OpenStackIdentityV3API(config: OpenstackIdentityV3Config, datastore: Datas
     }
   }
 
-  def getGroups(userId: String, tracingHeader: Option[String] = None, checkCache: Boolean = true): Try[List[Group]] = {
-    getFromCache[mutable.ArrayBuffer[Group]](GROUPS_KEY_PREFIX + userId) match {
+  def getGroups(userId: String, tracingHeader: Option[String] = None, checkCache: Boolean = true): Try[List[String]] = {
+    getFromCache[List[String]](GROUPS_KEY_PREFIX + userId) match {
```
dmnjohns: So glad Scala 2.11 provides serializable lists.

repose-aggregator/artifacts/filter-bundle/src/config/docs/DEPENDENCIES.txt
```diff
@@ -36,7 +36,6 @@ com.yammer.metrics                        metrics-core             2.2.0
 com.yammer.metrics                        metrics-ehcache          2.2.0           Apache License Version 2.0  http://metrics.dropwizard.io/2.2.0/maven/metrics-ehcache/license.html
 com.yammer.metrics                        metrics-graphite         2.2.0           Apache License Version 2.0  http://metrics.dropwizard.io/2.2.0/maven/metrics-graphite/license.html
 io.gatling                                jsonpath_2.11            0.6.4           Apache License Version 2.0  https://github.com/gatling/jsonpath
-io.spray                                  spray-json_2.11          1.2.6           Apache License Version 2.0  https://github.com/spray/spray-json
```
adrianjgeorge: good catch

repose-aggregator/components/filters/openstack-identity-v3-filter/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3Handler.scala
```diff
@@ -181,45 +179,40 @@ class OpenStackIdentityV3Handler(identityConfig: OpenstackIdentityV3Config, iden
...
+        request.replaceHeader(OpenStackIdentityV3Headers.X_TOKEN_EXPIRES, token.get.expiresAt)
         request.replaceHeader(OpenStackIdentityV3Headers.X_AUTHORIZATION.toString, OpenStackIdentityV3Headers.X_AUTH_PROXY) // TODO: Add the project ID if verified
-        token.get.user.name.foreach { user =>
+        token.get.userName foreach { user =>
           request.addHeader(OpenStackIdentityV3Headers.X_USER_NAME.toString, user)
           request.addHeader(PowerApiHeader.USER.toString, user, 1.0)
         }
-        token.get.roles.foreach { roles =>
-          request.addHeader(OpenStackIdentityV3Headers.X_ROLES, roles.map(_.name) mkString ",")
+        Option(token.get.roles.map(_.name) mkString ",").filter(_.nonEmpty) foreach { rolesString =>
```
adrianjgeorge: why were we doing a single header?

repose-aggregator/components/filters/openstack-identity-v3-filter/src/main/scala/org/openrepose/filters/openstackidentityv3/utilities/OpenStackIdentityV3API.scala
```diff
@@ -161,10 +168,33 @@ class OpenStackIdentityV3API(config: OpenstackIdentityV3Config, datastore: Datas
             // because we care to match on the status code of the response, if anything was set.
             validateTokenResponse.map(response => response.getStatus) match {
               case Some(statusCode) if statusCode == HttpServletResponse.SC_OK =>
-                val subjectTokenObject = jsonStringToObject[AuthResponse](inputStreamToString(validateTokenResponse.get.getData)).token
+                val json = Json.parse(inputStreamToString(validateTokenResponse.get.getData))
 
-                val expiration = new DateTime(subjectTokenObject.expires_at)
-                val identityTtl = safeLongToInt(expiration.getMillis - DateTime.now.getMillis)
+                val subjectTokenObject = ValidToken(
```
adrianjgeorge: Why are we doing this instead of just working with the json object from play directly?

Mario-Lopez: The json object isn't serializable, so we'd have to convert it to something else anyway to cache it.

adrianjgeorge: seriously? Why would you not make that serializable? It's pretty obvious what that looks like.

repose-aggregator/components/filters/openstack-identity-v3-filter/src/main/scala/org/openrepose/filters/openstackidentityv3/OpenStackIdentityV3Handler.scala
Mario-Lopez: No idea.  Now that I think about it, at the very least I could have done an appendHeader which would make this logic a lot cleaner.  Or would you prefer I do addHeader for each role?

adrianjgeorge: I don't mind that you kept it the way it was, i'm just curious what went in to that decision to begin with

dmnjohns: Pretty sure that was carried over from the old v2 filter.

dmnjohns: Nope. I'm a liar:
https://github.com/rackerlabs/repose/blob/repose-7.3.6.0/repose-aggregator/components/filters/client-auth/src/main/java/org/openrepose/filters/clientauth/openstack/OpenStackAuthenticationHeaderManager.java#L243

I've got nothing.

dmnjohns: Maybe. `getRoles()` actually returns a string, so I have no idea what is going on.

dmnjohns: Ah. There you go. Old auth n used to throw roles together. I imagine that's where the behavior originated.

PR #1551, opened 2016-06-08
REP-4028 Confirm Compression filter works as expected.

wdschei:
Added some tests and updated one place to make testing easier.
The end result is that we can confirm the new wrappers fixed the wonkiness happening with the compression filter in v6.x.

### DIFF
--- a/repose-aggregator/components/filters/compression-filter/src/test/scala/org/openrepose/filters/compression/CompressionFilterCompleteTest.scala
+++ b/repose-aggregator/components/filters/compression-filter/src/test/scala/org/openrepose/filters/compression/CompressionFilterCompleteTest.scala
@@ -91,7 +91,6 @@ class CompressionFilterCompleteTest extends FunSpec with BeforeAndAfterEach with
           val rxBody = servletResponse.getContentAsByteArray
           val expected = zippedContent.asInstanceOf[Array[Byte]]
           assert(rxBody sameElements expected)
-          assert(Option(servletResponse.getHeaderValue(HttpHeaders.CONTENT_LENGTH)).getOrElse("-1").toString.toInt == expected.length)
         } else {
           assert(servletResponse.getContentAsString == content)
         }
--- a/repose-aggregator/components/filters/compression-filter/src/test/scala/org/openrepose/filters/compression/CompressionFilterCompleteTest.scala
+++ b/repose-aggregator/components/filters/compression-filter/src/test/scala/org/openrepose/filters/compression/CompressionFilterCompleteTest.scala
@@ -1,0 +1,138 @@
+/*
+ * _=_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_=
+ * Repose
+ * _-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-
+ * Copyright (C) 2010 - 2015 Rackspace US, Inc.
+ * _-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


repose-aggregator/components/filters/compression-filter/src/main/scala/org/openrepose/filters/compression/CompressionFilter.scala
```diff
@@ -91,6 +93,7 @@ class CompressionFilter @Inject()(configurationService: ConfigurationService, co
           logger.error("Error with the CompressingFilter", e)
           httpResponse.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR)
       }
+      httpResponse.commitToResponse()
```
adrianjgeorge: i thought you didn't have to make any changes?

wdschei: It wasn't required since the powerfilter would eventually commit the stream and add the header. This is just more correct that the filter clean up after itself and not rely on something downstream to do it later.

wdschei: This also allowed us to test it in a unit test and not have to spin a Repose instance just to check it.

adrianjgeorge: The wrapper usage here is unnecessary since in our code we aren't making any changes to the response. By using it you are stating an intention to mutate the response in our code in some way, but you don't, that's handle entirely by the nested filter. All you've really done is introduced a hard break in the stream where one wasn't necessary.

wdschei: We are adding the correct header.

adrianjgeorge: are you sure its not coming out of internal filter correct now?

wdschei: Yes. It failed the unit test, but passed the functional test. This change made it pass the unit test also.

wdschei: I can remove this and the header value check in the unit test, but that seems a little wonky to me.

- https://github.com/rackerlabs/repose/blob/969dd2cb2665a8b4d9dfb80dd91a48a02f89ce00/repose-aggregator/components/filters/compression-filter/src/test/scala/org/openrepose/filters/compression/CompressionFilterCompleteTest.scala#L94

adrianjgeorge: weird. ok, then i agree leaving this in is reasonable so that the filter is properly self contained. I'd like to know where they failure to write is happening otherwise.

dmnjohns:
LGTM 👍

repose-aggregator/components/filters/compression-filter/src/main/scala/org/openrepose/filters/compression/CompressionFilter.scala
dmnjohns: So the following discussion kind of details why this change is "needed":
https://sourceforge.net/p/pjl-comp-filter/discussion/378427/thread/31dd22c2/

The PJL Compressing Filter does not (and, in fact, cannot) set the Content-Length header since it processes the body as a stream.

Our wrapper, on the other hand, will buffer the response until committed, allowing us to set the content-length at the time of commit. That is, our wrapper is where the content-length is actually being set. As @wdschei pointed out, the PowerFilter necessarily performs this function since parts of the response must be known for RMS to function.

The Content-Length header is not strictly required by HTTP spec, but if it is not present, either transfer-encoding should be set to chunked, or the connection should be closed after the body has been written. Some if not all containers will handle this behavior for the user.

I'm not sure if we need a unit test to check the content-length header or not -- the underlying filter does not set it, that is additional (optional) processing that we perform in the wrapping filter. It seems to me that we could leave that to the PowerFilter and save one instance of breaking the stream.

PR #1552, opened 2016-06-08
Do Not Merge - Dirty patch of valkyrie fixes REP-4069

adrianjgeorge:


wdschei:
This actually looks pretty clean comparing what I thought it was going to look like. The only thing that stands out is the use of magic numbers for the HTTP Status Codes instead of the values provided by `javax.servlet.http.HttpServletResponse`, but that is the way most of the v7.x stuff did it.

I left a comment on the Jira story for the commands I used to do a three-way diff between `repose-7.3.5.0`, `port-valk-changes`, `8.0.1.0`.

Also this PR is against master?
How is the automated testing going to work for this one?

dmnjohns:
I expect this won't be merged, but I pushed up the tests we are using on master for this feature just to prove that it works here as well.

adrianjgeorge:
no idea if your merge of the tests worked, because i'd already disabled the pull request build. So there is a question of whether we want to merge this or not. The code already exists on master, so there's no point other than not having lingering branches.

dmnjohns:
I noticed, but I ran the tests locally and got them to pass. That was good enough for me on this one.

Yeah, I wasn't sure if we wanted to merge or not. That's up to you, I think.

PR #1553, opened 2016-06-09
Removed all references to the RPM's requiring daemonize.

wdschei:
Since RPM platforms were the first to move to `systemd`, I think it is safe to remove System V support from them.

### DIFF
--- a/repose-aggregator/artifacts/valve/build.gradle
+++ b/repose-aggregator/artifacts/valve/build.gradle
@@ -248,7 +248,6 @@ buildRpm {
     }
 
     requires('java', '1.8.0', GREATER | EQUAL)
-    requires('daemonize')
     preInstall file('../src/config/scripts/preinst-rpm')
     postInstall('/sbin/chkconfig --add repose-valve')
 }


dmnjohns:
Changes are simple and look fine. The only question is whether or not we want to do this. It appears that [all relevant distributions][1] have been using systemd for over two years now, so it's probably a safe move. 

[1]:https://en.wikipedia.org/wiki/Systemd#Adoption_and_reception

adrianjgeorge:
Yeah, i think i'm ok with it. we are leaving all the hooks for the other system, so they just have to manually install daemonize if they want it. or at least i'm assuming the install scripts won't crap out without it

adrianjgeorge:
wait these changes look like they are removing the other way of doing things entirely. I'm not okay with that. That would have had to be done on 8

wdschei:
They are removed for the RPM's only.

JoelRizner:
I was able to install Repose 8.0.1.0 on Centos7 and ubuntu16 without having to install daemonize nor add the rpmforge repo

Mario-Lopez:
The manual testing was done in lieu of adding a functional test since we currently don't have the framework in place to automatically test this.

dmnjohns:
Merging.

PR #1554, opened 2016-06-10
Rep 3415

JoelRizner:
Updated atom-feed-service to utilize reverse-read feature


repose-aggregator/components/services/atom-feed-service/atom-feed-service-impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/actors/FeedReader.scala
```diff
@@ -92,7 +93,16 @@ class FeedReader(feedUri: String,
...
+        def getStream(): Stream[Entry] = {
+          AtomEntryStreamBuilder.build(feedUrl, AuthenticationRequestContextImpl(reposeVersion, requestId), authenticatedRequestFactory)
+        }
+
+        val entryStream = Option(order) match {
+          case Some(reverse) if reverse.value().equals("reverse-read") =>
+            getStream().reverse
+          case _ =>
+            getStream()
+        }
```
dmnjohns: Why use an `Option`? Order should always be defined as some value in the `EntryOrderType` enum. So while you shouldn't ever need a default case, you could still include one, but I would use it strictly for handling the weird case where the value of `order` is not something that should be possible. Actually, I wouldn't even do that, since the `MatchError` will be caught by the try block and logged anyway. Really, I would recommend just being explicit about the values `order` can have.

Also, for function calls with no side effect and no arguments, you can omit the `()`. Oh, and in this case, you really don't need to do a string comparison. 

Or, put a different way, some wild code appears!
```
val entryStream = order match {
  case EntryOrderType.READ =>
    getStream
  case EntryOrderType.REVERSE_READ =>
    getStream.reverse
}
```

repose-aggregator/components/services/atom-feed-service/atom-feed-service-impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/actors/FeedReader.scala
```diff
@@ -92,7 +93,16 @@ class FeedReader(feedUri: String,
       MDC.put(TracingKey.TRACING_KEY, requestId)
 
       try {
-        val entryStream = AtomEntryStreamBuilder.build(feedUrl, AuthenticationRequestContextImpl(reposeVersion, requestId), authenticatedRequestFactory)
+        def getStream(): Stream[Entry] = {
+          AtomEntryStreamBuilder.build(feedUrl, AuthenticationRequestContextImpl(reposeVersion, requestId), authenticatedRequestFactory)
+        }
```
dmnjohns: You can also drop the empty parameter list here (i.e., `getStream()` can become `getStream()`). IntelliJ will whine if you leave the empty parameter list there.

If you wanted, considering how simple this function is, you could even drop the return type and the curly braces. That comes down to personal preference, though.

Just for your reference:
http://docs.scala-lang.org/style/declarations.html#methods
http://docs.scala-lang.org/style/declarations.html#body

JoelRizner:
Cool I made those changes. Looks much more clean now.

dmnjohns:
LGTM :+1:

Mario-Lopez:
retest this please

PR #1555, opened 2016-06-15
REP-2163 Keystone v3 Cache Invalidation

dmnjohns:

PR #1557, opened 2016-06-17
REP-4010 Add Stripped Value into Links of JSON Response Bodies

dmnjohns:


dmnjohns:
Just the configuration changes so far.

repose-aggregator/commons/commons-configuration/src/test/resources/unmarshallerValidator/xsd/uri-stripper.xsd
```diff
@@ -65,14 +71,174 @@
...
+                </xs:attribute>
+            </xs:extension>
+        </xs:simpleContent>
+    </xs:complexType>
+
+    <xs:simpleType name="HttpMethodList">
+        <xs:list itemType="HttpMethod"/>
+    </xs:simpleType>
+
+    <xs:simpleType name="HttpMethod">
```
adrianjgeorge: We super need to export this to a common xsd

repose-aggregator/commons/commons-configuration/src/test/resources/unmarshallerValidator/xsd/uri-stripper.xsd
```diff
@@ -65,14 +71,174 @@
...
+            <xs:enumeration value="CONNECT"/>
+            <xs:enumeration value="TRACE"/>
+            <xs:enumeration value="ALL"/>
         </xs:restriction>
     </xs:simpleType>
 
+    <xs:simpleType name="LinkSegmentMismatchAction">
+        <xs:restriction base="xs:string">
+            <xs:enumeration value="keep"/>
+            <xs:enumeration value="remove"/>
```
adrianjgeorge: well that will add some ugly work

repose-aggregator/commons/commons-configuration/src/test/resources/unmarshallerValidator/xsd/uri-stripper.xsd
```diff
@@ -19,23 +19,29 @@
   =_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_=_
   -->
 
-
-<xs:schema xmlns:uri-stripper="http://docs.openrepose.org/repose/uri-stripper/v1.0" xmlns:html="http://www.w3.org/1999/xhtml"
+<xs:schema targetNamespace="http://docs.openrepose.org/repose/uri-stripper/v1.0"
```
adrianjgeorge: wait there are two schemas? that's bad. we should look at in a dev meeting

adrianjgeorge:
config looks good

repose-aggregator/commons/commons-configuration/src/test/resources/unmarshallerValidator/xsd/uri-stripper.xsd
dmnjohns: Agreed. I may cycle back to this when I get a minute.

repose-aggregator/commons/commons-configuration/src/test/resources/unmarshallerValidator/xsd/uri-stripper.xsd
dmnjohns: I removed this in #1555. That will likely get merged first, and this will be cleared up.

repose-aggregator/commons/commons-configuration/src/test/resources/unmarshallerValidator/xsd/uri-stripper.xsd
dmnjohns: It will. Still, the JSONPath or token index could be bad, at which point we have to make a decision. This just puts the ball in the user's court.

repose-aggregator/components/filters/uri-stripper-filter/src/main/resources/META-INF/schema/config/uri-stripper.xsd
```diff
@@ -65,14 +71,146 @@
...
+    <xs:complexType name="LinkPath">
+        <xs:simpleContent>
+            <xs:extension base="xs:string">
+                <xs:attribute name="token-index" type="NonNegativeInt" use="optional">
+                    <xs:annotation>
+                        <xs:documentation>
+                            <html:p>
+                                The URI Stripper filter will parse the links in the body of a response
+                                into tokens using the '/' character as a delimiter, re-insert the stripped token
+                                from the request URI, then rebuild the link. Index '0' if the first path segment.
```
Mario-Lopez: s/if/is/ ?

dmnjohns: Oops. Will fix.

repose-aggregator/components/filters/uri-stripper-filter/src/main/resources/META-INF/schema/config/uri-stripper.xsd
```diff
@@ -19,23 +19,29 @@
...
     <xs:complexType name="UriStripperConfig">
-
         <xs:annotation>
             <xs:documentation>
                 <html:p>Configuration for the URI Stripper Filter</html:p>
             </xs:documentation>
         </xs:annotation>
+
+        <xs:sequence>
+            <xs:element name="link-resource" type="LinkResource" minOccurs="1" maxOccurs="unbounded"/>
```
Mario-Lopez: Should be minOccurs="0".

repose-aggregator/components/filters/uri-stripper-filter/src/test/scala/org/openrepose/filters/uristripper/UriStripperFilterTest.scala
```diff
@@ -114,6 +118,216 @@ class UriStripperFilterTest extends FunSpec with BeforeAndAfterEach with Matcher
...
+           |}
+         """.stripMargin
+
+      filter.configurationUpdated(Marshaller.uriStripperConfigFromString(config))
+      request.setRequestURI("/v1/12345/foo")
+      setResponseBody(respBody, MimeType.APPLICATION_JSON.toString)
+
+      filter.doFilter(request, response, filterChain)
+
+      response.getContentAsString shouldEqual "{}"
```
Mario-Lopez: Already discussed on Slack - this isn't the expected behavior given the inputs.

dmnjohns: I edited the test, but since your comment is on this line specifically, I don't think GitHub recognizes the change.

So in this case, the link should not be removed. In the case where we fail because the replacement token-index is out of bounds, it should be. There are now tests for both cases.

repose-aggregator/tests/functional-tests/src/integrationTest/configs/features/filters/uristripper/linkresource/json/tokenindex/uri-stripper.cfg.xml
```diff
@@ -0,0 +1,18 @@
...
+        <json link-mismatch-action="keep" token-index="1">$.link-b</json>
+    </link-resource>
+    <link-resource>
+        <json link-mismatch-action="keep" token-index="5">$.link-c</json>
+    </link-resource>
+    <link-resource>
+        <json link-mismatch-action="keep">$.foo.bar.baz.link</json>
+    </link-resource>
+    <link-resource>
+        <json link-mismatch-action="keep">$.store..link</json>
```
Mario-Lopez: @dmnjohns Should I go ahead and remove the test for this?

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/uristripper/UriStripperLinkResourceJsonTokenIndexTest.groovy
```diff
@@ -0,0 +1,146 @@
...
+
+        when: "a request is made and the JSON response body is parsed"
+        def mc = deproxy.makeRequest(url: reposeEndpoint + requestUrl, defaultHandler: responseHandler)
+        def responseJson = jsonSlurper.parseText(mc.receivedResponse.body as String)
+
+        then: "the response body link is modified"
+        responseJson.foo.bar.baz.link == requestUrl
+    }
+
+    def "when a JSON path list is configured and the response body contains multiple instances, they are all updated"() {
```
Mario-Lopez: @dmnjohns For reference, this is the test for the `$.store..link` json path.

repose-aggregator/tests/functional-tests/src/integrationTest/configs/features/filters/uristripper/linkresource/json/tokenindex/uri-stripper.cfg.xml
dmnjohns: I'm about to look into this. I vaguely know what's going on, and how I might approach solving it, but I don't know what it looks like in code.

dmnjohns: This has proven to be somewhat tricky to implement. I suppose we can drop it for now.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
```diff
@@ -65,9 +72,9 @@ class UriStripperFilter @Inject()(configurationService: ConfigurationService)
     } else {
       val wrappedRequest = new HttpServletRequestWrapper(servletRequest.asInstanceOf[HttpServletRequest])
       val wrappedResponse = new HttpServletResponseWrapper(
-        servletResponse.asInstanceOf[HttpServletResponse], ResponseMode.MUTABLE, ResponseMode.PASSTHROUGH)
+        servletResponse.asInstanceOf[HttpServletResponse], ResponseMode.MUTABLE, ResponseMode.MUTABLE)
```
adrianjgeorge: We could handle this a little smarter, only going to mutable if it's a matching resource. It's one less spot where we break the stream. It won't make a huge difference though, since we tinker with headers, and the response can't actually be written till they are done.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
```diff
@@ -82,12 +89,15 @@ class UriStripperFilter @Inject()(configurationService: ConfigurationService)
...
-        wrappedRequest.setRequestURI(joinTokensIntoUri(uriTokens))
+        wrappedRequest.setRequestURI(joinTokensIntoPath(uriTokens))
       }
 
       filterChain.doFilter(wrappedRequest, wrappedResponse)
 
+      wrappedResponse.uncommit()
+
       val originalLocation = wrappedResponse.getHeader(CommonHttpHeader.LOCATION.toString)
+      // todo: What if the path is "/<tenant-id>"? The location header wouldn't be modified. Is that the correct behavior?
```
adrianjgeorge: doesn't seem correct to me, but it is current behavior. we should reach out to the original dev and see if he recalls why it was done this way

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
```diff
@@ -97,34 +107,129 @@ class UriStripperFilter @Inject()(configurationService: ConfigurationService)
...
               wrappedResponse.replaceHeader(
                 CommonHttpHeader.LOCATION.toString,
-                preUri.getOrElse("") + joinTokensIntoUri(locationTokens) + postUri.getOrElse(""))
+                replacePath(uri, joinTokensIntoPath(locationTokens)))
           }
         }
       }
 
+      val applicableLinkPaths = config.getLinkResource
+        .filter(resource => resource.getUriPathRegex.r.pattern.matcher(wrappedRequest.getRequestURI).matches)
```
adrianjgeorge: this still makes me sad, apparently we are the only people in the world doing this sort of thing, because otherwise this would be solved already. Wait, why don't we just solve this with an operator in commmons, and use the heck out of it? We could use `~=` like groovy.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
Mario-Lopez: Original dev is Meynard:
https://github.com/rackerlabs/repose/blob/e48463d4a20a776d728c2cf27c50d2068bfbd9bd/project-set/components/uri-stripper/src/main/java/com/rackspace/papi/components/uritranslation/UriStripperHandler.java

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
```diff
@@ -97,34 +107,129 @@ class UriStripperFilter @Inject()(configurationService: ConfigurationService)
...
+  private def getPathsForContentType(contentType: String, resource: LinkResource): List[LinkPath] = {
+    Option(contentType) match {
+      case Some(ct) if ct.toLowerCase.contains("json") => resource.getJson.toList
+      // todo: return xpath when xml is supported
+      case Some(ct) if ct.toLowerCase.contains("xml") => List.empty
+      case _ => List.empty
+    }
+  }
+
+  // todo: this function only provides support for named '$' and '.' JSONPath operators
```
adrianjgeorge: Would using the other library provide support for the `..` operator? If so you should do it

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
```diff
@@ -145,4 +250,10 @@ object UriStripperFilter {
 
   private final val UriDelimiter = "/"
   private final val QueryParamIndicator = "?"
+
+  case class MalformedJsonPathException(message: String, throwable: Throwable = null)
+    extends Exception(message, throwable)
+
+  case class LinkTransformException(message: String, throwable: Throwable = null)
```
adrianjgeorge: You aren't using the throwable anywhere, so there's no need for it.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
```diff
@@ -145,4 +250,10 @@ object UriStripperFilter {
 
   private final val UriDelimiter = "/"
   private final val QueryParamIndicator = "?"
+
+  case class MalformedJsonPathException(message: String, throwable: Throwable = null)
```
adrianjgeorge: same here

repose-aggregator/components/filters/uri-stripper-filter/src/test/scala/org/openrepose/filters/uristripper/Marshaller.scala
```diff
@@ -0,0 +1,51 @@
...
+import java.nio.charset.StandardCharsets
+
+import org.openrepose.commons.config.parser.jaxb.JaxbConfigurationParser
+import org.openrepose.commons.config.resource.ConfigurationResource
+import org.openrepose.commons.config.resource.impl.ByteArrayConfigurationResource
+import org.openrepose.filters.uristripper.config.UriStripperConfig
+
+import scala.reflect.ClassTag
+
+object Marshaller {
```
adrianjgeorge: ???

repose-aggregator/components/filters/uri-stripper-filter/src/test/scala/org/openrepose/filters/uristripper/UriStripperFilterTest.scala
```diff
@@ -114,6 +118,321 @@ class UriStripperFilterTest extends FunSpec with BeforeAndAfterEach with Matcher
...
+      request.setRequestURI("/v1/12345/foo")
+      setResponseBody(respBody, MimeType.APPLICATION_JSON.toString)
+
+      filter.doFilter(request, response, filterChain)
+
+      response.getStatus shouldEqual HttpServletResponse.SC_INTERNAL_SERVER_ERROR
+      response.getContentLength shouldEqual 0
+    }
+
+    it("should fail if the link's token index is out of bounds (fail)") {
```
adrianjgeorge: You've tested indexing failures, but what about if they haven't specified the index, and it can't figure out where to place it based on pre and post token?

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
dmnjohns: Yeah, I kind of figured since the headers mode was already mutable, we may as well just always set the body mode to mutable as well. I can make it "smarter" if you'd like, but the only way we could prevent this filter from breaking the stream is if we somehow knew before passing the response object along whether or not the response coming back contains a `Location` header.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
dmnjohns: Paging @malconis. Do you happen to remember this?

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
dmnjohns: It is a little strange that there isn't a better way of doing this. Either a `matches` method on Scala's `Regex` class, or an operator.

I'd be down for adding Groovy's `~=` operator to our commons library. If you write it as an `implicit` extension of `String` you could use it like so:
`"/my/pattern" ~= wrappedRequest.getRequestURI`

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
dmnjohns: The other library (https://github.com/josephpconley/play-jsonpath) really only helps retrieve a value given a JSONPath. Unfortunately, that doesn't help us transform the input JSON (the `JsValue` returned by the other library will not have a reference to its parent). To transform the JSON, we could use a combination of the other library with a JSON Patch library, but then things start to get a bit complex.

I think the ideal solution would be to find a way to map the library agnostic typing from the Gatling JSONPath parser to a Play JSON `JsPath`. The `parse` function from the other library does something somewhat similar, and may be helpful in figuring out what to do.

repose-aggregator/components/filters/uri-stripper-filter/src/test/scala/org/openrepose/filters/uristripper/UriStripperFilterTest.scala
dmnjohns: I'll add some tests for that case.

repose-aggregator/components/filters/uri-stripper-filter/src/test/scala/org/openrepose/filters/uristripper/Marshaller.scala
dmnjohns: This is something we should probably pull out to `commons-test`. It's used in a few places.

All this object does is convert from a `String` to a configuration object (one of the objects generated by JAXB). When testing the differences in behavior caused by different configurations, I find this method easier than hand crafting new objects with the specific values we want.

repose-aggregator/components/filters/uri-stripper-filter/src/main/resources/META-INF/schema/config/uri-stripper.xsd
```diff
@@ -65,14 +71,160 @@
...
+                    A specific ReST resource that contains one or more links which should have the stripped URI
+                    path segment re-inserted.
+                </html:p>
+            </xs:documentation>
+        </xs:annotation>
+
+        <xs:all>
+            <xs:element name="request" type="HttpMessage" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="response" type="HttpMessage" minOccurs="0" maxOccurs="1"/>
+        </xs:all>
```
Mario-Lopez: Having `link-resource` and `response` in this order seems like there would be more boiler plate.  Unless there's an advantage to being able to pair a request/response in a link-resource??

e.g.
```
<link-resource>
    <response>
        <json link-mismatch-action="continue" token-index="0">$.link-a</json>
    </response>
</link-resource>
<link-resource>
    <response>
        <json link-mismatch-action="continue" token-index="1">$.link-b</json>
    </response>
</link-resource>
<link-resource>
    <response>
        <json link-mismatch-action="continue" token-index="5">$.link-c</json>
    </response>
</link-resource>
<link-resource>
    <response>
        <json link-mismatch-action="continue">$.foo.bar.baz.link</json>
    </response>
</link-resource>
```

versus:
```
<response>
    <link-response>
        <json link-mismatch-action="continue" token-index="0">$.link-a</json>
    </link-response>
    <link-response>
        <json link-mismatch-action="continue" token-index="1">$.link-b</json>
    </link-response>
    <link-response>
        <json link-mismatch-action="continue" token-index="5">$.link-c</json>
    </link-response>
    <link-response>
        <json link-mismatch-action="continue">$.foo.bar.baz.link</json>
    </link-response>
</response>
```

dmnjohns: @Mario-Lopez It really depends on the configuration. In your first example, the `response` element is redundant. In your second example, the `link-response` element _would_ be redundant if you also defined transformations on the request for the same resources.

The funny thing is that I asked @adrianjgeorge about this exact thing this morning.

Having `request`/`response` be children of `link-resource` is consistent with what exists in the body-patcher filter where `request`/`response` are children of `change`.

Personally, I would have to see the common use case to be able to determine which is more appropriate.

Mario-Lopez: I forgot about the uri-path-regex.  👍

Mario-Lopez:
Build is happy.

repose-aggregator/components/filters/uri-stripper-filter/src/main/resources/META-INF/schema/config/uri-stripper.xsd
adrianjgeorge: So my take is, i think about things as a resource. I don't think of all my requests, and then all my responses. I think about the fact that i have this resource which has this input and this output, and this config aligns with that.

Mario-Lopez: Yup, totally agree.

Mario-Lopez:
I wrote a story to support more of JSON Path's features:
https://repose.atlassian.net/browse/REP-4224

PR #1558, opened 2016-06-21
Removing schema namespace translation tests

dmnjohns:


adrianjgeorge:
so many deletes, once it goes green, i'll merge it.

PR #1559, opened 2016-06-21
REP-3415: Atom feed functional tests updated to remove race conditions

Mario-Lopez:
I added a race condition in the functional tests for REP-3415.  This pull request is to fix that.

Code review task:
https://repose.atlassian.net/browse/REP-4162


dmnjohns:
So basically, fix the race condition by replacing global state with generated functions instead? How functional of you! I love it.

Merging.

PR #1560, opened 2016-06-28
Extracting regex operators to a commons utility

dmnjohns:
These match two of Groovy's regex operators, `=~` and `==~`. There are two ways to use this utility in Scala -- import the implicit conversion, or mix-in the trait. Technically, I suppose you could also instantiate the extended class, or do some other shenanigans to pull the implicit conversion into scope, but the two aforementioned methods are probably preferred.

Note that this branched is branched off of REP-4010, because I wanted to update that code to use the new operator.


repose-aggregator/components/filters/body-patcher-filter/src/main/scala/org/openrepose/filters/bodypatcher/BodyPatcherFilter.scala
```diff
@@ -123,7 +124,7 @@ class BodyPatcherFilter @Inject()(configurationService: ConfigurationService)
   def filterPathChanges(request: HttpServletRequest): List[ChangeDetails] = {
     val path: String = URLDecoder.decode(request.getRequestURI, StandardCharsets.UTF_8.toString)
     configuration.getChange.asScala.toList
-        .filter(_.getPath.r.pattern.matcher(path).matches)
+        .filter(_.getPath =~ path)
```
adrianjgeorge: love it

adrianjgeorge:
Once it goes green, i'll pull it in

adrianjgeorge:
retest this please

PR #1561, opened 2016-06-28
REP-3395 Atom Feed Consumption Service Should Use Akka Service Client

dmnjohns:


Mario-Lopez:
retest this please

repose-aggregator/components/services/atom-feed-service/atom-feed-service-api/src/main/java/org/openrepose/nodeservice/atomfeed/AuthenticatedRequestFactory.java
```diff
@@ -29,11 +27,11 @@
     /**
      * Mutates a request to adhere to some authentication scheme compatible with the Atom service.
      *
-     * @param atomFeedUrlConnection The unauthenticated feed {@link URLConnection} with the Atom service.
-     * @param context               A context object which contains information related to the request.
-     * @return The {@link URLConnection} with authentication mutations applied.
+     * @param feedReadRequest A mutable container for request data that will be sent to the Atom service.
```
adrianjgeorge: If it's mutable why are we returning it?

dmnjohns: So that null can be returned in the case of failure. That is, so that we have a way of identifying a failure.

Thinking about it, that was a lazy way out. We could add state to our request object indicating whether or not it's valid. Or rely on exception handling. Or something else.

repose-aggregator/components/services/atom-feed-service/atom-feed-service-impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomEntryStreamBuilder.scala
```diff
@@ -43,67 +46,84 @@ object AtomEntryStreamBuilder {
...
+            }
+          } else if (statusCode >= 400 && statusCode < 500) {
+            authenticator.foreach(_.onInvalidCredentials)
             if (firstAttempt) {
               // Inform the authenticator that the request failed and retry once
-              buildR(baseFeedUrl, context, authenticator, authenticationTimeLimit, firstAttempt = false)
+              buildR(feedURI, httpClient, context, authenticator, authenticationTimeout, firstAttempt = false)
             } else {
-              throw UnrecoverableIOException(ioe)
+              throw ClientErrorException
```
adrianjgeorge: I don't like losing the original exception, and neither will sonar if it ever figures out how to handle sonar.

repose-aggregator/components/services/atom-feed-service/atom-feed-service-impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomEntryStreamBuilder.scala
```diff
@@ -112,6 +132,10 @@ object AtomEntryStreamBuilder {
 
   case class UnrecoverableIOException(cause: Throwable) extends RuntimeException(cause)
 
+  case class ServerResponseException(message: String) extends Exception(message)
+
+  object ClientErrorException extends Exception
+
   object AuthenticationException extends Exception
```
adrianjgeorge: Not sure how i missed this the first time, but there are very few times when an exception has absolutely no arguments to a constructor.

repose-aggregator/components/services/atom-feed-service/atom-feed-service-impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/FeedReadRequestImpl.scala
```diff
@@ -0,0 +1,36 @@
...
+ */
+package org.openrepose.nodeservice.atomfeed.impl
+
+import java.net.URI
+import java.util
+
+import org.openrepose.nodeservice.atomfeed.FeedReadRequest
+
+class FeedReadRequestImpl(private var uri: URI, headers: util.Map[String, util.List[String]] = new util.TreeMap[String, util.List[String]](String.CASE_INSENSITIVE_ORDER))
+  extends FeedReadRequest {
```
adrianjgeorge: Why is this interfaced? It's just a normal data object. It should just be the class no interface and it should live in the api.

repose-aggregator/components/services/atom-feed-service/atom-feed-service-impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/auth/OpenStackIdentityV2AuthenticatedRequestFactory.scala
```diff
@@ -44,7 +45,7 @@ object OpenStackIdentityV2AuthenticatedRequestFactory {
 /**
   * Fetches a token from the OpenStack Identity service, if necessary, then adds the token to the request.
   */
-class OpenStackIdentityV2AuthenticatedRequestFactory(configuration: OpenStackIdentityV2AuthenticationType)
+class OpenStackIdentityV2AuthenticatedRequestFactory @Inject()(configuration: OpenStackIdentityV2AuthenticationType)
```
adrianjgeorge: this was supposed to also use the connection pool, with the same configured pool as it's feed

repose-aggregator/components/services/atom-feed-service/atom-feed-service-api/src/main/java/org/openrepose/nodeservice/atomfeed/AuthenticatedRequestFactory.java
adrianjgeorge: Yeah, this should be handled with exceptions, returning null as a failure is bad business.

repose-aggregator/components/services/atom-feed-service/atom-feed-service-impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomEntryStreamBuilder.scala
wdschei: There is no longer an exception caught here, rather it is just the response code.

repose-aggregator/components/services/atom-feed-service/atom-feed-service-impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomEntryStreamBuilder.scala
wdschei: These were updated with messages only since they are not used inside of `catch` blocks that could provide a `cause`.

repose-aggregator/components/services/atom-feed-service/atom-feed-service-impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomEntryStreamBuilder.scala
adrianjgeorge: Fair enough, i'd still like a message of some sort so we have tracability, also maybe a log.

wdschei: Updated.

repose-aggregator/components/services/atom-feed-service/atom-feed-service-impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomEntryStreamBuilder.scala
adrianjgeorge: This one doesn't even have a message. Or do you mean just now?

wdschei: Just now. Not even pushed yet.

wdschei: On second thought, the `ClientErrorException` and `AuthenticationException` are only thrown and also caught in one place each. Since there is an appropriate message logged when they are caught, do we really want to make these full [inner] Classes?

adrianjgeorge: Not sure what you mean by full inner classes, but just because the only usages right now have that doesn't mean that someone who picks it up in the future will do the right thing. I always assume that i'm not going to be as careful later as i am being right now. Messages cost us basically nothing to put on them, so message them up.

wdschei: It was late in the day and I was fading. Full `inner classes` not needed, just `case class`es instead of `objects`. C/W. Pushing now.

repose-aggregator/components/services/atom-feed-service/atom-feed-service-impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/FeedReadRequestImpl.scala
wdschei: If @dmnjohns has no inputs on this, then I will roll this up.

dmnjohns: Nope, no input. I split them because I didn't want to put an implemented class into the API package, but if we're okay with that, then do it up.

repose-aggregator/components/services/atom-feed-service/atom-feed-service-api/src/main/java/org/openrepose/nodeservice/atomfeed/AuthenticatedRequestFactory.java
wdschei: Updated.

repose-aggregator/components/services/atom-feed-service/atom-feed-service-impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/FeedReadRequestImpl.scala
wdschei: Rolled up.

repose-aggregator/components/services/atom-feed-service/atom-feed-service-impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/auth/OpenStackIdentityV2AuthenticatedRequestFactory.scala
wdschei: This was a bit more of a pain than I had anticipated when you mentioned it. I think it is mostly doing it, but there are a couple of tests I think I broke along the way. I will fix them after the Jenkins run.

repose-aggregator/components/services/atom-feed-service/atom-feed-service-api/src/main/java/org/openrepose/nodeservice/atomfeed/AuthenticatedRequestFactory.java
```diff
@@ -29,15 +27,24 @@
...
      * authenticateRequest method are found to be invalid.
      */
     void onInvalidCredentials();
+
+    /**
+     * Sets the configured Connection Pool Id.
+     * <p>
+     * NOTE: This default implementation does nothing and will need to be overridden if the value is required.
+     */
+    default void setConnectionPoolId(String poolId) {
```
adrianjgeorge: this isn't meant to be part of the interface for this, because not all implementations are going to need to make requests. The intent was to wire the feed's configuration into the auth component. Not the entire feed service config, just the feed itself.

repose-aggregator/components/services/atom-feed-service/atom-feed-service-api/src/main/java/org/openrepose/nodeservice/atomfeed/AuthenticationRequestException.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+package org.openrepose.nodeservice.atomfeed;
+
+public class AuthenticationRequestException extends Exception {
+
+    public AuthenticationRequestException(String message) {
+        super(message);
+    }
+
+    public AuthenticationRequestException(String message, Throwable cause) {
```
adrianjgeorge: I was going to ask if both of these are used, but since it's part of an extensible api it makes sense to provide both.

repose-aggregator/components/services/atom-feed-service/atom-feed-service-impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
```diff
@@ -132,31 +141,92 @@ class AtomFeedServiceImpl @Inject()(@Value(ReposeSpringProperties.CORE.REPOSE_VE
...
+                  throw new IllegalArgumentException(fqcn + " is not an AuthenticatedRequestFactory", cce)
+              }
+            }
+
+            authenticationContext.setParent(applicationContext)
+
+            // Register the feed's authentication configuration as a Spring Bean so that it can be auto-wired when the
+            // AuthenticatedRequestFactory is created.
+            authenticationContext.getBeanFactory.registerSingleton(id, config)
+            authenticationContext.refresh()
```
adrianjgeorge: So you'd add another line here adding the feeds configuration.

adrianjgeorge:
Looks good otherwise.

repose-aggregator/components/services/atom-feed-service/atom-feed-service-api/src/main/java/org/openrepose/nodeservice/atomfeed/AuthenticationRequestException.java
wdschei: BTW, they are both used since we don't want to lose the original exception when they are available.

repose-aggregator/components/services/atom-feed-service/atom-feed-service-api/src/main/java/org/openrepose/nodeservice/atomfeed/AuthenticatedRequestFactory.java
wdschei: This was removed.

repose-aggregator/components/services/atom-feed-service/atom-feed-service-impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/AtomFeedServiceImpl.scala
wdschei: Updated.

repose-aggregator/components/services/atom-feed-service/atom-feed-service-impl/src/main/scala/org/openrepose/nodeservice/atomfeed/impl/auth/OpenStackIdentityV2AuthenticatedRequestFactory.scala
```diff
@@ -67,59 +70,75 @@ class OpenStackIdentityV2AuthenticatedRequestFactory(configuration: OpenStackIde
...
+        Failure(new Exception("Failed to retrieve token from the identity service."))
     }
   }
 
-  private def parseTokenFromJson(tokenStream: InputStream): String = {
-    val contentString = Source.fromInputStream(tokenStream)(Codec.UTF8).mkString
+  override def onInvalidCredentials(): Unit = cachedToken = None
 
-    (Json.parse(contentString) \ "access" \ "token" \ "id").as[String]
+  def setConnectionPoolId(poolId: String): Unit = {
```
wdschei: As in it behaves appropriately, or that it is even needed?

adrianjgeorge: It seems that it's not needed, but is the behavior of the app right since it's not being called?

wdschei: Is no longer needed in the current way of doing things. Now that the parent feed's config is also being put in the context, it is currently only being set at instantiation time. So, yes, it can be removed and yes, it behaving properly.

PR #1568, opened 2016-07-15
Rep 1572 Include header that differentiates default project ID in keystone v3

JoelRizner:


wdschei:
You need to update the example config show the use of the new element and it's attributes.

Mario-Lopez:
Don't forget to update the example config in the artifact's directory, too.

JoelRizner:
is that the one in artifacts/filter-bundle/src/config/filters?

wdschei:
Yep. That is the second one.

PR #1571, opened 2016-08-01
Rep4304

JoelRizner:
This should be released as 7.3.7.0 i believe.

### DIFF
--- a/repose-aggregator/components/filters/herp/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
+++ b/repose-aggregator/components/filters/herp/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
@@ -53,9 +53,9 @@ class HerpFilter @Inject()(configurationService: ConfigurationService,
   extends Filter with HttpDelegationManager with UpdateListener[HerpConfig] with LazyLogging {
   private final val DEFAULT_CONFIG = "highly-efficient-record-processor.cfg.xml"
   private final val X_PROJECT_ID = "X-Project-ID"
-  private final val X_METHOD_LABEL: String = "X-METHOD-LABEL"
-  private final val X_PP_USER: String = "X-PP-User"
-  private final val X_PP_GROUPS: String = "X-PP-Groups"
+  private final val X_METHOD_LABEL = "X-METHOD-LABEL"
+  private final val X_PP_USER = "X-PP-User"
+  private final val X_PP_GROUPS = "X-PP-Groups"
 
   private var config: String = _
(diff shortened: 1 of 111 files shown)


adrianjgeorge:
you need to update the poms to be 7.3.7.0-SNAPSHOT

adrianjgeorge:
Looks fine, hoping it proves unneccessary

repose-aggregator/components/filters/herp/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
```diff
@@ -152,7 +154,9 @@ class HerpFilter @Inject()(configurationService: ConfigurationService,
       "clusterId" -> clusterId,
       "nodeId" -> nodeId,
       "requestorIp" -> Option(stripHeaderParams(httpServletRequest.getHeader(CommonHttpHeader.X_FORWARDED_FOR.toString)))
-        .getOrElse(httpServletRequest.getRemoteAddr)
+        .getOrElse(httpServletRequest.getRemoteAddr),
+      "userResp" -> Option(stripHeaderParams(httpServletResponse.getHeader(X_PP_USER))).orNull,
+      "groupsResp" -> Option(stripHeaderParams(httpServletResponse.getHeader(X_PP_GROUPS))).orNull
```
dmnjohns: Kind of odd. You shouldn't need to wrap anything in an `Option` here, if null is an acceptable value.

That is, if `stripHeaderParams(...)` returns null, well, that's what you want.

repose-aggregator/components/filters/herp/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
```diff
@@ -54,6 +54,8 @@ class HerpFilter @Inject()(configurationService: ConfigurationService,
   private final val DEFAULT_CONFIG = "highly-efficient-record-processor.cfg.xml"
   private final val X_PROJECT_ID = "X-Project-ID"
   private final val X_METHOD_LABEL: String = "X-METHOD-LABEL"
+  private final val X_PP_USER: String = "X-PP-User"
+  private final val X_PP_GROUPS: String = "X-PP-Groups"
```
dmnjohns: Considering this is a `final val` (i.e., a constant) of an obvious type (i.e., `String`), I would say that annotating the type is not really necessary. That's just personal preference, though.

adrianjgeorge:
This work has been superseded.

PR #1572, opened 2016-08-01
REP-4145 Added SSL/TLS Client Authentication to the Connection Pool service impl

wdschei:


repose-aggregator/components/services/http-client-service/http-client-service-api/src/main/resources/META-INF/schema/examples/http-connection-pool.cfg.xml
```diff
@@ -25,4 +25,13 @@
     <!-- Configuration for a custom pool that uses all default values -->
     <pool id="alldefaultspool" default="false"/>
 
+    <!-- Configuration for a connection pool that uses SSL/TLS Client Authentication -->
+    <pool id="clientAuthentication"
+          default="false"
+          keystore-filename="keystore.jks"
+          keystore-password="password"
+          key-password="secret"
+          truststore-filename="truststore.jks"
```
adrianjgeorge: man, do we really need both? ugh

wdschei: The `truststore` is not a requirement, but if we are setting this up we might as well do it now before someone asks.

adrianjgeorge: fair enough

repose-aggregator/components/services/http-client-service/http-client-service-impl/src/main/java/org/openrepose/core/services/httpclient/impl/HttpConnectionPoolProvider.java
```diff
@@ -83,6 +88,39 @@ public static HttpClient genClient(PoolType poolConf) {
         DefaultHttpClient client = new DefaultHttpClient(cm, params);
 
         SSLContext sslContext = ProxyUtilities.getTrustingSslContext();
+
+        if (poolConf.getKeystoreFilename() != null) {
+            SSLContextBuilder sslContextBuilder = SSLContexts.custom();
+            try {
+                File keystoreFile = new File(poolConf.getKeystoreFilename());
+                if (!keystoreFile.isAbsolute()) {
+                    keystoreFile = new File(configRoot, poolConf.getKeystoreFilename());
```
adrianjgeorge: you don't need to do this if, this line will always behave correctly, or at least that's what the docs say.

repose-aggregator/components/services/http-client-service/http-client-service-impl/src/main/java/org/openrepose/core/services/httpclient/impl/HttpConnectionPoolProvider.java
```diff
@@ -83,6 +88,39 @@ public static HttpClient genClient(PoolType poolConf) {
...
+                char[] keystorePassword = poolConf.getKeystorePassword() == null ? null : poolConf.getKeystorePassword().toCharArray();
+                char[] keyPassword = poolConf.getKeyPassword() == null ? null : poolConf.getKeyPassword().toCharArray();
+                sslContextBuilder = sslContextBuilder.loadKeyMaterial(keystoreFile, keystorePassword, keyPassword);
+
+                if (poolConf.getTruststoreFilename() == null) {
+                    sslContextBuilder = sslContextBuilder.loadTrustMaterial(keystoreFile, keystorePassword);
+                } else {
+                    File truststoreFile = new File(poolConf.getTruststoreFilename());
+                    if (!truststoreFile.isAbsolute()) {
+                        truststoreFile = new File(configRoot, poolConf.getTruststoreFilename());
```
adrianjgeorge: same here

repose-aggregator/components/services/http-client-service/http-client-service-impl/src/main/java/org/openrepose/core/services/httpclient/impl/HttpConnectionPoolProvider.java
wdschei: In practice it didn't work. It always pre-pended the `configRoot` to the filename.

adrianjgeorge: That's weird, it worked for me elsewhere....hmm......

adrianjgeorge:
a functional test that has two different pools one of which is secure seems like a good idea.

adrianjgeorge:
new test looks good.

repose-aggregator/components/services/http-client-service/http-client-service-impl/src/test/groovy/org/openrepose/core/services/httpclient/impl/HttpConnectionPoolProviderTest.groovy
```diff
@@ -107,8 +125,124 @@ class HttpConnectionPoolProviderTest {
...
+        def sslContextFactory = new SslContextFactory()
+        sslContextFactory.keyStoreResource = Resource.newClassPathResource("server.jks")
+        sslContextFactory.keyStorePassword = "password"
+        sslContextFactory.keyManagerPassword = "password"
+        sslContextFactory.needClientAuth = true
+        sslContextFactory.trustStoreResource = Resource.newClassPathResource("client.jks")
+        sslContextFactory.trustStorePassword = "password"
+
+        // SSL HTTP Configuration
+        def https_config = new HttpConfiguration()
```
dmnjohns: What's with the variable name here? What is this, Python?! 😄

wdschei: That was the name used in the original tests. I'll update it, because yeah. Standards. :-)

repose-aggregator/components/services/http-client-service/http-client-service-impl/src/test/groovy/org/openrepose/core/services/httpclient/impl/HttpConnectionPoolProviderTest.groovy
```diff
@@ -107,8 +125,124 @@ class HttpConnectionPoolProviderTest {
...
+                server,
+                new SslConnectionFactory(sslContextFactory, HttpVersion.HTTP_1_1.asString()),
+                new HttpConnectionFactory(https_config)
+        )
+        sslConnector.setPort(0)
+
+        // Start the server
+        def statusCode = HttpServletResponse.SC_OK
+        def responseContent = "The is the plain text test body data.\n".bytes
+        def contentType = "text/plain;charset=utf-8"
```
dmnjohns: Considering you're always declaring the charset to be `UTF-8`, which is totally legit, you also probably want to get the `String` as `Byte`s in the previous line using that charset. That is, I believe `"string".bytes` uses the platform default character encoding, which is usually, but not necessarily, `UTF-8`. So:
`def responseContent = "The is the plain text test body data.\n".getBytes(StandardCharsets.UTF_8)`

repose-aggregator/components/services/http-client-service/http-client-service-impl/src/test/groovy/org/openrepose/core/services/httpclient/impl/HttpConnectionPoolProviderTest.groovy
```diff
@@ -107,8 +125,124 @@ class HttpConnectionPoolProviderTest {
...
+        def serverPort = ((ServerConnector) server.connectors[0]).getLocalPort()
+
+        // Add the client creds to the connection pool
+        poolType.setKeystoreFilename(HttpConnectionPoolProviderTest.class.getResource("/client.jks").getFile())
+        poolType.setKeystorePassword("password")
+        poolType.setKeyPassword("password")
+        poolType.setTruststoreFilename(HttpConnectionPoolProviderTest.class.getResource("/server.jks").getFile())
+        poolType.setTruststorePassword("password")
+
+        def configRoot = new File(HttpConnectionPoolProviderTest.class.getResource("/client.jks").getFile()).getParent()
```
dmnjohns: I've seen this getResource call a bunch. Not that it makes a difference in the end, but it may be easier to read to just get these file locations up-front and bind them to some well-named variable.

dmnjohns: Also, setting the `configRoot` may be redundant here? Wouldn't the `setKeystoreFilename` and `setTruststoreFilename` calls set the full filename? In fact, I expect what you're actually passing to the `File` constructor in the `HttpConnectionPoolProvider` looks something like:
`new File("/some/path/to/resources", "/some/path/to/resources/client.jks")`

If that's the case, you're relying on `File` to handle absolute paths correctly. Which is should. Even so, it seems like we're passing redundant information.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/httpconnectionpool/HttpsClientAuthMultiStorePoolTest.groovy
```diff
@@ -0,0 +1,156 @@
...
+    def contentType = "text/plain;charset=utf-8"
+    @Shared
+    def identityEndpoint
+    @Shared
+    def MockIdentityV2Service fakeIdentityV2Service
+
+    def setupSpec() {
+        deproxy = new Deproxy()
+
+        reposeLogSearch.cleanLog()
```
dmnjohns: Huh. I feel like the `ReposeValveTest` should just do this as part if its `setupSpec`. Maybe.

dmnjohns:
I left a few comments, but they were mostly about things that just seemed kind of odd to me. Overall, this looks good. 👍

repose-aggregator/components/services/http-client-service/http-client-service-impl/src/test/groovy/org/openrepose/core/services/httpclient/impl/HttpConnectionPoolProviderTest.groovy
wdschei: Good call. Updated.

repose-aggregator/components/services/http-client-service/http-client-service-impl/src/test/groovy/org/openrepose/core/services/httpclient/impl/HttpConnectionPoolProviderTest.groovy
wdschei: Extracted the `getResource()` calls out to `static final` constants.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/httpconnectionpool/HttpsClientAuthMultiStorePoolTest.groovy
wdschei: Not a bad thought. It is something that is done for just about every test that uses this framework. Also creating the `Deproxy` object would probably be good too.
We should probably add a story to clean this up, or just add it to the one to parallelize the framework.

repose-aggregator/components/services/http-client-service/http-client-service-impl/src/test/groovy/org/openrepose/core/services/httpclient/impl/HttpConnectionPoolProviderTest.groovy
wdschei: The `configRoot` in this case is redundant since we know it is an absolute path being passed in.
Removed.

wdschei:
Since the Jenkins to GitHub hooks are currently buggered, this is the build for the 9ef7705 commit:

- https://jenkins.openrepose.org/job/repose-gradle-pull_request/595/

PR #1580, opened 2016-08-18
REP-4336 Overwrite target header in header translation

adrianjgeorge:


adrianjgeorge:
just the config so far

dmnjohns:
Config looks great. 👍

adrianjgeorge:
since the builder is busted, https://jenkins.openrepose.org/job/repose-gradle-pull_request/611/

repose-aggregator/components/filters/header-translation-filter/src/main/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilter.scala
```diff
@@ -74,10 +74,19 @@ class HeaderTranslationFilter @Inject()(configurationService: ConfigurationServi
...
-              if (quality.isEmpty) {
-                httpRequest.addHeader(newHeaderName, originalHeaderValue)
+              def writeHeaders(writeWithoutQuality: (String, String) => Unit,
+                               writeWithQuality: (String, String, Double) => Unit): Unit = {
+                if (quality.isEmpty) {
+                  writeWithoutQuality(newHeaderName, originalHeaderValue)
+                } else {
+                  writeWithQuality(newHeaderName, withoutQuality(originalHeaderValue), quality.get)
+                }
+              }
```
dmnjohns: I kind of wonder if it's worth breaking this out into a function. It's usage is pretty narrowly scoped, and whether we do it this way or the "Java" way, I think it comes down to the same thing -- we branch into four distinct paths. That is, the cyclomatic complexity of this method is the same either way, regardless of whether we nest if/then statements or create a 4-way if/else-if statement, where here you've done the former.

I guess I'm wondering what this approach buys us.

Just to be clear, I consider the "Java" way to be:
```
if (quality.isEmpty && sourceHeader.isOverwriteTarget) {
  httpRequest.replaceHeader(newHeaderName, originalHeaderValue)
} else if (quality.isEmpty && !sourceHeader.isOverwriteTarget) {
  httpRequest.addHeader(newHeaderName, originalHeaderValue)
} else if (!quality.isEmpty && sourceHeader.isOverwriteTarget) {
  httpRequest.replaceHeader(newHeaderName, originalHeaderValue, quality.get)
} else if (!quality.isEmpty && !sourceHeader.isOverwriteTarget) {
  httpRequest.addHeader(newHeaderName, originalHeaderValue, quality.get)
}
```

dmnjohns:
Left a comment on the code that more or less comes down to preference, I think.

Putting that aside, everything looks good. 👍

repose-aggregator/components/filters/header-translation-filter/src/main/scala/org/openrepose/filters/headertranslation/HeaderTranslationFilter.scala
adrianjgeorge: That if-else is pretty nasty. The other option is to nest an if that is identical inside both sides of the if else, which i didn't do because i didn't want the code duplication. This way shows the originals intention of doing it with or without the quality, but allows me to make the distinction between adding and replacing once and in a simple manner.

adrianjgeorge: So to say it clearer, the two things aren't really related and this allows me to separate them and deal with them individually.

dmnjohns: It totally does, but in doing so introduces another layer of abstraction (and another call on the stack). It seems like trading one minor form of complexity for another, which is why in the comment below I said that it probably comes down to preference.

Anywhoo, if you're happy with it, I have no reason not to be.

wdschei: I kind of like the nested function way of separating the functionality.

wdschei:
Looks good to me.

PR #1582, opened 2016-08-23
REP-4319 Add Client Authentication Support to the Distributed Datastore

dmnjohns:


dmnjohns:
Ah. I still need to add the servlet connection auth stuff to the configuration.

dmnjohns:
Actually, I'm going to make that a separate task.

adrianjgeorge:
first config change looks good so far

repose-aggregator/components/services/datastore-service/datastore-service-api/src/main/java/org/openrepose/core/services/datastore/DatastoreService.java
```diff
@@ -39,6 +40,7 @@
      * @return
      * @throws DatastoreUnavailableException if no datastore exists with the given datastoreName
      */
+    // TODO: Rename this to getDistributedDatastore for clarity?
     Datastore getDatastore(String datastoreName) throws DatastoreUnavailableException;
```
adrianjgeorge: theoritically this is going to strictly give you the dd, if you are in a non dd environment it should give you the local

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/main/resources/META-INF/schema/config/dist-datastore-configuration.xsd
```diff
@@ -46,6 +47,73 @@
...
+
+        <xs:attribute name="keystore-password" type="xs:string" use="optional">
+            <xs:annotation>
+                <xs:documentation>
+                    <html:p>The password for the client authentication keystore.</html:p>
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+
+        <xs:attribute name="key-password" type="xs:string" use="optional">
```
adrianjgeorge: I don't think we gave this option anywhere else. I think we forced them to be the same everywhere.

repose-aggregator/core/repose-core/src/test/groovy/org/openrepose/nodeservice/httpcomponent/RequestProxyServiceImplTest.groovy
```diff
@@ -161,4 +169,188 @@ class RequestProxyServiceImplTest extends Specification {
...
+
+        when:
+        requestProxyService.delete("http://www.google.com", "", Collections.emptyMap(), testPoolId)
+
+        then:
+        noAssertionError { verify(httpClientService).getClient(testPoolId) }
+    }
+
+    // TODO: Use Spock Mocking framework to avoid this hack (caused by Mockito mocks returning null)
+    def noAssertionError(Closure verification) {
```
adrianjgeorge: not sure why this was necessary. i haven't run into issues before jsut doing mockito verify's

Mario-Lopez: I'm not sure why Mockito is being used at all in a Spock test.

```
// prior to test:
HttpClient httpClient = Mock()
HttpClientService httpClientService = Mock()
// other stuff to set up all of those mocks for the code under test
```

```
given:
String testPoolId = "test-pool-id"
httpClient.execute(_ as HttpUriRequest) >> new BasicHttpResponse(HttpVersion.HTTP_1_1, 204, "OK")

when:
requestProxyService.delete("http://www.google.com", "", Collections.emptyMap(), testPoolId)

then:
1 * httpClientService.getClient(testPoolId)
```

adrianjgeorge: Probably because spock mocks are a hot mess, that conflates starting conditions and the test.

Mario-Lopez: Yes, sadly it does let you do that.  You don't have to, though.  Whatever works.

repose-aggregator/components/services/datastore-service/datastore-service-api/src/main/java/org/openrepose/core/services/datastore/DatastoreService.java
dmnjohns: Did you mean this isn't going to strictly give you the dd?

That's how the contract read to me, which is why I left the TODO. However, we currently only ever construct one local database, and only have a way of constructing a distributed database through the API. I guess it's a matter of contract vs implementation. If we plan to support getting of a local or DD datastore in the future, then this is still a good name.

I will remove the TODO.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/main/resources/META-INF/schema/config/dist-datastore-configuration.xsd
dmnjohns: This was actually mostly copy-paste configuration. I believe the key-password is optional in both the container configuration and the http connection pool configuration.

repose-aggregator/core/repose-core/src/test/groovy/org/openrepose/nodeservice/httpcomponent/RequestProxyServiceImplTest.groovy
dmnjohns: Well, Mockito was already being used in the test, so I was just going to continue using it.

The issue with Mockito's verify in Spock tests is that verify does not return a value; it only throws an error if the assertion is not satisfied. Well, Spock treats that as a failure. That is, `null`, `false`, and `0` are all treated as failure values by Spock in the `then` clause.

I could totally use Spock's Mocking framework, which is what the TODO I left suggests, I just didn't know if that's the path we wanted to take. I'm not sure we use Spock's Mocking framework anywhere else, and I would probably move this entire test over to it so that we aren't mixing Spock Mocks and Mockito Mocks.

Mario-Lopez: The only other option I can think of is ugly (using Spock's noExceptionThrown()):

```
given:
// stuff

when:
requestProxyService.delete("http://www.google.com", "", Collections.emptyMap(), testPoolId)

and:  // optional to add the "and"; either way the Mockito verification is in the "when" block
verify(httpClientService).getClient(testPoolId)

then:
noExceptionThrown()
```

dmnjohns: That also totally works.

I'm fine with whatever you guys want.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/main/resources/META-INF/schema/config/dist-datastore-configuration.xsd
adrianjgeorge: Fair enough, i just didn't remember it then. good job being consistent

repose-aggregator/core/repose-core/src/test/groovy/org/openrepose/nodeservice/httpcomponent/RequestProxyServiceImplTest.groovy
```diff
@@ -161,4 +168,220 @@ class RequestProxyServiceImplTest extends Specification {
...
+        when(httpClient.getParams()).thenReturn(new BasicHttpParams())
+
+        when:
+        requestProxyService.proxyRequest("http://www.google.com", new MockHttpServletRequest(), new MockHttpServletResponse())
+
+        and:
+        verify(httpClientService).getClient(null)
+
+        then:
+        noExceptionThrown()
```
adrianjgeorge: Urg, this is worse. go back to the old way but just change your method to be:
```def verification(Closure cl) {
     cl()
     true
}```

This will give us mockitos nice error message hopefully.

Comment (dmnjohns) [inline repose-aggregator/core/repose-core/src/test/groovy/org/openrepose/nodeservice/httpcomponent/RequestProxyServiceImplTest.groovy]: Fair enough.

File: repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/services/RequestProxyService.java [hunk 5]
```diff
@@ -29,19 +29,35 @@
 public interface RequestProxyService {
 
     //TODO: this is the most terrible return value ever
+    //TODO: Deprecate the methods which don't take a connection pool id?
```
wdschei: Do we want to go ahead and deprecate these for this PR?

dmnjohns: So we can, but the reason that I haven't done so yet is that I really just want to delete those other methods.

Alas, I don't think I can since this service API is fairly exposed and doing so would force the next release to be a second digit release.

Technically, I also changed some internal DD stuff in a non-backwards compatible way, but those really shouldn't be getting touched by anyone else (e.g., `CacheRequest`). That is, while developers could technically use that code, it isn't exposed as a service interface or anything like that.

So yeah, I'll deprecate those other methods unless someone tells me that we want to delete those methods instead. Or points out a way of doing what I did without changing the contract. ;)

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/datastore/DistDatastoreClientAuthMultiTest.groovy
```diff
@@ -0,0 +1,138 @@
...
+        Files.copy(clientFileOrig.toPath(), clientFileDest)
+        repose.configurationProvider.applyConfigs("features/services/datastore/clientauth", params)
+        repose.configurationProvider.applyConfigs("features/services/datastore/clientauth/truststore", params)
+
+        SSLContext sslContext = SSLContexts.custom()
+                .loadKeyMaterial(serverFile, "password".toCharArray(), "password".toCharArray())
+                .loadTrustMaterial(clientFile, "password".toCharArray())
+                .build()
+        CloseableHttpClient client = HttpClients.custom()
+                .setSSLSocketFactory(new SSLConnectionSocketFactory(sslContext, NoopHostnameVerifier.INSTANCE))
```
dmnjohns: We want to add a setting to the connection pool for hostname verification (allowing it to be disabled, at least). This will, of course, require additional documentation.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/main/java/org/openrepose/core/services/datastore/impl/distributed/CacheRequest.java
```diff
@@ -79,12 +79,20 @@ public static boolean isCacheRequestValid(HttpServletRequest request) {
...
     }
 
-    public static String urlFor(InetSocketAddress remoteEndpoint, String key) {
-        return new StringBuilder("http://").append(remoteEndpoint.getAddress().getHostAddress()).append(":").append(remoteEndpoint.getPort()).append(CACHE_URI_PATH).append(key).toString();
+    public static String urlFor(InetSocketAddress remoteEndpoint, String key, boolean useHttps) {
+        String proto = "http";
+        if (useHttps) {
+            proto = "https";
+        }
+        return new StringBuilder(proto).append("://").append(remoteEndpoint.getAddress().getHostAddress()).append(":").append(remoteEndpoint.getPort()).append(CACHE_URI_PATH).append(key).toString();
```
adrianjgeorge: whelp here's how we get an ip instead of a hostname

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/main/java/org/openrepose/nodeservice/distributed/jetty/DistributedDatastoreServer.java
```diff
@@ -85,4 +112,17 @@ public void stop() throws Exception {
     public int getPort() {
         return port;
     }
+
+    private File guardedAbsoluteFile(String parent, String child) {
```
adrianjgeorge: This needs to be extracted, we need it lots of places.

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/services/RequestProxyService.java
```diff
@@ -31,17 +31,38 @@
     //TODO: this is the most terrible return value ever
     int proxyRequest(String targetHost, HttpServletRequest request, HttpServletResponse response) throws IOException;
 
+    int proxyRequest(String targetHost, HttpServletRequest request, HttpServletResponse response, String connPoolId)
+            throws IOException;
+
     void setRewriteHostHeader(boolean value);
 
+    @Deprecated
```
adrianjgeorge: Why were these deprecated?

adrianjgeorge: Is the DD the only consumer of this object?

dmnjohns: All of the deprecated methods have an equivalent method that just also happens to take a connection pool ID. The deprecated methods just call through to the equivalent method passing null as the connection pool ID. The result is an interface with quite a few methods. The alternative would be an interface with fewer methods, but which expected null arguments in certain cases. It should be noted that the connection pool service API expects null in certain cases.

Initially I had left a TODO, as I wasn't sure whether or not we would want to deprecate these. There was a comment chain that started earlier this week asking about the TODO, so I just went ahead and did the TODO. I can drop the deprecations if you prefer the API as-is.

dmnjohns: Yes, the DD is currently the only Repose component which uses this API, however it is exposed as a "service" so there is the potential that a user has developed something which also uses this API.

dmnjohns:
**PASSED**
Here are the test results from the last run:
https://jenkins-proto.openrepose.org/job/repose-gradle-pull_request/631/

wdschei:
Since this passed all the tests and specifically the new functional test, I'm going to call QE good on this.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/datastore/DistDatastoreClientAuthMultiTest.groovy
dmnjohns: Turns out, hostname verification is currently disabled, so this change is not needed at this time.

wdschei: I created a story for this since it would be nice to enable it for Federated products:

- https://repose.atlassian.net/browse/REP-4414

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/main/java/org/openrepose/nodeservice/distributed/jetty/DistributedDatastoreServer.java
dmnjohns: Done.

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/services/RequestProxyService.java
adrianjgeorge: Given that, if there are absolutely no usages of the other methods just remove them, and we'll do the second digit again. Somebody remind to updat e all references to 8.1.1.0 in the documentation

dmnjohns: Will do.

dmnjohns:
Retest this please.

dmnjohns:
PASSED
Here are the test results from the last run:
https://jenkins-proto.openrepose.org/job/repose-gradle-pull_request/643/

PR #1583, opened 2016-08-30
REP-4030 Tenant ID removal from request body.

wdschei:


dmnjohns:
Here are the test results from the last run:
https://jenkins.openrepose.org/job/repose-gradle-pull_request/628/

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
```diff
@@ -102,7 +105,30 @@ class UriStripperFilter @Inject()(configurationService: ConfigurationService)
         wrappedRequest.setRequestURI(joinTokensIntoPath(uriTokens))
       }
 
-      filterChain.doFilter(wrappedRequest, wrappedResponse)
+      val requestLinkPaths = config.getLinkResource
+        .filter(resource => resource.getUriPathRegex =~ wrappedRequest.getRequestURI)
```
adrianjgeorge: I'm so glad we added this operator

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
```diff
@@ -102,7 +105,30 @@ class UriStripperFilter @Inject()(configurationService: ConfigurationService)
...
+        .filter(resource => isMatchingMethod(resource.getHttpMethods.toSet, wrappedRequest.getMethod))
+        .map(resource => getPathsForContentType(wrappedRequest.getContentType, resource.getRequest))
+        .fold(List.empty[LinkPath])(_ ++ _)
```
adrianjgeorge: what?

wdschei: @dmnjohns Why did you originally use the `map` & `fold` instead of just a `flatmap`?

dmnjohns: It looks like that last line compresses a `List[List[LinkPath]]` down to a `List[LinkPath]`.

The better way to do this would probably be to change the `map` line to a `flatMap` and then call `toList` at the end.

wdschei: So why you do it then man???
I fix.

dmnjohns: As for why, I have no idea.

Maybe because `flatMap` returns a `Buffer` in this case and my brain got confused? Or because I was just really into `fold` that day?

wdschei: Fixed.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
```diff
@@ -102,7 +105,30 @@ class UriStripperFilter @Inject()(configurationService: ConfigurationService)
...
+
+      val filterChainRequest = {
+        if (token.isDefined && requestLinkPaths.nonEmpty) {
+          Option(wrappedRequest.getContentType) match {
+            case Some(ct) if ct.toLowerCase.contains("json") => handleJsonRequestLinks(wrappedRequest, previousToken, nextToken, token, requestLinkPaths)
```
adrianjgeorge: what are all these extra tokens?

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
```diff
@@ -102,7 +105,30 @@ class UriStripperFilter @Inject()(configurationService: ConfigurationService)
...
+        } else {
+          Success(wrappedRequest)
+        }
+      }
+
+      filterChainRequest match {
+        case Success(request) =>
+          filterChain.doFilter(request, wrappedResponse)
+        case Failure(e) =>
+          wrappedResponse.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, e.getMessage)
```
adrianjgeorge: Should this strictly be a 500, there's a pretty good argument for it to be a 400.

adrianjgeorge: To be clear, I'm not necessarily saying a 500 is wrong either, it's a weird case.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
```diff
@@ -112,21 +138,22 @@ class UriStripperFilter @Inject()(configurationService: ConfigurationService)
         Try(wrappedResponse.replaceHeader(
           CommonHttpHeader.LOCATION.toString,
           replaceIntoPath(originalLocation, token.get, previousToken, nextToken, None))) match {
-          case Failure(ex: PathRewriteException) => logger.warn("Failed while trying to rewrite the location header", ex)
           case Success(_) => //don't care
+          case Failure(e: PathRewriteException) => logger.warn("Failed while trying to rewrite the location header", e)
+          case Failure(e) => throw e
```
adrianjgeorge: I suppose keeping the original exception instead of accidentally wrapping it in a failure to match exception makes sense. This clearly thinking about the problems that were expected then.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
wdschei: For this, they aren't used... yet. This was the potential to use them contextually like on the response in the future. It also let us easily reuse the response methods with only a minor tweak in the logic. I didn't want to throw away a bunch of references that we didn't NEED right now when there was a big potential they would be needed later. So since they were there and the reuse was easier with them, I kept them.

adrianjgeorge: I'm curious what reuse you are getting since we are taking opposite actions. I suppose i'll run into it soon.

wdschei: Mainly the exception and failure logic currently.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
```diff
@@ -135,9 +162,47 @@ class UriStripperFilter @Inject()(configurationService: ConfigurationService)
     }
   }
 
+  private def handleJsonRequestLinks(wrappedRequest: HttpServletRequestWrapper, previousToken: Option[String], nextToken: Option[String], token: Option[String], applicableLinkPaths: List[LinkPath]): Try[HttpServletRequestWrapper] = {
+    val tryParsedJson = Try(Json.parse(wrappedRequest.getInputStream))
+    applicableLinkPaths.foldLeft(tryParsedJson)(transformJsonLink(_, _, token.get, previousToken, nextToken, isRequest = true)) match {
```
adrianjgeorge: This has become a kind of junky method call at this point if you are going to reuse it. We have multiple parameters that aren't relevant to the other side. It seem like this would benefit from taking a function instead of some of these parameters. Can you still achieve the reuse you are trying to get here while factoring out a function that gives the specifics of request and another with the specifics of the response?

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
```diff
@@ -135,9 +162,47 @@ class UriStripperFilter @Inject()(configurationService: ConfigurationService)
...
+        transformer.setParameter("removedToken", token.getOrElse(""))
+        transformer.setParameter("prefixToken", previousToken.getOrElse(""))
+        transformer.setParameter("postfixToken", nextToken.getOrElse(""))
+        Try(transformer.transform(new StreamSource(in), new StreamResult(out))) match {
+          case Success(_) => new ServletInputStreamWrapper(new ByteArrayInputStream(out.toByteArray))
+          case Failure(e: SAXParseException) =>
+            if (linkPath.getLinkMismatchAction == FAIL) throw e
+            else new ServletInputStreamWrapper(new ByteArrayInputStream(out.toByteArray))
+          case Failure(e: PathRewriteException) =>
+            logger.warn("Failed while trying to rewrite the location header", e)
```
adrianjgeorge: not sure this log message makes sense

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
wdschei: Are we sure we aren't going to want them for the contextual removal on the request body like is used for the contextual insertion on the response body?

If we don't, then there is a lot of this that can be cleaned up and maybe even simplified.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
```diff
@@ -135,9 +162,47 @@ class UriStripperFilter @Inject()(configurationService: ConfigurationService)
...
+  }
+
+  private def handleXmlRequestLinks(wrappedRequest: HttpServletRequestWrapper, previousToken: Option[String], nextToken: Option[String], token: Option[String], applicableLinkPaths: List[LinkPath]): Try[HttpServletRequestWrapper] = {
+    try {
+      val result = applicableLinkPaths.foldLeft(wrappedRequest.getInputStream) { (in: InputStream, linkPath: LinkPath) =>
+        val out = new ByteArrayOutputStream()
+        val transformer = templateMapRequest(linkPath).newTransformer
+        transformer.asInstanceOf[Controller].addLogErrorListener
+        transformer.setParameter("removedToken", token.getOrElse(""))
+        transformer.setParameter("prefixToken", previousToken.getOrElse(""))
```
adrianjgeorge: Man passing these around everywhere really feels like planning for a change we may never even do. If we redo this filter i think we are more likely to go with explicitly telling us where you want things versus this grey maybe it'll end up somewhere smart that it does right now.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
```diff
@@ -135,9 +162,47 @@ class UriStripperFilter @Inject()(configurationService: ConfigurationService)
...
+            else new ServletInputStreamWrapper(new ByteArrayInputStream(out.toByteArray))
+          case Failure(e: PathRewriteException) =>
+            logger.warn("Failed while trying to rewrite the location header", e)
+            throw e
+          case Failure(e) => throw e
+        }
+      }
+
+      Success(new HttpServletRequestWrapper(wrappedRequest, new ServletInputStreamWrapper(result)))
+    } catch {
```
adrianjgeorge: Every now then you need a normal try, and i always have to go find the syntax.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
```diff
@@ -161,7 +226,10 @@ class UriStripperFilter @Inject()(configurationService: ConfigurationService)
           case Failure(e: SAXParseException) =>
             if (linkPath.getLinkMismatchAction == FAIL) throw e
             else new ByteArrayInputStream(out.toByteArray)
-          case Failure(e: PathRewriteException) => throw e
+          case Failure(e: PathRewriteException) =>
+            logger.warn("Failed while trying to rewrite the location header", e)
```
adrianjgeorge: I'm guessing this is the same copy pasta log message that didn't get updated.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
```diff
@@ -172,6 +240,25 @@ class UriStripperFilter @Inject()(configurationService: ConfigurationService)
...
+      case Success(parsedUri) =>
+        val pathTokens = splitPathIntoTokens(parsedUri.getRawPath)
+        Try(pathTokens.remove(tokenIndex)) match {
+          case Success(_) => //don't care
+          case Failure(e: IndexOutOfBoundsException) =>
+            logger.warn("Determined index outside of uri path length", e)
+            throw PathRewriteException("Determined index is out side the number of parsed path segments", e)
+          case Failure(e) => throw e
+        }
+        replacePath(parsedUri, joinTokensIntoPath(pathTokens))
```
adrianjgeorge: Mutable state confused me for a second. I thought about how you never caught the removal, but since this passes tests I'm assuming the the `splitPathIntoTokens` method returns a mutable collection of some sort.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
```diff
@@ -225,9 +313,9 @@ class UriStripperFilter @Inject()(configurationService: ConfigurationService)
       configuredMethods.contains(HttpMethod.fromValue(requestMethod))
 
   private def getPathsForContentType(contentType: String, resource: HttpMessage): List[LinkPath] = {
-    Option(contentType) match {
-      case Some(ct) if ct.toLowerCase.contains("json") => resource.getJson.toList
-      case Some(ct) if ct.toLowerCase.contains("xml") => resource.getXml.toList map(_.getXpath)
+    (Option(contentType), Option(resource)) match {
```
adrianjgeorge: Were you having some issue with resource being null? If so that seems like a testing problem, because I'm fairly certain it's not possible in usage. Of course the diff doesn't give me enough context to say for certain.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
```diff
@@ -287,59 +383,73 @@ class UriStripperFilter @Inject()(configurationService: ConfigurationService)
...
       setupTransformer.asInstanceOf[Controller].addLogErrorListener
       val updateXPathXSLTDomResult = new DOMResult()
-      setupTransformer.transform (new StreamSource(<ignore-input />), updateXPathXSLTDomResult)
+      setupTransformer.transform(new StreamSource(<ignore-input/>), updateXPathXSLTDomResult)
 
-      val config = saxonTransformFactory.asInstanceOf[TransformerFactoryImpl].getConfiguration()
-      config.registerExtensionFunction(new UrlPathTransformExtension("process-url","rax", "http://docs.rackspace.com/api",
-        xsltFunction(Option(xmlElement.getXpath.getTokenIndex).map(_.intValue()), xmlElement.getXpath.getLinkMismatchAction)))
+      val config = saxonTransformFactory.asInstanceOf[TransformerFactoryImpl].getConfiguration
+      config.registerExtensionFunction(new UrlPathTransformExtension("process-url", "rax", "http://docs.rackspace.com/api",
```
adrianjgeorge: There have been a couple of these, I'm surprised that they passed the scala style check.

adrianjgeorge: Originally that is.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
```diff
@@ -287,59 +383,73 @@ class UriStripperFilter @Inject()(configurationService: ConfigurationService)
...
     }
 
     config = uriStripperConfig
 
-    templateMap = (for (resource <- config.getLinkResource;
-                        xml <- Option(resource.getResponse).map(_.getXml.toList).getOrElse(List.empty))
-                        yield xml.getXpath -> setupTransformer(xml)).toMap
+    templateMapRequest = (for (resource <- config.getLinkResource;
+                               xml <- Option(resource.getRequest).map(_.getXml.toList).getOrElse(List.empty))
+      yield xml.getXpath -> setupTransformer(xml, isRequest = true)).toMap
```
adrianjgeorge: The change in formatting here made this harder to read

repose-aggregator/components/filters/uri-stripper-filter/src/test/scala/org/openrepose/filters/uristripper/UriStripperFilterRequestTest.scala
```diff
@@ -0,0 +1,1028 @@
...
+        request.setRequestURI("/v1/12345/foo")
+        request.setContent(requestBody.getBytes(CHARSET_UTF8))
+        request.setContentType(MimeType.APPLICATION_JSON.toString)
+
+        filter.doFilter(request, response, filterChain)
+
+        getPostFilterRequestBody shouldEqual requestBody
+      }
+
+      it("should not alter the body if the method does not match the configured resource type") {
```
adrianjgeorge: Maybe a name that better draws attention to the fact that the config is no longer on request, it took me a minute to figure out why this wouldn't fail.

repose-aggregator/components/filters/uri-stripper-filter/src/test/scala/org/openrepose/filters/uristripper/UriStripperFilterRequestTest.scala
```diff
@@ -0,0 +1,1028 @@
...
+        request.setContent(requestBody.getBytes(CHARSET_UTF8))
+        request.setContentType(MimeType.APPLICATION_JSON.toString)
+
+        filter.doFilter(request, response, filterChain)
+
+        getPostFilterRequestBody shouldEqual Json.parse(requestBody).toString()
+      }
+
+      it("should not alter the body if the link cannot be located (remove)") {
+        val config =
```
adrianjgeorge: I suppose fail forward make sense. We successfully removed nothing.

repose-aggregator/components/filters/uri-stripper-filter/src/test/scala/org/openrepose/filters/uristripper/UriStripperFilterRequestTest.scala
```diff
@@ -0,0 +1,1028 @@
...
+          """.stripMargin
+
+        filter.configurationUpdated(Marshaller.uriStripperConfigFromString(config))
+        request.setRequestURI("/v1/12345/foo")
+        request.setContent(requestBody.getBytes(CHARSET_UTF8))
+        request.setContentType(MimeType.APPLICATION_JSON.toString)
+
+        filter.doFilter(request, response, filterChain)
+
+        getPostFilterRequestBody shouldEqual Json.parse(requestBody).toString()
```
adrianjgeorge: why parse when you aren't changing it in theory?

repose-aggregator/components/filters/uri-stripper-filter/src/test/scala/org/openrepose/filters/uristripper/UriStripperFilterRequestTest.scala
```diff
@@ -0,0 +1,1028 @@
...
+
+        filter.configurationUpdated(Marshaller.uriStripperConfigFromString(config))
+        request.setRequestURI("/v1/12345/foo")
+        request.setContent(requestBody.getBytes(CHARSET_UTF8))
+        request.setContentType(MimeType.APPLICATION_JSON.toString)
+
+        filter.doFilter(request, response, filterChain)
+
+        response.getStatus shouldEqual HttpServletResponse.SC_INTERNAL_SERVER_ERROR
+        response.getContentLength shouldEqual 0
```
adrianjgeorge: A check to make sure the filter chain was never invoked would probably be good as well

repose-aggregator/components/filters/uri-stripper-filter/src/test/scala/org/openrepose/filters/uristripper/UriStripperFilterRequestTest.scala
```diff
@@ -0,0 +1,1028 @@
...
+  def getPostFilterRequestBody: String = {
+    val requestCaptor = ArgumentCaptor.forClass(classOf[HttpServletRequestWrapper])
+    verify(filterChain).doFilter(requestCaptor.capture(), any(classOf[ServletResponse]))
+    scala.io.Source.fromInputStream(requestCaptor.getValue.getInputStream)(scala.io.Codec.UTF8).mkString
+  }
+}
+
+object UriStripperFilterRequestTest {
+  val CHARSET_UTF8 = Charset.forName("UTF-8")
+}
```
adrianjgeorge: Given the size i understand why you made a new test class

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
wdschei: Updated this to extract the logic requiring the extra parameters out to local methods that can be passed in.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
wdschei: I agree. `BAD REQUEST`(400) is probably a better response when the failure occurs on the request processing.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
wdschei: Some of these were caught by the static analysis tools and some were updates on a full format check.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
wdschei: This is one of the ones updated by the reformat.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
wdschei: Yes, it was found in testing. And true, it shouldn't happen in production. I just hate to leave a known potential NPE just waiting for something to let it rear it's ugly head.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
wdschei: Updated logging message.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
wdschei: The `splitPathIntoTokens` call returns a `mutable.Buffer[String]`. I am guessing it was made mutable for just this case.

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
wdschei: Updated logging message.

repose-aggregator/components/filters/uri-stripper-filter/src/test/scala/org/openrepose/filters/uristripper/UriStripperFilterRequestTest.scala
wdschei: Updated the test names to be a bit more obvious.

repose-aggregator/components/filters/uri-stripper-filter/src/test/scala/org/openrepose/filters/uristripper/UriStripperFilterRequestTest.scala
wdschei: Yep. Needed to make sure we did just remove something to remove it. ;-)

repose-aggregator/components/filters/uri-stripper-filter/src/test/scala/org/openrepose/filters/uristripper/UriStripperFilterRequestTest.scala
wdschei: Maybe the tests shouldn't say `not alter`, but rather `not alter meaning`. During the processing, all unnecessary white space is removed including the `\n`'s in the `requestBody` value. So the `Json.parse().toString` accomplishes the same thing in a generic way for the test comparisons.

repose-aggregator/components/filters/uri-stripper-filter/src/test/scala/org/openrepose/filters/uristripper/UriStripperFilterRequestTest.scala
wdschei: Yep. The existing tests were very through in their coverage, so duplicating them for the request side just made sense.

wdschei:
https://jenkins.openrepose.org/job/repose-gradle-pull_request/641/

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
adrianjgeorge: please push it back, and the one under it

repose-aggregator/components/filters/uri-stripper-filter/src/test/scala/org/openrepose/filters/uristripper/UriStripperFilterRequestTest.scala
adrianjgeorge: fair enough

repose-aggregator/components/filters/uri-stripper-filter/src/main/scala/org/openrepose/filters/uristripper/UriStripperFilter.scala
wdschei: Updated.

wdschei:
https://jenkins.openrepose.org/job/repose-gradle-pull_request/642/

PR #1585, opened 2016-09-07
REP-4172 Configuring containers at cluster level in Valve

dmnjohns:


dmnjohns:
Just the configuration changes so far.

wdschei:
Config looks good as long as we take a patch as an override and not an additive to the base definition.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/nodeservice/request/RequestHeaderServiceImpl.java
```diff
@@ -118,7 +122,7 @@ public void setVia(HttpServletRequestWrapper request) {
         @Override
         public void configurationUpdated(ContainerConfiguration configurationObject) {
             if (configurationObject.getDeploymentConfig() != null) {
-                viaReceivedBy = configurationObject.getDeploymentConfig().getVia();
+                viaReceivedBy = containerConfigurationService.getVia().orElse(null);
```
dmnjohns: As @adrianjgeorge pointed out, this is technically broken. The order of updates is indeterminate, so this may just be re-applying the current config. Most of the time, it will catch up, but it's still wrong.

dmnjohns:
Jenkins PR job:
https://jenkins-proto.openrepose.org/job/repose-gradle-pull_request/660/

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
```diff
@@ -0,0 +1,146 @@
...
+        case ex: Exception =>
+          logger.error("Configuration update error. Reason: {}", ex.getLocalizedMessage)
+          logger.trace("", ex)
+      }
+    }
+  }
+
+  // Note: The container configuration should be read on init of this class. As such, this check
+  //       should not ever fail.
+  private def initializationCheck(): Unit =
```
adrianjgeorge: Doing this like this, if the inits are called out of order, those components that init before will never be subscribed. You should probably double check in the spring documentation when `PostConstruct` is called, and if there is any guarantee on order.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/DeploymentConfigPatchUtil.scala
```diff
@@ -0,0 +1,155 @@
...
+    Option(patch.getSslConfiguration) foreach { sslConfiguration =>
+      if (Option(baseClone.getSslConfiguration).isEmpty) {
+        baseClone.setSslConfiguration(new SslConfiguration())
+      }
+      patchSslConfiguration(baseClone.getSslConfiguration, sslConfiguration)
+    }
+    /* TODO: These calls are commented out as patching for those nodes is not yet supported. A future update may
+           add more patching support, so the code is being left in.
+
+    Option(patch.getDeploymentDirectory) foreach { deploymentDirectory =>
```
adrianjgeorge: Man this is a super useful one, but i can see where it would be problematic.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/DeploymentConfigPatchUtil.scala
```diff
@@ -0,0 +1,155 @@
...
+        baseClone.setLoggingConfiguration(new LoggingConfiguration())
+      }
+      patchLoggingConfiguration(baseClone.getLoggingConfiguration, loggingConfiguration)
+    }
+    */
+
+    baseClone
+  }
+
+  /* TODO: These functions are commented out as patching for those nodes is not yet supported. A future update may
```
adrianjgeorge: Needed this todo twice? also it's not really a todo, so maybe `INFO`?

adrianjgeorge: Ahh, I see you have another set of functions. Still not really todo.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/DeploymentConfigPatchUtil.scala
```diff
@@ -0,0 +1,155 @@
...
+    Option(patch.getExcludedProtocols).foreach(base.setExcludedProtocols)
+    Option(patch.getIncludedCiphers).foreach(base.setIncludedCiphers)
+    Option(patch.getExcludedCiphers).foreach(base.setExcludedCiphers)
+    Option(patch.isTlsRenegotiationAllowed).foreach(base.setTlsRenegotiationAllowed)
+    Option(patch.isNeedClientAuth).foreach(base.setNeedClientAuth)
+
+    base
+  }
+
+  private def deepCopy[T <: java.io.Serializable](obj: T): T = {
```
adrianjgeorge: Ugh, was there not a better way to accomplish this?

repose-aggregator/components/services/container-configuration-service/impl/src/test/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImplTest.scala
```diff
@@ -0,0 +1,272 @@
...
+import org.openrepose.core.container.config._
+import org.openrepose.core.services.config.ConfigurationService
+import org.scalatest.junit.JUnitRunner
+import org.scalatest.mock.MockitoSugar
+import org.scalatest.{BeforeAndAfterEach, FunSpec, Matchers}
+
+@RunWith(classOf[JUnitRunner])
+class ContainerConfigurationServiceImplTest extends FunSpec with Matchers with MockitoSugar with BeforeAndAfterEach {
+
+  final val DefaultClusterId = "defaultClusterId"
```
adrianjgeorge: Still not a fan of this being the standard for constants in scala.

repose-aggregator/components/services/container-configuration-service/impl/src/test/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImplTest.scala
```diff
@@ -0,0 +1,272 @@
...
+      }
+    }
+
+    it("should return an empty Optional if not configured") {
+      containerConfigurationService.configurationUpdated(minimalContainerConfiguration())
+
+      containerConfigurationService.getVia.isPresent shouldBe false
+    }
+
+    it("should return the configured via string") {
```
adrianjgeorge: This should be two tests, one to make sure you can get the base value, and one to make sure you can get a patched value.

repose-aggregator/components/services/container-configuration-service/impl/src/test/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImplTest.scala
```diff
@@ -0,0 +1,272 @@
...
+      }
+    }
+
+    it("should return an empty Optional if not configured") {
+      containerConfigurationService.configurationUpdated(minimalContainerConfiguration())
+
+      containerConfigurationService.getContentBodyReadLimit.isPresent shouldBe false
+    }
+
+    it("should return the configured content body read limit") {
```
adrianjgeorge: Same here, and for all the other patchable values.

repose-aggregator/components/services/container-configuration-service/impl/src/test/scala/org/openrepose/nodeservice/containerconfiguration/DeploymentConfigPatchUtilTest.scala
```diff
@@ -0,0 +1,237 @@
...
+      // TODO: This assertion is commented out as patching for this field not yet supported. A future update may
+      //       add more patching support, so the code is being left in.
+      // patchedConfig.getJmxResetTime shouldEqual 3000
+      patchedConfig.getSoLingerTime shouldEqual 4000
+      patchedConfig.getContentBodyReadLimit shouldEqual 5000L
+      patchedConfig.getVia shouldEqual "patched-via"
+    }
+  }
+
+  /* TODO: These tests are commented out as patching for those nodes is not yet supported. A future update may
```
adrianjgeorge: again not really todos.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/container/container-configuration.xsd
```diff
@@ -149,6 +218,20 @@
         </xs:simpleContent>
     </xs:complexType>
 
+    <xs:complexType name="ArtifactDirectoryPatch">
```
adrianjgeorge: Probably want to remove this for now, or at least comment it out so people aren't trying to use it

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/examples/container.cfg.xml
```diff
@@ -32,4 +32,10 @@
 -->
         </ssl-configuration>
     </deployment-config>
+
+    <cluster-config cluster-id="internal-cluster">
```
adrianjgeorge: What does the system do if you provide an id that isn't in the system model?

repose-aggregator/core/repose-core/build.gradle
```diff
@@ -15,6 +15,8 @@ dependencies {
     compile project(":repose-aggregator:components:services:http-client-service:http-client-service-api")
     compile project(":repose-aggregator:components:services:datastore-service:datastore-service-api")
     compile project(":repose-aggregator:components:services:atom-feed-service:atom-feed-service-impl")
+    compile project(":repose-aggregator:components:services:container-configuration-service:api")
```
adrianjgeorge: you shouldn't need the api, unless you are directly using the api contract in some way?

repose-aggregator/core/repose-core/src/main/java/org/openrepose/core/services/context/container/ContainerConfigurationService.java
```diff
@@ -21,6 +21,11 @@
 
 import java.util.Optional;
 
+/**
+ * This service has been deprecated. Use
+ * {@link org.openrepose.nodeservice.containerconfiguration.ContainerConfigurationService} instead.
+ */
+@Deprecated
```
adrianjgeorge: Kill it please.

repose-aggregator/core/repose-core/src/test/scala/org/openrepose/core/spring/NextVersionTimebombTest.scala
```diff
@@ -52,6 +52,12 @@ class NextVersionTimebombTest extends FunSpec with Matchers with TestFilterBundl
        * 2. Remove the functional tests for the above attributes:
        *    a. IdentityV3CacheOffSetOldTest
        *    b. IdentityV3NoCacheOffSetOldTest
+       *
+       * 3. Remove these attributes from container-configuration.xsd:
+       *    a. http-port
+       *    b. https-port
+       *
+       * 4. Remove org.openrepose.core.services.context.container.ContainerConfigurationService and Impl
```
adrianjgeorge: will already be gone, because of above.

repose-aggregator/artifacts/valve/src/test/scala/org/openrepose/nodeservice/test/FakeContainerConfigurationService.scala
```diff
@@ -0,0 +1,62 @@
...
+import java.lang.Long
+import java.util.Optional
+import javax.inject.Named
+
+import org.openrepose.commons.config.manager.UpdateListener
+import org.openrepose.core.container.config.{ArtifactDirectory, DeploymentConfiguration, DeploymentDirectory}
+import org.openrepose.nodeservice.containerconfiguration.ContainerConfigurationService
+
+@Named
+class FakeContainerConfigurationService extends ContainerConfigurationService {
```
dmnjohns: So this is actually part of the test Spring context. Spring manages it this way. We did more or less the same thing with the `FakeConfigService`.

I should also note that `ValveRunner` does have this bean injected -- it calls `getBean(...)` to get it. Does Spring's testing framework have support for this case?

repose-aggregator/artifacts/valve/src/test/scala/org/openrepose/valve/ReposeJettyServerTest.scala
```diff
@@ -24,14 +24,14 @@ import org.junit.runner.RunWith
...
+import org.scalatest.{FunSpec, Matchers}
 
 @RunWith(classOf[JUnitRunner])
-class ReposeJettyServerTest extends FunSpec with Matchers with BeforeAndAfterAll {
+class ReposeJettyServerTest extends FunSpec with Matchers {
 
-  override def beforeAll() = {
-    CoreSpringProvider.getInstance().reInitializeCoreContext("/config/root", false)
-  }
+  CoreSpringProvider.getInstance().reInitializeCoreContext("/config/root", false)
```
dmnjohns: Maybe I've just forgotten, but is there a reason to use the `beforeAll` clause rather than just putting that code in the constructor?

dmnjohns: Follow-up. I removed it to drop a trait because I didn't think it was necessary.

I could have been wrong, though. In which case I'll change this back.

repose-aggregator/components/services/container-configuration-service/api/build.gradle
```diff
@@ -0,0 +1,3 @@
+dependencies {
```
dmnjohns: Right?

Good work.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
```diff
@@ -0,0 +1,146 @@
...
+@Named
+class ContainerConfigurationServiceImpl @Inject()(@Value(ReposeSpringProperties.NODE.CLUSTER_ID) clusterId: String,
+                                                  configurationService: ConfigurationService)
+  extends ContainerConfigurationService with UpdateListener[ContainerConfiguration] with LazyLogging {
+
+  import ContainerConfigurationServiceImpl._
+
+  private val initialized = new AtomicBoolean(false)
+  private val patchedDeploymentConfiguration = new AtomicReference[DeploymentConfiguration]()
+  private val updateListeners = new AtomicReference[Set[UpdateListener[DeploymentConfiguration]]](Set.empty)
```
dmnjohns: It got super messy. I'm not that happy with it.

So subscription/unsubscription should be thread safe, and I thought about synchronizing, but I think the same problem still exists as it does with `AtomicReference`. I could have used `volatile` for everything except the `updateListeners` which I atomically update with a function. In the end, I just let everything be atomic. It's consistent in this class, and with what we've done in other places.

I don't know. I'm up for suggestions.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
```diff
@@ -0,0 +1,146 @@
...
+  override def subscribeTo(listener: UpdateListener[DeploymentConfiguration]): Unit = {
+    subscribeTo(listener, sendNotificationNow = true)
+  }
+
+  override def subscribeTo(listener: UpdateListener[DeploymentConfiguration], sendNotificationNow: Boolean): Unit = {
+    initializationCheck()
+    logger.debug("Subscribing listener with hash code: {}", listener.hashCode.toString)
+
+    // TODO: Clean this up with Scala 2.12 support for Scala function -> Java 8 function
+    updateListeners.updateAndGet(new UnaryOperator[Set[UpdateListener[DeploymentConfiguration]]] {
```
dmnjohns: Well, it will mostly get cleaned up when we move to Scala 2.12 and do this TODO.

Still, dropping the atomic references would clean things up. I'm not _that_ concerned with threading issues, but they could happen. But what's the chance of more than one configuration update being called at the same time? Or of subscription and unsubscription between called for the same listener across threads?

Open for suggestions here, as well.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/DeploymentConfigPatchUtil.scala
dmnjohns: Indeed.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/DeploymentConfigPatchUtil.scala
dmnjohns: Alright. I'll... INFO it?

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
```diff
@@ -0,0 +1,146 @@
...
+import org.springframework.beans.factory.annotation.Value
+
+import scala.collection.JavaConversions._
+
+@Named
+class ContainerConfigurationServiceImpl @Inject()(@Value(ReposeSpringProperties.NODE.CLUSTER_ID) clusterId: String,
+                                                  configurationService: ConfigurationService)
+  extends ContainerConfigurationService with UpdateListener[ContainerConfiguration] with LazyLogging {
+
+  import ContainerConfigurationServiceImpl._
```
dmnjohns: It "statically" imports the constants from the companion object.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/DeploymentConfigPatchUtil.scala
dmnjohns: So my other idea was just performing a shallow copy at every level when we apply the patch. With the way these methods are setup, it's relatively easy, but requires JAXB to make its generated classes `Cloneable`. Well that, or we always construct a new object, then populate the field either with the patch value, or the base value if the patch value does not exist.

repose-aggregator/components/services/container-configuration-service/impl/src/test/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImplTest.scala
dmnjohns: Since we have a mixed code base, there's an argument to be made the being consistent across our codebase is more important than following the Scala style guide. If you want to use the Java constant form, I wouldn't fight it.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/container/container-configuration.xsd
dmnjohns: Oops. I removed the other, but apparently missed this one. I'll remove it for now.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/examples/container.cfg.xml
dmnjohns: Nothing. It just wouldn't apply the patch. If we wanted to add a check, it would have to be in code... probably an explicit check in the configuration updated method of this service. I expect it would just log the issue. Do we want to do that?

repose-aggregator/core/repose-core/build.gradle
dmnjohns: It is. The `PowerFilter` gets the bean explicitly.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/core/services/context/container/ContainerConfigurationService.java
dmnjohns: Done.

repose-aggregator/core/repose-core/src/test/scala/org/openrepose/core/spring/NextVersionTimebombTest.scala
dmnjohns: Done.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/examples/container.cfg.xml
dmnjohns: Oh, that would also require this service listening listening for changes to the system model.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/container/container-configuration.xsd
dmnjohns: Done.

repose-aggregator/components/services/container-configuration-service/impl/src/test/scala/org/openrepose/nodeservice/containerconfiguration/DeploymentConfigPatchUtilTest.scala
dmnjohns: Done.

repose-aggregator/components/services/container-configuration-service/impl/src/test/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImplTest.scala
dmnjohns: Done.

repose-aggregator/components/services/container-configuration-service/impl/src/test/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImplTest.scala
dmnjohns: Done.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/DeploymentConfigPatchUtil.scala
dmnjohns: Done.

repose-aggregator/components/services/container-configuration-service/api/src/main/java/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationService.java
```diff
@@ -0,0 +1,84 @@
...
+     * @return the maximum size of the request body in bytes, if configured.
+     */
+    Optional<Long> getContentBodyReadLimit();
+
+    /**
+     * @return the full deployment configuration patched for the node this service is running on.
+     */
+    DeploymentConfiguration getDeploymentConfiguration();
+
+    // TODO: Should the service allow users to register callbacks? By doing so, the service could fully replace
```
dmnjohns: Done.

repose-aggregator/artifacts/valve/src/test/scala/org/openrepose/nodeservice/test/FakeContainerConfigurationService.scala
adrianjgeorge: Not directly, but i've got some stuff from prior team that will allow a mock to be the spring context. Don't worry about it for now.

repose-aggregator/artifacts/valve/src/test/scala/org/openrepose/valve/ReposeJettyServerTest.scala
adrianjgeorge: Leaving it in makes it more standard to every other testing framework. I suspect that there are also other semantics to it.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
adrianjgeorge: Can't that be done, but have it with the other imports? I'm not a fan of scoped imports if they aren't buying us something. Like when dealing with implicits scoping can make sure you don't accidentally trigger something unexpected.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
adrianjgeorge: We talked about this one yesterday.

repose-aggregator/components/services/container-configuration-service/impl/src/test/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImplTest.scala
adrianjgeorge: Ehh, I mostly fine with using the style of the language, i'm just disappointed that this is the standard, it just looks like you are doing something with a type. The obvious exception is when it's targeted at usage in java, then i'd suggest using the java style, so implementors recognize what it is.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/examples/container.cfg.xml
adrianjgeorge: This brings up an interesting issue, what happens when the container config updates before the system model when a node is being added? I think it'd be ok.

repose-aggregator/artifacts/valve/src/test/scala/org/openrepose/nodeservice/test/FakeContainerConfigurationService.scala
dmnjohns: Kk. So to this comment chain, I'll say...

Done.

repose-aggregator/artifacts/valve/src/test/scala/org/openrepose/valve/ReposeJettyServerTest.scala
dmnjohns: Fair enough, I'll put it back. Easy-peasy.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
dmnjohns: I'm actually not sure. That's an interesting question. I wonder if that doesn't work since the companion object is in the same file, and uses the same file scope imports.

Let me try it and see what happens.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/examples/container.cfg.xml
adrianjgeorge: Yeah, there's nothing right now that's really paying attention to both, and even if there was config ordering might be wrong. I'm just not a fan of silent failure, but there's not a great solution to that problem right now. Config read ordering is something we should consider for 9.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
dmnjohns: So it seems to work, although the import becomes the fully qualified companion object.

Honestly, though, I don't love that... We've followed this pattern elsewhere, and it's one that the Scala developers are aware of (http://docs.scala-lang.org/tutorials/tour/singleton-objects.html). I'd prefer to leave it as-is, but we can do whatever.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
dmnjohns: Done.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
adrianjgeorge: Also i think you get companion object references without the import?

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
dmnjohns: Still looking for documented evidence that stuff works as I would like it to...

repose-aggregator/components/services/container-configuration-service/impl/src/test/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImplTest.scala
dmnjohns: Sure. In which case...

Done.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/examples/container.cfg.xml
dmnjohns: Indeed.

Oh, and since this is a node service, the system model is guaranteed to have been read at least once already. After all, that's how we determined the cluster ID and node ID properties that are accessible to all node services. Since the cluster ID is wired in, and we subscribe to the container on `PostConstruct`, we should be guaranteed that the ordering there is always correct.

dmnjohns: Oh, and unless I missed something, I'll call this one...

Done.

repose-aggregator/core/repose-core/build.gradle
dmnjohns: Calling this one...

Done.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
dmnjohns: You mean `ContainerConfigurationServiceImpl.NotInitializedMessage`? That's how you would reference those fields without the import.

adrianjgeorge: Man why do i feel like i've successfully referenced it as just `NotInitializedMessage'. Leave it I guess.

repose-aggregator/artifacts/valve/src/test/scala/org/openrepose/valve/ReposeJettyServerTest.scala
dmnjohns: Done.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
dmnjohns: Done.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
dmnjohns: Removed the AtomicReference, so...

Done.

dmnjohns:
Last Jenkins run:
https://jenkins-proto.openrepose.org/job/repose-gradle-pull_request/666/

Looks like the Gradle build passed, but the GitHub hook failed. Not sure why it had the wrong commit SHA...

retest this please

adrianjgeorge:
pretty sure that sha is the merged sha, which obviously doesn't exist yet on github.

PR #1586, opened 2016-09-09
REP-3766 Updated the gradle-jaxb-plugin to be the newly published version in the Gradle Plugins repository.

wdschei:


adrianjgeorge:
totes legit, we'll let it pass though

adrianjgeorge:
meh, it's already successfully downloaded and started using it. no need to wait really, since there were no code changes.

PR #1587, opened 2016-09-20
REP-3653 Bypass Akka Cache Mechanism for Keystone v2

dmnjohns:


repose-aggregator/components/filters/keystone-v2-filter/src/test/scala/org/openrepose/filters/keystonev2/MockedAkkaServiceClient.scala
```diff
@@ -106,7 +110,10 @@ trait MockedAkkaServiceClient {
                       uri: String,
                       headers: util.Map[String, String],
                       payload: String,
-                      contentMediaType: MediaType): ServiceClientResponse = {
+                      contentMediaType: MediaType,
+                      checkCache: Boolean): ServiceClientResponse = {
+      checkCacheValues += checkCache
+
```
dmnjohns: This has nothing to do with my changes, but I really don't like that we built our own mock here. I may try and replace all of this with Mockito mocks.

dmnjohns:
retest this please

repose-aggregator/components/filters/keystone-v2-filter/src/test/scala/org/openrepose/filters/keystonev2/MockedAkkaServiceClient.scala
```diff
@@ -1,171 +0,0 @@
...
-  import java.util.concurrent.atomic.AtomicBoolean
-  import javax.ws.rs.core.MediaType
-
-  import com.typesafe.scalalogging.slf4j.LazyLogging
-  import org.openrepose.commons.utils.http.ServiceClientResponse
-  import org.openrepose.core.services.serviceclient.akka.{AkkaServiceClient, AkkaServiceClientException}
-
-  import scala.collection.mutable
-
-  class MockAkkaServiceClient extends AkkaServiceClient with LazyLogging {
```
adrianjgeorge: Good riddance to bad code.

repose-aggregator/components/services/akka-http-client-service/akka-http-client-service-api/src/main/java/org/openrepose/core/services/serviceclient/akka/AkkaServiceClient.java
```diff
@@ -29,9 +29,11 @@
  */
 public interface AkkaServiceClient {
 
-    ServiceClientResponse get(String token, String uri, Map<String, String> headers) throws AkkaServiceClientException;
+    // TODO: Should backwards compatible get and post methods be added? If so, the commit to use the new methods can be reverted.
 
-    ServiceClientResponse post(String requestKey, String uri, Map<String, String> headers, String payload, MediaType contentMediaType) throws AkkaServiceClientException;
+    ServiceClientResponse get(String token, String uri, Map<String, String> headers, boolean checkCache) throws AkkaServiceClientException;
```
adrianjgeorge: Yeah, here's where you started making a bunch of unnecessary changes, overload the methods. Then most of our calls can just stay as they are, and when we need it specifically we can call for a non-cache answer. Also obviously implementation wise the one just calls into the other with true.

repose-aggregator/components/services/akka-http-client-service/akka-http-client-service-impl/src/test/scala/org/openrepose/core/services/serviceclient/akka/impl/AkkaServiceClientImplReflectionTest.scala
```diff
@@ -0,0 +1,126 @@
...
+import org.openrepose.core.service.httpclient.config.{HttpConnectionPoolConfig, PoolType}
+import org.openrepose.core.services.config.ConfigurationService
+import org.openrepose.core.services.httpclient.{HttpClientContainer, HttpClientService}
+import org.scalatest.mock.MockitoSugar
+import org.scalatest.{BeforeAndAfterAll, BeforeAndAfterEach, FunSpec}
+
+import scala.collection.JavaConversions._
+
+// TODO: Remove this once the cache TTL (and other necessary fields) can be set by other means
+class AkkaServiceClientImplReflectionTest extends FunSpec with BeforeAndAfterAll with BeforeAndAfterEach with MockitoSugar {
```
adrianjgeorge: No need to call out the reflection in the test title. Also spring has some utils for working with reflection which may or may not be a help (I don't remember their specific use case right now).

repose-aggregator/components/services/akka-http-client-service/akka-http-client-service-api/src/main/java/org/openrepose/core/services/serviceclient/akka/AkkaServiceClient.java
dmnjohns: Easy enough. That basically answers the TODO I left in this file.

The hard part will be rolling back the changes elsewhere, but hopefully my other commits are good.

adrianjgeorge: yeah, i started seeing changes everywhere, and i was like what? Then realized that you probably forced the parameter in the contract.

repose-aggregator/components/services/akka-http-client-service/akka-http-client-service-impl/src/test/scala/org/openrepose/core/services/serviceclient/akka/impl/AkkaServiceClientImplReflectionTest.scala
dmnjohns: Well, `AkkaServiceClientImplTest` already exists, but I wanted to break this into its own test for a couple of reasons:

1. I don't love using reflection. This makes it clear that reflection is being used, and that once it is not longer needed, should not be used. Also, I'm not certain how the test runner runs tests, but I hoped that putting this in its own test class would prevent other tests from seeing the changes made to the class through reflection. I actually bet other tests do see the reflective change, but it shouldn't overly matter as none of our unit tests require that the timeout be 500ms.
2. The other test is kind of messy.

adrianjgeorge: That's fine. Whether other tests see it or not is mostly dependent on whether or not it's static.

dmnjohns:
Updated the Akka service API to be backwards compatible and reverted the changes that were showing up everywhere.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneRequestHandler.scala
```diff
@@ -186,7 +197,8 @@ class KeystoneRequestHandler(identityServiceUri: String, akkaServiceClient: Akka
       s"$identityServiceUri${GROUPS_ENDPOINT(forToken)}",
       (Map(CommonHttpHeader.AUTH_TOKEN.toString -> authenticatingToken,
         CommonHttpHeader.ACCEPT.toString -> MediaType.APPLICATION_JSON)
-        ++ traceId.map(CommonHttpHeader.TRACE_GUID.toString -> _)).asJava))
+        ++ traceId.map(CommonHttpHeader.TRACE_GUID.toString -> _)).asJava,
+      checkCache))
```
dmnjohns: I added the backwards compatible methods on the Akka service client, so I think this comment has been addressed.

repose-aggregator/components/services/akka-http-client-service/akka-http-client-service-impl/src/test/scala/org/openrepose/core/services/serviceclient/akka/impl/AkkaServiceClientImplReflectionTest.scala
dmnjohns: Unfortunately, it is static. Because of course it is.

So ultimately, what would you like done with this test?

adrianjgeorge: It's fine

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneRequestHandler.scala
adrianjgeorge: Not really, it's still all over this file. Does it really have to be passed through a million methods?

repose-aggregator/components/services/akka-http-client-service/akka-http-client-service-impl/src/test/scala/org/openrepose/core/services/serviceclient/akka/impl/AkkaServiceClientImplTest.scala
```diff
@@ -108,8 +108,8 @@ class AkkaServiceClientImplTest extends FunSpec with BeforeAndAfterEach with Mat
     methods.foreach { method =>
       describe(s"using the HTTP request method $method") {
         def akkaServiceClientImplDo(akkaServiceClientImpl: AkkaServiceClientImpl, headers: Map[String, String]): ServiceClientResponse = method match {
-          case "GET" => akkaServiceClientImpl.get(hashKey, uri, headers)
-          case "POST" => akkaServiceClientImpl.post(hashKey, uri, headers, BODY_STRING, MediaType.APPLICATION_XML_TYPE)
+          case "GET" => akkaServiceClientImpl.get(hashKey, uri, headers, true)
```
adrianjgeorge: Shouldn't these go back to what they were?

dmnjohns: Yes, they should. These were in the commit where I changed the service itself. Can't revert that one, so I'll just manually pull these out.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneRequestHandler.scala
dmnjohns: Oh, I see.

Right, so the `KeystoneV2Filter` is responsible for retrying (because it can access the datastore, while the handler cannot, and has the logic for self-validating tokens). The `KeystoneRequestHandler` just makes the requests and returns the responses. The distance between the component which triggers this and the actual passing of the argument is only one method. So it is one more layer than would be preferable, but eh. It would require a fair amount of refactoring to drop this parameter and handle the retry in the `KeystoneRequestHandler`.

With the ability to use default values for method calls, I thought this actually came out fairly clean.

dmnjohns:
Last test results (the PR building is broken again):
https://jenkins-proto.openrepose.org/job/repose-gradle-pull_request/678/

dmnjohns:
retest this please

dmnjohns:
retest this please

PR #1588, opened 2016-09-20
REP-4395 Updated the API-Checker version.

wdschei:
It is currently set to the published snapshot to test before final release.


wdschei:
https://jenkins.openrepose.org/job/repose-gradle-pull_request/669/

repose-aggregator/tests/functional-tests/build.gradle
```diff
@@ -1,3 +1,6 @@
+evaluationDependsOn(':repose-aggregator:tests:test-containers:test-container-glassfish')
+evaluationDependsOn(':repose-aggregator:tests:test-containers:test-container-tomcat')
```
dmnjohns: This may not be the cleanest way to do this. We could probably throw a bit of time at seeing if there's a better way of enforcing this.

For multi-project Gradle builds, "The default evaluation order of projects is alphanumeric (for the same nesting level)." [1] As a result, this Gradle build file runs before these two other build files. The copyArtifacts task
requires these other two build files to run first, as they declare configurations that are expected to be available..

[1] https://docs.gradle.org/current/userguide/multi_project_builds.html#sec:dependencies_which_dependencies

dmnjohns:
The transitive dependency in api-checker for wadl-tools was not making it onto the classpath. 1.0.33 is required by checker-core 2.0.2-SNAPSHOT, but only 1.0.32 was available. It seems that the dependency-recommender plugin was forcing 1.0.32 to be used, overriding the transitive dependency. A new version of that plugin changes the default behavior to respect transitive dependencies. See the README for the following versions:
https://github.com/nebula-plugins/nebula-dependency-recommender-plugin/tree/v3.3.0#producing-a-maven-bom-for-use-as-a-dependency-recommendation-source
https://github.com/nebula-plugins/nebula-dependency-recommender-plugin/tree/v3.6.3

So I upgraded the version we are using. For some reason, that caused the functional tests build to fail. Hence the other change to add `evaluationDependsOn` calls. I've added another comment on that line.

dmnjohns:
Last test results (the PR build is broken again):
https://jenkins-proto.openrepose.org/job/repose-gradle-pull_request/676/

Mario-Lopez:
I rebased, and the build appears to be happy again.  I have another pull request that will go into this branch and a pull request on API Checker pending.

PR #1589, opened 2016-09-28
REP-4350 Removing redundant cleanupSpec method from ReposeValveTests

dmnjohns:
I will likely clean up a few other things while I'm messing with this code, but I think this first commit covers the acceptance criteria for the story.

Also, I want to make sure I did not break anything.


repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/proxy/MisbehavingOriginTest.groovy
```diff
@@ -52,13 +46,7 @@ class MisbehavingOriginTest extends ReposeValveTest {
...
-    def cleanup() {
-        if (repose) {
-            repose.stop()
-        }
-        if (deproxy) {
-            deproxy.shutdown()
-        }
+    def cleanupSpec() {
         running = false
     }
```
Mario-Lopez: I don't remember... does Spock call the parent class `cleanupSpec()` as well, or do we need to specifically call it here?

dmnjohns: Spock will call the parent class `cleanupSpec()`.

From http://spockframework.org/spock/docs/1.1-rc-2/spock_primer.html :
> If fixture methods are overridden in a specification subclass then setup() of the superclass will run before setup() of the subclass. cleanup() works in reverse order, that is cleanup() of the subclass will execute before cleanup() of the superclass. setupSpec() and cleanupSpec() behave in the same way. There is no need to explicitly call super.setup() or super.cleanup() as Spock will automatically find and execute fixture methods at all levels in an inheritance heirarchy.

Mario-Lopez:
👍

PR #1590, opened 2016-09-30
Chunked Encoding Test

dmnjohns:
Adding test (that is expected to fail) to prove that Repose sends the "Transfer-Encoding: chunked" header if the user sends it, regardless of the value of the "chunked-encoding" attribute of the default connection pool.

Story: https://repose.atlassian.net/browse/REP-4490

### DIFF
--- a/repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/proxy/ContentLengthTest.groovy
+++ b/repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/proxy/ContentLengthTest.groovy
@@ -28,7 +28,7 @@ import spock.lang.Unroll
 
 class ContentLengthTest extends ReposeValveTest {
 
-    def timeBombDate = new GregorianCalendar(2017, Calendar.JANUARY, 5)
+    public static def TIME_BOMB_DATE = new GregorianCalendar(2017, Calendar.JANUARY, 5)
 
     def setupSpec() {
         deproxy = new Deproxy()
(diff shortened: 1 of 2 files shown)


adrianjgeorge:
Time bomb it so we can merge it in.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/proxy/ContentLengthTest.groovy
```diff
@@ -64,7 +65,33 @@ class ContentLengthTest extends ReposeValveTest {
         "PUT"   | null    | 0            | 1              | 0
         "TRACE" | "blah"  | 1            | 0              | 0
         "TRACE" | null    | 0            | 0              | 0
-
     }
 
+    @IgnoreIf({ new Date() < timeBombDate.getTime() })
```
dmnjohns: Yeah, I think we use the jUnit `Assume` elsewhere, but since this is in a Spock test, I kind of wanted to use a Spock construct. So here it is.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/proxy/ContentLengthTest.groovy
```diff
@@ -23,12 +23,14 @@ import framework.ReposeValveTest
 import org.rackspace.deproxy.Deproxy
 import org.rackspace.deproxy.Handling
 import org.rackspace.deproxy.MessageChain
+import spock.lang.IgnoreIf
 import spock.lang.Unroll
 
 class ContentLengthTest extends ReposeValveTest {
 
-    def setupSpec() {
+    def timeBombDate = new GregorianCalendar(2017, Calendar.JANUARY, 5)
```
adrianjgeorge: Your test failed, and this is why? You can't reference instance variables in annotations, only statics

adrianjgeorge:
I updated it

PR #1591, opened 2016-10-03
REP-4445 Forwarding authenticatedBy values as a header

dmnjohns:


repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -526,6 +526,13 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
                 request.addHeader(OpenStackServiceHeader.EXTENDED_AUTHORIZATION.toString, X_AUTH_PROXY)
             }
         }
+
+        // Construct and add authenticatedBy, if available
+        token.authenticatedBy foreach {
+          _ foreach {
+            request.addHeader(OpenStackServiceHeader.AUTHENTICATED_BY.toString, _)
```
dmnjohns: I wanted to point this out. It's not consistent with the rest of this filter which generally adds multiple headers on the same line. Perhaps, however, it is the way we should be doing it... since Repose splits headers coming in, and the merge-header filter exists if users want merged headers.

That is, this is inconsistent with the filter, but more consistent with Repose as a system, in my opinion.

Let me know what you want to do.

Mario-Lopez: For posterity, Jorge commented on the story that this is how we should do this.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneRequestHandler.scala
```diff
@@ -110,6 +110,10 @@ class KeystoneRequestHandler(identityServiceUri: String, akkaServiceClient: Akka
         val roles = (json \ "access" \ "user" \ "roles").as[JsArray].value
           .map(jsRole => Role((jsRole \ "name").as[String], (jsRole \ "tenantId").asOpt[String]))
           .toVector
+        val authenticatedBy = (json \ "access" \ "token" \ "RAX-AUTH:authenticatedBy").asOpt[JsArray]
+          .map(_.value)
+          .map(_.map(_.as[String]))
+          .map(_.toVector)
```
Mario-Lopez: Why keep it inside an Option instead of just letting it be an empty list?  (Aside from the fact that I doubt this would ever be an empty list, but it's easy enough to code for the possibility.)  Also, this can totally be a List now instead of a Vector.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/framework/mocks/MockIdentityV2Service.groovy
```diff
@@ -1054,7 +1054,11 @@ class MockIdentityV2Service {
             "tenant":{
                 "id": "\${tenantid}",
                 "name": "\${tenantname}"
-            }
+            },
+            "RAX-AUTH:authenticatedBy": [
+                "PASSWORD",
+                "MFA"
```
Mario-Lopez: You may want to change "MFA" to "PASSCODE" so that we don't think "MFA" is a valid value in the future.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/keystonev2/selfvalidation/KeystoneV2SelfValidationTest.groovy
```diff
@@ -74,6 +77,22 @@ class KeystoneV2SelfValidationTest extends ReposeValveTest {
...
+                if ("Password".equalsIgnoreCase(header.value)) {
+                    foundPassword = true
+                }
+                if ("Mfa".equalsIgnoreCase(header.value)) {
+                    foundMultiFactor = true
+                }
+            }
+        }
+        assertTrue(foundPassword)
+        assertTrue(foundMultiFactor)
```
Mario-Lopez: Actually, this whole section should be one or two lines.  You could do:
```
assert mc.handlings[0].request.headers.findAll("X-Authenticated-By").sort() == ["MFA", "PASSWORD"]
```

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneRequestHandler.scala
dmnjohns: Interesting, I actually started by treating an empty option as equivalent to an empty list.

In the end, I left it as an option as a way of being explicit with the typing. That is, this field is a Rackspace extension, so it's fairly likely that it simply will not be present. Technically it could even be an empty array, although that is unlikely (and maybe prohibited by whatever specification details this field). This way, all of that information is carried along with this value. Existence is determined by the state of the option, while the content of the vector mirror the contents of the JSON array.

So I guess the short and sweet answer is that this leaves open the possibility to modify behavior based on any possible state.

Oh, and yes, it can totally be a list. Yay Scala upgrades! I used a Vector for the sake of consistency. I would kind of like to drop the conversion to a vector from all of these fields, but I wasn't sure that this was the right time to do it. However, I will totally do it if no one has any issues with that. I'll note that the case class we stick all of these values into only requires that they be `Seq`s (and, implicitly, that they be Serializable).

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/framework/mocks/MockIdentityV2Service.groovy
dmnjohns: Fair (and easy) enough.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneRequestHandler.scala
Mario-Lopez: Either way works for me on both counts.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/keystonev2/selfvalidation/KeystoneV2SelfValidationTest.groovy
dmnjohns: Oh, @wdschei did this. In that case, JEEZ Bill! 😛

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/keystonev2/selfvalidation/KeystoneV2SelfValidationTest.groovy
```diff
@@ -79,20 +75,7 @@ class KeystoneV2SelfValidationTest extends ReposeValveTest {
...
-                    foundPassword = true
-                }
-                if ("Mfa".equalsIgnoreCase(header.value)) {
-                    foundMultiFactor = true
-                }
-            }
-        }
-        assertTrue(foundPassword)
-        assertTrue(foundMultiFactor)
+        mc.getHandlings().get(0).getRequest().getHeaders().findAll("X-Authenticated-By").sort() == ["MFA", "PASSWORD"]
```
Mario-Lopez: s/MFA/PASSCODE

Also, it doesn't rustle my jimmies or anything, but in Groovy, `mc.getHandlings().get(0)` is the same as `mc.handlings[0]` and `.getRequest()` the same as `.request`.

dmnjohns: I blame you for the MFA thing. I copypasta'd your line. You knew MFA was going to be replaced!

I'm aware. My problem with accessing fields directly is that access protection (e.g., private) mechanisms are ignored. If I remember correctly, Groovy will not generate an accessor method for restricted fields. So by always using accessor methods, whether they be natural or synthetically produced by Groovy, I am assured that I have access to the field in question. As far as indexing a collection goes, I don't really have a good reason to prefer `get(...)` to `[...]`. The latter is just syntactic sugar, right? I suppose the former could be copypasta'd into Java, and that's something...  Have I ever mentioned that I'm not really a fan of Groovy? 😛

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/keystonev2/selfvalidation/KeystoneV2SelfValidationTest.groovy
wdschei: :innocent:

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/keystonev2/selfvalidation/KeystoneV2SelfValidationTest.groovy
Mario-Lopez: LOL.

Yeah, it's all syntactic sugar.  Although to be clear, `handler.request` will call `handler.getRequest()`, not access the field directly.  To access the field, you would call `handler.@request`.  Also, I totally didn't know about "private" being meaningless in Groovy.  No wonder it lost the language war. 😔

Mario-Lopez:
All good!  Waiting on build to finish.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/keystonev2/selfvalidation/KeystoneV2SelfValidationTest.groovy
dmnjohns: That might make it even worse. So what is happening if I write `handler.request` and `request` is a private field? Groovy shouldn't be generating a `getRequest()` method.

```
groovy:000> class Foo {
groovy:001> private String foo = "foo"
groovy:002> String bar = "bar"
groovy:003> }
===> true
groovy:000> new Foo().foo
===> foo
groovy:000> new Foo().bar
===> bar
groovy:000> new Foo().getFoo()
ERROR groovy.lang.MissingMethodException:
No signature of method: Foo.getFoo() is applicable for argument types: () values: []
Possible solutions: getBar(), getAt(java.lang.String), setBar(java.lang.String), getClass(), notify(), every()
groovy:000> new Foo().getBar()
===> bar
groovy:000> new Foo().@foo
===> foo
groovy:000> new Foo().@bar
===> bar
```

and

```
groovy:000> class Foo {
groovy:001> private String privStr = "private"
groovy:002> public String pubStr = "public"
groovy:003> private String getPrivStr() {
groovy:004> println("getPrivStr")
groovy:005> return privStr
groovy:006> }
groovy:007> public String getPubStr() {
groovy:008> println("getPubStr")
groovy:009> return pubStr
groovy:010> }
groovy:011> }
===> true
groovy:000> new Foo().privStr
getPrivStr
===> private
groovy:000> new Foo().getPrivStr()
getPrivStr
===> private
groovy:000> new Foo().@privStr
===> private
groovy:000> new Foo().pubStr
getPubStr
===> public
groovy:000> new Foo().getPubStr()
getPubStr
===> public
groovy:000> new Foo().@pubStr
===> public
```

I'm an old man and I like my explicit method calls.

PR #1599, opened 2016-10-15
REP-4490 Not Honoring Chunked Transfer Encoding

dmnjohns:


repose-aggregator/components/services/http-client-service/http-client-service-api/src/main/resources/META-INF/schema/config/http-connection-pool.xsd
```diff
@@ -239,14 +239,22 @@
...
                         Determines whether or not this client should send chunked data when forwarding requests.
                     </html:p>
                     <html:p>
-                        WARNING: Setting this attribute will cause repose to attempt to retrieve the
+                        If set to true, chunked data will always be sent.
+                        If set to false, chunked data will never be sent.
+                        If set to auto, chunked data will only be sent if the original request to Repose was chunked.
+                        Note that the auto setting only has meaning for the default connection pool, and only when
+                        requests are forwarded to the origin service. A value of auto for non-default pools will
+                        behave identically to a value of true.
```
Mario-Lopez: 1. Is chunked encoding currently supported on all outgoing connections or just to the origin service?  If it's just to the origin service (which is what I would expect now that you mention it), I wouldn't change that behavior in this story.  I would just document this fact well and add a comment in https://repose.atlassian.net/browse/REP-3589 to address it.  Otherwise, what is the current behavior for non-origin service requests?
2. If you're talking about changing the behavior of non-origin service requests, then yes, I'd agree with you on it being a breaking change in a bad way, and I wouldn't do it here.
3. To your third point, this might be a breaking config change if anyone was hipster enough to configure a "1" or "0" in this field which is technically allowed.  ಠ_ಠ

dmnjohns: Ah. So indeed, it appears that only requests being made to the origin service check that flag. I hadn't really looked at the code, just the config, so I assumed chunked encoding applied to all requests. Oops. Good catch.
So my point 1 is moot. Although, that totally _could_ mean something to pools other than the default pool. Something to watch out for in the future, I guess.

repose-aggregator/components/services/http-client-service/http-client-service-api/src/main/resources/META-INF/schema/config/http-connection-pool.xsd
```diff
@@ -239,14 +239,22 @@
...
                         Determines whether or not this client should send chunked data when forwarding requests.
                     </html:p>
                     <html:p>
-                        WARNING: Setting this attribute will cause repose to attempt to retrieve the
+                        If set to true or 1, chunked data will always be sent.
+                        If set to false or 0, chunked data will never be sent.
+                        If set to auto, chunked data will only be sent if the original request to Repose was chunked.
+                        Note that the auto setting only has meaning for the default connection pool, and only when
+                        requests are forwarded to the origin service. A value of auto for non-default pools will
+                        behave identically to a value of true.
```
Mario-Lopez: Do we always use chunked encoding for third party APIs?

dmnjohns: It seems that we don't use chunked encoding for third party APIs. So I'll update this.

repose-aggregator/commons/commons-utilities/src/main/java/org/openrepose/commons/utils/io/stream/ServletInputStreamWrapper.java
```diff
@@ -40,6 +40,21 @@ public int read() throws IOException {
     }
 
     @Override
+    public int read(byte[] b) throws IOException {
```
adrianjgeorge: Were these all crap implementations before this?

repose-aggregator/components/services/http-client-service/http-client-service-api/src/main/resources/META-INF/schema/config/http-connection-pool.xsd
```diff
@@ -239,14 +239,22 @@
...
-                        Determines whether or not this client should send chunked data when forwarding requests.
+                        Determines whether or not this client should send chunked data when forwarding requests to
+                        the origin service.
                     </html:p>
                     <html:p>
-                        WARNING: Setting this attribute will cause repose to attempt to retrieve the
+                        If set to true or 1, chunked data will always be sent.
+                        If set to false or 0, chunked data will never be sent.
+                        If set to auto, chunked data will only be sent if the original request to Repose was chunked.
+                        Note that the auto setting only has meaning for the default connection pool, as that is the only
```
adrianjgeorge: I'd probably also mention that 1 and 0 are deprecated and slated for removal in 9, and then be sure to add that to our 9 timebomb list.

repose-aggregator/components/services/http-client-service/http-client-service-impl/src/main/java/org/openrepose/core/services/httpclient/impl/HttpConnectionPoolProvider.java
```diff
@@ -75,7 +75,7 @@ public static HttpClient genClient(String configRoot, PoolType poolConf) {
         params.setParameter(CoreConnectionPNames.MAX_HEADER_COUNT, poolConf.getHttpConnectionMaxHeaderCount());
         params.setParameter(CoreConnectionPNames.MAX_LINE_LENGTH, poolConf.getHttpConnectionMaxLineLength());
         params.setParameter(CoreConnectionPNames.SOCKET_BUFFER_SIZE, poolConf.getHttpSocketBufferSize());
-        params.setBooleanParameter(CHUNKED_ENCODING_PARAM, poolConf.isChunkedEncoding());
+        params.setParameter(CHUNKED_ENCODING_PARAM, poolConf.getChunkedEncoding());
```
adrianjgeorge: Seems like right here would be a good spot to filter out the 1 and 0, so your other code doesn't have to translate, but can just deal with values it understands.

repose-aggregator/commons/commons-utilities/src/main/java/org/openrepose/commons/utils/io/stream/ServletInputStreamWrapper.java
dmnjohns: Apparently.

repose-aggregator/components/services/http-client-service/http-client-service-impl/src/main/java/org/openrepose/core/services/httpclient/impl/HttpConnectionPoolProvider.java
adrianjgeorge: fair enough

repose-aggregator/components/services/http-client-service/http-client-service-api/src/main/resources/META-INF/schema/config/http-connection-pool.xsd
dmnjohns: Sure.

Done.

PR #1601, opened 2016-10-17
REP-4127 Socket timeout logs misleading message

dmnjohns:


repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilterRouterImpl.java
```diff
@@ -193,7 +193,8 @@ public void route(HttpServletRequestWrapper servletRequest, HttpServletResponse
                             LOG.error("Error reading request content", e);
                             servletResponse.sendError(HttpServletResponse.SC_REQUEST_ENTITY_TOO_LARGE, "Error reading request content");
                         } else {
-                            LOG.error("Connection Refused to {}", location.getUri(), e);
+                            LOG.error("Error communicating with {}", location.getUri());
+                            LOG.trace("", e);
```
dmnjohns: I think this is what we proposed. It's something that we do in a number of other places as well.

As for why, I couldn't really say. I mean, the idea must be to keep the log clean and separate concerns. The rationale may be that Repose will automatically "recover" so there is no need to inform the user of the nitty-gritty. On the other hand, though, this does seem like a rare error where having more information may be useful.

Which is my long-winded way of saying iunno.

Mario-Lopez: Would you be okay with adding the exception message to the error log?

https://github.com/rackerlabs/repose/blob/master/repose-aggregator/components/services/akka-http-client-service/akka-http-client-service-impl/src/main/java/org/openrepose/core/services/serviceclient/akka/impl/AkkaServiceClientImpl.java#L112-L113

dmnjohns: It literally makes no difference to me.

Sooo... @adrianjgeorge?

Mario-Lopez:
LGTM.  👍 

Waiting for build to finish.

PR #1602, opened 2016-10-17
REP-4563 Updated the Core Power Filter functional integration test that wasn't actually testing all the method types on the unroll.

wdschei:


adrianjgeorge:
Good catch

PR #1603, opened 2016-10-19
REP-4516 Fixed classloader test

adrianjgeorge:


repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/classloader/ClassLoaderTest.groovy
```diff
@@ -92,7 +82,8 @@ class ClassLoaderTest extends ReposeValveTest {
         repose.start(killOthersBeforeStarting: false,
                 waitOnJmxAfterStarting: false)
 
-        repose.waitForNon500FromUrl(url)
+        reposeLogSearch.awaitByString("Repose ready", 1, 30)
```
dmnjohns: It's a messy method, but you could use `waitUntilReadyToServiceRequests(checkLogMessage = true)` instead. It has a bit more logic that kicks in if the log message isn't found, but it's a method that gets used elsewhere, so as least we get that consistency and consolidation.

As a side note, we really need to fix up these methods that wait for Repose to be ready.

settings.gradle
```diff
@@ -68,6 +68,24 @@ include 'repose-aggregator',
...
+        'repose-aggregator:tests:test-bundles:filter-five',
+        'repose-aggregator:tests:test-bundles:classloader-bundle-one',
+        'repose-aggregator:tests:test-bundles:classloader-bundle-two',
+        'repose-aggregator:tests:test-bundles:classloader-bundle-three',
+        'repose-aggregator:tests:test-bundles:classloader-bundle-four',
+        'repose-aggregator:tests:test-bundles:classloader-bundle-five',
+        'repose-aggregator:tests:test-bundles:second-filter',
+        'repose-aggregator:tests:test-bundles:second-filter-bundle',
+        'repose-aggregator:tests:test-bundles:busted-application-name-ear',
+        'repose-aggregator:tests:test-bundles:busted-web-fragment-ear',
```
dmnjohns: That is a bunch of new subprojects.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/classloader/ClassLoaderTest.groovy
adrianjgeorge: On at least one of these I very specifically can't wait for a passing request, because a generic request will always fail. In another the test checks for a specific number of log messages that match successful requests which would also be broken by extra requests coming through. The middle test could do something else, but at that point i might as well be consistent within the test.

adrianjgeorge: Wait nevermind, i missed the short circuit line in the other method

adrianjgeorge: Also that method is a hot mess, i very much prefer different methods that do different ready checks to one that has all sorts of weird maybe behavior.

settings.gradle
dmnjohns: Fair enough.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/classloader/ClassLoaderTest.groovy
dmnjohns: All good observations. I would very much so like to fix up our wait methods, however, my immediate concern with this PR is that we are adding yet place where we wait for Repose to start.

So if in the future we change the "Repose ready" message, this gives us another place to update. Or if we come up with a foolproof solution, this is a place where we might miss updating.

As for how to handle that, I'm not sure. I'd love to clean up and consolidate (at least to a single class) the waiting methods that we have.

adrianjgeorge: So using that function breaks the tests, not entirely sure why, but i'm going to leave it as is for now.

dmnjohns: Fair enough.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/classloader/ClassLoaderTest.groovy
```diff
@@ -104,7 +95,6 @@ class ClassLoaderTest extends ReposeValveTest {
         then: "the request header should equal BAR"
         mc.handlings.size() == 1
         mc.receivedResponse.code == "200"
-        mc.handlings[0].request.headers.getFirstValue("FOO") == "BAR"
```
dmnjohns: Why remove this check? It seems like we do expect the "FOO" header to have a value of "BAR". Was this done because of the weird behavior you were seeing with Deproxy?

Also, removing this check makes the `then` comment inaccurate.

adrianjgeorge: So the header isn't actually being replaced, a ethod is being overwriten to make it look like it's replaced. There is no check we can of the deproxy objects that will fire off that method. Thankfully the check does still happen, but it happens in the filter after the one with the method. This is also why as as soon as i turned on intrafilter logging things started breaking. Intrafilter logging applies our own wrappers between filters, so i was losing the method call.

dmnjohns: Gotcha. Indeed, only one method was overridden... which assumes as a certain interaction with the wrapper.

I suppose you could replace that wrapper with our wrapper to get behavior you could test in Deproxy. Since you tied these projects into the rest of our build, you should be able to now. But it probably doesn't matter that much, since the next filter would throw an Exception if things weren't as expected. So fair enough.

dmnjohns:
Looks good to me. 👍

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/classloader/ClassLoaderTest.groovy
adrianjgeorge: Yeah, i considered that route, but it quickly spirals into more work than you would expect.

repose-aggregator/tests/test-bundles/filter-one/src/main/scala/org/openrepose/filters/core/filterone/ClassLoaderServletRequestWrapper.scala
```diff
@@ -19,12 +19,17 @@
...
+import org.openrepose.others.SimplicityDivine
+
 class ClassLoaderServletRequestWrapper(request: HttpServletRequest) extends HttpServletRequestWrapper(request) {
   override def getHeader(headerString: String): String = {
     println("\n\n\n")
     println(s"Requesting header ${headerString}")
     println("\n\n\n")
-    if (StringUtils.startsWith(headerString, "FOO")) {
+    if (StringUtils.startsWith(headerString, "FOO") && (super.getHeader(headerString) != null)) {
       return new SimplicityDivine().createBar
```
wdschei: Why all the return statements?
I get this was just a make it go task, so we might just want to make another task to actually make this production-like code.

adrianjgeorge: Ask @dkowis he did that junk.

dkowis:
SimplicityDevine was a dimtruck word. Perhaps its copy pasta from something he built? 

On October 21, 2016 10:56:20 AM CDT, William Scheidegger <notifications@github.com> wrote:
>wdschei commented on this pull request.
>
>
>
>>  class ClassLoaderServletRequestWrapper(request: HttpServletRequest)
>extends HttpServletRequestWrapper(request) {
>   override def getHeader(headerString: String): String = {
>     println("\n\n\n")
>     println(s"Requesting header ${headerString}")
>     println("\n\n\n")
>-    if (StringUtils.startsWith(headerString, "FOO")) {
>+    if (StringUtils.startsWith(headerString, "FOO") &&
>(super.getHeader(headerString) != null)) {
>       return new SimplicityDivine().createBar
>
>Must have been right after he started learning Scala.
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/rackerlabs/repose/pull/1603

-- 

PR #1605, opened 2016-10-24
Logrotate needs root ownership

bobross419:
[root@some-server ~]# cat /etc/redhat-release
CentOS Linux release 7.2.1511 (Core)

[root@some-server ~]# ls -lah /etc/logrotate.d/repose
-rw-r--r--. 1 repose repose 109 May  6 10:11 /etc/logrotate.d/repose

[root@some-server ~]# logrotate -v /etc/logrotate.d/repose
Ignoring /etc/logrotate.d/repose because the file owner is wrong (should be root).

Handling 0 logs
set default create context

### DIFF
--- a/repose-aggregator/artifacts/valve/build.gradle
+++ b/repose-aggregator/artifacts/valve/build.gradle
@@ -139,8 +139,8 @@ ospackage {
     from('../src/config/logrotate.d') {
         into '/etc/logrotate.d'
         createDirectoryEntry false
-        user 'repose'
-        permissionGroup 'repose'
+        user 'root'
+        permissionGroup 'root'
         fileMode 0644
     }
 
--- a/repose-aggregator/artifacts/web-application/build.gradle
+++ b/repose-aggregator/artifacts/web-application/build.gradle
@@ -112,8 +112,8 @@ ospackage {
     from('../src/config/logrotate.d') {
         into '/etc/logrotate.d'
         createDirectoryEntry false
-        user 'repose'
-        permissionGroup 'repose'
+        user 'root'
+        permissionGroup 'root'
         fileMode 0644
     }
 


bobross419:
Jenkins seems to be stuck on this:

features.core.classloader.ClassLoaderTest > test class loader three FAILED
    java.net.BindException at ClassLoaderTest.groovy:170

Mario-Lopez:
The Gradle build will only output the test failures while it runs the functional tests, so it should still be running.  If you're seeing a ClassLoaderTest failure, I wonder if you're branched off of the latest of master.

bobross419:
I forked and pulled from master branch.  I'll keep an eye on it and follow up later or tomorrow.  Thanks!

bobross419:
That test that I thought was hanging ended up as a failure.  Any ideas for what I might need to change/do different?

`:repose-aggregator:tests:functional-tests:integrationTest

features.core.classloader.ClassLoaderTest > test class loader three FAILED
    java.net.BindException at ClassLoaderTest.groovy:170

5470 tests completed, 1 failed, 7 skipped
:repose-aggregator:tests:functional-tests:integrationTest FAILED`

adrianjgeorge:
retest this please

adrianjgeorge:
looks like you mostly just had bad luck

bobross419:
I'm not dev enough to figure this piece out myself...  

Will the `/var/log/repose/*.log` declaration for logrotate be ok with the way log4j handles Rolling logs in current.log?  

Or should I include a change to specify the phone-home.log, stderr.log, and stdout.log log files?
Reference: https://github.com/rackerlabs/repose/blob/master/repose-aggregator/artifacts/src/config/logrotate.d/repose

bobross419:
retest this please

bobross419:
It looks like I will need to update the logrotate config to specify the stdout.log and stderr.log files since phone-home and current are both owned by log4j.  

That said, how would I tell the repose service to refresh it's file handles?  I tried a SIGHUP, but that ended up killing the entire repose process.

dmnjohns:
Sorry about the long delay, but we've been busy, and your question caused me to do some research around signals, logrotate, and log4j2.

To start, Java handles most signals by shutting down. At least, Oracle's HotSpot JVM does (http://www.oracle.com/technetwork/java/javase/signals-139944.html#gbzbl). Other JVM implementations may behave differently. There is a mechanism to perform arbitrary processing when receiving a signal, but only if that signal is causing the JVM to shutdown (http://docs.oracle.com/javase/8/docs/api/java/lang/Runtime.html#addShutdownHook-java.lang.Thread-). As you can see, using shutdown hooks for signal handling is a pretty sketchy thing to do, and since the JVM is shutting down, it's almost a moot point. Which is all to say, I would not expect signals to have much of an effect (apart from forcing process termination) on Repose or most Java applications.

So onto the question of file descriptors. From what I can tell, Java's standard libraries should manage these automatically. If a file was being read/written and the file descriptor became stale, then that would boil up in Java as some sort of exception which could be handled by whatever code was trying to read/write the file. Which brings us to...

Log4j2, the library Repose uses to log stuff! I actually don't know what, specifically, log4j2 would do in the case where an underlying file descriptor went stale. I would hope and guess that log4j2 would try to refresh the file descriptor using whatever information it used to get the file descriptor in the first place. That would probably be the path. As for what that means, I suppose I would say that a user really should not need to worry about file descriptors.

I got off on a tangent. I think the information is still useful, though, so I'm going to leave it. My point was just that if logrotate was moving log files that Repose was managing, things might just work. However, I still have not addressed your case specifically.

In your case, it sounds like you are redirecting the output of Repose to stdout.log and stderr.log (i.e., `java -jar repose.jar > stdout.log 2> stderr.log`). You're also allowing logrotate to manage those files. I'm assuming that, by default, logrotate moves the log files themselves, while leads to stale file descriptors which breaks your redirection. It's worth noting that all of that is outside of Repose. So that seems to be your problem.

As for the solution, the simplest thing would probably be to tell logrotate to copy and truncate the log file(s) rather than moving them. See:
http://serverfault.com/a/609768

So that's what I would recommend. Whew! All the reading for such a simple answer! If that's all you need, feel free to stop reading here. Otherwise, I would add that there may be some Unix-fu you could do to keep redirects working even when files move. Perhaps by using "additional" file descriptors 3-9. I may play with that and update this thread later. Another solution that is probably more reasonable would be to let log4j2 in Repose manage your logs rather than using redirects. Of course, then you run into permission issues since the `repose` user owns the file rather than the user running the logrotate process. So in the end, there would have to be some trick there to get it working. In which case, I would suggest not using logrotate. At least, as long as the `RollingFile` appender in log4j2 does what you need it to. Check out the documentation for that appender, because I have a feeling it might be enough! Remember that log4j2 is completely configurable within Repose. Pay particular attention to the triggering policy and rollover strategy:
https://logging.apache.org/log4j/2.x/manual/appenders.html#RollingFileAppender

Hope that helps!

dmnjohns:
I've noticed that the default logrotate configuration we ship does actually use `copytruncate`. So that's good, I suppose.

At any rate, I think I would like to see Repose moving away from relying on logrotate and towards using log4j2's capabilities instead. We'll talk about that a bit more as a team.

bobross419:
That's some good info, thanks.  With all that said then, maybe it isn't a good idea to merge this change since this would effectively enable repose logrotation on systems where it was previously failing with consequences that might not be easily or quickly apparent, especially if custom log4j2 configuration has been made.

I'm going to close this PR and solve the issue locally for now, but thanks again for your time on this.

PR #1606, opened 2016-10-24
REP-4582 Merging the 7.3.8.1 release into master

dmnjohns:
I updated the code to use the new way of handling streams. That is the only change to the content of the merge that was modified. All of the tests were just stuck in the right spot.

Once this is merged, we can delete the release branch (`REP-4575_FixInputStreamAssumptions`).

### DIFF
--- a/repose-aggregator/components/filters/rackspace-auth-user-filter/src/main/scala/org/openrepose/filters/rackspaceauthuser/RackspaceAuthUserFilter.scala
+++ b/repose-aggregator/components/filters/rackspace-auth-user-filter/src/main/scala/org/openrepose/filters/rackspaceauthuser/RackspaceAuthUserFilter.scala
@@ -27,6 +27,7 @@ import javax.servlet.http.{HttpServletRequest, HttpServletResponse}
 import com.typesafe.scalalogging.slf4j.LazyLogging
 import org.openrepose.commons.config.manager.UpdateListener
 import org.openrepose.commons.utils.http.PowerApiHeader
+import org.openrepose.commons.utils.io.BufferedServletInputStream
 import org.openrepose.commons.utils.servlet.http.HttpServletRequestWrapper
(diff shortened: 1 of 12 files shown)


repose-aggregator/components/filters/rackspace-auth-user-filter/src/main/scala/org/openrepose/filters/rackspaceauthuser/RackspaceAuthUserFilter.scala
```diff
@@ -62,7 +63,11 @@ class RackspaceAuthUserFilter @Inject()(configurationService: ConfigurationServi
         // this filter only operates on POST requests that have a body to parse
         filterChain.doFilter(servletRequest, servletResponse)
       } else {
-        val wrappedRequest = new HttpServletRequestWrapper(httpServletRequest)
+        val rawRequestInputStream = httpServletRequest.getInputStream
+        val requestInputStream =
+          if (rawRequestInputStream.markSupported) rawRequestInputStream
```
adrianjgeorge: Why not just wrap it always, it's not that expensive?

dmnjohns: So this was mostly copying the behavior of what we released in 7.3.8.1.

However, always wrapping this still concerns me a bit. Especially making a pattern out of that. That may very well leave us with multiple copies of the same request body if multiple filters that read the body are buffering it.

Interestingly, while a `BufferedInputStream` will buffer up to 8192 bytes by default, that number is configurable. I wonder what happens on large request bodies. Luckily, we know the format of requests going through this filter (mostly) so that may not be an issue here.

dmnjohns:
retest this please

PR #1611, opened 2016-10-28
REP-4570 Bumping Sonar Gradle plugin version

dmnjohns:


adrianjgeorge:
I assume you at least ran gradle tasks to make sure the configuration still worked?

dmnjohns:
I did. Also because I didn't remember the name of the task that the plugin makes.

I even ran that test, but it failed because I don't have the proper Sonar host system configuration set.

adrianjgeorge:
Normally they have to mostly match in version, so it's not likely to work by itself

PR #1614, opened 2016-10-31
REP-4555 Fix header name casing being modified.

wdschei:


adrianjgeorge:
Could you add a test hat includes a header from the excluded set, and make sure it does the right thing. Otherwise looks good.

wdschei:
Test added.

adrianjgeorge:
The tests that are failing ar looking for the header names with a lower case, so they need fixing

wdschei:
That was a little more painful than anticipated to fix correctly. The failing tests are now Case Insensitive.

adrianjgeorge:
retest this please

PR #1616, opened 2016-11-02
Updated the Release Verification Fake Services to Node v6 LTS.

wdschei:
This still requires some rework in the Vagrant files.


dmnjohns:
LGTM. I don't need no stinking tests!

I'll fix anything that breaks on the other branch.

PR #1619, opened 2016-11-04
REP-4256 Adding our versioned docs to the landing page

adrianjgeorge:


index.html
```diff
@@ -2,6 +2,23 @@
...
+
+    <section class="top-bar-section">
+        <ul class="right">
+            <li class="has-dropdown">
+                <a href="#">Documentation</a>
+                <ul class="dropdown">
+                    {% include versions.html %}
+                </ul>
+            </li>
+        </ul>
```
dmnjohns: How this becomes that top bar, I'll never know. Fancy formatting magic.

index.html
```diff
@@ -2,6 +2,23 @@
 layout: default
 title: Repose
 ---
+<nav class="top-bar" data-topbar role="navigation">
```
adrianjgeorge: It's not. I tried doing that, but it broke the menu functionality, and it wasn't important enough to try to fix

dmnjohns: Fair enough.

dmnjohns:
I think I have to chalk all of this up to magic.

I super do not know enough about the Foundation framework or other front-end things. If this is what was previewed, though, then that's good enough for me! 👍

adrianjgeorge:
I can walk you through it on monday if you'd like, or we could add it to the next dev meeting.

PR #1621, opened 2016-11-04
REP-4604 Create and Document Repose Docker Deployment

dmnjohns:
Merging onto the release verification branch, since that sets up a few of the dependencies needed by this work.


repose-aggregator/artifacts/docker/build.gradle
```diff
@@ -0,0 +1,53 @@
...
+        resolutionStrategy {
+            force 'de.gesellix:unix-socket-factory:2016-04-06T22-21-19'
+            force 'com.fasterxml.jackson.core:jackson-databind:2.6.4'
+            force 'com.fasterxml.jackson.core:jackson-core:2.6.4'
+            force 'com.fasterxml.jackson.core:jackson-annotations:2.6.4'
+        }
+    }
+}
+
+String reposeVersion = project.hasProperty('repose-version') ? project.property('repose-version') as String : version as String
```
adrianjgeorge: Why allow setting of version outside of the version that is project wide?

repose-aggregator/artifacts/docker/build.gradle
```diff
@@ -0,0 +1,53 @@
...
+
+docker {
+    url = "unix:///var/run/docker.sock"
+
+    registryCredentials {
+        username = project.hasProperty('dockerhub.username') ? project.property('dockerhub.username') : 'Need to provide a Docker Hub username'
```
adrianjgeorge: I'll assume that there is or was a pr for these credentials on the infrastructure management project

repose-aggregator/artifacts/docker/src/docker/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+RUN touch /var/log/repose/current.log
+USER root
+
+# Expose the default Repose service port for host port forwarding.
+# If the port in the user's system model differs from this port, the user will have to map it manually using the
+# "-p" flag with the Docker run command.
+EXPOSE 8080
+
+# Start Repose.
+CMD ["java", "-jar", "/usr/share/repose/repose-valve.jar", "-c", "/etc/repose"]
```
adrianjgeorge: super don't like that they can't run the service like everybody else

repose-aggregator/artifacts/docker/build.gradle
dmnjohns: It enabled doing what I did a couple of days ago -- releasing Docker Hub images for versions of Repose that we've already cut. Or for versions that we patch (that came before these tasks).

repose-aggregator/artifacts/docker/build.gradle
dmnjohns: There was.

https://github.com/rackerlabs/repose-infrastructure-ng/pull/37

dmnjohns:
No CentOS version... yet. We talked about that a bit, and I still haven't come up with a great reason for having both (other than the fact that we could use both for verification).

I actually wrote a blurb about that on the Docker Hub repository page under the "Base Image" header:
https://hub.docker.com/r/repose/repose/

repose-aggregator/artifacts/docker/build.gradle
adrianjgeorge: i figured as much

adrianjgeorge:
Logs, people will want the logs (we'll want them to have the logs). I assume their is a mounting you can do their as well?

dmnjohns:
Of course! So the cop out answer would be that is a Docker thing, go read their docs.

However, I've distilled some of the more useful information here:
https://repose.atlassian.net/wiki/display/REPOSE/Docker#Docker-DockerContainer

Basically, you can make any directory in the container a `Volume` which is shared with the host. See:
https://docs.docker.com/engine/tutorials/dockervolumes/#/adding-a-data-volume

You could also run the container in attached mode and use Unix shell redirection to put logs wherever you want them. That's not very elegant, though.

dmnjohns:
Oh, and you can always launch a shell in the Docker container if you just want to view the logs. Of course, I imagine most users want logs to persist, so using a `Volume` is probably the way to go. Again, though, it's just Docker stuff.

PR #1623, opened 2016-11-11
REP-4681 Sonar cleanup

adrianjgeorge:


repose-aggregator/commons/commons-utilities/src/main/java/org/openrepose/commons/utils/reflection/ReflectionTools.java
```diff
@@ -38,24 +38,7 @@ private ReflectionTools() {
...
-                }
-
-                if (!constructorParameters[i].isAssignableFrom(parameters[i])) {
-                    suitable = false;
-                    break;
-                }
-            }
-
-            if (suitable) {
+            if (parametersMatch(constructorParameters, parameters)) {
```
Mario-Lopez: Seems like a waste to create the `constructorParameters` variable now.

You could also replace the method with:
```
return Arrays.stream((Constructor<T>[]) clazz.getConstructors())
        .filter(constructor -> parametersMatch(constructor.getParameterTypes(), parameters))
        .findFirst()
        .orElseThrow(() -> new NoSuchMethodException("No constructor found with expected signature for class: " + clazz.getCanonicalName()));
```

dmnjohns: Oh right. Java 8.

I'll use the stream approach, then.

dmnjohns:
LGTM.

PR #1629, opened 2016-11-25
REP-4734 Updated jython dependency to include jython stdlib

Mario-Lopez:
When using python in the scripting filter, we currently don't have access to the stdlib (e.g. the `json` module).  I updated the dependency from using jython to jython-standalone so we'd get the stdlib as well.

I updated the scripting filter functional test to ensure we got access to the `json` module, and I added another functional test to verify we could base64 decode and JSON parse the X-Catalog header created by the Keystone v2 filter using Python.


thomasem:
:+1

dmnjohns:
The Gradle changes look good. 👍

PR #1631, opened 2016-11-30
REP-4726 HERP should log the Tenant ID from OS response if it wasn't already in the request.

wdschei:


repose-aggregator/components/filters/herp-filter/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
```diff
@@ -122,7 +122,11 @@ class HerpFilter @Inject()(configurationService: ConfigurationService,
     }
 
     val projectIds = httpServletRequest.getHeaders(OpenStackServiceHeader.TENANT_ID.toString).asScala
-      .++(httpServletRequest.getHeaders(X_PROJECT_ID).asScala).toTraversable
+      .++(httpServletRequest.getHeaders(X_PROJECT_ID).asScala).toTraversable match {
+      case x if x.nonEmpty => x
```
dmnjohns: You could simplify this a bit by dropping the guard. I often see this pattern expressed in the following form:
```
myCollection match {
  case _ :: _ => // do something for a non-empty collection
  case _ => // do something for an empty collection
}
```

repose-aggregator/components/filters/herp-filter/src/test/scala/org/openrepose/filters/herp/HerpFilterTest.scala
```diff
@@ -387,6 +428,20 @@ class HerpFilterTest extends FunSpec with BeforeAndAfterEach with Matchers {
...
+
+      // when:
+      herpFilter.configurationUpdated(herpConfig)
+      herpFilter.doFilter(servletRequest, servletResponse, filterChain)
+
+      // then:
+      def logEvents = listAppenderPre.getEvents
+      logEvents.size shouldBe 1
+      logEvents.get(0).getMessage.getFormattedMessage should include("\"ProjectID\" : [  \"foo\"  ,\"bar\"  ]")
+    }
```
dmnjohns: I'd love to see a test around the behavior if either the `X-Tenant-Id` or `X-Project-Id` header is present in both the request and response.

wdschei: Good call. I'll add that one.

wdschei: There is one in the functional tests.

dmnjohns:
I did not look at your functional tests (leaving that for the QE review task) but everything else (that was not commented on) looks good to me. 👍

repose-aggregator/components/filters/herp-filter/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
dmnjohns: Also, I realize you might be using the guard so that you can bind x. Of course, you could always return `myCollection`.

In the end, it might be a preference thing... I just don't really like unnecessary guards. Thinking about it, though, it may actually be more efficient in this case.

dmnjohns: You know what, just ignore this entire comment chain.

I cannot come up with a way of doing this that I really like. I kind of want a method like `getOrElse` on Option but for  collections. I want to treat an empty collection as being distinct from a non-empty collection.

wdschei: I did update these a bit to abstract out to the common `HeaderInteractor` interface on the new `Wrapper`'s. That way some of the copy/paste could be eliminated. Lines wise I think it was the same, but it looks nicer now.

repose-aggregator/components/filters/herp-filter/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
```diff
@@ -121,8 +123,17 @@ class HerpFilter @Inject()(configurationService: ConfigurationService,
       else stripHeaderParams(allProjectIds.maxBy(getQuality))
     }
 
-    val projectIds = httpServletRequest.getHeaders(OpenStackServiceHeader.TENANT_ID.toString).asScala
-      .++(httpServletRequest.getHeaders(X_PROJECT_ID).asScala).toTraversable
+    def getSplitHeaders(headerInteractor: HeaderInteractor, headers: List[String]): Traversable[String] = {
+      headers.foldLeft(Traversable[String]()) { (accumulator, current) =>
```
dmnjohns: I'm not sure if we have a real convention, but I've always preferred the `Traversable.empty[String]` style. It just seems clearer to me (since you avoid the parenthesis pile-up).

repose-aggregator/components/filters/herp-filter/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
```diff
@@ -121,8 +123,17 @@ class HerpFilter @Inject()(configurationService: ConfigurationService,
...
+        accumulator.++(headerInteractor.getSplittableHeaders(current).asScala)
```
dmnjohns: Hm, I wonder if you still need to invoke the `++` function using dot notation or if you can just use it in infix notation...

repose-aggregator/components/filters/herp-filter/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
```diff
@@ -121,8 +123,17 @@ class HerpFilter @Inject()(configurationService: ConfigurationService,
...
+      }
+    }
+
+    val tenantProjectHeaders = List(OpenStackServiceHeader.TENANT_ID.toString, X_PROJECT_ID)
+    val projectIds = getSplitHeaders(new HttpServletRequestWrapper(httpServletRequest), tenantProjectHeaders) match {
+      case x if x.nonEmpty => x
+      case _ => getSplitHeaders(httpServletResponse, tenantProjectHeaders)
+    }
```
dmnjohns: Now that this is much more manageable, we probably could drop the match for a simple if/else conditional. Again, that's probably a preference thing, but something about using pattern matching to evaluate a conditional expression feels off to me.

```
val reqProjectIds = getSplitHeaders(new HttpServletRequestWrapper(httpServletRequest), tenantProjectHeaders)
val projectIds = if (reqProjectIds.nonEmpty) reqProjectIds else getSplitHeaders(httpServletResponse, tenantProjectHeaders)
```

repose-aggregator/src/docs/asciidoc/filters/herp.adoc
```diff
@@ -1,3 +1,213 @@
-= Highly Efficient Record Processor Filter
+= Highly Efficient Record Processor (HERP) Filter
```
dmnjohns: Oh yeah. I probably should have waited for documentation before code reviewing this... I'll take a look tomorrow. 😛

repose-aggregator/components/filters/herp-filter/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
```diff
@@ -121,8 +123,17 @@ class HerpFilter @Inject()(configurationService: ConfigurationService,
...
 
-    val projectIds = httpServletRequest.getHeaders(OpenStackServiceHeader.TENANT_ID.toString).asScala
-      .++(httpServletRequest.getHeaders(X_PROJECT_ID).asScala).toTraversable
+    def getSplitHeaders(headerInteractor: HeaderInteractor, headers: List[String]): Traversable[String] = {
+      headers.foldLeft(Traversable[String]()) { (accumulator, current) =>
+        accumulator.++(headerInteractor.getSplittableHeaders(current).asScala)
+      }
+    }
+
+    val tenantProjectHeaders = List(OpenStackServiceHeader.TENANT_ID.toString, X_PROJECT_ID)
```
dmnjohns: Also, why a `List` (here and in the test)? If anything, I would think of this as a `Set`. `List` seems odd because it's a specific collection type -- not like `Traversable` -- which makes me think there must be a reason for using it... but I'm not sure there is?

repose-aggregator/components/filters/herp-filter/src/test/scala/org/openrepose/filters/herp/HerpFilterTest.scala
```diff
@@ -355,10 +392,60 @@ class HerpFilterTest extends FunSpec with BeforeAndAfterEach with Matchers {
...
+      logEvents.get(0).getMessage.getFormattedMessage should include("\"DefaultProjectID\" : \"bar\"")
+    }
+
+    val hdrNames = List("X-Tenant-Id", "X-Project-Id")
+    val reqValues = List(List("reqFoo;q=0.5", "reqBar;q=1.0"), List("reqFoo;q=0.5,reqBar;q=1.0"))
+    val resValues = List(List("resFoo;q=0.5", "resBar;q=1.0"), List("resFoo;q=0.5,resBar;q=1.0"))
+    val headers = for {
+      reqHdr <- hdrNames
+      resHdr <- hdrNames
+    } yield (reqHdr, resHdr)
```
dmnjohns: Ah, the classic permutation with replacement. It's a little strange that Scala doesn't provide an easy function for generating this like it does for permutations without replacement (i.e., `permutations`). This is pretty clean, though. Better than the solution I came up with when I ran across this awhile back, which I'll put below for your amusement:
```
hdrNames.flatMap(List.fill(hdrNames.length)(_)).combinations(hdrNames.length).flatMap(_.permutations).toList
```

I will note, however, that you could single line this. Not sure if it's better or worse, but it is an option.
```
val headers = for (reqHdr <- hdrNames; resHdr <- hdrNames) yield (reqHdr, resHdr)
```

repose-aggregator/components/filters/herp-filter/src/test/scala/org/openrepose/filters/herp/HerpFilterTest.scala
```diff
@@ -355,10 +392,60 @@ class HerpFilterTest extends FunSpec with BeforeAndAfterEach with Matchers {
...
+    val values = for {
+      reqVal <- reqValues
+      resVal <- resValues
+    } yield (reqVal, resVal)
+    headers.foreach { case (reqHdr, resHdr) =>
```
dmnjohns: The Scala style guide recommends dropping the dot from the method invocation here (i.e., use infix notation). I tend to agree.

http://docs.scala-lang.org/style/method-invocation.html

repose-aggregator/components/filters/herp-filter/src/test/scala/org/openrepose/filters/herp/HerpFilterTest.scala
```diff
@@ -355,10 +392,60 @@ class HerpFilterTest extends FunSpec with BeforeAndAfterEach with Matchers {
...
+      values.foreach { case (reqVal, resVal) =>
+        it(s"should log the default tenant id header from the Request even if it is available on the Response (REQ=$reqHdr:$reqVal)(RES=$resHdr:$resVal)") {
```
dmnjohns: Good ol' generated tests. 👍

repose-aggregator/components/filters/herp-filter/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
wdschei: Infix does work and again looks cleaner.
Fixed.

repose-aggregator/components/filters/herp-filter/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
wdschei: The `if`/`else` way causes the storage of a `value` whereas the `match`/`case` doesn't. This is probably a very moot point and is optimized appropriately by the compiler anyway and the first value would be as transient as the `x` variable. I don't feel strongly either way on this one.
Fixed.

repose-aggregator/components/filters/herp-filter/src/test/scala/org/openrepose/filters/herp/HerpFilterTest.scala
wdschei: I to am surprised there is no easier way to do this, but I have used your other way before too. Again I don't have a strong opinion on the single line vs the multiline and both are equally readable.
Fixed.

repose-aggregator/components/filters/herp-filter/src/test/scala/org/openrepose/filters/herp/HerpFilterTest.scala
wdschei: Fixed.

repose-aggregator/components/filters/herp-filter/src/test/scala/org/openrepose/filters/herp/HerpFilterTest.scala
wdschei: It was this style that led us to find the Xerces bug last year using API Checker. It is a great way to ensure all of the permutations are hit without having to write them all manually.

repose-aggregator/components/filters/herp-filter/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
dmnjohns: Looks like we're still using `List`. Is that what we want to stay with?

repose-aggregator/components/filters/herp-filter/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
dmnjohns: That's true. In this case, I'm pretty sure the pattern match is actually compiled to the same bytecode as a simple conditional would have been. If you really wanted, you could put both lines into a code block to reduce the scope of the variable, but as you said, there probably isn't a reason to. Also, that gets messy again, at which point using pattern matching may have been clearer choice anyway.

If you like the pattern match, you can totally keep using it. I have weird preferences sometimes.

repose-aggregator/src/docs/asciidoc/filters/herp.adoc
```diff
@@ -1,3 +1,213 @@
...
+
+The Highly Efficient Record Processor (HERP) filter logs an event for each API request.
+These logs provide information regarding the processing of each request.
+The logs can be inspected to view the usage of a service and to perform auditing.
+Events are recorded even if the API request is rejected by Repose before reaching the origin service.
+
+== General filter information
+* *Filter name:* HERP
+* *Filter configuration:* highly-efficient-record-processor.cfg.xml
+* *Released:* v7.0.0.1
```
dmnjohns: I'm curious what the reason is for including the release version. That's probably a discussion that's easier to have in person. I'll ask at lunch, probably. 😛

repose-aggregator/src/docs/asciidoc/filters/herp.adoc
```diff
@@ -1,3 +1,213 @@
...
+    </filterOut>
+</highly-efficient-record-processor>
+----
+
+[NOTE]
+The `{{methodLabel}}` variable value is provided by the `api-validator` filter.
+If you aren't using that filter, `{{methodLabel}}` will return empty string.
+
+=== User Access Event Recording With Flume
+For information about using the Flume with the HERP filter for user access event recording, see link:../recipes/api-event-flume.html[CF Flume Sink].
```
dmnjohns: This should be a cross-reference rather than a link.

See:
http://asciidoctor.org/docs/user-manual/#inter-document-cross-references

repose-aggregator/src/docs/asciidoc/filters/herp.adoc
```diff
@@ -1,3 +1,213 @@
...
+==== Required headers
+This filter has no required headers.
+
+==== Required preceding filters
+This filter has no dependencies on other filters and can be placed wherever it is needed in the filter chain.
+
+[CAUTION]
+In order for the HERP filter to process and record every request, no preceding filter in the filter chain may write a response.
+To prevent short-circuiting without changing behavior, delegation has been added to many Repose filters.
+See: Delegation in Repose
```
dmnjohns: I suspect this should be a cross-reference?

repose-aggregator/src/docs/asciidoc/filters/herp.adoc
```diff
@@ -1,3 +1,213 @@
...
+
+== Prerequisites & Postconditions
+=== Input
+==== Required headers
+This filter has no required headers.
+
+==== Required preceding filters
+This filter has no dependencies on other filters and can be placed wherever it is needed in the filter chain.
+
+[CAUTION]
```
dmnjohns: We may want to come up with something for the style guide with respect to admonitions. We could either prefer the paragraph style and only use the block style where necessary, or we could always use the block style.

In your case, a paragraph style admonition would work:
> CAUTION: 

You used a block style admonition even though the content does not appear to be complex.

See: http://asciidoctor.org/docs/user-manual/#admonition

repose-aggregator/src/docs/asciidoc/filters/herp.adoc
wdschei: Yep. I missed this one.
Fixed.

repose-aggregator/src/docs/asciidoc/filters/herp.adoc
```diff
@@ -1,3 +1,213 @@
...
+----
+
+[NOTE]
+The `{{methodLabel}}` variable value is provided by the `api-validator` filter.
+If you aren't using that filter, `{{methodLabel}}` will return empty string.
+
+=== User Access Event Recording With Flume
+For information about using the Flume with the HERP filter for user access event recording, see link:../recipes/api-event-flume.html[CF Flume Sink].
+
+=== Configurable parameters
```
dmnjohns: I don't think we want this section at all? Example and the schema link should suffice.

dmnjohns: I'm not really sure that this belongs under the `Examples` section header.

repose-aggregator/src/docs/asciidoc/filters/herp.adoc
```diff
@@ -1,3 +1,213 @@
...
+
+1. Edit the `highly-efficient-record-processor.cfg.xml` file.
+1. Add the HERP filter to your system model.
+1. Configure the logger-name attribute to note the SLF4j logger
+1. Configure the service-code attribute to note the service being accessed.
+1. Configure the region attribute to note the region of the service.
+1. Configure the data-center attribute to note the data center of the service.
+1. Configure the template element to define the format of the messages to be logged by the HERP filter.
+
+=== Optional user-configurable filtering
```
dmnjohns: I'm not really sure that this belongs under the `Examples` section header.

repose-aggregator/src/docs/asciidoc/filters/herp.adoc
```diff
@@ -1,3 +1,213 @@
...
+        <!-- Implicit AND -->
+        <match field="dataCenter" regex="DFW"/>
+    </filterOut>
+    <!-- Implicit OR -->
+    <filterOut>
+        <match field="userName" regex=".*[bB]ar.*"/>
+        <!-- Implicit AND -->
+        <match field="parameters.abc" regex="123"/>
+    </filterOut>
+</highly-efficient-record-processor>
```
dmnjohns: This is a legit example, but I would add callouts. And remove comments.

repose-aggregator/src/docs/asciidoc/filters/herp.adoc
```diff
@@ -1,3 +1,213 @@
...
+   For example, a regex of `"racker"` would match a value of `"rackerTwo"`.
+   To strictly match, use a more explicit regex such as `"^racker$"`.
+1. Optionally, you can:
+ * Specify a `pre-filter-logger-name` that logs all events to it prior to performing any filter logic.
+ * Specify a `post-filter-logger-name` that logs all events to it that are not filtered out.
+
+[source,xml]
+.highly-efficient-record-processor.cfg.xml
+----
+<?xml version="1.0" encoding="UTF-8"?><highly-efficient-record-processor xmlns="http://docs.openrepose.org/repose/highly-efficient-record-processor/v1.0"
```
dmnjohns: The XML document header and root element are weirdly on the same line.

repose-aggregator/src/docs/asciidoc/filters/herp.adoc
```diff
@@ -1,3 +1,213 @@
...
+
+[cols="2", options="header,autowidth"]
+|===
+| Supported helpers          | Description
+| cadfOutcome <responseCode> | Converts a HTTP response code into either "success" or "failure".
+| cadfTimestamp <timestamp>  | Converts a timestamp into the CADF format.
+| cadfMethod <requestMethod> | Converts a HTTP method into the CADF format.
+|===
+
+=== Recommendation for managing disk space
```
dmnjohns: I'm not really sure that this belongs under the `Examples` section header.

repose-aggregator/src/docs/asciidoc/filters/herp.adoc
wdschei: As per the DevMeeting, we'll leave the released version in the template.

repose-aggregator/src/docs/asciidoc/filters/herp.adoc
wdschei: Fixed.

repose-aggregator/src/docs/asciidoc/filters/herp.adoc
wdschei: Moved to the new `Additional Information` section.
Fixed

wdschei:
The six remaining  test failures seem to be linked to the bug identified here:
* https://repose.atlassian.net/browse/REP-4764

PR #1632, opened 2016-12-02
REP-4674 - CORS update for Chrome

Mario-Lopez:
I updated the CORS Filter to handle Chrome/Safari same-origin requests correctly.

**Background:**
Previously, the CORS Filter was following the CORS spec and doing server-side request rejection on top of that.  The spec does not say one way or another whether or not the server should be rejecting invalid requests, but the spec does assume you don't do rejection because it specifies that the presence of an Origin header initiates the CORS flow.  Unfortunately, the Origin header is technically permissible on all requests (both non-CORS and CORS) which Chrome and Safari do, and this becomes a problem when a same-origin request is treated like a CORS request and gets rejected.  This isn't a problem if the CORS Filter is configured to let a small number of trusted origins use the API.  It becomes a problem when the filter is configured to tightly restrict the API because several untrusted origins are being allowed to make CORS requests to the server.

There is not a lot of literature on how to get around this, but I did find a couple of recommendations to compare the Host header with the Origin header in order to determine if the request was a CORS request or a same-origin request.  This comparison is tightly constrained by the Origin spec (RFC 6454), so a simple string comparison is not sufficient.

There were also recommendations to check the X-Forwarded-Host header which, if present, would contain the original Host header value if the request had gone through a reverse-proxy server.  This header is not covered by any spec, but it seems to be treated the same as the Host header except that it could have multiple values (say in the case of going through multiple proxy servers).

**Updates to the filter:**
The CORS Filter will continue to treat requests without an Origin header as non-CORS requests.  Any request with an Origin header, a preflight header, and a method of OPTIONS is immediately treated as a CORS Preflight request since no web browser should be sending a preflight header for a same-origin request.

After checking for those two possibilities, the CORS Filter first checks for an X-Forwarded-Host header that it can successfully parse, otherwise it uses the Host header (which is already parsed and attained through the request object).  It then normalizes and converts the host and Origin to a `java.net.URI` that can then be used to compare them to each other.  If they match, then the request is considered a non-CORS request (same-origin).  If they don't match, then it'll be considered a CORS Actual request.  If the Origin header is malformed (i.e. can't be parsed as a `java.net.URI`), then the filter returns a 400.

Unfortunately, I didn't find any library that did exactly what I needed for a lot of this.  `java.net.URI` parsed the strings for me, but it didn't give me the default ports for the "http" and "https" schemes nor did it handle normalizing IPv6 addresses.  Converting it to a `java.net.URL` did way too much and incorrectly considered host names with different subdomains as the same (i.e. the `java.net.URL` equality rules differed from the ones I needed in RFC 6454).  Hence the IP address and default port normalizing I'm doing in the code.

Also, my understanding is parsing a `java.net.URI` isn't cheap, so I avoid doing so if I can.  For example, I don't actually care if the Origin is malformed in a CORS Preflight request, so I don't bother validating it in that case.

Since not everyone full understands CORS, I added additional logging that can help users understand why the filter might be doing what it's doing.  The DEBUG level will indicate when a request is considered a CORS request and why it was rejected.  The TRACE level will give even more information for every single request including if it was CORS or not and if it was allowed or rejected with the reason why.


repose-aggregator/components/filters/cors-filter/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
```diff
@@ -166,9 +159,42 @@ class CorsFilter @Inject()(configurationService: ConfigurationService)
...
+    lazy val hostUri = getHostUri(request)
+
+    (Option(originHeader), isOptionsRequest, Option(preflightRequestedMethod)) match {
+      case (None, isOptions, _) =>
+        logger.trace("Request is Non-CORS request because it does not have an Origin header.")
+        NonCorsRequest(isOptions)
+      case (Some(origin), true, Some(requestedMethod)) =>
+        logger.debug("Request is CORS Preflight request because it is an OPTIONS request with Origin and preflight headers.")
+        PreflightCorsRequest(origin, requestedMethod)
+      case (Some(_), isOptions, _) if originUri.isFailure =>
```
dmnjohns: I wonder if it would be worth inlining the `getOriginUri` function call here. `lazy val`s do have a small cost since they have to be thread safe, though I think the Scala compiler folks have made it pretty negligible at this point.

dmnjohns: Derp. Just realized it's used more than once. Ignore me. Inlining would be gross.

repose-aggregator/components/filters/cors-filter/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
```diff
@@ -178,23 +204,71 @@ class CorsFilter @Inject()(configurationService: ConfigurationService)
         Nil
     })
   }
+
+  def getHostUri(request: HttpServletRequestWrapper): URI = {
+    val maybeForwardedHost = request.getSplittableHeaderScala(CommonHttpHeader.X_FORWARDED_HOST).headOption
+    val maybeForwardedHostUri = Try(maybeForwardedHost
+      .map(forwardedHost => new URIBuilder(s"${request.getScheme}://$forwardedHost"))
+      .map(uri => uri.setPort(normalizePort(uri.getPort, uri.getScheme)).setHost(normalizeUriHost(uri.getHost)).build()))
+    lazy val hostUri = new URIBuilder().setScheme(request.getScheme).setHost(normalizeHost(request.getServerName))
```
dmnjohns: Is `request.getServerName`guaranteed to be in-sync with the `Host` header? I would hope so, but I'm not certain.

Actually, I know that isn't the case in our wrappers. Server name can be set independently of the `Host` header and vice-versa. We should probably fix that.

Mario-Lopez: The contract seems to say it is.

http://docs.oracle.com/javaee/6/api/javax/servlet/ServletRequest.html#getServerName()

repose-aggregator/components/filters/cors-filter/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
```diff
@@ -178,23 +204,71 @@ class CorsFilter @Inject()(configurationService: ConfigurationService)
         Nil
     })
   }
+
+  def getHostUri(request: HttpServletRequestWrapper): URI = {
+    val maybeForwardedHost = request.getSplittableHeaderScala(CommonHttpHeader.X_FORWARDED_HOST).headOption
```
Mario-Lopez: Good catch, but it's a moot point.

1. This filter rejecting a request shouldn't be relied upon for anything at all.
2. Leaving off the `Origin` header has the same effect as making the `Origin` header match the `Host` header and as you mentioned making it match the `X-Forwarded-Host` header, although...
3. ...the `X-Forwarded-Host` header is different because it's not on the [forbidden list of headers](https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name) that you can set in a request within a web browser.  Luckily, this is a useless superpower.  If a request tries to set the `X-Forwarded-Host` header which is not a [CORS simple header](https://www.w3.org/TR/cors/#simple-header), the web browser will first send a CORS Preflight Request to ask the third-party server if it's allowed to send the `X-Forwarded-Host` header.  It will do so with the original Origin and Host header which would be rejected by the CORS filter.

dmnjohns: Ah, I see. So the user must be using a browser that conforms with the CORS specification to get any benefit from it. However, using such a browser is what makes CORS "secure" as it were.

repose-aggregator/components/filters/cors-filter/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
dmnjohns: On second thought, perhaps I shouldn't. In this case, we really aren't concerned with where the request is going, but rather with where it is now. If the two are kept in-sync, then we lose that distinction (due to the nature of being a reverse proxy).

repose-aggregator/components/filters/cors-filter/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
```diff
@@ -178,23 +204,71 @@ class CorsFilter @Inject()(configurationService: ConfigurationService)
...
+
+  def normalizePort(port: Int, scheme: String): Int = (port, scheme.toLowerCase) match {
+    case (p, _) if p > 0 => p
+    case (_, s) if s == "http" => 80
+    case (_, s) if s == "https" => 443
+    case _ => port
+  }
+
+  def getHeaderNamesToExpose(response: HttpServletResponseWrapper): List[String] =
+    (DefaultExposeHeaders.filterNot(response.containsHeader) ++ response.getHeaderNames.asScala).distinct
```
dmnjohns: Filtering then calling `distinct` seems redundant? Also, `distinct`is not case-insensitive, so I'm not sure how much it's buying you here.

repose-aggregator/components/filters/cors-filter/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
```diff
@@ -44,18 +48,18 @@ class CorsFilter @Inject()(configurationService: ConfigurationService)
...
 
-  private var configurationFile: String = DEFAULT_CONFIG
+  private var configurationFile: String = _
   private var initialized = false
-  private var allowedOrigins: Iterable[Regex] = _
-  private var allowedMethods: Iterable[String] = _
-  private var resources: Iterable[Resource] = _
+  private var allowedOrigins: Seq[Regex] = _
+  private var allowedMethods: Seq[String] = _
+  private var resources: Seq[Resource] = _
```
dmnjohns: Out of a curiosity, was there a particular reason for this change? I'd like to thing something came out of our discussion on the subject.

repose-aggregator/components/filters/cors-filter/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
```diff
@@ -178,23 +204,71 @@ class CorsFilter @Inject()(configurationService: ConfigurationService)
...
+  private final val DefaultExposeHeaders = List(CommonHttpHeader.CONTENT_LENGTH.toString)
 
   implicit def autoHeaderToString(hc: HeaderConstant): String = hc.toString
 
-  sealed trait RequestType
-  object NonCorsRequest extends RequestType
-  object PreflightRequest extends RequestType
-  object ActualRequest extends RequestType
+  sealed trait RequestType {
+    def isOptions: Boolean
```
dmnjohns: Also, traits can totally hold 'val's. That may have been more appropriate for what you were trying to express -- namely, that every instance of this trait should have a readable `isOptions` value. Granted, the generated bytecode will probably look very similar if not identical.

repose-aggregator/components/filters/cors-filter/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
Mario-Lopez: Pretty much.

repose-aggregator/components/filters/cors-filter/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
Mario-Lopez: Yeah...  the only potential benefit is not getting the two `Date` headers I always see in the response.  I'll remove it.

repose-aggregator/components/filters/cors-filter/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
```diff
@@ -68,39 +72,24 @@ class CorsFilter @Inject()(configurationService: ConfigurationService)
...
 
       validationResult match {
-        case Pass =>
+        case Pass(validMethods) =>
           requestType match {
-            case NonCorsRequest => filterChain.doFilter(httpServletRequest, httpServletResponse)
-            case PreflightRequest =>
-              logger.trace("Allowing preflight request.")
+            case NonCorsRequest(_) => filterChain.doFilter(httpServletRequest, httpServletResponse)
+            case InvalidCorsRequest(_, _) => // this is not possible, but adding this line to remove a compiler warning
```
Mario-Lopez: Legit.  I'll do that instead.

repose-aggregator/components/filters/cors-filter/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
Mario-Lopez: It mostly stemmed from wanting to change this:
`validMethods.exists(requestedMethod == _)`
to this:
`validMethods.contains(method)`
where `requestedMethod` was renamed to `method`.

From there it allowed me to propagate the change throughout.  I vaguely remember having another reason, but that was the big one.

repose-aggregator/components/filters/cors-filter/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
Mario-Lopez: Legit.  There was what I had in mind and then there was what I managed to do.  I apparently didn't go back to verify that what I managed to do was worth doing at all.  I fix.

repose-aggregator/components/filters/cors-filter/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
dmnjohns: Fair enough.

dmnjohns:
Looks good to me. 👍

Mario-Lopez:
Cool.  I'm still working on the docs.  I think I'm just going to update the docs to be up to date and then create another story to add a fancy troubleshooting section since that's not explicitly required by this story.

repose-aggregator/src/docs/asciidoc/filters/cors.adoc
```diff
@@ -1,20 +1,27 @@
...
-The presence of certain headers will indicate what kind of request is being made.
+Configuring CORS improperly could expose https://en.wikipedia.org/wiki/Cross-site_request_forgery[Cross-Site Request Forgery (CSRF)] vulnerabilities in the origin service that were previously hidden by not supporting CORS.
+====
+
+== General filter information
+* *Name:* cors
+* *Default Configuration:* cors.cfg.xml
+* *Released:* v7.2.0.0
+* *Bundle:* repose-filter-bundle
+* *Schema:* link:../schemas/cors-configuration.xsd[cors-configuration.xsd]
```
dmnjohns: If we're including the schema in the general information block, I feel like we should at least put it next to the `Default Configuration` item. I'd probably put those two back-to-back at the end of this list. Perhaps even clarify that it's the configuration schema?

Mario-Lopez: Yeah, my thinking was the "General Filter Information" would include the "Name" of the filter, and the section we're now calling "Examples" would be called "Configuration" with the "Default filename" and "Schema" listed there.  I'm guessing we decided against that based on what I saw in the Migrate Doc story which is why I went with this.  I'll make the changes you mentioned so we can feel it out.

repose-aggregator/src/docs/asciidoc/filters/cors.adoc
```diff
@@ -1,20 +1,27 @@
...
-Configuring CORS improperly could expose CSRF vulnerabilities in the origin service that were previously blocked by not supporting CORS.
-
-== Headers
-
-=== Required
-The presence of certain headers will indicate what kind of request is being made.
+Configuring CORS improperly could expose https://en.wikipedia.org/wiki/Cross-site_request_forgery[Cross-Site Request Forgery (CSRF)] vulnerabilities in the origin service that were previously hidden by not supporting CORS.
+====
+
+== General filter information
```
dmnjohns: Hm, that's a question. On some pages we capitalize all words other than articles while on others we only capitalize the first word. We should probably try to be consistent on section titles, so which do we prefer? That should also be in the style guide.

Mario-Lopez: Totally agree.  I have not been able to figure out which way we're going to go.

repose-aggregator/src/docs/asciidoc/filters/cors.adoc
```diff
@@ -26,18 +33,45 @@ All of these headers will be supplied by the client making the request.
...
 a| * `Origin`
    * `Access-Control-Request-Method`
-a| * `Origin`
+a| * `Origin`*
 |===
 
-=== Created
-The CORS filter wil not add any headers to the response for non-CORS requests.
-The filter will add the following headers to the response for these scenarios when the request is allowed to proceed:
+*{sp}If the `Origin` header matches the original `Host` header in the request, it will be considered a non-CORS request.
```
dmnjohns: This is kind of odd. I get that the {sp} attribute is so that this isn't made into a list, but...

Would a footnote work here? I mean, it would work, but is it displayed in a way that you find acceptable?
http://asciidoctor.org/docs/user-manual/#user-footnotes

dmnjohns: So it renders the footnote within the table cell. How annoying. It seems I'm not alone in thinking that.
https://github.com/asciidoctor/asciidoctor/issues/1705

However, that could probably be customized in our CSS?

Mario-Lopez: Not with any CSS that I know of, but if you can make a smiley face with CSS, I suppose anything's possible?

repose-aggregator/src/docs/asciidoc/filters/cors.adoc
Mario-Lopez: I prefer capitalizing the first letter of each word.  In fact, you can tell which headings I wrote by which ones are written like that.  I should have noticed they weren't consistent, though.  In any case, should I capitalize the first letter of each word for now?  We can change it later, but for now I'd love to be consistent within the same page at least.

repose-aggregator/src/docs/asciidoc/filters/cors.adoc
dmnjohns: Nope. The HTML itself is all wonky (as you would probably expect). So the Asciidoctor HTML engine probably needs to be changed. Well, or we get straight up cray-cray with CSS. But no thanks.

repose-aggregator/src/docs/asciidoc/filters/cors.adoc
```diff
@@ -290,64 +337,57 @@ the specified URI will be treated as a Java regular expression.
...
 
-WARNING: CORS is not a security feature.
-It is a mechanism for informing clients (e.g. web browsers) of conditions when client-side security can be slightly relaxed in certain circumstances.
-That is, the security lies completely within the client.
-Simply leaving out the `Origin` header in the request completely bypasses the CORS spec (and thus this filter).
-You should continue securing your API in other ways using proper authentication and authorization.
+[CAUTION]
+====
+If a proxy server sitting between the client and Repose rewrites the `Host` without updating the `X-Forwarded-Host` header with the original value,
+Repose will not be able to correctly identify same-origin requests coming from that client.
```
Mario-Lopez: Bah!  I had been putting a new line after the comma before, but in this changeset, I started undoing that since I think we decided one sentence per line, no more, no less.  To make it easy, I went line by line looking for lines starting with lowercase letters.  I missed this line because it started with "Repose".  I will put it on the previous line.

Mario-Lopez: Of note, searching for `,\n` in IntelliJ works when you enable regex.  That would have saved me a couple of minutes.

dmnjohns:
Documentation also looks good. 👍

Mario-Lopez:
Ok, everything should be addressed pending QE review.

PR #1633, opened 2016-12-05
REP-4764 Fixing sendError method in HttpServletResponseWrapper

dmnjohns:


repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -465,6 +498,13 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
       originalResponse.flushBuffer()
     }
 
+    if (sentError) {
```
adrianjgeorge: I think this needs to be higher, because as it stands you've potentially already committed the other body.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -86,6 +86,7 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
   private var reason: Option[String] = None
   private var committed: Boolean = false
   private var flushedBuffer: Boolean = false
+  private var sentError: Boolean = false
```
adrianjgeorge: this needs to be public or have a getter

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: Ah. So the thing is, since both `flushBuffer` and `sendError` commit the response, both flags should not be set at the same time. However, if a user calls `uncommit` between calling those methods, you could end up in that state. I spent a bunch of time trying to figure out all of the possibilities, but apparently I missed that one.

Since both of these calls commit the underlying response, I'll just use an if/else. I'll give priority to `sendError` since it can be explicitly unset (and should be if the intent is not to send an error but just to flush the buffer).

adrianjgeorge: Or I'm missing something right now.

dmnjohns: `writeBody` actually only writes the bytes from the buffer output stream to the underlying output stream. Technically, I suppose, if the underlying buffer filled up those bytes could be written all the way out to the client (although, not actually, because we buffer at the `PowerFilter` level anyway and Repose would run out of memory before writing those bytes to the client... which is another issue entirely). Otherwise, `sendError` should, by contract, clear the buffer. So calling `sendError` on the underlying response should negate that `commit` you see here.

I figured this was safe to do since if you were dealing with a raw container response, I think you would have the same issue? What happens if you write a bunch to the body of a Jetty response and then call `sendError`?

adrianjgeorge: I think you are fine. I missed where commit was our method not part of the original contract, so i was mentally applying the meaning of commit of the response to the stream contract.

PR #1638, opened 2016-12-19
REP-4841 Added an additional GUID for following the Tracing header's value

wdschei:
This insures a truly unique value while still retaining the tracing header's value as part of the UAE's key.

### DIFF
--- a/repose-aggregator/components/filters/herp-filter/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
+++ b/repose-aggregator/components/filters/herp-filter/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
@@ -156,7 +156,7 @@ class HerpFilter @Inject()(configurationService: ConfigurationService,
       "responseCode" -> httpServletResponse.getStatus,
       "responseMessage" -> Try(HttpStatus.valueOf(httpServletResponse.getStatus).name).getOrElse("UNKNOWN"),
       "guid" -> Option(stripHeaderParams(TracingHeaderHelper.getTraceGuid(httpServletRequest.getHeader(CommonHttpHeader.TRACE_GUID.toString))))
-        .getOrElse(java.util.UUID.randomUUID.toString),
+        .getOrElse("").concat(":").concat(java.util.UUID.randomUUID.toString),
       "serviceCode" -> serviceCode,
       "region" -> region,
       "dataCenter" -> dataCenter,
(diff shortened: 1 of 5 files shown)


repose-aggregator/components/filters/herp-filter/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
```diff
@@ -156,7 +156,7 @@ class HerpFilter @Inject()(configurationService: ConfigurationService,
       "responseCode" -> httpServletResponse.getStatus,
       "responseMessage" -> Try(HttpStatus.valueOf(httpServletResponse.getStatus).name).getOrElse("UNKNOWN"),
       "guid" -> Option(stripHeaderParams(TracingHeaderHelper.getTraceGuid(httpServletRequest.getHeader(CommonHttpHeader.TRACE_GUID.toString))))
-        .getOrElse(java.util.UUID.randomUUID.toString),
+        .getOrElse("").concat(":").concat(java.util.UUID.randomUUID.toString),
```
adrianjgeorge: I thought we decided to not lead with the colon if we didn't have anything? Maybe we explicitly say we couldn't find the transaction id? So something like `NO_TRANSACTION_ID:{guid}`?

repose-aggregator/components/filters/herp-filter/src/test/scala/org/openrepose/filters/herp/HerpFilterTest.scala
```diff
@@ -138,7 +139,7 @@ class HerpFilterTest extends FunSpec with BeforeAndAfterEach with Matchers with
       // then:
       val logEvents = listAppenderPre.getEvents
       logEvents.size shouldBe 1
-      logEvents.get(0).getMessage.getFormattedMessage should include regex "\"ServiceCode\" : \"\".*\"URL\" : \"\""
+      logEvents.get(0).getMessage.getFormattedMessage should include regex """"ServiceCode" : "".*"URL" : """""
```
adrianjgeorge: SO many quotes. Nothing wrong , it's just funny.

adrianjgeorge:
Of course now it needs a test to prove that's what it's supposed to do.

adrianjgeorge:
Looks good, I think I'll leave the QE task though so we can get another set of eyes on it.

wdschei:
I was still updating the test. It is there now.

PR #1639, opened 2016-12-20
REP-4756 Clearing Sonar violations.

wdschei:


adrianjgeorge:
I think i'd mostly suppress these string violations. I feel like they make the intent harder to spot not easier. Anyone else have an opinion?

Mario-Lopez:
I agree about suppressing most of the string violations.  It doesn't really help anything and adds an extra step to see what's going on in the code for no reason I can think of.  Plus, I'm pretty sure the Java compiler will already consolidate repeated string literals to one string in the class.

I like the MimeType update, though.  I would also fix the repeated "text" string, too.

Mario-Lopez:
The test failures are because I killed those Jenkins jobs.

PR #1640, opened 2016-12-29
REP-4771 Automatically Publish Latest Tag to Dockerhub

dmnjohns:


repose-aggregator/artifacts/docker/build.gradle
```diff
@@ -33,21 +37,89 @@ docker {
...
-project.tasks.getByPath(':release').dependsOn pushImage
+static int compareVersions(String v1, String v2) {
+    List<String> v1Parts = v1.tokenize('.')
+    List<String> v2Parts = v2.tokenize('.')
+
+    int minLen = Math.min(v1Parts.size(), v2Parts.size())
+
+    // Compare each number from the versions until either the numbers do not match,
+    // or the end of a version string is hit.
+    for (int i = 0; i < minLen; ++i) {
```
adrianjgeorge: why prefix ++?

dmnjohns: No particular reason. I actually got into the habit of preferring the prefix over the suffix notation when I was writing stuff in C++. If the logic doesn't dictate which to use, the prefix notation is ever-so-slightly more performant. See:
http://stackoverflow.com/a/1813008

adrianjgeorge: Also, to be clear, i'm not that worried about it. I was just wondering why.

PR #1643, opened 2017-01-06
REP-4886 SAML Orchestration

adrianjgeorge:


adrianjgeorge:
Config ready for another look.

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/resources/META-INF/schema/examples/saml-policy.cfg.xml
```diff
@@ -1,9 +1,20 @@
...
+    </policy-bypass-issuers>
+    <policy-acquisition>
+        <keystone-credentials uri="http://keystone.somewhere.com"
+                              username="aUsername"
+                              password="somePassword"
+                              connection-pool-id="default"/>
+        <policy-endpoint uri="http://keystone.somewhere.com"/>
+        <cache ttl="300" atom-feed-id="identity-policy"/>
+    </policy-acquisition>
+    <signature-credentials keystore="thing.ks"
```
Mario-Lopez: It might be clearer if the example was `thing.jks`.

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
```diff
@@ -21,21 +21,184 @@
...
+      val request = new HttpServletRequestWrapper(servletRequest.asInstanceOf[HttpServletRequest])
+      var response = servletResponse.asInstanceOf[HttpServletResponse]
+      val samlResponse = decodeResponse(request)
+      val rawDocument = readToDom(samlResponse)
+
+      val version = determineVersion(rawDocument)
+      val finalDocument = version match {
+        case 1 =>
+          request.addHeader("Identity-API-Version", "1.0")
+          rawDocument
```
adrianjgeorge: i considered calling do filter twice, once here, and once in the other flow, but i don't know. I like having only one way into the chain, but having to do work afterwards in the one flow really makes consider going the other way. Thoughts?

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
```diff
@@ -21,21 +21,184 @@
...
+  }
+
+  /**
+    * Gets the SAMl response from the post encoded body, and decodes it into a plain String.
+    *
+    * @param request the servlet request that has an encoded body
+    * @return the decoded saml response
+    * @throws SamlPolicyException if decoding fails
+    */
+  def decodeResponse(request: HttpServletRequest): String = ???
```
Mario-Lopez: This method name keeps throwing me off.  Could we name it to `decodeSamlResponse`?

adrianjgeorge: Sure thing

Mario-Lopez: Also, I don't know if `String` is the best thing to return here.  Base64 decoding will return a `byte[]` (which you can then turn into a `String`, of course), but could we skip that step to get to a DOM (and maybe also avoid potential encoding issues as well)?

Mario-Lopez: To be clear, I'm wondering if we should return something else (maybe a `byte[]`, maybe a stream of some sort) that the `readToDom` method would take as a parameter.

adrianjgeorge: I'm good either way, i just assumed you had a string coming out of the methods you were calling

adrianjgeorge: will scala seamlessly convert from java `byte[]` to scala `Array[byte]`?

adrianjgeorge: Might be better off with an `InputStream`

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
```diff
@@ -21,21 +21,184 @@
...
+        case 2 =>
+          response = new HttpServletResponseWrapper(response, ResponseMode.PASSTHROUGH, ResponseMode.MUTABLE)
+          request.addHeader("Identity-API-Version", "2.0")
+          val issuer = validateResponseAndGetIssuer(rawDocument)
+          val translatedDocument = translateResponse(rawDocument, cache.get(issuer))
+          signResponse(translatedDocument)
+      }
+      val inputStream = convertDocumentToStream(finalDocument)
+
+      request.replaceHeader(CONTENT_TYPE.toString, MediaType.APPLICATION_XML)
```
dmnjohns: Should the `Content-Length` header also be replaced? Decoding the response will likely change that value, and if I remember correctly, a request requires either the `Content-Length` header or `chunked` encoding.

Not sure how easy that will be since you're dealing in a DOM then a stream.

adrianjgeorge: I agree we should probably change account for the size, that means the ServletInputStream implementation is going to have to be smart about what it returns from available. Chunked encoding might be the smarter way to go.

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
Mario-Lopez: Looks legit.
```
import java.io.ByteArrayInputStream
import org.apache.commons.codec.binary.Base64

val test = Base64.decodeBase64("YmFuYW5hcGhvbmU=")
test.isInstanceOf[Array[Byte]]

val testString = new String(test)
val testInputStream = new ByteArrayInputStream(test)
```

Output:
```
import java.io.ByteArrayInputStream
import org.apache.commons.codec.binary.Base64

test: Array[Byte] = Array(98, 97, 110, 97, 110, 97, 112, 104, 111, 110, 101)
res0: Boolean = true

testString: String = bananaphone
testInputStream: java.io.ByteArrayInputStream = java.io.ByteArrayInputStream@4e1ce6e7
```

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
dmnjohns: If you did go the other way, wouldn't you also have to duplicate all the header manipulation stuff? That seems like a fair amount of duplication all told. I like the way you did it, personally.

If we were going to have to support even more versions, I might consider other approaches. Namely, creating a new type with functions to process the request and response. Then you only end up doing the version check once. That doesn't really matter right now, though.

adrianjgeorge: Yeah, there was only one header being monkeyed with before, and i could still potentially move it up higher. But it's fine where it is for now.

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
```diff
@@ -20,22 +20,188 @@
...
+    * Stores the configuration and marks the filter as initialized.
+    * I'm going to have to initialize the cache and atom feed listener here, but it's not in the critical path for the moment.
+    *
+    * @param newConfiguration
+    */
+  override def configurationUpdated(newConfiguration: SamlPolicyConfig): Unit = super.configurationUpdated(newConfiguration)
+}
+
+case class SamlPolicyException(statusCode: Int, message: String, cause: Throwable) extends Exception(message, cause) {
+  def this(statusCode: Integer, message: String) = this(statusCode, message, null)
```
Mario-Lopez: Unnecessary boxing of Integer?

repose-aggregator/components/filters/saml-policy-translation-filter/src/test/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilterTest.scala
```diff
@@ -49,4 +54,48 @@ class SamlPolicyTranslationFilterTest extends FunSpec with BeforeAndAfterEach wi
       verify(chain).doFilter(request, response)
     }
   }
+
+  describe("decodeResponse") {
```
Mario-Lopez: `decodeSamlResponse`

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
```diff
@@ -20,22 +20,188 @@
...
+
+      request.replaceHeader(CONTENT_TYPE.toString, MediaType.APPLICATION_XML)
+      request.removeHeader(CONTENT_LENGTH.toString)
+      request.replaceHeader("Transfer-Encoding", "chunked")
+
+      chain.doFilter(new HttpServletRequestWrapper(request, inputStream), response)
+
+      if (version == 2) {
+        doResponseStuff()
+        response.asInstanceOf[HttpServletResponseWrapper].commitToResponse()
```
Mario-Lopez: It's unfortunate we have to cast a variable we know will always be one thing (when it's used) and that it's declared using `var`, but I don't know what the alternative would be with all of the scoping since we're calling `doFilter` the way we are (to avoid a bunch of duplication, I know, I know...).  ¯\_(ツ)_/¯

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
adrianjgeorge: Yep, i thought i'd caught this already

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
```diff
@@ -20,22 +20,188 @@
...
+    * Stores the configuration and marks the filter as initialized.
+    * I'm going to have to initialize the cache and atom feed listener here, but it's not in the critical path for the moment.
+    *
+    * @param newConfiguration
+    */
+  override def configurationUpdated(newConfiguration: SamlPolicyConfig): Unit = super.configurationUpdated(newConfiguration)
+}
+
+case class SamlPolicyException(statusCode: Int, message: String, cause: Throwable) extends Exception(message, cause) {
+  def this(statusCode: Int, message: String) = this(statusCode, message, null)
```
dmnjohns: Rather than define another constructor, you may just want to provide a default for `cause` in the case class. That is:
`case class SamlPolicyException(statusCode: Int, message: String, cause: Throwable = null) extends Exception(message, cause)`

I think that still captures what you want to do, but it generates the appropriate `apply` method allowing users to write:
```
SamlPolicyException(200, "OK", new Exception())
SamlPolicyException(200, "OK")
```

Whereas the way it stands, a user must write:
```
SamlPolicyException(200, "OK", new Exception())
new SamlPolicyException(200, "OK")
```

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
```diff
@@ -20,22 +20,201 @@
...
 
+import java.io.InputStream
 import javax.inject.{Inject, Named}
-import javax.servlet.{FilterChain, ServletRequest, ServletResponse}
+import javax.servlet.http.HttpServletResponse.SC_INTERNAL_SERVER_ERROR
+import javax.servlet.http.{HttpServletRequest, HttpServletResponse}
+import javax.servlet.{FilterChain, ServletInputStream, ServletRequest, ServletResponse}
+import javax.ws.rs.core.MediaType
 
+import com.google.common.cache.{Cache, LoadingCache}
```
dmnjohns: The `Cache` class does not seem strictly necessary at this point. The import seems to be unused.

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
```diff
@@ -20,22 +20,210 @@
...
+      atomFeedService.unregisterListener(feedId.get)
+      feedId = None
+    }
+
+    if (feedId.isEmpty && requestedFeedId.nonEmpty) {
+      feedId = Option(atomFeedService.registerListener(requestedFeedId.get, this))
+    }
+
+    if (Option(configuration).map(_.getPolicyAcquisition.getCache.getTtl) != Option(newConfiguration.getPolicyAcquisition.getCache.getTtl)) {
+      cache = CacheBuilder.newBuilder()
```
adrianjgeorge: Anybody got a good idea on how to test this? My best ideas are to expose the cache reflectively to check if the if is working correctly, and do an inheritance of this class to replace getPolicy to test that the ttl and get are working. These are terrible, but as i said my "best" ideas.

adrianjgeorge: so slightly less worse, for the second part i wait till the method is implemented and i watch for the appropriate calls when i do a get on the cache which i now have since i hacked it out reflectively.

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
dmnjohns: This got hidden for some reason. ☹️ 

Just calling attention to it again.

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
```diff
@@ -20,22 +20,225 @@
...
+                            override def load(key: String): XsltExecutable = getPolicy(key)
+                          })
+    }
+
+    super.configurationUpdated(newConfiguration)
+  }
+}
+
+case class SamlPolicyException(statusCode: Int, message: String, cause: Throwable) extends Exception(message, cause) {
+  def this(statusCode: Int, message: String) = this(statusCode, message, null)
```
Mario-Lopez: Interface is still inconsistent.  The current options are:

```
SamlPolicyException(500, "Nope", new Exception())
new SamlPolicyException(500, "Nope")
```

Should be:
```
case class SamlPolicyException(statusCode: Int, message: String, cause: Throwable = null) extends Exception(message, cause)
```

Which would let you:
```
SamlPolicyException(500, "Nope", new Exception())
SamlPolicyException(500, "Nope")
```

repose-aggregator/src/docs/asciidoc/filters/saml-policy.adoc
```diff
@@ -15,15 +15,19 @@ It'll attempt to locate a policy to use for the translation based on the asserti
...
+None
 
 === Request headers created
+`Content-Type` - Will be changed to `application/xml`
+`Content-Length` - Will be removed if present
+`Transfer-Encoding` - Will be set to `chunked`
 `Identity-API-Version` - `1.0` if the assertion issuer is in the configured list and `2.0` otherwise.
 
 === Request body changes
+The body will be decoded and replaced with and xml version.
```
Mario-Lopez: "with an xml"

repose-aggregator/src/docs/asciidoc/filters/saml-policy.adoc
```diff
@@ -15,15 +15,19 @@ It'll attempt to locate a policy to use for the translation based on the asserti
...
+`Content-Type` - Needs a value of `application/x-www-form-urlencoded`
 
 === Required preceding filters
-If using the https://en.wikipedia.org/wiki/SAML_2.0#HTTP_POST_Binding[post form binding], you'll need to use the <<saml-binding.adoc#,SAML Binding Translation Filter>> to decode the request first.
+None
 
 === Request headers created
+`Content-Type` - Will be changed to `application/xml`
+`Content-Length` - Will be removed if present
+`Transfer-Encoding` - Will be set to `chunked`
```
Mario-Lopez: Should we put these in a table or maybe a list?  Right now they're in a paragraph, and it doesn't look that great.  I'm not sure if you'd also want to update the "Required headers" section for consistency, although it's only one header under that section.

repose-aggregator/src/docs/asciidoc/filters/saml-policy.adoc
```diff
@@ -41,20 +45,49 @@ None.
 `401` - If a policy that matches the assertion issuer can not be found.
```
Mario-Lopez: Should make this a table.

repose-aggregator/src/docs/asciidoc/filters/saml-policy.adoc
```diff
@@ -41,20 +45,49 @@ None.
...
+    </policy-acquisition>
+    <signature-credentials keystore-filename="thing.jks" <!--10-->
+                           keystore-password="banana" <!--11-->
+                           key-name="thingy" <!--12-->
+                           key-password="phone"/> <!--13-->
+</saml-policy>
 ----
-<1> Specifies foo.
-<2> Specifies bar.
+<1>  The list of 1.0 issuers.
```
Mario-Lopez: Should you add that it's optional?

repose-aggregator/src/docs/asciidoc/filters/saml-policy.adoc
```diff
@@ -41,20 +45,49 @@ None.
...
+<2>  The keystone endpoint to get an authentication token from.
+<3>  The username of the user to use for policy acquisition.
+<4>  The password of the user to use for policy acquisition.
+<5>  The id of the connection pool to use when making the requests.
+     It's optional and will use the default if unspecified.
+<6>  The endpoint to use for policy acquisition.
+<7>  The connection pool to use when acquiring the policy.
+     It's optional and will use the default if unspecified.
+<8>  The time to live in the cache for policies.
+<9>  The atom feed id of an atom feed that contains policy events to be used for cache evictions.
```
Mario-Lopez: Should you add that it's optional?

repose-aggregator/src/docs/asciidoc/filters/saml-policy.adoc
```diff
@@ -41,20 +45,49 @@ None.
...
+<10> The location of the keystore that contains the certificates for signing the SAMl response.
```
Mario-Lopez: s/SAMl/SAML/

repose-aggregator/src/docs/asciidoc/filters/saml-policy.adoc
```diff
@@ -41,20 +45,49 @@ None.
...
+<11> The password fir using the keystore.
```
Mario-Lopez: s/fir/for/

repose-aggregator/src/docs/asciidoc/filters/saml-policy.adoc
```diff
@@ -41,20 +45,49 @@ None.
...
+<12> The name the certificate is under in the keystore.
+<13> The password for the certificate inthe keystore.
```
Mario-Lopez: s/inthe/in the/

Mario-Lopez:
Only build failure is the functional test which is expected.
I have one comment on the Exception case class and the rest are comments on the documentation.
Other than that, looks good.

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/resources/META-INF/schema/examples/saml-policy.cfg.xml
```diff
@@ -1,9 +1,21 @@
...
+                              connection-pool-id="default"/>
+        <policy-endpoint uri="http://keystone.somewhere.com"
+                         connection-pool-id="different"/>
+        <cache ttl="300" atom-feed-id="identity-policy"/>
+    </policy-acquisition>
+    <signature-credentials keystore-filename="thing.jks"
+                           keystore-password="banana"
+                           key-name="thingy"
+                           key-password="phone"/>
 </saml-policy>
```
Mario-Lopez: We might want to stick a copy of this config file in the appropriate artifacts directory before we forget.

repose-aggregator/src/docs/asciidoc/filters/saml-policy.adoc
adrianjgeorge: Weird when i previewed it before they looked nice. I fix.

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/resources/META-INF/schema/examples/saml-policy.cfg.xml
adrianjgeorge: This is a pretty targeted config, i don't think we want to ship it with everything else.

Mario-Lopez:
REP-4804 missed updating the dependencies list:
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/artifacts/extensions-filter-bundle/src/config/docs/DEPENDENCIES.txt

dmnjohns:
I handled updating the dependencies list, since I missed it.

dmnjohns:
Retest this, please.

dmnjohns:
All tests passed, so I'm merging. 👍

PR #1646, opened 2017-01-11
REP-4898 decodeSamlResponse implementation

dmnjohns:


repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
```diff
@@ -97,13 +100,25 @@ class SamlPolicyTranslationFilter @Inject()(configurationService: ConfigurationS
...
+    * Gets the SAMl response from the post encoded body, and decodes it.
     *
     * @param request the servlet request that has an encoded body
     * @return the decoded saml response
     * @throws SamlPolicyException if decoding fails
     */
-  def decodeSamlResponse(request: HttpServletRequest): InputStream = ???
+  def decodeSamlResponse(request: HttpServletRequest): InputStream = {
+    try {
+      Option(request.getParameter("SAMLResponse"))
```
dmnjohns: So this is interesting. `getParameter` actually returns a string, but the content of `SAMLResponse` could be in the body. How does `getParameter` know the encoding of that body? Well, `application/x-www-form-urlencoded` does not actually define the `charset` parameter. In fact, the content is always encoded as `US-ASCII`. To reason about that, consider that the same content could be transmitted over HTTP as a query parameter. Therefore, it's always safe and correct for this method to return a `String` (assuming a correct implementation of `HttpServletRequest`).

Now you might be wondering how exotic characters can be sent using HTML's `application/x-www-form-urlencoded` over HTTP. Well, a `_charset_` parameter can be included as a parameter and can specify the encoding of the content after decoded using `US-ASCII`.

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
```diff
@@ -97,13 +100,25 @@ class SamlPolicyTranslationFilter @Inject()(configurationService: ConfigurationS
...
+        .map(Base64.getDecoder.decode)
```
dmnjohns: Also interesting, Java's `Base64` decoder that I get here uses `ISO-8859-1` to decode when passed a `String`. That works just fine for us, because `US-ASCII` is a subset of `ISO-8859-1`.

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
Mario-Lopez: Will the `getParameter` method properly handle the URL encoding of the value?

For example, would this work if the request coming in the door had parameters that looked like this:

```
SAMLResponse=PHNhbWxwOkxvZ291dFJlc3BvbnNlIHhtbG5zOnNhbWxwPSJ1cm46b2FzaXM6bmFtZXM6dGM6U0FNTDoyLjA6cHJvdG9jb2wiIHhtbG5zPSJ1cm46b2FzaXM6bmFtZXM6dGM6U0FNTDoyLjA6YXNzZXJ0aW9uIg0KICAgIElEPSJiMDczMGQyMWI2MjgxMTBkOGI3ZTAwNDAwNWIxM2EyYiIgSW5SZXNwb25zZVRvPSJkMmI3YzM4OGNlYzM2ZmE3YzM5YzI4ZmQyOTg2NDRhOCINCiAgICBJc3N1ZUluc3RhbnQ9IjIwMDQtMDEtMjFUMTk6MDA6NDlaIiBWZXJzaW9uPSIyLjAiPg0KICAgIDxJc3N1ZXI%2BaHR0cHM6Ly9TZXJ2aWNlUHJvdmlkZXIuY29tL1NBTUw8L0lzc3Vlcj4NCiAgICA8c2FtbHA6U3RhdHVzPg0KICAgICAgICA8c2FtbHA6U3RhdHVzQ29kZSBWYWx1ZT0idXJuOm9hc2lzOm5hbWVzOnRjOlNBTUw6Mi4wOnN0YXR1czpTdWNjZXNzIi8%2BDQogICAgPC9zYW1scDpTdGF0dXM%2BDQo8L3NhbWxwOkxvZ291dFJlc3BvbnNlPg%3D%3D&RelayState=token
```

Mario-Lopez: Of note, I did a quick google and checked the JavaDocs of the method, and I didn't see an answer.  This may just have to wait until I write a functional test for it.

dmnjohns: I kind of assume so. I certainly hope so. It's definitely a good thing to test.

Mario-Lopez: I would think so, too, but I would have figured it would mention it in the JavaDoc.  Either it's so obvious it would it's not worth mentioning, or it isn't. &nbsp;&nbsp;&nbsp; Guess we'll see.

wdschei:
When the build finishes and the conflicts are resolved I'll merge.

dmnjohns:
As far as the test failures go, `SamlPolicyTranslationTest` is expected to fail for the moment, and the other two... well, I don't know why those failed. They are passing locally, and aren't even related to the code changes.

 I'll let Jenkins run through the job again, and if those tests fail again, I'll go digging. I'm hoping things work a little better this time.

wdschei:
That works.

dmnjohns:
That looks better. Now only the new Saml test is failing!

wdschei:
Perfect. I will merge in the AM.

PR #1647, opened 2017-01-13
REP-4804 Use SAML Library to do Translation

dmnjohns:
I'm just all around not happy with this. The method itself is basically a one-line call through to the SAML library with some exception translation wrapped around it. The SAML library is already expected to be tested, so I didn't want to write tests proving that method call works, but I also cannot easily mock it since it's a Scala `object`. PowerMock does not play nice with Scala, and there does not seem to be a Scalatest native mocking framework with the power I would need. So in the end, I need a valid `Document` and `XsltExecutable`. Those are a bit of a pain to create, and in the end, I didn't care about the content, so I wrote dumb placeholders. I actually started by stealing a test case from the `AttributeMapper` library, but going from the Identity mapping `String` to a `XsltExecutable` had me calling `AttributeMapper.generateXSLExec(...)` which would not generate a failing XSLT. I decided that if I was going to have to write custom XSLT, I would just make them simple, stupid ones that always fail/succeed. Same goes for the SAML Response, actually.

So basically, the code I wrote just translates exceptions, but it was incredibly aggravating to set up tests for that.

Please, if you have suggestions, let me know.

### DIFF
--- a/repose-aggregator/components/filters/saml-policy-translation-filter/src/test/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilterTest.scala
+++ b/repose-aggregator/components/filters/saml-policy-translation-filter/src/test/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilterTest.scala
@@ -21,10 +21,14 @@
 package org.openrepose.filters.samlpolicy
 
 import java.util.Base64
+import java.io.StringReader
 import javax.servlet.FilterChain
 import javax.servlet.http.HttpServletResponse.SC_BAD_REQUEST
 import javax.servlet.http.{HttpServletRequest, HttpServletResponse}
+import javax.xml.parsers.DocumentBuilderFactory
+import javax.xml.transform.stream.StreamSource
 
+import net.sf.saxon.s9api.Processor
 import org.junit.runner.RunWith
 import org.mockito.Mockito._
 import org.mockito.{Matchers => MM}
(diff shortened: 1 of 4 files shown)


repose-aggregator/components/filters/saml-policy-translation-filter/src/test/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilterTest.scala
```diff
@@ -114,7 +119,45 @@ class SamlPolicyTranslationFilterTest extends FunSpec with BeforeAndAfterEach wi
...
+      .compile(new StreamSource(new StringReader(workingXslt)))
+
+    it("should throw a SamlPolicyException(400) if the translation fails") {
+      intercept[SamlPolicyException] {
+        filter.translateResponse(document, brokenXsltExec)
+      }.statusCode shouldEqual SC_BAD_REQUEST
+    }
+
+    it("should return the translated document without throwing an exception") {
+      filter.translateResponse(document, workingXsltExec)
```
adrianjgeorge: Maybe check that you have different documents? Or better yet, have the xslt actually do something, and look for that in the resulting document.

dmnjohns: Checking for a different documents is legit.

I thought about verifying the content of the translated document, but that seems like more of an integration test than a unit test. All of the code that actually performs the translation is in the library, the code I wrote just manages the exception translation (including adding the response code). 🤷‍♂️

dmnjohns:
This is just a reminder to myself to update any documentation necessary with the response code here.

repose-aggregator/components/filters/saml-policy-translation-filter/src/test/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilterTest.scala
adrianjgeorge: fair enough

PR #1655, opened 2017-01-25
REP-4811 Feed Invalidation

wdschei:


repose-aggregator/core/repose-core-api/src/main/scala/org/openrepose/core/filter/AbstractConfiguredFilter.scala
```diff
@@ -74,38 +75,42 @@ abstract class AbstractConfiguredFilter[T: ClassTag](val configurationService: C
...
   }
 
   /**
+    * Called after the configuration reference is updated.
+    *
+    * @param oldConfiguration
+    */
+  def doConfigurationUpdated(oldConfiguration: T): Unit = {
+    logger.trace("{} Default doConfigurationUpdated ...", this.getClass.getSimpleName)
+  }
```
dmnjohns: Adding a comment for history.

This method makes sense to me for one purpose -- to provide seamless extension of the processing performed by `configurationUpdated`. Specifically, this method means that `configurationUpdated` need not be overridden, so you don't encounter the logging issue where `configurationUpdated` is overridden in the subclass of `AbstractConfiguredFilter` then `super.configurationUpdated` is called.

However, I'm not sure I see the need to store a local variable above and pass the `oldConfiguration` to this method. If you replaced the temporary local variable line with the `doConfigurationUpdated(...)` line, the `doConfigurationUpdated` would still have a reference to both the old and the new configurations. In fact, passing the new configuration as an argument makes more sense to me because it is consistent with the configuration service API, and it is the new configuration is the catalyst for the method invocation, so it seems natural that the new configuration would be parameter to `configurationUpdated`.

Actually, I might also make this method do nothing if you really want seamless extension. Why log the occurrence of literally nothing? Leave meaningful logging to the thing doing meaningful work.

wdschei: Inverted to be consistent with the abstract parent's method.

repose-aggregator/core/repose-core-api/src/main/scala/org/openrepose/core/filter/AbstractConfiguredFilter.scala
```diff
@@ -119,14 +123,31 @@ abstract class AbstractConfiguredFilter[T: ClassTag](val configurationService: C
...
+    configurationService.unsubscribeFrom(configFile, this)
+    doDestroy()
+    logger.trace("{} destroyed.", this.getClass.getSimpleName)
+  }
+
+  /**
+    * Called immediately after the configFile is unsubscribed from.
+    */
+  def doDestroy(): Unit = {
+    logger.trace("{} Default postDestroy ...", this.getClass.getSimpleName)
```
dmnjohns: I would also have this do nothing for the same reason as the other. I mean, it's a trace log, so not overly offensive... I just don't know why it's valuable, either.

dmnjohns:
I left a couple of comments about logging, but I honestly don't care what comes from them.

Overall, I like this. Clean, simple, functional. 👍

repose-aggregator/core/repose-core-api/src/main/scala/org/openrepose/core/filter/AbstractConfiguredFilter.scala
wdschei: The main value was to see in the logs that it was a default being called and not something actually doing anything. Which was the reason for making it a trace as you noted.

dmnjohns: Right, but does anyone actually care if the default is called, considering it doesn't do anything?

Also consider that if a subclass did override this method, that subclass could actually log the exact same message using the exact same logger. So having this message in the logs is not strictly proof that a subclass did not override this method.

dmnjohns: I was overruled on this one. 😢

dmnjohns:
Tests (other than the saml tests) passed. Merging!

PR #1656, opened 2017-01-26
REP-4810 Get Policy from Identity

dmnjohns:


adrianjgeorge:
ohh man you have changes that overlap @wdschei. that's going to get ugly

dmnjohns:
Yeah, I've had changes that overlap every SAML story that has been completed so far. I've gotten pretty good at rebasing this stuff. I don't mind having this merged after @wdschei's stuff.

There were a few things that bugged me, so I changed them even though they weren't strictly part of the `getPolicy` work. Those changes did, however, make a few things nicer for me when I was doing the actual `getPolicy` work.

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlIdentityClient.scala
```diff
@@ -0,0 +1,262 @@
...
+              case f: Exception =>
+                throw GenericIdentityException("Token could not be parsed from response from Identity", f)
+            }
+          case SC_REQUEST_ENTITY_TOO_LARGE | SC_TOO_MANY_REQUESTS =>
+            Failure(OverLimitException(buildRetryValue(serviceClientResponse), "Rate limited when getting token"))
+          case statusCode =>
+            Failure(UnexpectedStatusCodeException(statusCode, "Unexpected response from Identity"))
+        }
+      case Failure(f) =>
+        Failure(GenericIdentityException("Failure communicating with Identity when getting token", f))
```
adrianjgeorge: I assume you are handling all these exceptions further down, but if not, you need to be.

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
```diff
@@ -77,20 +84,36 @@ class SamlPolicyTranslationFilter @Inject()(configurationService: ConfigurationS
...
+    )
+  }
+
   override def doWork(servletRequest: ServletRequest, servletResponse: ServletResponse, chain: FilterChain): Unit = {
     try {
       val request = new HttpServletRequestWrapper(servletRequest.asInstanceOf[HttpServletRequest])
       var response = servletResponse.asInstanceOf[HttpServletResponse]
       val samlResponse = decodeSamlResponse(request)
       val rawDocument = readToDom(samlResponse)
+      val traceId = Option(request.getHeader(CommonHttpHeader.TRACE_GUID.toString)).filter(_ => sendTraceHeader)
```
adrianjgeorge: Took me a second to figure out what you were trying to do with the filter.

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
```diff
@@ -364,29 +450,52 @@ class SamlPolicyTranslationFilter @Inject()(configurationService: ConfigurationS
...
+    )
   }
 
   /**
     * Unsubscribe from the Atom Feed service.
     */
   override def doDestroy(): Unit = {
+    configurationService.unsubscribeFrom(SystemModelConfig, SystemModelConfigListener)
+
     if (feedId.nonEmpty) {
```
adrianjgeorge: Not your code, but this totally doesn't need to be an if. it can just be a foreach

dmnjohns: Legit. I'll update it anyway.

wdschei: Yep. That was mine. Thanks for the assist.

repose-aggregator/components/filters/saml-policy-translation-filter/src/test/scala/org/openrepose/filters/samlpolicy/SamlIdentityClientTest.scala
```diff
@@ -0,0 +1,939 @@
...
+
+  override def beforeEach(): Unit = {
+    akkaServiceClient = mock[AkkaServiceClient]
+    akkaServiceClientFactory = mock[AkkaServiceClientFactory]
+
+    samlPolicyProvider = new SamlIdentityClient(akkaServiceClientFactory)
+
+    when(akkaServiceClientFactory.newAkkaServiceClient())
+      .thenReturn(akkaServiceClient)
+    when(akkaServiceClientFactory.newAkkaServiceClient(MM.anyString()))
```
adrianjgeorge: I'm a terrible person. I do this same aliasing, but it just occurred to me that we could instead alias `anyString` to `neString`. Please don't do that, because it's terrible. I feel like its the code equivalent of a pun.

repose-aggregator/components/filters/saml-policy-translation-filter/src/test/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilterTest.scala
```diff
@@ -754,37 +1147,37 @@ class SamlPolicyTranslationFilterTest extends FunSpec with BeforeAndAfterEach wi
     }
 
     it("should initialize the cache when given a config") {
-      ReflectionTestUtils.getField(filter, "cache") shouldBe null
+      ReflectionTestUtils.getField(filter, "org$openrepose$filters$samlpolicy$SamlPolicyTranslationFilter$$policyCache") shouldBe null
```
adrianjgeorge: I'm scared this will randomly start failing again when the compiler decides it wants to change the name again.

dmnjohns: You and me both... I may poke around at it some more when I get the chance.

I dug around for about an hour and couldn't find the reason for the compiler naming some fields this way. I really want a better solution, but I couldn't come up with one...

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/filter/AbstractConfiguredFilterTest.scala
```diff
@@ -101,6 +101,18 @@ class AbstractConfiguredFilterTest
       verify(configurationService).subscribeTo(anyString(), anyString(), any(classOf[URL]),
         any(classOf[UpdateListener[String]]), any(classOf[Class[String]]))
     }
+
+    it("should call doInit") {
```
adrianjgeorge: could also have checked the logs. That's kind of a pain in the butt though.

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/filter/AbstractConfiguredFilterTest.scala
```diff
@@ -143,6 +167,18 @@ class AbstractConfiguredFilterTest
 
       filter.getInitialized shouldBe true
     }
+
+    it("should call doConfigurationUpdated") {
```
adrianjgeorge: These are pretty basic tests. I'm more concerned with that they are calling in the right spot in the lifecycle versus them being called in general. Does that make sense?

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/filter/AbstractConfiguredFilterTest.scala
dmnjohns: It super is. I went back and forth a bit, but ended on the thought that if these methods do nothing (with the exception of logging), then there isn't really a reason to test that nothing is being done. I suppose the argument would be that logging is something, but honestly, I don't know that it's worth proving the logging call is made. For one, proving that the log line actually appears is testing part of the logging framework. Proving only that the call is made, on the other hand, would require digging around the class a bit to mock out the logger or some such thing.

I can add a test to check that the logger logs the line if you'd like, but I feel pretty meh about it.

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/filter/AbstractConfiguredFilterTest.scala
dmnjohns: Hm... as in being called after running the abstract filter code for the related method? I can certainly add a test for that, but do we actually care? I suppose we might if you want to be able to, say unsubscribe immediately from the configuration that the abstract filter subscribed to. Although in that case, I think a user would probably be better off just overriding the primary (non-extension) method.

I feel like I'm missing something here...

adrianjgeorge: so in the case of the configuration updated method when it is called is super important, because the end of that method is what marks the filter as ready. We have no test making sure that some one doesn't accidentally go and move the call to the new method to after that and potentially fubar up the filter running.

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlIdentityClient.scala
dmnjohns: I handle most cases, but there are some that I didn't handle. I was going to let the `doWork` method handle those and a return a 500. One example would be a connection timeout in Akka. The result is that the exception chain has all of the information, and much of that will be logged in the warning that accompanies the 500. The only potential issue with that approach is that the message sent back to the user is somewhat vague.

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/filter/AbstractConfiguredFilterTest.scala
dmnjohns: Legit. I added a test for that specific case. Would you also like tests for the ordering of the other methods?

repose-aggregator/components/filters/saml-policy-translation-filter/src/test/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilterTest.scala
Mario-Lopez: It's getting this name now because `policyCache` is being used in an anonymous function (what we're passing in to `.foreach(policyCache.invalidate(_))` in the onNewAtomEntry method) which is getting converted into the inner class:
```
public final INNERCLASS org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter$$anonfun$onNewAtomEntry$2
```

If you comment out that line, the field name goes back to normal.

Of note, `feedId` also has the long name now, too, because it's in a foreach in the doDestroy method.  I mention this because you made the comparison between the two at some point.

wdschei: So the foreach loop is the trigger?
That could be why it was originally in an if clause instead.

Mario-Lopez: Well anything that would put it in a function would do it, e.g. using it with `map`, `filter`, `fold`, `reduce`, `sortBy`, etc.

But yes, in this case adding the `foreach` loop was the trigger.

dmnjohns: Huh. The more you know. Well that's kind of annoying, but it should be consistent, at least. I certainly wouldn't want to start using Java branching code just because of the implementation details of the Scala compiler.

Mario-Lopez: I agree with wanting to avoid going back to the Java branching on this one.  Also because it would be 1. for test code 2. that's using reflection.

I wonder if there's a more appropriate reflection library that'll work better with Scala classes.

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/filter/AbstractConfiguredFilterTest.scala
adrianjgeorge: I don't know of any specific needs for the others

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/filter/AbstractConfiguredFilterTest.scala
adrianjgeorge: Yeah, i was mostly worried about the other cases.

PR #1657, opened 2017-01-26
REP-4999 Fix container.cfg.xml whitespace bug

wdschei:


repose-aggregator/core/repose-core/src/main/java/org/openrepose/core/services/deploy/ContainerConfigurationListener.java
```diff
@@ -56,14 +56,11 @@ public synchronized void configurationUpdated(ContainerConfiguration configurati
                 if (ad.getCheckInterval() > 0) {
                     dirWatcher.updateCheckInterval(ad.getCheckInterval());
                 }
-
-                dirWatcher.updateArtifactDirectoryLocation(new File(ad.getValue()));
-
-                deploymentDirectory = new File(dd.getValue());
+                dirWatcher.updateArtifactDirectoryLocation(new File(ad.getValue().trim()));
+                deploymentDirectory = new File(dd.getValue().trim());
```
dmnjohns: It looks like these directories do not, technically, have to be absolute. Since spaces (which `trim()` removes) are valid directory name characters (on Mac OS X, at least), I'm worried that these lines limit support on certain file systems.

While I would be happy to assist the user where possible, in this case doing so limits our versatility. Ultimately, I want to say that using whitespace to format XML text nodes is not a good idea, and it's on the user to make sure they are providing the value they want.

@adrianjgeorge Thoughts?

wdschei: Having a fully qualified path that starts our ends with a space is bad practice on any OS. The `trim()` doesn't effect whitespace in the middle, so it shouldn't limit Repose in any way.

dmnjohns: Well we don't necessarily have a fully qualified path, do we? And sure, of course it's bad practice, but it's also technically valid in some filesystems.

I always struggle when it comes to imposing artificial constraints on our users. Especially when, on the other hand, artificial constraints may provide some convenience to most users.

wdschei: I guess you are correct that it doesn't have to be a fully qualified path, but it should be. We should probably update the docs to say that. The reason being, you can technically start the JVM from any directory as long as all the right arguments are provided. This is not the typical scenario, so some behavior may be assumed, but you know how that old saying goes.

And if the only thing we are preventing someone from doing is starting or ending a file name with a space, then I'm pretty ok with that.

dmnjohns: Fair enough. Having clear documentation about the fully qualified path and handling of leading/trailing spaces would make me happy. Let me know once we've got those? I'll merge this PR then.

PR #1658, opened 2017-01-30
REP-4816 SAML Attribute Mapping Tests

Mario-Lopez:
These are the start of the attribute mapping tests and subsequently the extended attributes testing, the translations going to the request and the response and how the cache may impact all of that.

### DIFF
--- a/repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/samlpolicy/SamlAttributeMappingTest.groovy
+++ b/repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/samlpolicy/SamlAttributeMappingTest.groovy
@@ -1,0 +1,206 @@
+/*
+ * _=_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_=
+ * Repose
+ * _-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-
+ * Copyright (C) 2010 - 2015 Rackspace US, Inc.
+ * _-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/samlpolicy/SamlAttributeMappingTest.groovy
```diff
@@ -0,0 +1,429 @@
...
+
+    @FailsWith(ConditionNotSatisfiedError)
+    def "the saml:response will be translated before being sent to the origin service"() {
+        given: "a mapping policy with a literal value and a path-based value in addition to the standard attributes"
+        def extAttribLiteral = "banana"
+        def extAttribLiteralValue = "phone"
+        def extAttribPath = "adventure"
+        def extAttribPathValue = "Mordor"
+        def mappingPolicy = createMappingJsonWithValues(
+                userExtAttribs: [(extAttribLiteral): extAttribLiteralValue, (extAttribPath): "{0}"],
```
dmnjohns: Having to escape the keys seems so silly. All just so that string keys can be used without quotations. But why would you even want that? The quotations provide a consistent indicator that something is a string. Furthermore, it seems just as likely that you would use a variable for the key as a string, so why prefer the string? This is one of those times where I'm like, man, I'm just not a big fan of Groovy.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/samlpolicy/SamlAttributeMappingTest.groovy
```diff
@@ -0,0 +1,429 @@
...
+                remote: [[path: $/\/saml2p:Response\/saml2:Assertion\/saml2:Subject\/saml2:NameID\/@SPProvidedID/$]])
+
+        and: "a saml:response with a value at the path specified by the mapping policy"
+        def samlIssuer = generateUniqueIssuer()
+        def attribName = "Frodo Baggins"
+        def attribRole = "nova:cool_cat"
+        def attribDomain = "193083"
+        def attribEmail = "frodo.baggins@nowhere.abc"
+        def saml = samlResponse(issuer(samlIssuer) >> status() >> assertion(
```
dmnjohns: Yay function composition! Admittedly, it threw me for a look at first since you have methods creating `Closure`s which are then composed, and the methods are the things that take parameters. Also, I hadn't seen the syntax for this in Groovy yet.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/samlpolicy/SamlAttributeMappingTest.groovy
Mario-Lopez: I do super miss Groovy's string literals when writing Scala.

Mario-Lopez:
By the way, I updated the tests to remove the `@FailsWith` so we'll know which tests are still failing.  I can add that back for now if you want, but I figured we're close enough to done to want the info now.

PR #1660, opened 2017-01-30
REP-4857 SAML filter Modify Response

wdschei:
This includes the `readToDom()` method too since it didn't have a happy home yet.


repose-aggregator/components/filters/saml-policy-translation-filter/src/test/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilterTest.scala
```diff
@@ -184,7 +186,19 @@ class SamlPolicyTranslationFilterTest extends FunSpec with BeforeAndAfterEach wi
...
 
   describe("readToDom") {
-    pending
+    it("should return a Document if the stream is able to be parsed") {
+      val document = filter.readToDom(new BufferedServletInputStream(new ByteArrayInputStream(samlResponseStr.getBytes(UTF_8))))
+      document should not be null
+    }
+
+    it("should throw an exception if the stream is not able to be parsed") {
+      val exception = intercept[SamlPolicyException] {
```
dmnjohns: So this is 100% legit Scalatest. It comes in with all of the default assertions.
http://www.scalatest.org/user_guide/using_assertions

However, I've been using the ScalaTest Matchers DSL to be consistent with our other assertions.
http://www.scalatest.org/user_guide/using_matchers#checkingForExpectedExceptions

My vote is to use the Matchers DSL, but whichever we choose to use, we should at least be consistent within this test class.

adrianjgeorge: Pretty sure we already have a ton of the intercept. if you want to go back and change the others that's fine.

adrianjgeorge: I do like the way the other one looks though, it's mostly cleaner than the ugly intercept way.

adrianjgeorge:
Well, now i'm concerned. Shouldn't this have made some of the tests that were flagged as supposed to be failing pass?

wdschei:
Maybe not if the tests are still failing for other reasons.

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
```diff
@@ -97,6 +99,8 @@ class SamlPolicyTranslationFilter @Inject()(configurationService: ConfigurationS
   private var keyEntry: KeyStore.PrivateKeyEntry = _
   private var keyInfo: KeyInfo = _
   private var legacyIssuers: List[URI] = List.empty
+  private val documentBuilderFactory = DocumentBuilderFactory.newInstance()
+  documentBuilderFactory.setNamespaceAware(true)
```
RackerWilliams: Please consider using XML Parser pool instead of doing this from scratch. 

For a couple of reasons:

1. It's more efficient to reuse a builder factory.
2. You need to take special security issues into mind when parsing XML.  There are attacks that are possible in the normal configuration.
3.  There are potential class loader issues in different deployment if you simply do ```newInstance()``` on a document builder factory  also some versions of ```newInstance()``` scan jars which makes them really slow.


You can find parsing pool here.

https://github.com/rackerlabs/api-checker/blob/master/util/src/main/scala/com/rackspace/com/papi/compenents/checker/util/XMLParserPool.scala#L25

It handles all of the issues above and has been battle tested.

Use ```borrowParser``` to get ```DocumentBuilder``` and ```returnParser``` when you're done -- usually do this in a ```try {} finally {} ``` 

Here's a simple test that  shows the pattern:

https://github.com/rackerlabs/api-checker/blob/master/util/src/test/scala/com/rackspace/com/papi/components/checker/util/XMLParserPoolSuite.scala#L27

wdschei: @RackerWilliams This has now been addressed in #1664.

PR #1662, opened 2017-01-30
Updated url to always point to the matching versioned doc.

adrianjgeorge:


build.gradle
```diff
@@ -101,6 +100,10 @@ allprojects {
             url "https://maven.research.rackspacecloud.com/content/groups/aggregate/"
         }
     }
+
+    ext {
+        docsHome = "http://www.openrepose.org/versions/$version/index.html".toString()
+    }
```
dmnjohns: What is the difference between putting the external property on every project versus putting it on the root project? Out of curiosity.

dmnjohns: Oh, I see. Was this just done to get the version in scope? Does the version not get inherited from the parent? Could that possible be extracted out of the `allprojects` block?

adrianjgeorge: I originally had it above, but version is not yet defined at that point. And for other reasons version can't by in the outer scope, so it has to be down here.

dmnjohns: Fair enough.

PR #1663, opened 2017-01-31
REP-4892 Fix Search Functionality in Documentation

dmnjohns:


adrianjgeorge:
All you needed was to do a bunch of deletes?

dmnjohns:
Heh, yes, in the code. I apparently wrote the stuff for both on-the-fly and custom search when I was looking at this originally. I messed around a bit in the Google Custom Search Engine management, because the generated script pulls stuff from there at runtime, but that's all.

PR #1667, opened 2017-01-31
REP-5019 getParameter Not Returning SAMLResponse

dmnjohns:
The issue was that the PowerFilterChain was setting a request attribute with the value of the parameter map. Since `getParameterMap` now reads the body for certain requests, we need to make sure that the body gets reset when it does. The request attribute in question is used in the HERP filter.

As I wrote that description, I was left wondering why we do that in the first place rather than using the parameter map in the request that we have when processing in the HERP filter. Does anyone remember?

I also left in some TODOs that I intend to work on for other things I noticed while digging around.

### DIFF
--- a/repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
+++ b/repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
@@ -366,8 +366,9 @@ public class PowerFilter extends DelegatingFilterProxy {
         //        .map(l -> new LimitedReadInputStream(l, request.getInputStream()))
         //        .orElseGet(request.getInputStream());
 
+        final BufferedServletInputStream bufferedInputStream = new BufferedServletInputStream(requestBodyInputStream);
         final HttpServletRequestWrapper wrappedRequest = new HttpServletRequestWrapper((HttpServletRequest) request,
-                new BufferedServletInputStream(requestBodyInputStream));
+                bufferedInputStream);
         final HttpServletResponseWrapper wrappedResponse = new HttpServletResponseWrapper((HttpServletResponse) response,
                 ResponseMode.MUTABLE,
                 ResponseMode.MUTABLE);
@@ -413,6 +414,13 @@ public class PowerFilter extends DelegatingFilterProxy {
                     LOG.info("Tracing header: {}", TracingHeaderHelper.decode(tracingHeader));
                     wrappedResponse.addHeader(TRACE_GUID.toString(), tracingHeader);
                 }
+
+                // Since getParameterMap may read the body, we must reset the InputStream so that we aren't stripping
+                // the body when form parameters are sent.
+                bufferedInputStream.mark(Integer.MAX_VALUE);
+                wrappedRequest.setAttribute("http://openrepose.org/queryParams", wrappedRequest.getParameterMap());
+                bufferedInputStream.reset();
+
                 requestFilterChain.startFilterChain(wrappedRequest, wrappedResponse);
             }
         } catch (InvalidMethodException ime) {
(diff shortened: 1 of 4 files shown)


adrianjgeorge:
You shouldn't have had to reset it then? I thought i saw some stuff about getting the parameters from the wrapped request? Because that needs to be happening otherwise, things get messy, or we should just reset it like @wdschei was doing originally, which we didn't want because that gets super weird.

That got rambally, i hope you understand.

dmnjohns:
I'm not sure I did (understand, that is).

So we wrap the request in the `PowerFilter`, and that request has access to the parameter map as you would expect. However, it consumes the body. So when we wrap the request again in the SAML filter and get the parameter map, we try to parse the query string and body, but there is no body. So we get nothing.

Unfortunately, our wrapper is not looking at the wrapped request's parameters, it is only parsing the query string and body (for certain requests). Figuring out if we can also grab the wrapped requests parameters is one of my goals.

Even so, we probably do want to reset the body if we are reading it here. Otherwise, we are always modifying posted form requests.

Of course, if we don't need to set that attribute, that last point is moot.

adrianjgeorge:
Ugh, yeah, we super wanted to consume it from the upstream request if he already read them, otherwise we always have to reset the body, because the person down stream won't know be able to read them. We were trying to mimic the behavior of basically everyone else when you call get parameters which leaves the body read. We are in a weird position though, as sort of consumers but not really, so i guess you should just reset it after reading it like @wdschei was originally doing. Crap, that's no good either though, somebody writing a filter might expect us to consume it like everybody else does. We super need this behavior to be intuitive, and i don't know that there is an intuitive solution. I would really like get Parameters to always work down stream without manually resting the stream between every layer, but i'm not seeing an easy way to do that right now. If you've got something i'd like to hear it, otherwise we need to document the shit out of this. Or just say screw it and reset the stream after reading the parameters out of it, we aren't mimicking everyone else's behavior, but that gives us better behavior downstream.

TL;DR
Just make the part where we read body parameters out reset the stream.

dmnjohns:
That basically describes the conundrum we are in. I don't really like any way of doing it (that I've thought of), but we are kind of stuck in an odd position because of the way the servlet spec works. In my mind, the best solution is to allow for both possibilities. That is, we consume the body, but if a user wants to wrap again and provide the body after consumption, that's fine. Leave it to the user to get the behavior they want, if what they want is not what is specified by the servlet spec. That's why I made the change I did -- manually reset the stream after `getParameterMap` consumes it in the `PowerFilter`.

Regardless, I absolutely agree that we should get the parameters from the wrapped request since it may have already consumed the body. That's somewhat tangential, though. If we did that, we should always have an accurate parameter map, but that still doesn't answer the question of what to do with the body (either always reset it, or force the user to reset it).

dmnjohns:
Also, I think this is the quick fix. There is very likely a better fix, and I would like to keep hacking at that, but I'm not sure I'll be able to get there in the time frame we need for the SAML stuff...

adrianjgeorge:
Yeah, i suppose we should leave the manual way right now. Resetting the stream won't reset the flag that says they asked for the stream though.

dmnjohns:
D'oh. Good point. I'll have to re-wrap, then.

adrianjgeorge:
Another argument for a setter for the input stream

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
```diff
@@ -366,11 +366,21 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
                 ResponseMode.MUTABLE,
                 ResponseMode.MUTABLE);
+        final BufferedServletInputStream bufferedInputStream = new BufferedServletInputStream(requestBodyInputStream);
+        HttpServletRequestWrapper wrappedRequest = new HttpServletRequestWrapper((HttpServletRequest) request,
+                bufferedInputStream);
+
+        // Since getParameterMap may read the body, we must reset the InputStream so that we aren't stripping
+        // the body when form parameters are sent.
+        bufferedInputStream.mark(Integer.MAX_VALUE);
+        wrappedRequest.setAttribute("http://openrepose.org/queryParams", wrappedRequest.getParameterMap());
```
adrianjgeorge: They aren't even necessarily the query params. What do we use this for?

dmnjohns: I legitimately don't know. I figured we must have had a reason, but I certainly can't think of what it was...

Mario-Lopez:
retest this please

adrianjgeorge:
Hey nice, the number of failures went down.

PR #1670, opened 2017-02-01
REP-4816 SAML origin service response test

Mario-Lopez:
I added a few test cases to the Flow 1.0 and Flow 2.0 validation tests around:
- Identity returning non-200 results (in both JSON and XML)
- a 200 response in a valid format that's neither JSON nor XML
- a 200 response that can't be parsed as the claimed format (in both JSON and XML)

The results are mostly the same except for the "can't be parsed" case where it'll still be a 200 in the Flow 1.0 case (since we really don't care) and a 502 in the Flow 2.0 case (since something went wrong when parsing something).

I also moved away from using my own content type strings and started using the `javax.ws.rs.core.MediaType` strings, and I started using a shared method to do execute the saml call where I could (i.e. tests without too many special cases).

### DIFF
--- a/repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/samlpolicy/SamlFlow10Test.groovy
+++ b/repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/samlpolicy/SamlFlow10Test.groovy
@@ -31,7 +31,6 @@ import spock.lang.Unroll
 
 import static features.filters.samlpolicy.util.SamlPayloads.*
 import static features.filters.samlpolicy.util.SamlUtilities.*
-import static framework.mocks.MockIdentityV2Service.createMappingJsonWithValues
 import static javax.servlet.http.HttpServletResponse.SC_OK
 
 /**
@@ -64,7 +63,11 @@ class SamlFlow10Test extends ReposeValveTest {
     }
 
     def setup() {
-        fakeIdentityV2Service.resetCounts()
+        fakeIdentityV2Service.with {
+            resetCounts()
+            getIdpFromIssuerHandler = null
+            getMappingPolicyForIdpHandler = null
+        }
     }
 
     @Unroll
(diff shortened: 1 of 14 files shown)


repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/samlpolicy/SamlFlow10Test.groovy
```diff
@@ -267,4 +260,104 @@ class SamlFlow10Test extends ReposeValveTest {
...
+        and: "the XML sent by the origin service and received by the client are similar enough"
+        new Diff(mc.handlings[0].response.body as String, mc.receivedResponse.body as String).similar()
+
+        where:
+        code                     | fault           | message
+        SC_BAD_REQUEST           | "badRequest"    | "Saml response issueInstant cannot be in the future."
+        SC_INTERNAL_SERVER_ERROR | "identityFault" | "Internal Server Error"
+    }
+
+    def "an unsupported content-type from the origin service will be returned to the client unaltered"() {
```
adrianjgeorge: Hopefully @RackerWilliams will get back to us on whether this is the preferred behavior. It makes sense to me to be what we do.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/samlpolicy/SamlFlow10Test.groovy
```diff
@@ -267,4 +260,104 @@ class SamlFlow10Test extends ReposeValveTest {
...
+
+        then: "the client receives the response code sent by the origin service"
+        mc.receivedResponse.code as Integer == SC_OK
+
+        and: "the response was not altered by Repose"
+        mc.receivedResponse.body as String == responseBody
+    }
+
+    @Unroll
+    def "a malformed response with content-type #contentType from the origin service will be returned to the client unaltered"() {
```
adrianjgeorge: should maybe call out that it's a legacy issuer in the title

adrianjgeorge: nevermind, this is the 1.0 test

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/samlpolicy/SamlFlow10Test.groovy
Mario-Lopez: Yeah, although I suspect that only matters for Flow 2.0?

PR #1674, opened 2017-02-02
REP-4816 SAML Atom config live update test

Mario-Lopez:
Tests to confirm the filter supports updating the atom feed configuration while Repose is running


Mario-Lopez:
ok to test

PR #1675, opened 2017-02-02
REP-4886 Encoding the issuer in the query string

dmnjohns:


repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlIdentityClient.scala
```diff
@@ -237,7 +238,7 @@ object SamlIdentityClient {
   final val PolicyRequestKey: (String) => String = (idpId: String) => s"SAML:POLICY:$idpId"
   final val TokenPath: String = "/v2.0/tokens"
   final val IdpPath: (String) => String =
-    (issuer: String) => s"/v2.0/RAX-AUTH/federation/identity-providers?issuer=$issuer"
+    (issuer: String) => s"/v2.0/RAX-AUTH/federation/identity-providers?issuer=${URLEncoder.encode(issuer, StandardCharsets.UTF_8.name())}"
```
dmnjohns: I kind of want to add a comment explaining that `URLEncoder` encodes more aggressively than necessary, but it makes this block of constants ugly...

PR #1678, opened 2017-02-09
REP-5051 Add version to footer of docs

dmnjohns:
So this was more of a pain than I was initially expecting. The commits tell a story. Or half a story, anyway.

I initially attempted to stick the version in a `docinfo-footer.html` filter. Unfortunately, I found that the `docinfo-footer` content was being placed below the `footer` div with the formatting. So that didn't work. My test `docinfo-footer.html` file had these contents:
`Version: {project-version}`

So next up was messing with the CSS. I noticed that the theme we're using builds a stylesheet that provides `::before` and `::after` hooks into the footer. Well, it was easy enough to get text in the right place by modifying the CSS to exploit that, but asciidoctor attributes were not available. That is, getting the version in the CSS would have required more work than I wanted to do. This was my test CSS file:
```
div > div#footer-text::before {
  content: "{project-version}";
}
```

So then I found out about the engines being used by asciidoctor, and in turn, the templates those engines were using. I figured out that we were using the HAML engine, and the HAML template accepts a `revnumber` attribute that it sticks in the footer, so I just populated that attribute. Alas, that attribute also gets stuck in the header, and it looked terrible. While the Gradle Asciidoctor plugin makes it easy to disable both the header and footer, disabling one or the other is not so easy. So fine, maybe I shouldn't go that route. That was my first commit.

My second commit was a hack. Rather than introducing the `revnumber` attribute, I just modified the `last-update-label` to include the version. That actually looks and works fine, but it is totally a hack.

So for my final commit, I just told Asciidoctor to use my custom template instead of the default template. I actually copied the default template, then modified it to add a proper version field to the footer. This way, going forward, we can relatively easily modify the template as needed. It's a bit better than having to mess with the CSS in certain cases, and it provides a bit more power.

### DIFF
--- a/repose-aggregator/build.gradle
+++ b/repose-aggregator/build.gradle
@@ -28,7 +28,7 @@ asciidoctor {
                'toc': 'left',
                'icons': 'font',
                'docinfo': 'shared',
-               'revnumber': project.version
+               'last-update-label': 'Version: ' + project.version + '<br>Last Updated:'
 
     resources {
         from 'src/docs/resources'
--- a/repose-aggregator/build.gradle
+++ b/repose-aggregator/build.gradle
@@ -27,7 +27,8 @@ asciidoctor {
                'stylesheet': 'repose.css',
                'toc': 'left',
                'icons': 'font',
-               'docinfo1': ''
+               'docinfo': 'shared',
+               'revnumber': project.version
 
     resources {
         from 'src/docs/resources'
--- a/repose-aggregator/build.gradle
+++ b/repose-aggregator/build.gradle
@@ -22,13 +22,13 @@ task copyXsds(type: Copy) {
 
 asciidoctor {
     requires 'asciidoctor-diagram'
+    options template_dirs: ["${project.projectDir}/src/docs/asciidoc/_backends/haml"]
     attributes 'source-highlighter': 'coderay',
                'stylesdir': "${project.projectDir}/src/docs/asciidoc/_stylesheets",
                'stylesheet': 'repose.css',
                'toc': 'left',
                'icons': 'font',
-               'docinfo': 'shared',
-               'last-update-label': 'Version: ' + project.version + '<br>Last Updated:'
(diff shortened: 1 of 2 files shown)


repose-aggregator/build.gradle
```diff
@@ -22,12 +22,13 @@ task copyXsds(type: Copy) {
...
 asciidoctor {
     requires 'asciidoctor-diagram'
+    options template_dirs: ["${project.projectDir}/src/docs/asciidoc/_backends/haml"]
     attributes 'source-highlighter': 'coderay',
                'stylesdir': "${project.projectDir}/src/docs/asciidoc/_stylesheets",
                'stylesheet': 'repose.css',
                'toc': 'left',
                'icons': 'font',
-               'docinfo1': ''
+               'docinfo': 'shared'
```
dmnjohns: I also changed this. `docinfo1` is just an alias for what I replaced it with, but since the attributes were added later as a replacement (since they are more descriptive) I switched to using those.

repose-aggregator/src/docs/asciidoc/_backends/haml/html5/document.html.haml
```diff
@@ -0,0 +1,122 @@
...
+        .footnote{:id=>['_footnote', fn.index]}
+          = succeed ". #{fn.text}" do
+            %a(href="#_footnoteref_#{fn.index}")=fn.index
+  - unless nofooter
+    #footer
+      #footer-text
+        - if attr? 'project-version'
+          = precede %(#{attr 'version-label'} #{attr 'project-version'}) do
+            - if (attr? :revnumber) || (attr? 'last-update-label')
+              %br
```
dmnjohns: The four lines above this comment are what I added to the default template.

I should note that I'm re-using the `version-label` attribute, which will only be a problem if, in the future, we want to provide a `revnumber` and label it differently than we label the project version.

PR #1680, opened 2017-02-12
REP-4947 Figure out how to link to the latest version of the docs

dmnjohns:


dmnjohns:
Just pushed the commit for those changes. I personally think that the code is clearer, even if there is a little more of it. Also, the abstraction is better, and leaves open opportunities for the future.

PR #1683, opened 2017-02-16
Make sure a compatible version of Java is installed by the Docker image

dmnjohns:
Do this all in one line, because if any of it fails, the image should not be built successfully.


adrianjgeorge:
This change already made it to master.

PR #1685, opened 2017-02-17
REP-4024 Header Normalization HTTP Response additions

wdschei:


repose-aggregator/components/filters/header-normalization-filter/src/main/resources/META-INF/schema/config/header-normalization-configuration.xsd
```diff
@@ -20,32 +20,56 @@
...
         <xs:annotation>
             <xs:documentation>
-                <html:p></html:p>
+                <html:p>The root config type for the Header Normalization filter configuration file.</html:p>
             </xs:documentation>
         </xs:annotation>
 
         <xs:all>
-            <xs:element name="header-filters" type="hnorm:HeaderFilterList" minOccurs="1" maxOccurs="1"/>
+            <!-- TODO for v9.0.0.0: remove this deprecated element. -->
```
adrianjgeorge: this is probably a removal in 10 not 9

repose-aggregator/components/filters/header-normalization-filter/src/main/resources/META-INF/schema/config/header-normalization-configuration.xsd
```diff
@@ -20,32 +20,56 @@
...
+                <xs:annotation>
+                    <xs:documentation>
+                        <html:p>
+                            Deprecated: This element is deprecated. Use request-header-filters instead.
+                        </html:p>
+                    </xs:documentation>
+                </xs:annotation>
+            </xs:element>
+            <xs:element name="request-header-filters" type="hnorm:HeaderFilterList" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="response-header-filters" type="hnorm:ResponseHeaderFilterList" minOccurs="0" maxOccurs="1"/>
```
adrianjgeorge: Why did this need a separate type?

repose-aggregator/components/filters/header-normalization-filter/src/main/resources/META-INF/schema/config/header-normalization-configuration.xsd
```diff
@@ -82,27 +108,67 @@
...
+
+    <xs:complexType name="ResponseTarget">
+        <xs:complexContent>
+            <xs:extension base="hnorm:Target">
+                <xs:annotation>
+                    <xs:documentation>
+                        <html:p>Defines a Response target with a list of allowed and/or disallowed HTTP Headers.</html:p>
+                    </xs:documentation>
+                </xs:annotation>
+                <xs:attribute name="code-regex" type="xs:string" use="optional">
```
adrianjgeorge: Where did this come from?

repose-aggregator/components/filters/header-normalization-filter/src/main/resources/META-INF/schema/config/header-normalization-configuration.xsd
wdschei: Fair enough.

repose-aggregator/components/filters/header-normalization-filter/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
```diff
@@ -71,22 +73,42 @@ class HeaderNormalizationFilter @Inject()(configurationService: ConfigurationSer
...
+    configResponse find { target =>
+      // find the first "target" config element that matches this request (if any)
+      target.url.matcher(wrappedRequest.getRequestURI).matches &&
+        (target.methods.contains(wrappedRequest.getMethod) || target.methods.contains(AllHttpMethods))
+    } foreach { target =>
+      // figure out which headers to remove, and remove them
+      (target.access match {
+        // The Response Wrapper doesn't return a case-insensitive collection of Header Names at this time.
+        case WhiteList => (new TreeSet[String]()(caseInsensitiveOrdering) ++
+          wrappedResponse.get.getHeaderNames.asScala).diff(target.headers)
```
dmnjohns: So for this bit, I think @adrianjgeorge pointed out that you could just call `.toLowercase` everything before you do the comparison. That is, lowercase the configured header names on configuration updated, and the response header names when you get them. Then the comparison would look something like:
`wrappedResponse.get.getHeaderNames.asScala.filterNot(h => target.headers.contains(h.toLowerCase))`

If you'd rather make use of Ordering to try and keep things clean (and maintain `Set` semantics) you could just make the `Set` you create for the configured header names a `TreeSet` up front. I'll add more comments after this one to show the changes necessary for that approach.

dmnjohns: For the Ordering approach, these lines become:
```
val headerNames = wrappedResponse.get.getHeaderNames.asScala.toSet
headerNames -- headerNames.intersect(target.headers)
```

or

```
val headerNames = wrappedResponse.get.getHeaderNames.asScala.toSet
headerNames.diff(headerNames.intersect(target.headers))
```

However, neither of these approaches seem particularly clear to me, since they rely on behavior of the `target.headers` `Set`. I'm slightly surprised that there is not a method to perform the negation of a diff. That would work perfectly here. Instead, we're applying DeMorgan's law.

repose-aggregator/components/filters/header-normalization-filter/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
```diff
@@ -96,18 +118,56 @@ class HeaderNormalizationFilter @Inject()(configurationService: ConfigurationSer
...
-        Option(target.getUriRegex).getOrElse(".*").r.pattern,       // if not configured, default is ".*"
-        target.getHttpMethods.asScala.map(_.toString).padTo(1, AllHttpMethods).toSet,  // default is "ALL"
-        access,
-        headers)
+    def getTarget(target: ConfigTarget, accessList: AccessList, headers: HttpHeaderList): Option[Target] = {
+      Option(Target(
+        Option(target.getUriRegex).getOrElse(".*").r.pattern, // if not configured, default is ".*"
+        target.getHttpMethods.asScala.map(_.toString).padTo(1, AllHttpMethods).toSet, // default is "ALL"
+        accessList,
+        headers.getHeader.asScala.map(_.getId).toSet))
```
dmnjohns: For the Ordering approach, this line becomes:
`TreeSet(headers.getHeader.asScala.map(_.getId): _*)(caseInsensitiveOrdering)))`

repose-aggregator/components/filters/header-normalization-filter/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
dmnjohns: A third option would be to write a case class for header names and implement the `equals` and `hashCode` methods to be case-insensitive. That's what we did in the wrappers before taking the Ordering approach.

adrianjgeorge: i thought there was a union method?

repose-aggregator/components/filters/header-normalization-filter/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
```diff
@@ -71,22 +73,42 @@ class HeaderNormalizationFilter @Inject()(configurationService: ConfigurationSer
 
     val wrappedRequest = new HttpServletRequestWrapper(servletRequest.asInstanceOf[HttpServletRequest])
 
-    config find { target =>
+    configRequest find { target =>
       // find the first "target" config element that matches this request (if any)
       target.url.matcher(wrappedRequest.getRequestURI).matches &&
```
adrianjgeorge: Could probably use our new operator here?

repose-aggregator/components/filters/header-normalization-filter/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
adrianjgeorge: `wrappedResponse.get.getHeaderNames.asScala.map(_.toLowerCase)).diff(target.headers.map(_.toLowerCase))`
Is another option

dmnjohns: Oh, right. `map` then `diff` is probably cleaner than a `filterNot`. Although, I would still probably lowercase the configured header names as soon as you get them so that you aren't doing redundant work on every request.

repose-aggregator/components/filters/header-normalization-filter/src/test/scala/org/openrepose/filters/headernormalization/HeaderNormalizationConfigurationSchemaTest.scala
```diff
@@ -0,0 +1,290 @@
...
+            |    </target>
+            |</header-normalization>""".stripMargin
+        val exception = intercept[SAXParseException] {
+          validator.validateConfigString(config)
+        }
+        exception.getLocalizedMessage should include("Cannot define the deprecated header-filters at this level along side a target")
+      }
+
+      listTypes.foreach { listType =>
+        it(s"should successfully validate when a deprecated header-filters' target defines two $listType") {
```
adrianjgeorge: Shouldn't this fail? You have a test above that says it should. were you previously able to specify both a white list and black list? I assumed we were using a single type for target, if not we probably should be.

repose-aggregator/components/filters/header-normalization-filter/src/test/scala/org/openrepose/filters/headernormalization/HeaderNormalizationConfigurationSchemaTest.scala
```diff
@@ -0,0 +1,290 @@
...
+               |    <target>
+               |        <$filterType>
+               |        </$filterType>
+               |    </target>
+               |</header-normalization>""".stripMargin
+          validator.validateConfigString(config)
+        }
+      }
+
+      it(s"should fail to validate when a deprecated header-filters' target defines both a whitelist and blacklist") {
```
adrianjgeorge: see this test says that other one shouldn't validate

repose-aggregator/components/filters/header-normalization-filter/src/test/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilterTest.scala
```diff
@@ -372,21 +516,64 @@ class HeaderNormalizationFilterTest extends FunSpec with BeforeAndAfterEach with
...
 
-  case class ConfigTarget(headers: List[String], access: AccessList, uri: Option[String], methods: Option[Iterable[String]])
+  case class ConfigTarget(targetType: TargetType, access: AccessList, headers: List[String], uri: Option[String], methods: Option[Iterable[String]])
+
+  private def getStyle(newStyle: Boolean): String = {
+    if (newStyle) "new style"
+    else "old style"
+  }
+
+  def reqres(targetType: TargetType, newStyle: Boolean): String = targetType match {
```
adrianjgeorge: can we give these a more descriptive name?

repose-aggregator/components/filters/header-normalization-filter/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
wdschei: Updated.

repose-aggregator/components/filters/header-normalization-filter/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
wdschei: I did put the `.toLowerCase` in here so that it was done at update time instead of every check.

repose-aggregator/components/filters/header-normalization-filter/src/test/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilterTest.scala
wdschei: Updated.

repose-aggregator/components/filters/header-normalization-filter/src/test/scala/org/openrepose/filters/headernormalization/HeaderNormalizationConfigurationSchemaTest.scala
wdschei: This is a side effect of being able to define multiple whitelists and multiple blacklists on the same target. I have added TODO's that will make this fail, but currently you can define a whitelist and a blacklist on the same target.

wdschei: That was wrong. You can define multiple of one type, but not both types.

repose-aggregator/components/filters/header-normalization-filter/src/test/scala/org/openrepose/filters/headernormalization/HeaderNormalizationConfigurationSchemaTest.scala
wdschei: You can define multiple of one type, but not both types.

repose-aggregator/components/filters/header-normalization-filter/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
wdschei: Updated.

wdschei: I'm not sure we would want to in this case. As it is, the `Pattern` is only compiled to a `Matcher` one time when the config is updated. The `RegexStringOperators` `trait` doesn't look like caches the `Matcher` unless I'm missing something. So with the `trait` it would compile it every time it tried to match.

wdschei: Nevermind. That didn't make any since. Updating it to use the `RegexStringOperators`.

wdschei: What I meant to say was the `Pattern` isn't cached, but it doesn't look like that is a big deal.

dmnjohns: You're right that the `Pattern` is not cached, but going from `String` to `Pattern` should be pretty cheap.

If you wanted, you could use the `RegexString` class directly. To provide maximum flexibility, the `RegexStringOperators` file allows you to either import the implicit conversion to a regex string, mix-in the implicit conversion to a regex string, or instantiate a regex string. So you could build your regex strings on configuration updated, but...

repose-aggregator/components/filters/header-normalization-filter/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
```diff
@@ -71,22 +72,47 @@ class HeaderNormalizationFilter @Inject()(configurationService: ConfigurationSer
...
+
+    val wrappedResponse = if (configResponse.isEmpty) None else Option(
+      new HttpServletResponseWrapper(servletResponse.asInstanceOf[HttpServletResponse], MUTABLE, PASSTHROUGH))
+
+    filterChain.doFilter(wrappedRequest, wrappedResponse.getOrElse(servletResponse.asInstanceOf[HttpServletResponse]))
+
+    var responseHeaders: Option[Set[String]] = None
+
+    def getResponseHeaders: Set[String] = {
+      if (responseHeaders.isEmpty) responseHeaders = Option(wrappedResponse.get.getHeaderNames.asScala.map(_.toLowerCase).toSet)
```
adrianjgeorge: This line is hard to process

repose-aggregator/components/filters/header-normalization-filter/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
```diff
@@ -71,22 +72,47 @@ class HeaderNormalizationFilter @Inject()(configurationService: ConfigurationSer
...
+      responseHeaders.get
     }
 
-    filterChain.doFilter(wrappedRequest, servletResponse)
+    configResponse find { target =>
+      // find the first "target" config element that matches this request (if any)
```
adrianjgeorge: I hate the way this works. If it's not already a story for a later version we should add one to make this apply all that match.

PR #1686, opened 2017-02-17
REP-5100 Rate Limiting Bug Fix

adrianjgeorge:


repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
```diff
@@ -318,6 +318,10 @@ object HttpServletRequestWrapper {
     parsedParameterMap.toMap
   }
 
+  def parseQueryString(queryString: String): java.util.Map[String, Array[String]] = {
```
dmnjohns: This confuses me. I know I'm not on code review, but...

Why make this a method rather than a companion object function?

Also, I would probably simplify this a bit. I would add the null-safety part to the `parseParameterString` function, then have this function simply call through to `parseParameterString` then convert the result to Java.

dmnjohns: What am I even talking about? So ignore that bit about it being a companion object function, which is clearly is.

I guess the rest is still a thing that could be done.

repose-aggregator/components/filters/rate-limiting-filter/src/main/java/org/openrepose/filters/ratelimiting/RateLimitingServiceHelper.java
```diff
@@ -68,7 +69,7 @@ public MimeType queryCombinedLimits(HttpServletRequest request, MimeType preferr
     }
 
     public void trackLimits(HttpServletRequest request, int datastoreWarnLimit) throws OverLimitException {
-        service.trackLimits(getPreferredUser(request), getPreferredGroups(request), decodeURI(request.getRequestURI()), request.getParameterMap(), request.getMethod(), datastoreWarnLimit);
+        service.trackLimits(getPreferredUser(request), getPreferredGroups(request), decodeURI(request.getRequestURI()), HttpServletRequestWrapper$.MODULE$.parseQueryString(request.getQueryString()), request.getMethod(), datastoreWarnLimit);
```
dmnjohns: Huh. That is not a pretty static method call in Java.

dmnjohns: Actually, this call appears to work without the `$.MODULE$` madness. Can that be dropped?

This gives a brief explanation, and seems to state that it can be:
http://lampwww.epfl.ch/~michelou/scala/using-scala-from-java.html

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
adrianjgeorge: So i can do the null check in the other function, but it's not strictly necessary there, because of some craziness from boxing and unboxing of types. Or at least that's what i was seeing.

repose-aggregator/components/filters/rate-limiting-filter/src/main/java/org/openrepose/filters/ratelimiting/RateLimitingServiceHelper.java
adrianjgeorge: I'll check this is what i was finding in autocomplete

PR #1687, opened 2017-02-20
Re-enable Docker Image Publishing

dmnjohns:
Ignore the incorrect branch name. This just re-enables Docker image publishing. We should be good to go on that front with the most recent Puppet changes.


adrianjgeorge:
Do it against the branch instead of against master.

PR #1398, opened 2015-10-20
Rep 2233 fix

dkowis:
This fixes REP-2233 by enforcing via XSD validation that all limit IDs across all limit-groups are unique.

EDIT: I think this is only a partial fix. I'm going to write another test to verify. I think the bug will resurface if multiple `rate-limiting-filter`s are used, there could be conflicting IDs in those files (since the logic happens in the RateLimitingService, not the filter.

EDIT: It's not a partial fix, I added another test to it to prove that it behaves well. Yay.

PR #1696, opened 2017-02-28
REP-4754 Update Rate Limiting filter to handle unsupported media type in Accept header

dmnjohns:


repose-aggregator/commons/commons-utilities/src/main/java/org/openrepose/commons/utils/http/media/MediaRangeProcessor.java
```diff
@@ -52,18 +52,4 @@ public MediaType process(HeaderValue headerValue) {
 
         return new MediaType(mediaTypeWithParametersStripped, mediaType, headerValue.getParameters());
     }
-
-    public static List<MimeType> getMimeTypesFromHeaderValues(List<String> headerValues) {
```
dmnjohns: This was only being used by the rate limiting filter. This felt like an overly specific method that was also in the wrong spot, so I modified it a bit and moved it into `MimeType`.

repose-aggregator/commons/commons-utilities/src/main/java/org/openrepose/commons/utils/http/media/MimeType.java
```diff
@@ -24,6 +24,8 @@
 @SuppressWarnings("PMD.AvoidDuplicateLiterals")
 public enum MimeType {
 
+    // TODO: This list isn't even close to being exhaustive.
+    //       An enum is probably not the right structure for this data since it cannot represent unlisted MIME types.
```
dmnjohns: I really don't like that this enum even exists, but replacing it proved to be a fair bit of work that is out of scope for this story. I left this comment to express my dislike of this code -- I'd rather use something like JAX-RS' `MediaType`.

repose-aggregator/commons/commons-utilities/src/main/java/org/openrepose/commons/utils/http/media/MimeType.java
```diff
@@ -73,17 +75,30 @@ public static MimeType guessMediaTypeFromString(String mimeType) {
                     return ct;
                 }
             }
-
-            // this is unreachable code, and at this point I'm too afraid to fix it
-            for (MimeType ct : values()) {
```
dmnjohns: None of the unit tests broke, and this really was unreachable code... so I removed it. Hopefully no other tests break, either.

repose-aggregator/components/filters/rate-limiting-filter/src/main/java/org/openrepose/filters/ratelimiting/RateLimitingHandler.java
```diff
@@ -109,9 +108,30 @@ private boolean requestHasExpectedHeaders(HttpServletRequest request) {
...
+    }
+
     private FilterAction describeLimitsForRequest(HttpServletRequestWrapper request, HttpServletResponseWrapper response) {
         if (originalPreferredAccept == MimeType.UNKNOWN) {
+            // If the accept header is set to a media type that the rate limiting filter cannot handle,
+            // the rate limiting filter should return a 406, or a body with some default media type
+            // (ignoring the accept header). HTTP/1.0 spec dictates that the former approach be taken,
+            // while HTTP/1.1 leaves it as an implementation decision. We chose to comply with the former.
             response.setStatus(SC_NOT_ACCEPTABLE);
+            // TODO: Set the supported accept types in the body? Provide a link to a supported types page?
```
dmnjohns: I wasn't sure what we wanted to do here. There does not seem to be specified (by RFC) behavior for automatic content-negotiation with regards to accept type. It is suggested, however, that the response payload contain the supported media types. Of course, the format is not specified, so... 

> The server SHOULD generate a payload containing a list of available
>  representation characteristics and corresponding resource identifiers
> from which the user or user agent can choose the one most
> appropriate.  A user agent MAY automatically select the most
>  appropriate choice from that list.  However, this specification does
> not define any standard for such automatic selection, as described in
> Section 6.4.1.

Reference: https://tools.ietf.org/html/rfc7231#section-6.5.6

repose-aggregator/components/filters/rate-limiting-filter/src/main/java/org/openrepose/filters/ratelimiting/RateLimitingHandler.java
```diff
@@ -42,16 +42,22 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static javax.servlet.http.HttpServletResponse.*;
 
 /* Responsible for handling requests and responses to rate limiting, also tracks and provides limits */
 public class RateLimitingHandler {
 
     private static final Logger LOG = org.slf4j.LoggerFactory.getLogger(RateLimitingHandler.class);
+    private static final Set<MimeType> SUPPORTED_MIME_TYPES = Stream.of(
```
adrianjgeorge: Not really a fan of this initialization.

dmnjohns: I'm with you, but I don't really know of a better way. If you come up with something, let me know!

Mario-Lopez:
```
com.google.common.collect.Sets.newHashSet(
    MimeType.APPLICATION_JSON,
    MimeType.APPLICATION_XML);
```
or
```
com.google.common.collect.ImmutableSet.of(
    MimeType.APPLICATION_JSON,
    MimeType.APPLICATION_XML);
```

Both would require adding `com.google.guava:guava` to the filter's dependencies, though.

dmnjohns: That is totally an option, but I'm not super eager to add a Guava dependency just for that. Unless someone particularly wants that solution. 

After looking through this code again, I'd really love to rewrite the rate limiting filter in Scala. Eventually, the service, too.

repose-aggregator/commons/commons-utilities/src/main/java/org/openrepose/commons/utils/http/media/MimeType.java
wdschei: Those were my sentiments as well when I was cleaning up the `enum`'s last week.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/ratelimiting/RateLimitingTest.groovy
```diff
@@ -29,13 +29,17 @@ import org.rackspace.deproxy.Response
...
 /**
  * Rate limiting tests ported over from python and JMeter
  *  update test to get limits response in json to parse response and calculate
  *  since often get inconsistent xml response fro limits cause Rate Limiting Tests flaky.
  */
 class RateLimitingTest extends ReposeValveTest {
-    final handler = { return new Response(200, "OK") }
+    final handler = { return new Response(SC_OK, "OK") }
+
+    static final int SC_TOO_MANY_REQUESTS = 429
```
adrianjgeorge: Pretty sure we have this defined somewhere else

adrianjgeorge: ExtendedStatusCodes

Mario-Lopez: Fixed.

repose-aggregator/commons/commons-utilities/src/main/java/org/openrepose/commons/utils/http/media/MimeType.java
dmnjohns: So I stopped using this class in the Rate Limiting filter, at least. I switch to Spring Web's MediaType instead.

dmnjohns:
I just pushed a huge update, so this should probably be reviewed all over again.

When I ran the tests, 4 failed. I think those failures actually have to do with the tests rather than the code. The reason for the failures if that, given the header Accept: application/*;q=0.8, application/xml;q=0.4, application/xml content is returned. The reason for that is application/xml is more specific than application/*, so while application/* has a higher quality, application/xml is preferred. Technically, application/* includes application/xml, so returning application/xml seems like the right approach. The Accept specification is a bit vague on this point.

repose-aggregator/components/filters/rate-limiting-filter/src/main/java/org/openrepose/filters/ratelimiting/write/LimitsResponseMimeTypeWriter.java
```diff
@@ -20,28 +20,44 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class LimitsResponseMimeTypeWriter {
 
+    public static final Set<MediaType> SUPPORTED_MEDIA_TYPES = Stream.of(
+            MediaType.APPLICATION_JSON,
+            MediaType.APPLICATION_XML
+            // TODO: Add support for MediaType.TEXT_XML
+    ).collect(Collectors.toSet());
```
dmnjohns: I moved the supported media types to this class, since it does the actual writing to the body.

repose-aggregator/components/filters/rate-limiting-filter/src/main/java/org/openrepose/filters/ratelimiting/RateLimitingHandler.java
```diff
@@ -109,29 +115,76 @@ private boolean requestHasExpectedHeaders(HttpServletRequest request) {
...
+            removes any media types that are not supported by this filter (since they will not be returned anyway),
+            then sorts the remaining media types by specificity (primarily) and quality (secondarily).
+              */
+            List<MediaType> sortedRelevantAcceptTypes = acceptValues.stream()
+                    .map(RateLimitingHandler::parseMediaType)
+                    .filter(Optional::isPresent)
+                    .map(Optional::get)
+                    .filter(mediaType -> SUPPORTED_MEDIA_TYPES.stream().anyMatch(mediaType::isCompatibleWith))
+                    .sorted(new CompoundComparator<>(MediaType.SPECIFICITY_COMPARATOR, MediaType.QUALITY_VALUE_COMPARATOR))
+                    .collect(Collectors.toList());
```
dmnjohns: I'm not happy about collecting this `Stream` into a `List` just to make it a stream again later, so if someone comes up with a better solution, please let me know!

repose-aggregator/components/filters/rate-limiting-filter/src/main/java/org/openrepose/filters/ratelimiting/write/LimitsResponseMimeTypeWriter.java
```diff
@@ -20,28 +20,44 @@
...
+            return MediaType.APPLICATION_JSON;
+        } else if (MediaType.APPLICATION_XML.isCompatibleWith(mediaType)) {
             outputStream.write(readableContents);
-            return MediaType.APPLICATION_XML_TYPE;
+            return MediaType.APPLICATION_XML;
         } else {
             // default to json for now
             responseTransformer.streamAsJson(new ByteArrayInputStream(readableContents), outputStream);
-            return MediaType.APPLICATION_JSON_TYPE;
+            return MediaType.APPLICATION_JSON;
```
dmnjohns: I hate that code is duplicated in the explicit JSON case and the default case. Technically, the default case should never be reached (at least, in rate limiting processing), but concerns are split between this class and the handler class, so that may not be clear. I couldn't come up with a quick solution given the methods available on `MediaType`, and I didn't want to do even more invasive refactoring, so this is how I left it.

I thought about extracting the JSON code to a method, but that's not much better. It certainly isn't very flexible is, say, we wanted to switch the default media type at some point.

repose-aggregator/components/filters/rate-limiting-filter/src/main/java/org/openrepose/filters/ratelimiting/RateLimitingHandler.java
```diff
@@ -109,29 +115,76 @@ private boolean requestHasExpectedHeaders(HttpServletRequest request) {
...
+            Parses media types from the "Accept" header (dropping any that cannot be parsed),
+            removes any media types that are not supported by this filter (since they will not be returned anyway),
+            then sorts the remaining media types by specificity (primarily) and quality (secondarily).
+              */
+            List<MediaType> sortedRelevantAcceptTypes = acceptValues.stream()
+                    .map(RateLimitingHandler::parseMediaType)
+                    .filter(Optional::isPresent)
+                    .map(Optional::get)
+                    .filter(mediaType -> SUPPORTED_MEDIA_TYPES.stream().anyMatch(mediaType::isCompatibleWith))
+                    .sorted(new CompoundComparator<>(MediaType.SPECIFICITY_COMPARATOR, MediaType.QUALITY_VALUE_COMPARATOR))
```
Mario-Lopez: I'm guessing this is the recommended way of comparing these since there's a `sortBySpecificityAndQuality` method in `MediaType` that does essentially the same thing, but do you by chance know why?  It looks like the `SPECIFICITY_COMPARATOR` looks at the quality, then compares the number of header parameters (so in theory `*/*;what=ever` would be more important than `application/xml`), and it looks like `QUALITY_VALUE_COMPARATOR` does everything you would need.  I'm guessing I'm missing something?

dmnjohns: I actually hadn't looked too much at the implementation of those `Comparator`s. I just used the JavaDocs, which made specificity sound like it had more to do with wildcards than qualities or parameters. Looking at the implementation, you appear to be right... even though `MimeType.SpecificityComparator` behaves as I would have expected. I should point out that `MimeType` is a superclass of `MediaType`. It seems like there should be some call from the `Comparator` in `MediaType` to the one in `MimeType` but I'm not seeing it. I may try replacing the `Comparator`s in this line to judge the behavior.

Reference:
http://docs.spring.io/spring/docs/4.1.4.RELEASE/javadoc-api/

dmnjohns: Derp. The `MediaType` comparator is a constant defined as:
`new SpecificityComparator<MediaType>() {
...
}`

In other words, it's an anonymous class that extends the `MimeType` `Comparator`. It doesn't override the `compare` method, only the `compareParameters` method. Then the overridden method calls through in some cases.

dmnjohns: So the difference I'm seeing between the specificity and quality comparators is that the quality comparator checks quality first (then specificity, then parameters, if applicable), whereas the specificity comparator checks specificity first (then quality, then parameters, if applicable).

Prioritizing specificity still makes sense to me, I think. It seems like we don't actually need to use a `CompoundComparator`, but their implementation did, so I did.

Mario-Lopez: Ahhhh.  I see.  I wondered how on Earth the `SPECIFICITY_COMPARATOR` was doing that until I accidently expanded the definition and saw that it was extending `SpecificityComparator` (like you said).

Mario-Lopez:
For the `application/*;q=0.8,application/xml;q=0.4` case, I think RFC 7231 indicates that for this specific case, `application/json` should be returned according to the last couple of paragraphs of the Accept section (5.3.2, bottom of page 39 and top of page 40).

https://tools.ietf.org/html/rfc7231#section-5.3

The example header is:
```
Accept: text/*;q=0.3, text/html;q=0.7, text/html;level=1,
        text/html;level=2;q=0.4, */*;q=0.5
```
Which does not have `text/plain`, but the table indicates that `text/plain` would be given a quality of `0.3` since the highest precedence match (`text/*`) has a quality of `0.3` (and presumably the server can supply the media type which is the reason for its inclusion in the table and consideration).

That leads me to believe that the following header:
```
Accept: application/*;q=0.8,application/xml;q=0.4
```
would result in the following table:
```
+-------------------+---------------+
| Media Type        | Quality Value |
+-------------------+---------------+
| application/json  | 0.8           |
| application/xml   | 0.3           |
+-------------------+---------------+
```

dmnjohns:
I wish that table was accompanied by a textual explanation like the previous example... 

Fair enough, though. I buy that.

To explain a bit more, I ended up using the `sortBySpecificityAndQuality` as kind of a guideline. My logic was that Spring Web provided a combination sort for specificity first, but not for quality first (although, after looking at the code, the quality `Comparator` does consider specificity anyway). That assumes not only the reason for that method, but also that the reason is correct. Perhaps not a valid assumption.

Side note, I think in your table `application/xml` should have a quality of `0.4` rather than `0.3`. 😉

Mario-Lopez:
The reason for the `0.3` quality in the table is due to RFC 2795 section 1, paragraph 2:
https://tools.ietf.org/html/rfc2795#section-1

😓

adrianjgeorge:
a couple of the diagrams you brought in as images can be made in the doc, with t the diagram plugin we use. in fact they look like that was how they were originally made.

dmnjohns:
You are 100% correct. In fact, by setting the `imagesdir` directory, that's where those generated images get stuck. I added the `imagesdir` property because it makes it much easier to reference images within documentation, but this seems to be an unintended side effect. Generating the docs still seems to work, so I wasn't that worried about it...

repose-aggregator/src/docs/asciidoc/filters/rate-limiting.adoc
```diff
@@ -1,52 +1,339 @@
...
-.???.cfg.xml
+.rate-limiting.cfg.xml
 ----
-<??? xmlns="http://docs.openrepose.org/repose/???/v1.0">
-    <???> <!--1-->
-    </???>
-</???>
+<rate-limiting xmlns="http://docs.openrepose.org/repose/rate-limiting/v1.0"
+    overLimit-429-responseCode="true" <!--1-->
+    datastore="distributed/hash-ring" <!--2-->
```
adrianjgeorge: this is a thing?

repose-aggregator/src/docs/asciidoc/filters/rate-limiting.adoc
```diff
@@ -1,52 +1,339 @@
...
+    datastore-warn-limit="1200" <!--3-->
```
adrianjgeorge: and this?

repose-aggregator/src/docs/asciidoc/filters/rate-limiting.adoc
dmnjohns: It totally is:
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/components/filters/rate-limiting-filter/src/main/java/org/openrepose/filters/ratelimiting/RateLimitingFilter.java#L163-L190

repose-aggregator/src/docs/asciidoc/filters/rate-limiting.adoc
```diff
@@ -1,52 +1,339 @@
...
+<1> Specifies whether this filter should return a 413 or 429 response status code when a request is limited.
+<2> Specifies which <<../services/datastores.adoc#, datastore>> should be used.
+<3> Specifies the upper bound for the number of user-specific limits to store in the datastore before a warning is logged.
+<4> Specifies whether or not the content captured by regular expression capture groups should distinguish multiple, separate limits.
+<5> Specifies that the this filter should provide an endpoint where limit data can be queried.
+<6> Specifies that the origin service will supply absolute limits which this filter will use to enrich the response to limit data queries.
+<7> Specifies the regular expression used (i.e., matched against the URI) to determine whether or not a request is a limit data query request.
+<8> Specifies <<Global Limits>>, which are applied irrespective of user.
+<9> Specifies a unique ID which identifies this limit.
+No two limits may have the same ID within a limit group, including the global limit group.
```
adrianjgeorge: I think this ended up needing to completely unique at one point

repose-aggregator/src/docs/asciidoc/filters/rate-limiting.adoc
dmnjohns: Also a thing:
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/components/services/rate-limiting-service/src/main/java/org/openrepose/core/services/ratelimit/cache/ManagedRateLimitCache.java#L64-L66

repose-aggregator/src/docs/asciidoc/filters/rate-limiting.adoc
Mario-Lopez: Are you referring to this release?

https://repose.atlassian.net/wiki/display/REPOSE/Version+7x+release+notes#Version7xreleasenotes-Release7.2.1.0(11/06/2015,Valkyrieacount_adminenhancement)

Specifically this update?
https://repose.atlassian.net/browse/REP-2233

dmnjohns: Pretty sure you're right, but I think we fixed that. The XML schema does not assert that they must be globally unique, and the limit key now uses both.

Ah, here's the commit where we changed it:
https://github.com/rackerlabs/repose/commit/415950eec543e130ad0c499469228235b235601f

Mario-Lopez: https://github.com/rackerlabs/repose/blob/master/repose-aggregator/components/services/rate-limiting-service/src/test/scala/org/openrepose/core/services/ratelimit/config/RateLimitingSchemaTest.scala#L104-L118

dmnjohns: Fair enough. I see the assert in the schema now. I'm... not sure that's necessary anymore?

Mario-Lopez: Maybe not.  We should write a test to verify.  Perhaps write a story so we can know if the schema assertion can be removed?

We might be able to repurpose one of the tests that was updated in the JIRA that made the original assertion:
https://repose.atlassian.net/browse/REP-2233
https://github.com/rackerlabs/repose/pull/1398

dmnjohns: I wrote a story -- feel free to edit it:
https://repose.atlassian.net/browse/REP-5190

PR #1700, opened 2017-03-08
Do Not Merge: REP-4221 Support for running functional tests in parallel

Mario-Lopez:
I want to know if this branch builds.  By default, it will not run in parallel.

To run in parallel, you can pass in:
```
-PintegrationTestMaxParallelForks=4
```
For example:
```
gradle cleanIntegrationTest :repose-aggregator:tests:functional-tests:integrationTest \
  -PintegrationTestMaxParallelForks=4 \
  -PrunAll \
  --tests 'features.filters.samlpolicy.*'
```

Although there is at least one more required change to make parallel builds work that is not on this branch, and that's to update the default `killOthersBeforeStarting` behavior in `ReposeValveLauncher` from `true` to `false`.  I haven't put that on this branch (yet) since that would probably impact non-parallel builds.

### DIFF
--- a/repose-aggregator/tests/functional-tests/build.gradle
+++ b/repose-aggregator/tests/functional-tests/build.gradle
@@ -134,7 +134,7 @@ integrationTest {
         }
     }
 
-    maxParallelForks = project.findProperty('integrationTestMaxParallelForks') ?: '1' as int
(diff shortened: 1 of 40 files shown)


adrianjgeorge:
Superseded by #1726

PR #1702, opened 2017-03-11
DO NOT MERGE: REP-5175 Metric Service Updates Spike

wdschei:
This still has one crucial set of unit tests that are currently ignored, but I wanted to see how the functional integration tests turned out.


repose-aggregator/components/filters/header-normalization-filter/src/test/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilterTest.scala
```diff
@@ -48,21 +49,36 @@ class HeaderNormalizationFilterTest extends FunSpec with BeforeAndAfterEach with
...
+    reset(meter)
 
     when(filterConfig.getInitParameterNames).thenReturn(List.empty[String].toIterator.asJavaEnumeration)
-    when(metricsService.newMeterByCategorySum(any(), any(), any(), any())).thenReturn(metricsMeters)
+    when(metricRegistry.meter(anyString())).thenReturn(meter)
+
+    metricsService = new MetricsService {
+      override def getRegistry: MetricRegistry = metricRegistry
+
+      override def name(className: String, names: String*): String = MetricRegistry.name(JMX_PREFIX + className, names:_*)
```
wdschei: I just wanted to point out that this is the Scala syntax, `names:_*`, to pass an incoming Java varargs along to another Java method.

repose-aggregator/components/filters/api-validator-filter/src/main/java/org/openrepose/filters/apivalidator/ApiValidatorHandler.java
```diff
@@ -153,9 +146,16 @@ public void doFilter(HttpServletRequest request, HttpServletResponse response, F
                 }
 
                 if (!isValid) {
-                    if (mbcsInvalidRequests != null) {
+                    if (Optional.ofNullable(metricsService).isPresent()) {
+                        MetricRegistry metricRegistry = metricsService.getRegistry();
                         for (String s : matchedRoles) {
-                            mbcsInvalidRequests.mark(s);
+                            metricRegistry.meter(
```
dmnjohns: Remember that this call potentially invokes a bunch of synchronization in the Dropwizard metrics library. Since we want this meter for every request, we may as grab it on initialization and keep track of it. Following that pattern, we also only ever need to call `metricsService.getRegistry()` once, during initialization, when we grab all of the metrics we are going to need.

repose-aggregator/components/filters/api-validator-filter/src/main/java/org/openrepose/filters/apivalidator/ApiValidatorHandler.java
```diff
@@ -153,9 +146,16 @@ public void doFilter(HttpServletRequest request, HttpServletResponse response, F
                 }
 
                 if (!isValid) {
-                    if (mbcsInvalidRequests != null) {
+                    if (Optional.ofNullable(metricsService).isPresent()) {
```
dmnjohns: Do we still need the null check on the metrics service? I feel like that had to do with the way we were loading things before Spring, but with Spring, I'm not sure how it would ever have a null value.

repose-aggregator/components/filters/api-validator-filter/src/main/java/org/openrepose/filters/apivalidator/ApiValidatorHandler.java
dmnjohns: Oops. I was being careless. This is a dynamic metric that were were passing a string to before, because we had those category classes.

repose-aggregator/components/filters/api-validator-filter/src/main/java/org/openrepose/filters/apivalidator/ApiValidatorHandler.java
wdschei: In Spring Boot, if an appropriate `@Autowired` value can't be found, doesn't it pass a `null`?

repose-aggregator/components/filters/api-validator-filter/src/main/java/org/openrepose/filters/apivalidator/ApiValidatorHandler.java
wdschei: Although that may have been the initial reason for those category classes.

repose-aggregator/components/filters/api-validator-filter/src/main/java/org/openrepose/filters/apivalidator/ApiValidatorHandler.java
```diff
@@ -153,9 +146,16 @@ public void doFilter(HttpServletRequest request, HttpServletResponse response, F
...
+                        MetricRegistry metricRegistry = metricsService.getRegistry();
                         for (String s : matchedRoles) {
-                            mbcsInvalidRequests.mark(s);
+                            metricRegistry.meter(
+                                    metricsService.name(
+                                            "org.openrepose.core.filters.ApiValidator",
+                                            "api-validator", // TODO replace "api-validator" with filter-id or name-number in sys-model
```
dmnjohns: We could probably satisfy this TODO by replacing `api-validator` with the stored value of `FilterConfig.getFilterName()`. We're passed the `FilterConfig` in the `init(...)` method of all `Filter`s.

wdschei: Yep. I completely agree. There was just already enough going on with updating the metrics library, I was just trying to keep it all functionally equivalent.

repose-aggregator/components/filters/destination-router-filter/src/test/scala/org/openrepose/filters/destinationrouter/DestinationRouterFilterTest.scala
```diff
@@ -38,13 +38,19 @@ import org.springframework.mock.web.{MockFilterChain, MockFilterConfig, MockHttp
...
+  private val metricsService = mock[MetricsService]
+  private val metricRegistry = mock[MetricRegistry]
+  private val meter = mock[Meter]
 
-  override def beforeEach() = {
+  override def beforeEach(): Unit = {
     reset(configurationService)
     reset(metricsService)
+    reset(metricRegistry)
+    reset(meter)
```
dmnjohns: I know that you're following the pattern this test was already using, but I super don't like having to call `reset` on mock objects. Then you have to rely on mutable state being in the right state across tests. I much prefer just creating new mocks for each test.

For more reasoning, I like this answer:
http://stackoverflow.com/a/11386867

repose-aggregator/components/filters/header-normalization-filter/src/test/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilterTest.scala
dmnjohns: It sure is. We've had to do this a few other places, primarily, I think, when dealing with Mockito mocks.

dmnjohns: Although, I think we generally put a space between the colon and the type? So `names: _*` rather than `names:_*`.

wdschei: Really. This was the first time I came across a need for it.

repose-aggregator/components/filters/api-validator-filter/src/main/java/org/openrepose/filters/apivalidator/ApiValidatorHandler.java
dmnjohns: Maybe, but I'm still glad you got rid of those classes. ;)

Really, those classes were just saving us from reconstructing the prefix for the metric name (i.e., the non-dynamic bit). All of the thread-safety is still here. In Scala, I think I can even give you the savings for the metric name. Let me play with that...

repose-aggregator/components/filters/api-validator-filter/src/main/java/org/openrepose/filters/apivalidator/ApiValidatorHandler.java
dmnjohns: Do we actually use Spring Boot to support any part of Repose? For the auto-wiring, we use Spring's IOC API directly (See: `CoreSpringProvider`). I'm not sure if Spring Boot give you another way to auto-wire things, but I don't think we're using it anyway.

To get back on topic, I think you're about Spring wiring in `null`, but the metrics service is a core service. Which is to say, it should never be null. While it isn't strictly necessary for Repose to run, it might be better to treat it like it is. Otherwise, a user might not get the metrics they are expecting. That is, drop the `null` check and let the exception bubble up if we are in a case we should never be in. We don't check that other service beans are non-null, do we?

repose-aggregator/components/filters/api-validator-filter/src/main/java/org/openrepose/filters/apivalidator/ApiValidatorHandler.java
dmnjohns: Yeah, that's probably legit. I just have to comment on things when they come to mind, else there's a very real likelihood that I'll forget my comment(s). At least this way they are recorded somewhere.

repose-aggregator/components/filters/api-validator-filter/src/main/java/org/openrepose/filters/apivalidator/ApiValidatorHandler.java
wdschei: My mistake. I meant just Spring, not Spring Boot.

I'm not sure if we check on other services currently or not. However, since we are moving towards all services being first class citizens of the Repose environment, a user could conceivably turn something off and then there wouldn't be something to pass in.

repose-aggregator/components/filters/api-validator-filter/src/main/java/org/openrepose/filters/apivalidator/ApiValidatorHandler.java
wdschei: Totally valid.

repose-aggregator/components/filters/uri-normalization-filter/src/main/scala/org/openrepose/filters/urinormalization/UriNormalizationFilter.scala
```diff
@@ -66,7 +59,14 @@ class UriNormalizationFilter @Inject()(configurationService: ConfigurationServic
...
       if (request.getParameterMap.nonEmpty) {
         queryStringNormalizers.find(_.normalize(request))
-          .foreach(queryStringNormalizer => mbcsUriNormalizations.mark(queryStringNormalizer.getLastMatch.toString + "_" + request.getMethod))
+          .foreach(queryStringNormalizer =>
+            metricsServiceOption.foreach(metricService =>
+              metricService.getRegistry.meter(metricService.name(
+                "org.openrepose.core.filters.UriNormalization",
+                "uri-normalization",
+                "Normalization",
+                queryStringNormalizer.getLastMatch.toString + "_" + request.getMethod))
```
dmnjohns: Ew, concatenation of dynamic strings.

repose-aggregator/components/filters/versioning-filter/src/main/scala/org/openrepose/filters/versioning/VersioningFilter.scala
```diff
@@ -115,6 +108,16 @@ class VersioningFilter @Inject()(@Value(ReposeSpringProperties.NODE.CLUSTER_ID)
       val configurationData = new ConfigurationData(destinations, serviceVersionMappings)
 
       try {
+        def markMeter(name: String): Unit = {
```
dmnjohns: This seems like a really odd place to define this function. We don't really need to redefine it for each request since it's parameterized anyway, so it can probably be extracted.

Oh, here's an opportunity to show the Scala trick I wanted to use!

So first, pull out the commonalities.

```
val metricRegistry = metricService.getRegistry
val meterName: String => Meter  = metricService.name(
    "org.openrepose.core.filters.Versioning",
    "versioning", // todo: replace "versioning" with filter-id or name-number in sys-model,
    "VersionedRequest",
    _: String
)
```

Then when you want to mark a meter:

```
metricRegistry.meter(meterName(targetOriginService.getMapping.getId)).mark()
```

repose-aggregator/components/filters/api-validator-filter/src/main/java/org/openrepose/filters/apivalidator/ApiValidatorHandler.java
dmnjohns: So after a quick bit of research, it appears that Spring will actually blow up if a dependency is not satisfied. Although, that behavior can be controlled using the `required` parameter on the `Autowired` annotation.

Anyway, technically, core services as we have defined them are always available. Also technically, the metrics service is currently a core service. So right now it's probably safe to drop the null checks, but once we move to services as first class citizens, then all bets are off. I hope we can come up with a design where we don't have `Optional`s all over the place to handle optional services. I mean, that isn't the worst, but it's also not the best. The only general solution I can think of involves reflection and dynamic method call routing, but that's about as nasty as it sounds. Like, worse than the JAX-RS `RuntimeDelegate` nasty, because at least that has a static interface.

dmnjohns:
You have no idea how happy seeing this refactor makes me.

repose-aggregator/components/filters/destination-router-filter/src/test/scala/org/openrepose/filters/destinationrouter/DestinationRouterFilterTest.scala
wdschei: We should probably discuss this at a Dev Meeting and then clean up this behavior across the board.

repose-aggregator/components/filters/versioning-filter/src/main/scala/org/openrepose/filters/versioning/VersioningFilter.scala
wdschei: :+1:

repose-aggregator/components/filters/api-validator-filter/src/main/java/org/openrepose/filters/apivalidator/ApiValidatorHandler.java
wdschei: Cool. That sounds like a winner. Removing all of the `if (Optional.ofNullable(metricsService).isPresent()) {` in the Java files along with the `val metricsServiceOption = Option(metricsService)` and `metricsServiceOption.foreach(metricsService =>` in the Scala files should fix it.

dmnjohns: You know what? I take that back. `Optional`s are totally the best. Turns out, Spring supports Java 8's `Optional` when it auto-wires. So I propose a solution very similar to yours, distilled into the following example:

```
@Named
class MyFilter @Inject()(Optional[MetricsService] metricsService) extends Filter {

  private val metricRegistry: Optional[MetricsRegistry] = metricsService.map(MetricsService::getRegistry)

  ...

  override def doFilter(request: ServletRequest, response: ServletResponse, chain: FilterChain): Unit = {
    metricsRegistry.ifPresent(mr -> mr.meter("myMeter").mark())
  }

  ...

}
```

Note that, yes, I am using a Java `Optional` in Scala.

adrianjgeorge:
work has been incorporated into #1707

PR #1707, opened 2017-03-26
REP-5204 Update metrics from Yammer v2.2.0 to Dropwizard v3.2.0

dmnjohns:
Obsoletes #1702.


repose-aggregator/components/filters/api-validator-filter/src/main/java/org/openrepose/filters/apivalidator/ApiValidatorFilter.java
```diff
@@ -43,15 +44,15 @@
     private static final String SCHEMA_FILE_NAME = "/META-INF/schema/config/validator-configuration.xsd";
 
     private final ConfigurationService configurationService;
-    private final MetricsService metricsService;
+    private final Optional<MetricsService> metricsService;
```
dmnjohns: I should note that IntelliJ whines about having `Optional` fields and constructor parameters. Not a big deal, but something to mention.

adrianjgeorge: I have some creative words for it then.

repose-aggregator/components/filters/api-validator-filter/src/test/groovy/org/openrepose/filters/apivalidator/ApiValidatorHandlerFactoryTest.groovy
```diff
@@ -42,132 +42,132 @@ import static org.mockito.Mockito.*
 public class ApiValidatorHandlerFactoryTest {
 
     public static class WhenCreatingHandlers {
-        private final String wadl = "default.wadl";
```
adrianjgeorge: Way to make the diff on this file worthless.

repose-aggregator/components/filters/destination-router-filter/src/main/scala/org/openrepose/filters/destinationrouter/DestinationRouterFilter.scala
```diff
@@ -95,7 +91,13 @@ class DestinationRouterFilter @Inject()(configurationService: ConfigurationServi
             logger.error("The destination could not be added -- the destinations attribute was of an unknown type")
         }
 
-        mbcsRoutedReponse.mark(target.getId)
+        metricsServiceOption.foreach(metricsService =>
```
adrianjgeorge: I kind of feel like we'd be better served by having an optional of the metric itself. these blocks feel needlessly complicated as you interact with the registry every time. I see you are calling a mark method without an arguement, is that no longer available when dealing with the raw objects?

dmnjohns: I agree about the complexity. In this case, the target ID is part of the metric, so the metric itself would have to be fetched when the configuration is updated. That's no issue. In other cases, though, there is a dynamic component to the metric name that is only available when processing the request. I used this form everywhere to be consistent, but it's true that it isn't necessarily the best form to use in all cases.

As for the mark method without an argument, that is all the underlying library provides. At least, mark with a `String` parameter is not supported by the underlying metrics library. That is what our custom `ByCategory` metric classes were buying us. In the end we get the same metrics, it's just a matter of what the call looks like.

Let me try to clean up this class a little bit, then you can tell me what you think.

dmnjohns: Updated the code. I'll point out some stuff.

repose-aggregator/components/filters/destination-router-filter/src/main/scala/org/openrepose/filters/destinationrouter/DestinationRouterFilter.scala
```diff
@@ -20,36 +20,34 @@
...
-    "destination-router",
-    "Routed Response",
-    TimeUnit.SECONDS)
-
   private var initialized: Boolean = false
   private var configurationFileName: String = _
   private var configuration: DestinationRouterConfiguration = _
+  private var routedResponseMetric: Option[Meter] = None
+  private val metricsServiceOption = Option(metricsService.orElse(null))
+  private val getRoutedResponseMetricName = MetricRegistry.name(classOf[DestinationRouterFilter], "Routed Response", _)
```
dmnjohns: I'm not sure if this is any better, but Scala allows us to partially apply a function. As a result, we can reduce duplication when it comes to metric names in Scala. In this case, we aren't really reducing duplication, just giving a name to the function that specifically creates a metric name from a target ID.

repose-aggregator/components/filters/destination-router-filter/src/main/scala/org/openrepose/filters/destinationrouter/DestinationRouterFilter.scala
adrianjgeorge: So i haven't looked at what you changed it to yet, but I thought about it more last night, and maybe a method on the service that took the arguments and did all the calls? It seems reasonable since you do the exact same thing on all of these calls. So it'd be `metricsService.mark(whatever.class, bunchOfStringsVarArg)` I think there was another arguement in there, but i don't remember.

repose-aggregator/components/filters/header-normalization-filter/src/test/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilterTest.scala
```diff
@@ -48,23 +50,29 @@ class HeaderNormalizationFilterTest extends FunSpec with BeforeAndAfterEach with
   var filter: HeaderNormalizationFilter = _
   var servletRequest: MockHttpServletRequest = _
   var servletResponse: MockHttpServletResponse = _
-  var filterChain: FilterChain = _
-  var filterConfig: FilterConfig = _
-  var metricsService: MetricsService = _
-  var metricsMeters: MeterByCategorySum = _
+  private val filterChain = mock[FilterChain]
```
adrianjgeorge: Why make this change? It seems to me you made this test less safe, now if someone adds another mock they need to be sure manually reset it, with the other way it didn't matter.

repose-aggregator/components/filters/versioning-filter/src/main/scala/org/openrepose/filters/versioning/VersioningFilter.scala
```diff
@@ -115,6 +110,17 @@ class VersioningFilter @Inject()(@Value(ReposeSpringProperties.NODE.CLUSTER_ID)
       val configurationData = new ConfigurationData(destinations, serviceVersionMappings)
 
       try {
+        def markMeter(name: String): Unit = {
```
adrianjgeorge: So this is a bit better, but i think we can still take it farther.

adrianjgeorge:
If you are referencing the dropwizard classes directly please add the dependency to the gradle build for that module. Seems like you did in some places, but not others.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/core/services/reporting/metrics/MetricsJmxObjectNameFactory.java
```diff
@@ -0,0 +1,73 @@
...
+    public ObjectName createName(String type, String domain, String name) {
+        try {
+            /*
+             Split the name on every period and give each segment its own property.
+             Doing so should give us unlimited nested buckets in JConsole.
+             Since the name argument is provided by the user, we always quote it.
+              */
+            int keyIndex = STARTING_KEY_INDEX;
+            Hashtable<String, String> objectNameProperties = new Hashtable<>();
+            for (String nameSegment : name.split(KEY_SEGMENT_DELIMITER)) {
```
adrianjgeorge: This doesn't seem to be what we agreed upon? I thought we were going to try to pick out the actual classname, and treat everything after that as the final bucket? Because can't regexes be part of the name which potentially includes `.`s

repose-aggregator/src/docs/asciidoc/services/metrics.adoc
```diff
@@ -0,0 +1,48 @@
+= Metrics Service
+
+include::../_includes/in-progress.adoc[]
```
adrianjgeorge: If this doc is complete, drop this.

repose-aggregator/src/docs/asciidoc/services/metrics.adoc
```diff
@@ -0,0 +1,48 @@
...
+----
+<1> Enables the service.
+If this attribute is set to `false`, then metrics will not be reported, +
+Defaults to `true`.
+<2> Holds the Graphite server definitions.
+<3> Declares that a Graphite server running on the `graphite.example.com` host should receive published metrics.
+<4> Declares the port that the Graphite server is listening on.
+<5> Defines the polling period, in seconds, of the Graphite reporter.
+In this case, new metrics will be sent to the Graphite server every `10` seconds.
+<6> Defines the prefix which will be prepended to the names of any metrics being sent to the Graphite server.
```
adrianjgeorge: This kind of implies to me that the hostname is not going to end up on any of the metrics going to graphite. Is that accurate?

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/powerfilter/DestinationRouterJMXTest.groovy
```diff
@@ -62,32 +59,4 @@ class DestinationRouterJMXTest extends ReposeValveTest {
         then:
         repose.jmx.getMBeanAttribute(DESTINATION_ROUTER_TARGET, "Count") == (target + 1)
     }
-
-
-    def "when requests match destination router target URI, should increment DestinationRouter mbeans for all endpoints"() {
```
adrianjgeorge: This suggests that we removed a metric, is that right?

adrianjgeorge: I see a similar change to a bunch of tests here.

versions.properties
```diff
@@ -25,10 +25,10 @@ com.typesafe.akka:akka-testkit_2.11=$akkaVersion
 com.typesafe.akka:akka-http-core-experimental_2.11=1.0
 com.typesafe.play:play-json_2.11=2.5.2
 com.typesafe.scala-logging:scala-logging-slf4j_2.11=2.1.2
-metricsVersion=2.2.0
-com.yammer.metrics:metrics-core=$metricsVersion
-com.yammer.metrics:metrics-ehcache=$metricsVersion
-com.yammer.metrics:metrics-graphite=$metricsVersion
+metricsVersion=3.2.0
+io.dropwizard.metrics:metrics-core=$metricsVersion
```
adrianjgeorge: These are in the wrong spot now

versions.properties
```diff
@@ -67,6 +67,9 @@ org.apache.tomcat.embed:tomcat-embed-jasper=$tomcatVersion
 org.apache.tomcat:tomcat-jasper-el=$tomcatVersion
 org.apache.tomcat.embed:tomcat-embed-logging-juli=$tomcatVersion
 org.asciidoctor:asciidoctorj-diagram=1.5.0
+aspectjVersion=1.8.10
```
adrianjgeorge: didn't end up using these

versions.properties
```diff
@@ -122,6 +125,7 @@ org.slf4j:slf4j-api=$slf4jVersion
 org.slf4j:jcl-over-slf4j=$slf4jVersion
 org.spockframework:spock-core=0.7-groovy-2.0
 springVersion=4.1.4.RELEASE
+org.springframework:spring-aop=$springVersion
```
adrianjgeorge: or this

repose-aggregator/components/filters/header-normalization-filter/src/test/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilterTest.scala
dmnjohns: I honestly don't know how or why this change got in here. I've actually been trying to go the other way everywhere I see `reset`. I agree that creating new instances per test is the way to go.

I fix.

dmnjohns: Oh, `git blame` tells me @wdschei made this change. Sending the bus his way! (That's for those OpenShift comments)

repose-aggregator/src/docs/asciidoc/services/metrics.adoc
dmnjohns: So as written, yes, because we expose the Graphite prefix as a configuration item but not the JMX domain. Which is weird to me, but is what we have been doing.

I would actually like to include the hostname in what we send to Graphite. So it would probably become `${HOSTNAME}.${PREFIX}.${METRIC-NAME}` in Graphite. The prefix is optional, so it would not be added if the user does not configure it. For reference, it's `${HOSTNAME}:${METRIC-NAME}` in JMX.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/powerfilter/DestinationRouterJMXTest.groovy
dmnjohns: That is right. We no longer have the `ALL` metrics that were being managed by the custom metric classes that were removed. I was actually surprised to see that the Dropwizard metrics library doesn't seem to support any metric aggregation across metrics.

So, since that aggregation is just another metric, we could still manually report it, or leave it to the user to aggregate the metrics themselves. Right now, that's gone, so I removed the tests.

If we still want the `ALL` metrics, I would suggest not forcing them onto the user, but providing them as an option. So write aggregation classes for each type of metric and make them available, but don't make them the only way to report metrics (because that limits our support for the metrics library).

repose-aggregator/components/filters/header-normalization-filter/src/test/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilterTest.scala
adrianjgeorge: There were a couple others too, so be sure to look for them

repose-aggregator/src/docs/asciidoc/services/metrics.adoc
adrianjgeorge: in 9

repose-aggregator/components/filters/destination-router-filter/src/main/scala/org/openrepose/filters/destinationrouter/DestinationRouterFilter.scala
dmnjohns: Actually, there's no need to use that magic. Just create a name prefix, then call the name method again with the specific metric name. That will be consistent across JVM languages, and does not require Scala-specific knowledge.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/core/services/reporting/metrics/MetricsJmxObjectNameFactory.java
dmnjohns: Oh, I thought we decided to go with this simpler approach. We could totally try and discern the class name, then stick the remaining portion of the name in a name bucket, but I'm hesitant to do so. For one, you lose specific buckets, which may come into play in a few places where the last portion of the name is dynamic. In addition, that approach assumes you always start with a FQCN, which may be an unnecessary restriction. Also, it complicates this logic a small bit, which isn't a big deal, but if there's a simpler, equally viable solution, I would go with that.

repose-aggregator/src/docs/asciidoc/services/metrics.adoc
dmnjohns: Oops. Totally meant to, but forgot.

repose-aggregator/components/filters/header-normalization-filter/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
```diff
@@ -17,38 +17,41 @@
...
 import org.openrepose.filters.headernormalization.config.{HeaderNormalizationConfig, HttpHeaderList, HttpMethod, Target => ConfigTarget}
 
 import scala.collection.JavaConverters._
 
 @Named
-class HeaderNormalizationFilter @Inject()(configurationService: ConfigurationService, metricsService: MetricsService)
+class HeaderNormalizationFilter @Inject()(configurationService: ConfigurationService, optMetricsService: Optional[MetricsService])
   extends Filter with UpdateListener[HeaderNormalizationConfig] with LazyLogging {
 
+  private final val NormalizationMetricPrefix = MetricRegistry.name(classOf[HeaderNormalizationFilter], "Normalization")
```
adrianjgeorge: I am still not a fan of this being the scala style standard for constant naming.

repose-aggregator/components/filters/header-normalization-filter/src/main/scala/org/openrepose/filters/headernormalization/HeaderNormalizationFilter.scala
```diff
@@ -83,7 +84,13 @@ class HeaderNormalizationFilter @Inject()(configurationService: ConfigurationSer
         case BlackList => target.headers
       }).foreach(wrappedRequest.removeHeader)
 
-      metricsMeter.mark(s"${target.url.pattern.toString}_${wrappedRequest.getMethod}_request")
+      metricsService foreach {
+        _.createSummingMeterFactory(RequestNormalizationMetricPrefix)
+          .createSummingMeter(MetricRegistry.name(
+            wrappedRequest.getMethod,
+            target.url.pattern.toString.replace('.', '_')))
```
adrianjgeorge: Do we maybe want a util method in the metrics service to do this kind of replacement for us?

dmnjohns: I like that idea. So as discussed, I'll create a utility class in the API project.

repose-aggregator/components/filters/versioning-filter/src/main/scala/org/openrepose/filters/versioning/VersioningFilter.scala
```diff
@@ -127,7 +123,7 @@ class VersioningFilter @Inject()(@Value(ReposeSpringProperties.NODE.CLUSTER_ID)
               httpResponse.setStatus(HttpServletResponse.SC_OK)
               httpResponse.addHeader("Content-Type", getPreferredMediaRange(wrappedRequest).getMimeType.getName)
 
-              mbcVersionedRequests.mark(targetOriginService.getMapping.getId)
+              markVersionedRequestMeter(targetOriginService.getMapping.getId)
```
adrianjgeorge: Lost the all metric?

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/services/reporting/metrics/SummingMeterFactory.java
```diff
@@ -19,22 +19,20 @@
...
- * factory class.
+ * A factory used to create {@link Meter}s. The {@link Meter}s created by this
+ * factory will have their values tracked and summed. The summed value will be
+ * reported as its own {@link Meter}.
+ * <p>
+ * This factory is hierarchical. Child factory names will be prefixed by the name
+ * of their parent factory. The overall structure of all factories is a tree.
  */
-public interface TimerByCategory {
+public interface SummingMeterFactory {
```
adrianjgeorge: I'd change the name of this to AggregateMeterFactory, and the implementation to SummingMeterFactory. Method names will have to change to reflect that. Which may mean a change in the name of the above all metric?
Hmm, that starts to make me think we may want to change the api in a more general sense. Maybe you ask us for an aggregate and tell us what type when you ask? I'm mostly thinking about the create child factory method and what we would do if you have one type at one level and another another layer in?
Crap, what happens when you want both types? Summing and this as of yet undefined other type?

You know what, ignore all of this for now. This would require more significant changes, and i don't want to go down that rabbit hole right now.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/core/services/reporting/metrics/MetricsJmxObjectNameFactory.java
adrianjgeorge: Stick it in the bucket

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/services/reporting/metrics/SummingMeterFactory.java
dmnjohns: I'll go ahead and change the names of things to open the way for other implementations, at least. Leaving the combination of these things in the future sounds good, though.

repose-aggregator/components/filters/versioning-filter/src/main/scala/org/openrepose/filters/versioning/VersioningFilter.scala
dmnjohns: Not in this case. That is, there wasn't an all metric being reported by the `VersioningFilter`.

The `mbc` prefix here is calling out that this metric used to be a `MeterByCategory` metric. The `MeterByCategory` metric only ever help the 9/10th name. `MeterByCategorySum` is the metric that also created the all metric. That would be designated by a `mbcs` prefix in the old code.

dmnjohns:
I added the utility class, so that should probably be looked at. I also renamed some stuff.

PR #1708, opened 2017-03-28
fix header markdown

aspiers:
In GitHub Markdown, headers with hashes on both sides require spaces to format correctly.


Mario-Lopez:
Thanks for the pull request!

PR #1709, opened 2017-04-03
REP-4725 Don't add a Server header.

wdschei:


adrianjgeorge:
looks like there were some other tests that were looking at the `Server` value

dmnjohns:
@wdschei I cleaned up your `ServerHeaderTest` a bit. A lot of what the code was doing is already done by `ReposeValveTest`, which your test extends. I also ran a formatter on things. Let me know if you disagree with any of that.

Otherwise, you made this QE review easy. I can think of one more test I want to add, but otherwise, this looks good to me.

wdschei:
That all looks good to me.

PR #1710, opened 2017-04-05
REP-5214 Via header configuration

wdschei:


repose-aggregator/core/repose-core-api/build.gradle
```diff
@@ -8,6 +8,7 @@ dependencies {
 
     provided "org.glassfish:javax.servlet"
 
+    testCompile project(":repose-aggregator:core:repose-core")
```
dmnjohns: At first I was like, oh no! Cyclic dependency!

Then I was like, oh, nevermind, it's a `testCompile` dependency.

Getting the Repose version really is more of a pain than I would like... but I'll leave that for another time.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/container/container-configuration.xsd
```diff
@@ -95,11 +95,14 @@
             </xs:annotation>
         </xs:attribute>
 
-        <!-- Check to make sure we really need this to be a Name -->
+        <!-- TODO for v9.0.0.0: The DeploymentConfigurationBase element's via attribute should be removed. -->
         <xs:attribute name="via" type="xs:string" use="optional">
```
dmnjohns: > A new element will exist that is mutually exclusive of the existing attribute

I think this is missing an assertion or something that enforces mutual exclusivity.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/container/container-configuration.xsd
```diff
@@ -252,6 +257,55 @@
...
+        <xs:simpleContent>
+            <xs:extension base="papic:EmptyString">
+                <xs:attribute name="enabled" type="xs:boolean" use="optional" default="true">
+                    <xs:annotation>
+                        <xs:documentation>
+                            <html:p>Indicates whether or not the Via header is created.</html:p>
+                        </xs:documentation>
+                    </xs:annotation>
+                </xs:attribute>
+                <xs:attribute name="content" type="xs:string" use="optional">
```
dmnjohns: I think these attributes are slightly different than discussed.

So it looks to me like `content` will be the full value of the `Via` header. Is that correct? If so, I believe we only want it to be part of the `Via` header, the way the custom, configurable part of the `Via` header is now. That may just be a matter of naming.

The `enabled` attribute also seems to indicate that it turns `Via` header creation on/off. I believe what it's supposed to do is turn the `(Repose/99.99)` part of the `Via` header on/off. Again, that may just be in the name.

So with these attributes, I _believe_ I can configure the following cases:
* No Via header created
* Via header created with Repose value
* Via header created with custom value

However, the following case is not supported, but should be:
* Via header created with Repose and custom value combined

The way I would achieve that is by having your boolean attribute dictate whether or not to include the `(Repose/99.99)` bit, and the string attribute dictate whether or not to add a prefix. Then, to disable the via header, I would simply set the boolean attribute to false and omit the string attribute. To send only the Repose portion, I would set the boolean attribute to true and omit the string attribute. To send only custom content, I would set the boolean attribute to false and configure the string attribute. To send a mix of content, I would set the boolean value to true and configure the string attribute.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/container/container-configuration.xsd
```diff
@@ -252,6 +257,55 @@
...
+                    </xs:annotation>
+                </xs:attribute>
+            </xs:extension>
+        </xs:simpleContent>
+    </xs:complexType>
+
+    <xs:complexType name="ViaConfigurationPatch">
+        <xs:simpleContent>
+            <xs:extension base="papic:EmptyString">
+                <xs:attribute name="enabled" type="xs:boolean" use="optional" default="true">
```
dmnjohns: You probably don't want to provide a default value here. By not providing a default, we can discern whether or not this attribute was meant to be patched based on its existence. When a default is provided, we will always see a value returned by the generated object.

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/container/config/ContainerSchemaTest.scala
```diff
@@ -166,55 +212,102 @@ class ContainerSchemaTest extends FunSpec with Matchers {
...
     }
   }
+
+  private val coreSpringProvider = CoreSpringProvider.getInstance()
+  coreSpringProvider.initializeCoreContext("/etc/repose", false)
+  private val reposeVersion = coreSpringProvider.getCoreContext.getEnvironment.getProperty(
+    ReposeSpringProperties.stripSpringValueStupidity(ReposeSpringProperties.CORE.REPOSE_VERSION))
+  describe("deprecated schema validation") {
+    if (reposeVersion.startsWith("8.")) {
+      it("should reject a config where cluster IDs are not unique across all patches") {
```
dmnjohns: This test seems redundant? It's included in the `describe` above.

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/container/config/ContainerSchemaTest.scala
```diff
@@ -142,17 +143,62 @@ class ContainerSchemaTest extends FunSpec with Matchers {
...
+          |        <ssl-configuration>
+          |            <key-password>other-password</key-password>
+          |        </ssl-configuration>
+          |    </cluster-config>
+          |</repose-container>""".stripMargin
+      validator.validateConfigString(config)
+    }
+  }
+
+  describe("new schema validation with via-configuration element") {
```
dmnjohns: Why break these out into a new `describe` block? It seems like these tests are still just testing the schema as it exists.

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/container/config/ContainerSchemaTest.scala
```diff
@@ -166,55 +212,102 @@ class ContainerSchemaTest extends FunSpec with Matchers {
...
+            |
+            |    <cluster-config cluster-id="foo" via="test"/>
+            |</repose-container>""".stripMargin
+        val exception = intercept[SAXParseException] {
+          validator.validateConfigString(config)
+        }
+        exception.getLocalizedMessage should include("When defining per-cluster configuration, all cluster IDs must be unique")
+      }
+
+      it("should accept a config where cluster IDs are unique across all patches") {
```
dmnjohns: Also seems redundant.

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/container/config/ContainerSchemaTest.scala
```diff
@@ -166,55 +212,102 @@ class ContainerSchemaTest extends FunSpec with Matchers {
...
+            |    </deployment-config>
+            |
+            |    <cluster-config cluster-id="foo" via="test"/>
+            |
+            |    <cluster-config cluster-id="bar" via="test"/>
+            |</repose-container>""".stripMargin
+        validator.validateConfigString(config)
+      }
+    } else {
+      it("should reject a config that uses the old via attribute") {
```
dmnjohns: Of the three, this test seems like the one that is valuable. That said, I would add removing this test to the timebomb test notes as well, since it doesn't have value once we remove the `via` attribute.

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/container/config/ContainerSchemaTest.scala
wdschei: These were the original tests. I was just keeping them in the suite until they were obsoleted by age. These were the only tests that originally used the old `via` attribute.

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/container/config/ContainerSchemaTest.scala
wdschei: Same.

repose-aggregator/core/repose-core-api/build.gradle
wdschei: Yes. I wish it was a bit easier to obtain.

repose-aggregator/components/filters/header-normalization-filter/src/test/scala/org/openrepose/filters/headernormalization/HeaderNormalizationConfigurationSchemaTest.scala
```diff
@@ -101,7 +101,7 @@ class HeaderNormalizationConfigurationSchemaTest extends FunSpec with Matchers {
   private val reposeVersion = coreSpringProvider.getCoreContext.getEnvironment.getProperty(
     ReposeSpringProperties.stripSpringValueStupidity(ReposeSpringProperties.CORE.REPOSE_VERSION))
 
-  if (reposeVersion != "10") {
+  if (reposeVersion.startsWith("8.") || reposeVersion.startsWith("9.")) {
```
wdschei: That works. Added.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/container/container-configuration.xsd
wdschei: Yep. It sure is. On it.

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/container/config/ContainerSchemaTest.scala
wdschei: Yeah. No sense in segregating what is now/will be part of the schema.

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/container/config/ContainerSchemaTest.scala
wdschei: Added to the timebomb removal list.

Mario-Lopez: Of note, we've also been creating stories with a fixed version of `9.0.0.0` for the version timebombs.  Looks like all of the ones written so far are in the REP-499x range.

https://repose.atlassian.net/issues/?jql=fixVersion%20in%20(9.0.0.0%2C%20Major)%20ORDER%20BY%20created%20DESC%2C%20key%20ASC

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/container/config/ContainerSchemaTest.scala
dmnjohns: Sure, but the thing being tested is tested elsewhere. Sure, the `via` attribute is used, but it's just sitting there as extraneous text.

Also, I'm not sure we need or even want to have unit tests for every XML node we define. Enforcing those constraints is on the validation engine. Otherwise we end up in the situation where we are testing every node that exists. Perhaps even every permutation of every node. Perhaps even nodes that are not supposed to exist. That is a lot of testing to double check the validation engine.

In my mind, what we absolutely ought to test is our assertions, since those are:
1. Potentially complex.
1. Enforcing arbitrary logic.

The rest just needs to be declared correctly.

Although, testing that we can validate a full example configuration is probably legit, just to prove that user configuration will look the way we expect it to. Also, testing the removal of things is legit, too, to prove that things that were are no longer.

That's how I see it, anyway.

/me Gets off high horse.

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/container/config/ContainerSchemaTest.scala
wdschei: Story added here:
 - https://repose.atlassian.net/browse/REP-5326

And here for the v10.0.0.0 stuff:
 - https://repose.atlassian.net/browse/REP-5327

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/container/config/ContainerSchemaTest.scala
wdschei: I agree that testing assertions should be the main focus, but at the same time typical acceptable configurations are good tests too.

I have no issues removing the tests, it just felt weird removing tests while they were still valid. That is why I just guarded them in the version logic along with the one to ensure that it was no longer available.

dmnjohns: Nah, don't worry about it. There's a good chance that my logic is flawed, but even if it isn't, these tests will be removed in the next version anyway.

I just love wasting time in theory land. Or maybe hypothetical land.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/container/container-configuration.xsd
wdschei: The `content` attribute was intended to be the user supplied portion of the `Via` header like it is now. I am definitely open to a new name.

Is it supposed to only disable the version? The confusion seems to be here:
> New element will have an optional boolean attribute that defaults to true to enable/disable the *Via header entirely*

and

> New element will have an optional String attribute to specify the prefix of the Via header
>  - no prefix and *repose value* is off, no via header will be present

(emphasis added by me)

wdschei: So as it is currently you can configure the following cases:
* No Via header created
* Via header created with only the Repose value
* Via header created with custom value and the Repose value combined

The unsupported case is:
* Via header created with custom value only

wdschei: In other words the `boolean` attribute enables/disables the `Via` header entirely not just the content of the Repose version.

dmnjohns: Got it. Man, I suck at names. I am not the one to ask. `prefix` or `value-prefix` seem legit. `content` just seems like it would be the entire value to me, especially alongside `enabled`.

It was my understanding that what is currently the `enabled` attribute would enable/disable the Repose _portion_ of the `Via` header, not the entire `Via` header. That way, the user has complete control over what ends up in the `Via` header, and whether or not the `Via` header is set at all. I also like that approach better. So I choose to accept the second part you have quoted and reject the first part. 😉 

Summoning @adrianjgeorge for final clarification.

adrianjgeorge: So the boolean was meant to just be a flag for the repose portion of the via header.

wdschei: If we want to change the names/intent of the values to be as you described, then that works too. So in that way, the `boolean` attribute would enable/disable the Repose value only and the content would still be user portion only.

content | `boolean` | result
--------|-----------|-------
Stuff | T | _Stuff (Repose/v0.0.0.0)_
Stuff | F | _Stuff_
**Undefined** | T | _(Repose/v0.0.0.0)_
**Undefined** | F | **<NO `Via` HEADER>**

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/container/container-configuration.xsd
wdschei: You would be correct. This will just force it to be a `Boolean` instead of a `boolean` in the XJC class.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/container/container-configuration.xsd
```diff
@@ -252,6 +257,67 @@
...
+                                If undefined and the repose-version is false, then no Via header will be added.
+                            </html:p>
+                        </xs:documentation>
+                    </xs:annotation>
+                </xs:attribute>
+                <xs:attribute name="repose-version" type="xs:boolean" use="optional" default="true">
+                    <xs:annotation>
+                        <xs:documentation>
+                            <html:p>
+                                Indicates whether or not the Via header is created.
```
dmnjohns: It seems like this actually indicates whether or not to include the Repose version in the Via header.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/container/container-configuration.xsd
```diff
@@ -121,6 +124,7 @@
                     <xs:element name="artifact-directory" type="papic:ArtifactDirectory" minOccurs="1" maxOccurs="1"/>
                     <xs:element name="ssl-configuration" type="papic:SslConfiguration" minOccurs="0" maxOccurs="1"/>
                     <xs:element name="logging-configuration" type="papic:LoggingConfiguration" minOccurs="0" maxOccurs="1"/>
+                    <xs:element name="via-configuration" type="papic:ViaConfiguration" minOccurs="0" maxOccurs="1"/>
```
Mario-Lopez: I'm curious why you went with the name `via-configuration`.  I would have expected something like `via-header` which would be consistent with the `tracing-header` element in the System Model config.

As an aside, this schema is wildly inconsistent.  Sometimes it uses `config` while other times it spells out `configuration`.  Sometimes `configuration` simply implies a configuration file, whereas other times it's a section of configuration.  Sometimes the names are camel case but most times they are dash-separated.  But I digress.

wdschei: I was just using the naming convention used in this particular configuration file. I good with calling it `via-header` if that is what we want to call it. A name is a name to me unless there is some compelling reason for it not to be named something. I do not feel strongly either way on this one, so I will leave that for the team to decide.

Mario-Lopez:
The Acceptance Criteria for the story says, "Do not change the request." If I remember correctly, we decided on that so we wouldn't have to worry about updating all of the request code as well. If that's the case, what is the expected behavior when the element is specified and a prefix is specified?

If the old attribute was specified, I believe (although I could be wrong) the behavior would impact both the request and the response (e.g. `Via: 1.1 Custom String (Repose/8.4.1.0)`). If we're not updating the request, then using the element to specify the prefix doesn't allow the request custom string to be set.

If we are updating the request Via header, I would point out that the spec says we must always populate the Via header on the request side, whereas it's optional on the response side. Would we ignore the "no Via header" case for the request?

Thoughts?

Mario-Lopez:
Link to spec for convenience:
https://tools.ietf.org/html/rfc7230#section-5.7.1

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/container/container-configuration.xsd
dmnjohns: I don't really have an opinion here.

Now that Mario pointed out the inconsistencies, though, I cannot unsee them.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/container/container-configuration.xsd
wdschei: Updated.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/container/container-configuration.xsd
wdschei: Updated to `via-header`.

wdschei:
The truth table below is from a previous `Outdated` comment about how the configuration effects the Response:

prefix | repose-version | result
--------|-----------|-------
Stuff | T | _Stuff (Repose/v0.0.0.0)_
Stuff | F | _Stuff_
**Undefined** | T | _(Repose/v0.0.0.0)_
**Undefined** | F | **<NO `Via` HEADER>**​

Are we just looking at assuming the `repose-version` is `true` when adding the `Via` header on the Request before sending it to the Origin Service?

dmnjohns:
The last configuration changes look good to me. 👍

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/container/container-configuration.xsd
wdschei: Done.

Mario-Lopez:
I was going to put this in one table, but it got way too big.

Assuming the request and response are `HTTP/1.1`, the host is `localhost:10010`, and the Repose version is the latest released version:

For this table, values for request-prefix, response-prefix, repose-version do not matter.

via attribute | via element | Request header | Response header
--------------|-------------|----------------|----------------
Undefined | Undefined | `1.1 localhost:10010 (Repose/8.4.1.0)` | `1.1 Repose (Repose/8.4.1.0)`
`Stuff` | Undefined | `1.1 Stuff (Repose/8.4.1.0)` | `1.1 Stuff (Repose/8.4.1.0)`
`Stuff` | **Defined** | N/A | N/A

For this table, the via attribute is undefined and the via element is defined.

request-prefix | response-prefix | repose-version | Request header | Response header
---------------|-----------------|----------------|----------------|----------------
Undefined | Undefined | Undefined | `1.1 localhost:10010 (Repose/8.4.1.0)` | `1.1 (Repose/8.4.1.0)`
`Potato` | Undefined | Undefined | `1.1 Potato (Repose/8.4.1.0)` | `1.1 (Repose/8.4.1.0)`
Undefined | `Salad` | Undefined | `1.1 localhost:10010 (Repose/8.4.1.0)` | `1.1 Salad (Repose/8.4.1.0)`
`Potato` | `Salad` | Undefined | `1.1 Potato (Repose/8.4.1.0)` | `1.1 Salad (Repose/8.4.1.0)`
Undefined | Undefined | `true` | `1.1 localhost:10010 (Repose/8.4.1.0)` | `1.1 (Repose/8.4.1.0)`
`Potato` | Undefined | `true` | `1.1 Potato (Repose/8.4.1.0)` | `1.1 (Repose/8.4.1.0)`
Undefined | `Salad` | `true` | `1.1 localhost:10010 (Repose/8.4.1.0)` | `1.1 Salad (Repose/8.4.1.0)`
`Potato` | `Salad` | `true` | `1.1 Potato (Repose/8.4.1.0)` | `1.1 Salad (Repose/8.4.1.0)`
Undefined | Undefined | `false` | `1.1 localhost:10010 (Repose/8.4.1.0)` | No header *added*
`Potato` | Undefined | `false` | `1.1 Potato (Repose/8.4.1.0)` | No header *added*
Undefined | `Salad` | `false` | `1.1 localhost:10010 (Repose/8.4.1.0)` | `1.1 Salad`
`Potato` | `Salad` | `false` | `1.1 Potato (Repose/8.4.1.0)` | `1.1 Salad`

wdschei:
Those tables define it as I understood it to be. Consequently, that is also the intended behavior of the current branch.

Mario-Lopez:
Per conversation with Bill, I updated the above table to add `Repose` to the `Via` response header when no prefix is specified in order to match the spec.

https://tools.ietf.org/html/rfc7230#section-5.7.1
```
     Via = 1#( received-protocol RWS received-by [ RWS comment ] )

     received-protocol = [ protocol-name "/" ] protocol-version
                         ; see Section 6.7
     received-by       = ( uri-host [ ":" port ] ) / pseudonym
     pseudonym         = token
```

https://tools.ietf.org/html/rfc7230#section-3.2.6
```
     token          = 1*tchar

     tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
                    / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
                    / DIGIT / ALPHA
                    ; any VCHAR, except delimiters

     comment        = "(" *( ctext / quoted-pair / comment ) ")"
     ctext          = HTAB / SP / %x21-27 / %x2A-5B / %x5D-7E / obs-text
```

https://tools.ietf.org/html/rfc7230#section-3.2.3
```
     OWS            = *( SP / HTAB )
                    ; optional whitespace
     RWS            = 1*( SP / HTAB )
                    ; required whitespace
     BWS            = OWS
                    ; "bad" whitespace
```

repose-aggregator/core/repose-core/src/main/java/org/openrepose/nodeservice/ReqResHeaderService.java
```diff
@@ -17,27 +17,21 @@
...
-        this.viaReceivedBy = StringUtilities.isBlank(configuredViaReceivedBy) ? "Repose" : configuredViaReceivedBy;
-    }
-
-    @Override
-    protected String getViaValue(HttpServletRequest request) {
-        StringBuilder builder = new StringBuilder(" ");
-
-        return builder.append(viaReceivedBy).append(" (Repose/").append(reposeVersion).append(")").toString();
+public interface ReqResHeaderService {
+    default void appendProtocolVersion(final StringBuilder builder, final HttpServletRequest request) {
```
dmnjohns: Oh, and I don't love that you're passing in a `StringBuilder`. I'm never a fan of mutable state, but passing around a mutable object just feels wrong.

dmnjohns: The more I think about it, the more I want to suggest just creating the header services as distinct core services without a common ancestor or interfaces. As for this method, I'm thinking it's just a utility method. Take a request, return the protocol as a `String`. You still get to reuse the code, but you aren't passing around mutable state or dealing with inheritance shenanigans. Just my 2c.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
```diff
@@ -63,9 +63,29 @@ class ContainerConfigurationServiceImpl @Inject()(@Value(ReposeSpringProperties.
...
+      // TODO for v9.0.0.0: This will need updated.
+      .getOrElse(Optional.ofNullable(patchedDeploymentConfiguration.getVia))
+      //.getOrElse(Optional.empty())
+  }
+
+  override def includeViaReposeVersion: Boolean = {
+    initializationCheck()
+    Option(patchedDeploymentConfiguration.getViaHeader)
+      .map(_.isReposeVersion)
+      .orElse(Option(true)).get
```
wdschei: IntelliJ had even a better suggestion to take it to a `.forall(_.isReposeVersion)` after making the `.getOrElse(true)` update.

dmnjohns: Legit.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
```diff
@@ -63,9 +63,27 @@ class ContainerConfigurationServiceImpl @Inject()(@Value(ReposeSpringProperties.
     configurationService.unsubscribeFrom(ContainerConfigurationFilename, this)
   }
 
-  override def getVia: Optional[String] = {
+  override def getRequestVia: Optional[String] = {
     initializationCheck()
-    Optional.ofNullable(patchedDeploymentConfiguration.getVia)
+    Option(patchedDeploymentConfiguration.getViaHeader)
+      .map(viaHdr => Optional.ofNullable(viaHdr.getRequestPrefix))
```
wdschei: I like the third one best and it makes adding the interim deprecated one easy too.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
```diff
@@ -126,10 +146,10 @@ class ContainerConfigurationServiceImpl @Inject()(@Value(ReposeSpringProperties.
...
   //       should not ever fail.
   private def initializationCheck(): Unit =
-    if (!isInitialized) {
-      logger.error(NotInitializedMessage)
-      throw new IllegalStateException(NotInitializedMessage)
-    }
+  if (!isInitialized) {
+    logger.error(NotInitializedMessage)
+    throw new IllegalStateException(NotInitializedMessage)
+  }
```
wdschei: Yep. Extra curly's it is.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/container/container-configuration.xsd
```diff
@@ -156,6 +160,10 @@
                         </xs:documentation>
                     </xs:annotation>
                 </xs:attribute>
+                <xs:assert vc:minVersion="1.1"
+                           test="count((@via, ./papic:via-header)) le 1"
```
wdschei: I was just following the existing, but since we're gutting things we might as well do them right.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/DeploymentConfigPatchUtil.scala
```diff
@@ -148,6 +156,22 @@ object DeploymentConfigPatchUtil {
...
 
+  /**
+    * This function does not return a copy of the base configuration, but instead mutates it directly.
+    *
+    * @param base  the base configuration object
+    * @param patch the patch configuration object, the values of which will override the base
+    * @return the base configuration object with the patch applied
+    */
+  def patchViaHeader(base: ViaHeader,
+                            patch: ViaHeaderPatch): ViaHeader = {
```
wdschei: I have no idea what happened on that one, because the formatter fixed it as expected. Weird.

repose-aggregator/core/repose-core/src/test/scala/org/openrepose/nodeservice/response/ResponseHeaderServiceImplTest.scala
```diff
@@ -0,0 +1,80 @@
...
+
+            responseHeaderServiceImpl.setVia(request, response)
+
+            if ((prefix == null || prefix.length == 0) && includeVersion) {
+              verify(response).setHeader(CommonHttpHeader.VIA, s"$protocol Repose${versionString(includeVersion)}")
+            } else if ((prefix != null && prefix.length > 0) || includeVersion) {
+              verify(response).setHeader(CommonHttpHeader.VIA, s"$protocol${prefixToString(prefix)}${versionString(includeVersion)}")
+            } else {
+              verify(response, never).setHeader(anyString, anyString)
+            }
```
Mario-Lopez: **DAMP** (Descriptive And Meaningful Phrases) and **DRY**:
http://stackoverflow.com/a/11837973/3448480

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/via/ViaHeaderOldAttributeConfigTest.groovy
```diff
@@ -0,0 +1,200 @@
...
+    @Shared String reposePort
+    @Shared String reposeVersion
+    @Shared CustomizableSocketServerConnector socketServerConnector
+
+    def setupSpec() {
+        deproxy = new Deproxy()
+        Endpoint originService = deproxy.addEndpoint(name: 'origin service', connectorFactory: { Endpoint endpoint ->
+            new CustomizableSocketServerConnector(endpoint, properties.targetPort)
+        })
+        socketServerConnector = originService.serverConnector as CustomizableSocketServerConnector
```
Mario-Lopez: I'm fetching the output of the closure after Deproxy has fed it an Endpoint.

The alternative is this:
```
deproxy.addEndpoint(name: 'origin service', connectorFactory: { Endpoint endpoint ->
    socketServerConnector = new CustomizableSocketServerConnector(endpoint, properties.targetPort)
    socketServerConnector
})
```

dmnjohns: OIC. I guess I just glanced at this and thought you were passing the endpoint an actual connector and not a factory. Nevermind me then. My bad.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/nodeservice/ReqResHeaderService.java
Mario-Lopez: It may not be an issue since there really shouldn't be any shared logic between the two (i.e. request and response). Checking the HTTP protocol of the request for both cases (which is what we were doing before this story) looks out of spec to me.

https://tools.ietf.org/html/rfc7230#section-5.7.1
```
5.7.1.  Via

   The "Via" header field indicates the presence of intermediate
   protocols and recipients between the user agent and the server (on
   requests) or between the origin server and the client (on responses),
   similar to the "Received" header field in email (Section 3.6.7 of
   [RFC5322]).  Via can be used for tracking message forwards, avoiding
   request loops, and identifying the protocol capabilities of senders
   along the request/response chain.

     Via = 1#( received-protocol RWS received-by [ RWS comment ] )

     received-protocol = [ protocol-name "/" ] protocol-version
                         ; see Section 6.7
----- snip -----

   For each intermediary, the received-protocol indicates the protocol
   and protocol version used by the upstream sender of the message.
   Hence, the Via field value records the advertised protocol
   capabilities of the request/response chain such that they remain
   visible to downstream recipients; this can be useful for determining
   what backwards-incompatible features might be safe to use in
   response, or within a later request, as described in Section 2.6.
```

For the response, we should be looking at the HTTP protocol of the response we get back from the origin service.

Mario-Lopez: I should also note, I don't see how it's possible to get the HTTP protocol of the response from the origin service. I don't see anything in the HttpServletResponse that would provide it, and when I debugged Repose, the only mention of HTTP protocol was in the original response object's `_channel` field (`org.eclipse.jetty.server.HttpChannel`) and then in that's `_version` field.

That is, from `ResponseHeaderServiceImpl.setVia(...)`, it's in:
```
response
    .originalResponse  // private final
    .getHttpChannel()  // protected
    .getHttpVersion()  // public
```

On top of that, it's not even the correct value. It doesn't appear to actually get set to anything for responses, so it keeps its default `HTTP/1.1` value anyway.

This is why I was in WireShark earlier to verify that Repose was in fact getting an `HTTP/1.0` response. There doesn't seem to be a way for us to know that from within Repose, though, despite it being perfectly normal to send an `HTTP/1.1` request and get an `HTTP/1.0` response.

dmnjohns: That interesting, but I can't say I am overly surprised. The Servlet specification always seemed to want to be general, but as a result, does not provide the best support for HTTP (or body manipulation, or response manipulation).

I guess for now we should keep current behavior, even if it is wrong. I'm not sure what the correct solution would be, given that we don't seem to have access to the information we need from the response. I wonder if perhaps we can get that information from the Apache client when we make the request to the origin service and stick it in a response parameter or something like that. That's neither here nor there, though.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/nodeservice/request/RequestHeaderServiceImpl.java
```diff
@@ -75,63 +71,42 @@ public RequestHeaderServiceImpl(ConfigurationService configurationService,
...
-        }
-
-        @Override
-        public boolean isInitialized() {
-            return isInitialized;
+        final StringBuilder builder = new StringBuilder();
+        this.appendProtocolVersion(builder, request);
+        final Optional<String> requestVia = containerConfigurationService.getRequestVia();
+        if (requestVia.isPresent() && StringUtils.isNotBlank(requestVia.get())) {
+            builder.append(" ");
```
wdschei: Yep. This was a change that obviously didn't get fully cleaned up. Fixed.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/nodeservice/ReqResHeaderService.java
wdschei: Converted this over to a static helper instead of an interface with a default method.

dmnjohns:
@wdschei The recent changes look good. Would you mind adding an item to the release notes as well?

wdschei:
No problem. The only thing that isn't currently addressed is the `StringUtils.isNotBlank` modifications.

Mario-Lopez:
Just to put it out there that I'm not done with the functional tests...

Mario-Lopez:
The only test failure was an unrelated functional test where iteration 262 out of 500 did not get back a 200 response.

Otherwise, this is good to go.

PR #1716, opened 2017-04-12
REP-5214 Take 2 - Moving the isNotBlank() check further back in the processing.

wdschei:
I'm not sure we want to do this or not, but it was mentioned in the original PR code review.

### DIFF
--- a/repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
+++ b/repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
@@ -25,6 +25,7 @@ import javax.annotation.{PostConstruct, PreDestroy}
 import javax.inject.{Inject, Named}
 
 import com.typesafe.scalalogging.slf4j.LazyLogging
+import org.apache.commons.lang3.StringUtils
 import org.openrepose.commons.config.manager.UpdateListener
 import org.openrepose.core.container.config.{ContainerConfiguration, DeploymentConfiguration}
 import org.openrepose.core.services.config.ConfigurationService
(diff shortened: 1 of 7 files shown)


repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
```diff
@@ -68,8 +69,9 @@ class ContainerConfigurationServiceImpl @Inject()(@Value(ReposeSpringProperties.
     Optional ofNullable {
       Option(patchedDeploymentConfiguration.getViaHeader)
         .map(_.getRequestPrefix)
+        .filter(StringUtils.isNotBlank)
         // TODO for v9.0.0.0: Remove the .orElse().
-        .orElse(Option(patchedDeploymentConfiguration.getVia))
+        .orElse(Option(patchedDeploymentConfiguration.getVia).filter(StringUtils.isNotBlank))
```
dmnjohns: Couldn't you just `filter` after the `orElse` so that you only have to do it once?

dmnjohns:
Had I thought about it sooner, the same could be accomplished in the schema itself. That would probably actually be the best place for it. Validate the input rather than coerce it.

http://stackoverflow.com/a/20499484

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
wdschei: If you don't filter the first one (e.g. `.map()`), then will not reach the `.orElse()` if the `.getViaHeader.getRequestPrefix` is present but blank.

wdschei:
Yeah. That is true and still very do'able.

repose-aggregator/components/services/container-configuration-service/impl/src/main/scala/org/openrepose/nodeservice/containerconfiguration/ContainerConfigurationServiceImpl.scala
dmnjohns: That's right. Fair enough.

wdschei:
This is merged and that might be the better change.
I'll fire up another branch with that mod.

PR #1722, opened 2017-04-14
REP-5115 Adds a task between building and testing of artifacts and their publishing

adrianjgeorge:
This adds a central task when publishing that requires all artifacts build and test before anything gets released. This adds a check that will prevent some partial releases.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,6 +40,8 @@ ext {
     scalaMajDotMin = "$scalaMajor.$scalaMinor"
 }
 
+task('buildAndTestAll', description: 'Builds all the sub-projects as well as runs there tests.', group: 'release')
+
 task('publishToPackageRepo', group: 'publishing', description: 'Publishes the debian and rpm packages to the package repo.',
         dependsOn: [':repose-aggregator:artifacts:cli-utils:buildDeb',
                     ':repose-aggregator:artifacts:cli-utils:buildRpm',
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -40,6 +40,8 @@ ext {
     scalaMajDotMin = "$scalaMajor.$scalaMinor"
 }
 
+task('buildAndTestAll', description: 'Builds all the sub-projects as well as runs there tests.', group: 'release')
```
adrianjgeorge: I'm not super happy with this name, but I haven't thought of anything better yet.

adrianjgeorge:
retest this please

build.gradle
```diff
@@ -361,8 +365,10 @@ subprojects {
         sign configurations.archives
     }
 
+    def publishGate = project.tasks.getByPath(':buildAndTestAll')
```
dmnjohns: Rather than getting the task by path, couldn't you just use the task directly? So drop this line, and on the following line do:
`buildAndTestAll.dependsOn build`

build.gradle
dmnjohns: `buildAndCheckAll`? 😛 

I honestly don't know what else to call this task. It doesn't do anything on its own, but running it would run `build` on all sub-projects due to dependencies. I suppose you could call it `publishGate`, but that is more descriptive of what it is than what it does.

build.gradle
```diff
@@ -361,8 +365,10 @@ subprojects {
...
+    publishGate.dependsOn build, check
```
adrianjgeorge: huh, that has not been my experience in the past. I wonder if they changed it?
Ahh, I know what it is. Publish doesn't depend on build, it depends on assemble.

build.gradle
adrianjgeorge: that's a maybe. I'll try it. because of scoping it might get weird.

PR #1726, opened 2017-04-18
REP-5193 Work on parallelization for functional tests

adrianjgeorge:
This isn't ready for merge yet, I've opened this so i can get the failing tests.
Also this rebases and supersedes #1700

### DIFF
--- a/repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/config/DefaultDestinationTest.groovy
+++ b/repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/config/DefaultDestinationTest.groovy
@@ -7,9 +7,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -38,11 +38,9 @@ class DefaultDestinationTest extends ReposeValveTest {
     def setupSpec() {
         deproxy = new Deproxy()
         deproxy.addEndpoint(properties.targetPort)
-
     }
 
     def setup() {
-
         int dataStorePort = PortFinder.instance.getNextOpenPort()
         params = properties.getDefaultTemplateParams()
 
@@ -73,7 +71,6 @@ class DefaultDestinationTest extends ReposeValveTest {
 
         // start repose
         when: "starting Repose with more or less than one default destination endpoint"
-        reposeLogSearch.cleanLog()
         repose.start([waitOnJmxAfterStarting: false])
 
         then: "error should be logged"
--- a/repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/config/DefaultDestinationTest.groovy
+++ b/repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/config/DefaultDestinationTest.groovy
@@ -19,11 +19,11 @@
  */
 package features.core.config
 
+import framework.PortFinder
 import framework.ReposeValveTest
 import framework.category.Slow
 import org.junit.experimental.categories.Category
 import org.rackspace.deproxy.Deproxy
-import org.rackspace.deproxy.PortFinder
 import spock.lang.Unroll
 
 import static org.linkedin.groovy.util.concurrent.GroovyConcurrentUtils.waitForCondition
@@ -43,7 +43,7 @@ class DefaultDestinationTest extends ReposeValveTest {
 
     def setup() {
 
-        int dataStorePort = PortFinder.Singleton.getNextOpenPort()
+        int dataStorePort = PortFinder.instance.getNextOpenPort()
         params = properties.getDefaultTemplateParams()
 
         params += [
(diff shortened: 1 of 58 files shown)


repose-aggregator/tests/functional-tests/src/integrationTest/groovy/framework/PortFinder.groovy
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+package framework
+
+import groovy.util.logging.Log4j
+import org.springframework.util.SocketUtils
+
+@Log4j
+@Singleton
+class PortFinder {
```
Mario-Lopez: I would also point out that the latest Deproxy (that we're not using) does do what you mentioned where it can, but also like you mentioned, we still have the Repose side to worry about.

dmnjohns: Great! Then this comment can just be ignored entirely. Although we should totally upgrade Deproxy at some point.

dmnjohns: Oh, we haven't even release a version of Deproxy with that change. Well then.

Mario-Lopez: In fact, the Deproxy build doesn't even work right now!

At least it didn't when I checked a few months ago.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/valveSelfConfigure/MultiClusterMultiNodeTest.groovy
```diff
@@ -126,15 +99,5 @@ class MultiClusterMultiNodeTest extends Specification {
         then:
         thrown(ConnectException)
     }
-
-    def cleanup() {
```
adrianjgeorge: You are still counting on the inherited method then?

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/config/DefaultDestinationTest.groovy
```diff
@@ -19,87 +19,61 @@
...
-        )
-        repose.enableDebug()
-        reposeLogSearch = new ReposeLogSearch(properties.getLogFile())
+        repose.configurationProvider.applyConfigs("features/core/config/default-dest", params)
         reposeLogSearch.cleanLog()
 
-
+        // start repose
         when: "starting Repose with more or less than one default destination endpoint"
+        reposeLogSearch.cleanLog()
```
adrianjgeorge: I have no idea how that got added. Bad copy pasta?

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/reposelint/ReposeLintWRoleConfigTest.groovy
```diff
@@ -42,16 +42,12 @@ class ReposeLintWRoleConfigTest extends Specification {
...
     static String allowedWithoutAuthorizationDesc = "Users with the '${configrole}' Identity role WILL pass through this component BUT authorization checks will not be performed"
-    def
     static String allowedWithAuthorizationDesc = "Users with the '${configrole}' Identity role WILL pass through this component IF AND ONLY IF their Identity service catalog contains an endpoint required by the authorization component"
-    def static String allowedDesc = "Users with the '${configrole}' Identity role WILL pass through this component"
-    def
     static String notAllowedDesc = "Users with the '${configrole}' Identity role WILL NOT pass through this component"
 
     def setupSpec() {
-        this.testProperties = new TestProperties()
+        this.testProperties = new TestProperties(this.getClass().canonicalName.replace('.', '/'))
```
adrianjgeorge: Indeed

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/valveSelfConfigure/MultiClusterMultiNodeTest.groovy
dmnjohns: Absolutely. One of my complaints with our performance framework is that the lifecycle of certain things aren't as encapsulated as I would like. For example, `deproxy` has to be initialized in every test, but that declaration is in `ReposeValveTest` and the `cleanupSpec` block handles its shutdown. That's a bit tangential, though.

In this case, we are performing a `setup`, but expecting the `cleanupSpec` block to do the cleanup. So the lifecycle doesn't really match. By matching the lifecycle, I think we gain two things:

1. Clarity. `setupSpec` is cleaned up by `cleanupSpec`. So if I want to know what is handling the cleanup, if I had a `setupSpec`, I would first go look for a `cleanupSpec`.
2. Safety. This one is a bit iffy, because someone could change the `setupSpec` to `setup` and add more tests, but my hope is that by forcing the developer to make that change, the developer realizes that they will also need a `cleanup` block. Left as is, the developer could just add mores tests and things would appear to work. However, the old Repose instances wouldn't be shut down, only the newest instance would when `cleanupSpec` gets run.

If we wanted to match the lifecycle, we would either have to keep this `cleanup` block or change `setup` to `setupSpec`. This cleanup block is redundant code that the inherited `cleanupSpec` already contains, so to keep DRY, I would just change `setup` to `setupSpec`.

A lot of rationale for a tiny change, I know.

dmnjohns:
Looks like a codenarc violation failed the build.

adrianjgeorge:
yeah, i saw that.

PR #1727, opened 2017-04-19
REP-5337 Ship Performance Data to InfluxDB

dmnjohns:


repose-aggregator/tests/performance-tests/roles/gatling/files/gatling.conf
```diff
@@ -0,0 +1,132 @@
...
+    file {
+      #bufferSize = 8192                # FileDataWriter's internal data buffer size, in bytes
+    }
+    leak {
+      #noActivityTimeout = 30  # Period, in seconds, for which Gatling may have no activity before considering a leak may be happening
+    }
+    graphite {
+      #light = false                    # only send the all* stats
+      host = "influxdb.openrepose.org"  # The host where the Carbon server is located
+      port = 13002                      # The port to which the Carbon server listens to (2003 is default for plaintext, 2004 is default for pickle)
```
dmnjohns: I don't like having the port exposed here.

i'm going to look into using [Ansible's Vault](http://docs.ansible.com/ansible/playbooks_vault.html) instead.

repose-aggregator/tests/performance-tests/roles/gatling/files/gatling.conf
```diff
@@ -0,0 +1,132 @@
+#########################
+# Gatling Configuration #
```
dmnjohns: I should note that this file is a copy of the default configuration file installed with Gatling. The only change was adding and configuring a Graphite writer.

repose-aggregator/tests/performance-tests/roles/gatling/files/gatling.conf
```diff
@@ -0,0 +1,132 @@
...
+    #noReports = false       # When set to true, don't generate HTML reports
+    #maxPlotPerSeries = 1000 # Number of points per graph in Gatling reports
+    #useGroupDurationMetric = false  # Switch group timings from cumulated response time to group duration.
+    indicators {
+      #lowerBound = 800      # Lower bound for the requests' response time to track in the reports and the console summary
+      #higherBound = 1200    # Higher bound for the requests' response time to track in the reports and the console summary
+      #percentile1 = 50      # Value for the 1st percentile to track in the reports, the console summary and Graphite
+      #percentile2 = 75      # Value for the 2nd percentile to track in the reports, the console summary and Graphite
+      #percentile3 = 95      # Value for the 3rd percentile to track in the reports, the console summary and Graphite
+      #percentile4 = 99      # Value for the 4th percentile to track in the reports, the console summary and Graphite
```
Mario-Lopez: This is outside the scope of this story, but I've never been a huge fan of Gatling's default percentiles. If we were ever going to change them, now would be the time. Do you have any thoughts on this?

Before finalizing on anything, I would want to see what other folks are doing and which percentiles give us the most valuable data. What I've thought in the past is that the 75th percentile wasn't terribly useful. Instead, we could have a 50/95/99/99.9 or maybe 50/90/95/99.

Of note, I've configured 50/95/99/99.9 in the past, and I thought that data was pretty interesting. The max value and the 99th percentile are usually pretty far apart, and the max value could easily represent just one data point. The 99.9th percentile provided more value IMO.

Mario-Lopez: Actually, this is irrelevant if InfluxDB or Grafana are the ones that are going to be doing that analysis.

dmnjohns: So this is the data that Gatling will push into InfluxDB. Grafana will simply query InfluxDB, so if data isn't there, we aren't going to see it. Gatling itself is crunching all the numbers.

That being said, we might be able to get InfluxDB to do some processing. Anything that we can do given the data in the DB and InfluxQL. The Grafana docs actually give a basic example:
http://docs.grafana.org/features/datasources/influxdb/#query-editor

Here are the data fields currently being collected:
`time, count, max, mean, min, percentiles50, percentiles75, percentiles95, percentiles99, repose_version, status, stdDev`

The catch is that we can only store/analyze the data that Gatling is sending out. Here is a snippet of the non-user data sent by Gatling:
>>> gatling.computerworld.getComputers.ko.count 0 1492629621
gatling.computerworld.getComputers.all.count 2 1492629621
gatling.computerworld.getComputers.all.min 192 1492629621
gatling.computerworld.getComputers.all.max 219 1492629621
gatling.computerworld.getComputers.all.mean 205 1492629621
gatling.computerworld.getComputers.all.stdDev 13 1492629621
gatling.computerworld.getComputers.all.percentiles50 192 1492629621
gatling.computerworld.getComputers.all.percentiles75 219 1492629621
gatling.computerworld.getComputers.all.percentiles95 219 1492629621
gatling.computerworld.getComputers.all.percentiles99 219 1492629621

The Puppet PR defines the templates that are used to transform those lines of data that Gatling sends into data in a DB measurement (table).

dmnjohns: Oh, as for percentiles, I don't have much of an opinion. `50/95/99/99.9` seems legit to me.

repose-aggregator/tests/performance-tests/roles/gatling/files/gatling.conf
dmnjohns: I won't worry about this for now. @Mario-Lopez proposed using a Virtual Cloud Network instead to provide better isolation and security for our hosts. If time permits, I will look into the solution, instead.

The form that using Ansible Vault would take is that we encrypt the port either in the roles default variables files, a file alongside it, or as a task variable on the template task. We then use the `template` module to template and copy the Gatling configuration file to the hosts that will run Gatling. As part of that process, the decrypted port will be templated into the configuration file. Of course, when Gatling is run, the encryption password must be provided either on the command line (by passing a password file), or by setting an environment variable. There are a few ways to accomplish that, but we would likely use Puppet to store the Vault password (in eyaml) and pass the password to the Ansible master host (whether it be the Jenkins slaves themselves or other cloud servers).

dmnjohns: To capture this concern, I have created:
https://repose.atlassian.net/browse/REP-5412

repose-aggregator/tests/performance-tests/roles/gatling/files/gatling.conf
dmnjohns: Now with percentiles changed as well.

repose-aggregator/src/docs/asciidoc/architecture/performance.adoc
```diff
@@ -48,6 +48,59 @@ As Gatling runs, it will pipe the performance data it captures to InfluxDB.
...
+
+....
+gatling.*.*.*.* measurement.simulation.request.status.field
+....
+
+Resulting in a database full of entries like the following:
+
+....
+name: gatling # <1>
+time                count max mean min percentiles50 percentiles75 percentiles95 percentiles99 request                  simulation    status stdDev
```
Mario-Lopez: Ugh...  this table looks horrible when rendered due to how large the font is and how little of the screen space gets used.

While I would like us to stop wasting so much screen space and use a decent sized font, I would be fine with using a smaller font and adding a horizontal scrollbar if need be (and possible).

Thoughts?

repose-aggregator/src/docs/asciidoc/architecture/performance.adoc
```diff
@@ -48,6 +48,59 @@ As Gatling runs, it will pipe the performance data it captures to InfluxDB.
...
+----                ----- --- ---- --- ------------- ------------- ------------- ------------- -------                  ----------    ------ ------
+1492629621000000000 2     219  205 192 192           219           219           219           getComputers           computerworld all    12
+....
+<1> The name of the measurement.
+
+To learn more about how InfluxDB handles Graphite input, see the https://github.com/influxdata/influxdb/blob/master/services/graphite/README.md[Graphite Input README].
```
Mario-Lopez: Is it redundant to put "README" in the output text of the link?

repose-aggregator/tests/performance-tests/roles/gatling/files/gatling.conf
Mario-Lopez: I've been torn about whether or not we should include the whole config file. I like having all of the possible config options right here, but it makes it difficult to see what we've configured (i.e. changed from the default value).

If that's just me and you think it's a great idea to leave everything here, then that's fine with me.

repose-aggregator/tests/performance-tests/roles/gatling/files/gatling.conf
```diff
@@ -0,0 +1,132 @@
...
+    #noReports = false       # When set to true, don't generate HTML reports
+    #maxPlotPerSeries = 1000 # Number of points per graph in Gatling reports
+    #useGroupDurationMetric = false  # Switch group timings from cumulated response time to group duration.
+    indicators {
+      #lowerBound = 800      # Lower bound for the requests' response time to track in the reports and the console summary
+      #higherBound = 1200    # Higher bound for the requests' response time to track in the reports and the console summary
+      percentile1 = 50      # Value for the 1st percentile to track in the reports, the console summary and Graphite
+      percentile2 = 95      # Value for the 2nd percentile to track in the reports, the console summary and Graphite
+      percentile3 = 99      # Value for the 3rd percentile to track in the reports, the console summary and Graphite
+      percentile4 = 99.9    # Value for the 4th percentile to track in the reports, the console summary and Graphite
```
Mario-Lopez: This might exacerbate a bug we have in the Ansible stuff.

https://github.com/rackerlabs/repose/blob/master/repose-aggregator/tests/performance-tests/roles/gatling/defaults/main.yml#L18

Even my TO DO comment is misleading. What it really does is look at the `percentile4` value.

https://github.com/rackerlabs/repose/blob/master/repose-aggregator/tests/performance-tests/roles/gatling/files/simulations/filters/saml/SamlSimulation.scala#L120

At the very least, I recommend changing the `500` value to something larger, maybe `3000`. The big reason is because if it fails to meet this, the `gatling` command returns a non-negative status code which causes the Ansible run to abruptly stop.

Slightly better solution would be to change all of the simulations to use `percentile3` instead.

Best would include fixing the name of the Ansible variable. I never did because I wasn't 100% sure what the point of it was. My initial guess was it would be a red flag in the HTML report if the expectation wasn't met, but in reality, it seems to be treated as a "if this isn't met, something terrible must have happened and the results are useless, so throw them away."

repose-aggregator/src/docs/asciidoc/architecture/performance.adoc
dmnjohns: Is it? i'm not sure what the redundancy is. I don't mind changing it, though. What were you thinking?

repose-aggregator/src/docs/asciidoc/architecture/performance.adoc
```diff
@@ -48,6 +48,59 @@ As Gatling runs, it will pipe the performance data it captures to InfluxDB.
...
+==== Gatling
+
+Data will be transported from Gatling to InfluxDB using the Graphite line protocol.
+The data being transferred will include numerous statistics calculated for a distinct time.
+To get a better idea of the data in question, a sample has been provided which captures the number of request/response pairs sent at time `1492629621` as well as various statistics calculated from the amount of time taken to service those requests.
+Note that the sample data only captures a single second of the Gatling simulation.
+Much more data of the same form would be expected for every second that the Gatling simulation ran.
+
+....
+gatling.computerworld.getComputers.ok.count 2 1492629621
```
dmnjohns: I imagine so.

repose-aggregator/src/docs/asciidoc/architecture/performance.adoc
dmnjohns: So I see that there's a way to [add a horizontal scrollbar](http://asciidoctor.org/docs/user-manual/#to-wrap-or-to-scroll). I haven't found a way to change the font size for just this block yet.

Changing the style for just this block does make me a little nervous, though. I worry that, since it doesn't behave like other blocks, it will stick out and be counter-intuitive. That is, I may not think to scroll in this block when every other block wraps. Or I may have to increase my font size to read this block then decrease it again to read the rest of the document comfortably.

I agree that this block renders pretty terribly. So maybe instead, I'll just translate this into a table. It really is table data. The way it's formatted is what you would see as output from InfluxDB, but we don't necessarily need to keep that format. How does that sound?

repose-aggregator/tests/performance-tests/roles/gatling/files/gatling.conf
dmnjohns: A fair concern. So if we weren't to copy the entire config file, how would we configure the parts we care about? That seems like it would be a post-installation step to modify the default, installed `gatling.conf` file. However, then it seems like we either need to have a tool that understands the format and allows us to modify parts, or take a more brutal approach with something like sed.

There might be a better approach that my braining isn't thinking of right now, though.

repose-aggregator/src/docs/asciidoc/architecture/performance.adoc
Mario-Lopez: tl;dr: A table is totally the way to go. Silly me for not seeing it sooner.

Regarding the other stuff...

**Scrollbar**
I think we may still end up needing to do this with a table. The rendered horizontal scrollbar should be indication enough that one can scroll the block. The only problem would be Mac users who have "Show scroll bars" set to hide them, but I think that's on them.

**Font Size**
I don't like the font size of our new docs in general, and I was convinced the font size of the fixed-width text was larger. I even put together some screenshots to prove it, but apparently it's not the case. In fact, it seems the line height (i.e. line spacing) of the fixed-width text might even be smaller. I'm guessing the bold look and fixed-width part of the text led me to believe the font size was larger.

In fact, while the font size in general is still larger than a lot of other docs I see, the screenshots (included below) also showed that apparently I really dislike the line height, too. This is, of course, totally out of scope of this, but I thought I would point out what I was going on about.

The sections are:
(first row) our old docs, Jenkins docs, Puppet firewall docs, Rackspace blog, our new docs
(second row) our new docs, our new docs fixed-width text

font sizes

repose-aggregator/src/docs/asciidoc/architecture/performance.adoc
Mario-Lopez: Just "Graphite Input" or maybe "The Graphite Input" to match the title of the linked page. Adding "README" at the end seemed similar to adding "Documentation" or "link" to the end. I figure it's implied.

repose-aggregator/tests/performance-tests/roles/gatling/files/gatling.conf
Mario-Lopez: I meant copying over a `gatling.conf` that looked like this:
```
#########################
# Gatling Configuration #
#########################

gatling {
  charting {
    indicators {
      percentile1 = 50      # Value for the 1st percentile to track in the reports, the console summary and Graphite
      percentile2 = 95      # Value for the 2nd percentile to track in the reports, the console summary and Graphite
      percentile3 = 99      # Value for the 3rd percentile to track in the reports, the console summary and Graphite
      percentile4 = 99.9    # Value for the 4th percentile to track in the reports, the console summary and Graphite
    }
  }
  data {
    writers = [console, file, graphite] # The list of DataWriters to which Gatling write simulation data (currently supported : console, file, graphite, jdbc)
    graphite {
      host = "influxdb.openrepose.org"  # The host where the Carbon server is located
      port = 12345                      # The port to which the Carbon server listens to (2003 is default for plaintext, 2004 is default for pickle)
    }
  }
}
```

Assuming that's allowed. I don't know if it's picky about needing the empty section names. That is:
```
gatling {
  core {
    extract {
      regex { }
    }
  }
...
```

Of note, the idiomatic way for the brutal approach is Ansible's `lineinfile` module, although its use is highly discouraged in favor of copying files and templates over instead.
http://docs.ansible.com/ansible/lineinfile_module.html

repose-aggregator/src/docs/asciidoc/architecture/performance.adoc
dmnjohns: Ah. See, when I was writing it, the README was the thing I was actually linking to, so I included the word "README". "See the Graphite Input" doesn't make much sense to me, because the "Graphite Input" is not defined in that context. Is it an actual input? What even is an input? A readme, however, is not as context bound. That is, most people reading the architecture page probably know what a readme is. Of course, I would also probably try to justify including "documentation" or "link" in a link. In the end, I want the information to be clear and available for the user. For example, I would totally be fine with something like, "To learn more... follow http://...[this link]."

Having said all of that, I'm also totally fine dropping the word "README". I'll also drop the leading "the". That way, it reads "see Gatling Input", where "Gatling Input" is the title of a document. Sound good?

repose-aggregator/tests/performance-tests/roles/gatling/files/gatling.conf
dmnjohns: Oh, I see. That makes sense. So we will still be copying the file, just not including all of the defaults. Totally legit. I'll do that.

Last night my brain was stuck on trying to modify the installed configuration instead, which I wasn't a fan of.

repose-aggregator/src/docs/asciidoc/architecture/performance.adoc
Mario-Lopez: Good point on also potentially going with "see [Graphite Input]." That being said, I can also see what you're saying, too. I'm fine with leaving it as is.

dmnjohns: NOPE! IT'S TOO LATE. THIS IS HAPPENING.

Mario-Lopez: OH NOES! I WAS NOT READY.

repose-aggregator/tests/performance-tests/roles/gatling/files/gatling.conf
dmnjohns: A quick test showed that leaving out that defaults does not break anything (which was expected). It also seems that we do not need the empty sections.

repose-aggregator/tests/performance-tests/roles/gatling/files/gatling.conf
```diff
@@ -0,0 +1,21 @@
...
+  charting {
+    indicators {
+      percentile1 = 50      # Value for the 1st percentile to track in the reports, the console summary and Graphite
+      percentile2 = 95      # Value for the 2nd percentile to track in the reports, the console summary and Graphite
+      percentile3 = 99      # Value for the 3rd percentile to track in the reports, the console summary and Graphite
+      percentile4 = 99.9    # Value for the 4th percentile to track in the reports, the console summary and Graphite
+    }
+  }
+  data {
+    writers = [console, file, graphite] # The list of DataWriters to which Gatling write simulation data (currently supported : console, file, graphite, jdbc)
```
Mario-Lopez: Should this be commented out, too? Or will Gatling not care about not reaching the Graphite (protocol) server?

dmnjohns: Gatling shouldn't care. iptables on the InfluxDB host is setup to drop packets by default, so the Gatling writer should be able to establish a TCP connection, but all of the packets it sends through that connection will be dropped. The Graphite line protocol is an exceptionally simple application-level protocol -- it doesn't respond to the data it receives in any way.

dmnjohns: To verify, I ran a test simulation with this Gatling configuration and it had no problems.

Mario-Lopez: If it's dropping all of the packets, it won't be able to establish a TCP connection. I'm guessing Gatling gives zero cares about being able to send Graphite data then if you got back a zero return code.

Establishing a TCP connection:

create-tcp-connection

Of note, I tried connecting to InfluxDB on ports 13001 and 13002 this morning, and I monitored the network using Wireshark. In both cases, my computer sent a SYN packet to establish a connection (retrying a few times) and never got anything back.

Mario-Lopez: To be clear, since you tested it, I'm good with this staying as is.

dmnjohns: Ah, I see. My misunderstanding was not with how to establish a TCP connection, but rather with the behavior of iptables. I guess I thought iptables would handle the connection with the foreign host, then create a new connection with the local service if allowable based on the rules. It appears that iptables does not take such a complex approach (and I'm not even sure if it's a feasible approach), and instead literally just drops all packets. So iptables returns nothing. This is in contrast to the same server not running iptables, which would return a RST/SYN if a port was not bound or a SYN/ACK if it was.

Of course, iptables can be configured to return a RST/SYN or some other ICMP packet if desired (using REJECT instead of DROP). But that is neither here nor there.

So I was thinking iptables pulled a Repose. That is, Repose manages a connection to the client (through whatever Servlet container) and a separate connection to the origin service (through Apache HTTPClient). Of course, iptables would have to do that at the transport layer, not the application layer. Again, I'm not sure if that's possible... maybe, since iptables uses the kernel firewall.

I  am really not a network person.

PR #1729, opened 2017-04-26
REP-5234 Remote Datastore 

wdschei:


repose-aggregator/src/docs/asciidoc/services/datastores.adoc
```diff
@@ -3,4 +3,293 @@
...
+That is, the distributed datastore may be thought of as a server, and clients in the connection pool as clients.
+Both the client and server need to be aware of how to communicate, and so they both must be configured with the appropriate secrets.
+
+For managing keystores and truststores, the aptly named keytool can be used.
+
+For more details, see:
+
+* http://www.eclipse.org/jetty/documentation/current/configuring-ssl.html#loading-keys-and-certificates
+* https://docs.oracle.com/javase/8/docs/technotes/tools/unix/keytool.html
+* https://docs.oracle.com/javase/8/docs/technotes/guides/security/crypto/CryptoSpec.html#KeystoreImplementation
```
dmnjohns: These links should probably be given shorter link text to make them more human-friendly.

For example:
`http://www.eclipse.org/jetty/documentation/current/configuring-ssl.html#loading-keys-and-certificates[Loading Keys and Certificates]`

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
```diff
@@ -3,4 +3,293 @@
...
+
+=== Distribution
+The distributed datastore shares key-space with all of the enabled cluster nodes.
+Key-space is determined by the maximum value of the distributed datastore's hashing algorithm.
+At current the only supported hashing algorithm is MD5.
```
dmnjohns: Awkward phrasing. `Currently` would work better. Or `Presently`. Or a million other things.

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
```diff
@@ -3,4 +3,293 @@
...
+becomes
+
+`GET /powerapi/dist-datastore/objects/cecda330-5a61-26cd-1a71-d5fe34a8e302`
+
+Obscuring key-space is not a function of the distributed datastore service.
+This functionality is only present in the internally consumed java cache client.
+If an external application puts an object into the distributed datastore, the object will be stored under the value of the key given.
+
+=== Remote Management
+The repose distributed datastore component is a filter component that hosts a simple RESTful API that can be contacted to perform remote object store operations.
```
dmnjohns: A "filter component"? That isn't true anymore.

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
```diff
@@ -3,4 +3,293 @@
...
+If the object does not exist, then a new one created.
+Return the modified/new object.
+The object must be Patchable.
+
+=== Remote Fail-Over
+In the event that a node with in a datastore cluster falls off line or is unable to respond to requests, it is removed from the node's cluster membership for a period of time.
+During this time, the online node will then re-address its key-space in order to continue operation.
+After certain periods of rest, the node may attempt to introduce the damaged cluster member into its cluster membership.
+A damaged cluster member must go through several validation passes where the member is introduced back into the addressing algorithm before it can be considered online.
+In order to keep healthy nodes from attempting to route requests to the damaged node, a participating node may tell it's destination that the destination may not route the request and must store, delete or get the value locally.
```
dmnjohns: "store, delete or get". Not a fan of that phrasing. That seems to try and enumerate the methods that can be performed with a value, but it does not include patch. Even if you read it more generically, patch does not store "the value" locally in all cases. It modifies the existing value and stores that. I would love to see that phrase replaced with something like "handle". So "... and must handle the value locally."

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
```diff
@@ -3,4 +3,293 @@
...
+During times of load this can affect performance and data integrity as when one node cannot communicate with another it will mark that node damaged and store/create information locally.
+One way this can happen is if the user running repose hits their open file limit.
+Luckily this can be mitigated by increasing the open file limit for the user running Repose.
+
+=== JMX Reporting
+Currently Repose instances do not report Distributed Datastore information to JMX.
+This is something that has been done in the past, but an upgrade to the metrics library used has made this capability incompatible with the current codebase.
+
+=== Authentication
+The authentication components will use the local datastore to cache token, role, and group information.
```
dmnjohns: This seems like it doesn't really belong here. Sure, authentication components consume the datastore service, but increasingly so do other components. This has nothing to do with the datastore service, which is data-agnostic, and everything to do with the client component. Therefore, I feel like this should be called out on those other components' respective pages. Same with the rate limiting bit below.

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
```diff
@@ -3,4 +3,293 @@
...
+=== Configuration
+* *Default Configuration:* remote-datastore.cfg.xml
+* *Released:* v8.5.1.0
+* link:../schemas/remote-datastore.xsd[Schema]
+
+At times, Repose instances may need to share information between nodes that are unaware of each other.
+An example of this is a dynamic containerized environment like OpenShift or other 12-Factor environment.
+
+The Remote Datastore Service allows dynamic isolated Repose instances to use a single static Repose instance's object store.
+The Remote Datastore Service communicates to the configure host through the configured port.
```
dmnjohns: `configure` -> `configured`

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
```diff
@@ -3,4 +3,293 @@
...
+             connection-pool-id="remote-datastore"/> <!--5-->
+</remote-datastore>
+----
+<1> Defines the id of the cluster this configuration is for.
+<2> Defines the host providing the remote datastore.
+<3> Defines the port on which the remote datastore is listening (1-65535).
+<4> Defines if the remote datastore is expecting a secure connection request (Default: True).
+<5> Http Connection Pool ID to use when connecting to the Remote Datastore.
+
+NOTE: Refer to the <<http-connection-pool.adoc#,http-connection-pool.cfg.xml>> documentation for setting up Client Authentication on the connection pool used to connect to the Remote Datastore.
```
dmnjohns: Admonition block that doesn't follow the style guide.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/main/java/org/openrepose/core/services/datastore/impl/distributed/CacheRequest.java
```diff
@@ -65,34 +59,16 @@ private static String getCacheKey(HttpServletRequest request) {
...
         return request.getRequestURI().startsWith(CACHE_URI_PATH);
     }
 
     public static String urlFor(InetSocketAddress remoteEndpoint, String key, boolean useHttps) {
-        String proto = "http";
-        if (useHttps) {
-            proto = "https";
-        }
-        return new StringBuilder(proto).append("://").append(remoteEndpoint.getAddress().getHostAddress()).append(":").append(remoteEndpoint.getPort()).append(CACHE_URI_PATH).append(key).toString();
+        return "http" + (useHttps ? "s" : "") + "://" + remoteEndpoint.getAddress().getHostAddress() + ":" + remoteEndpoint.getPort() + CACHE_URI_PATH + key;
```
dmnjohns: Not sure I like doing straight string concatenation here. This probably won't be a method that gets called a bunch, but we may as well write it like it is. That is, there are dynamic components to the string, so I'm pretty sure this line will result in allocating a bunch of intermediate strings.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/main/java/org/openrepose/core/services/datastore/impl/remote/RemoteDatastoreManager.java
```diff
@@ -22,19 +22,19 @@
...
 import org.openrepose.core.services.datastore.Datastore;
+import org.openrepose.core.services.datastore.DatastoreManager;
 import org.openrepose.core.services.datastore.hash.MD5MessageDigestFactory;
-import org.openrepose.core.services.datastore.impl.AbstractDatastoreManager;
 import org.openrepose.core.services.datastore.impl.distributed.remote.RemoteCommandExecutor;
 
 import java.net.InetSocketAddress;
 
-public class RemoteDatastoreManager extends AbstractDatastoreManager {
+public class RemoteDatastoreManager implements DatastoreManager {
```
dmnjohns: Yay!

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/main/java/org/openrepose/core/services/datastore/impl/distributed/CacheRequest.java
Mario-Lopez: Other than the proto/useHttps change, the two lines should result in the same bytecode. The only difference is what we want to see in the code.

dmnjohns: Ah. You're totally right about this creating the same bytecode (I reread [this](http://www.pellegrino.link/2015/08/22/string-concatenation-with-java-8.html), which I think BIll linked before, and is a great read).

I would point out two things that result in relying on compiler optimization, though:

1. Changes to the compiler, or using different compilers, may result in different behavior. This is mostly a moot point because why would we ever not use `javac`? I don't even know of an alternative, actually. As for changes to the compiler, I expect that would only make things better.
2. Utilizing string concatenation here could lead future development to use string concatenation when it is inappropriate. That is, I wouldn't want a future developer (especially a more junior developer) to see/copy this without understanding why it was allowed in this context.

So yeah. I guess, in the end, it doesn't matter too much. There are pros and cons to either approach. I tend to prefer being explicit because it takes away some of the magic (which is why I'm all about static, strongly-typed languages). In this case, that magic is the compiler optimization which just results in using what is essentially a `StringBuilder` anyway. Of course, on the other hand, letting the compiler work its magic means that if the compiler gets better magic in the future, we don't have to change this.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/datastore/RemoteDatastoreServiceTest.groovy
```diff
@@ -0,0 +1,172 @@
...
+        reposeValveLauncher.start(false, false, "repose", type)
+        def artifacts = (new File(runDir, "artifacts")).listFiles(
+                ((FilenameFilter) new WildcardFileFilter("filter-bundle-*.ear")))
+        artifacts.each { file ->
+            def fileOrig = new File(testProperties.reposeHome, "artifacts/${file.name}")
+            def fileNew = new File(runDir, "$subName/artifacts/${file.name}")
+            fileNew.parentFile.mkdirs()
+            def fileDest = new FileOutputStream(fileNew)
+            Files.copy(fileOrig.toPath(), fileDest)
+        }
```
wdschei: The `TestProperties` class was updated to support the parallel testing so that `repose.home` from the `test.properties` is not used if the `testRunDirectory` is supplied.

* https://github.com/rackerlabs/repose/blob/64740bf885967304b4422550e342b95e013c766d/repose-aggregator/tests/functional-tests/src/integrationTest/groovy/framework/TestProperties.groovy#L75

So the short answer is that yes they will look in `${repose.home}/artifacts`, but that isn't where you think it is since we override it for each new Repose instance.

With that said, with the current way we extract all of the artifacts, each Repose instance needs it's own copy of the them. I was talking with @adrianjgeorge a little about this yesterday and he was say the two of you found that we didn't even really need to explode them anymore. If that ends up being the case, then this won't be needed anymore.

repose-aggregator/components/services/datastore-service/datastore-service-api/src/main/java/org/openrepose/core/services/datastore/DatastoreService.java
```diff
@@ -19,13 +19,18 @@
...
 import org.openrepose.core.services.datastore.distributed.DistributedDatastore;
 
+import java.net.InetSocketAddress;
+
 /**
  * DatastoreService - service that manages the lifecycle and configuration of {@link Datastore}s
  */
-@SuppressWarnings("squid:RedundantThrowsDeclarationCheck") //We're defining the contract here, so the extra explicitness is warranted
+@SuppressWarnings("squid:RedundantThrowsDeclarationCheck")
+//We're defining the contract here, so the extra explicitness is warranted
```
wdschei: I don't know. The formatter just didn't want it on the same line. I'm good either way. Since you obviously have a preference, I'll change it out.

repose-aggregator/components/services/datastore-service/datastore-service-api/src/main/java/org/openrepose/core/services/datastore/DatastoreService.java
```diff
@@ -36,34 +41,34 @@
     /**
      * Get a datastore associated with the provided datastore name
      *
-     * @param datastoreName
-     * @return
+     * @param datastoreName unique name for the Datastore
```
wdschei: I'm good with that. Fixed.

repose-aggregator/components/services/datastore-service/datastore-service-api/src/main/java/org/openrepose/core/services/datastore/DatastoreService.java
```diff
@@ -36,34 +41,34 @@
...
     void destroyDatastore(String datastoreName);
 
     /**
      * Create and return a distributed datastore using the provided configuration.  The created
      * datastore can be retrieved by the same name provided using getDatastore(datastoreName)
      *
-     * @param datastoreName
-     * @param configuration
-     * @return
+     * @param datastoreName unique name for this Datastore
```
wdschei: Yep.

repose-aggregator/components/services/datastore-service/datastore-service-api/src/main/java/org/openrepose/core/services/datastore/DatastoreService.java
```diff
@@ -73,15 +78,33 @@ DistributedDatastore createDatastore(String datastoreName, ClusterConfiguration
...
+     * Create and return a remote datastore using the provided configuration.  The created
+     * datastore can be retrieved by the same name provided using getDatastore(datastoreName)
+     *
+     * @param datastoreName    unique name for this Datastore
+     * @param proxyService     Proxy Service for making remote calls
+     * @param encodingProvider Encoding Provider
+     * @param target           the IP Socket Address of the Datastore
+     * @param connPoolId       the name of the pool to borrow a connection from
+     * @param useHttps         indicates if SSL/TLS should be used
+     * @return the newly created Datastore
```
wdschei: Yep. That is how the formatter works. Each type is lined up.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/main/java/org/openrepose/core/services/datastore/impl/remote/RemoteDatastore.java
```diff
@@ -0,0 +1,244 @@
...
+        this.localDatastore = localDatastore;
+        this.remoteCommandExecutor = remoteCommandExecutor;
+        if (localDatastore == null) {
+            LOG.warn("Local datastore is null");
+        }
+    }
+
+    private Object performAction(String name, DatastoreAction action) {
+        return performRemote(name, action);
+    }
```
wdschei: Since nothing is ever being performed locally, I'll keep the `performRemote` name.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/main/java/org/openrepose/nodeservice/remote/RemoteDatastoreLauncherService.java
```diff
@@ -0,0 +1,216 @@
...
+        this.configurationListener = new RemoteDatastoreConfigurationListener();
+    }
+
+    @PostConstruct
+    public void init() {
+        //Subscribe to the system model to know if we even want to turn on...
+        // (If we do want to turn on, we should probably only then subscribe to the remote datastore config)
+        URL systemModelXSD = getClass().getResource(SYSTEM_CONFIG_SCHEMA);
+        configurationService.subscribeTo(SYSTEM_CONFIG_NAME, systemModelXSD, systemModelListener, SystemModel.class);
+        //If and only if we're going to be turned on, should we subscribe to the other one
```
wdschei: Yep. Fixed in both.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/main/java/org/openrepose/nodeservice/remote/RemoteDatastoreLauncherService.java
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    private class SystemModelListener implements UpdateListener<SystemModel> {
+
+        @Override
+        public void configurationUpdated(SystemModel configurationObject) {
+            SystemModelInterrogator smi = new SystemModelInterrogator(clusterId, nodeId);
+            Optional<ReposeCluster> clusterOption = smi.getLocalCluster(configurationObject);
+            if (clusterOption.isPresent()) {
+                boolean listed = smi.getServiceForCluster(configurationObject, "remote-datastore").isPresent();
```
wdschei: Sure. Updated.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/main/java/org/openrepose/nodeservice/remote/RemoteDatastoreLauncherService.java
```diff
@@ -0,0 +1,216 @@
...
+        }
+    }
+
+    private class RemoteDatastoreConfigurationListener implements UpdateListener<RemoteDatastoreConfiguration> {
+        private boolean initialized = false;
+
+        @Override
+        public void configurationUpdated(RemoteDatastoreConfiguration configurationObject) {
+            RemoteDatastoreConfiguration currentRemoteDatastoreConfiguration = currentConfiguration.get();
+            Optional<RemoteClusterConfiguration> oldClusterConfigOpt = currentRemoteDatastoreConfiguration == null ?
```
wdschei: Ok. Updated.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/main/resources/META-INF/schema/remote-datastore/remote-datastore.xsd
```diff
@@ -0,0 +1,101 @@
...
+        </xs:annotation>
+        <xs:sequence>
+            <xs:element name="cluster" type="RemoteClusterConfiguration" minOccurs="1" maxOccurs="unbounded"/>
+        </xs:sequence>
+    </xs:complexType>
+
+    <xs:complexType name="RemoteClusterConfiguration">
+        <xs:annotation>
+            <xs:documentation>
+                <html:p>The root config type for the Remote Datastore configuration file.</html:p>
```
wdschei: Yep. Fixed.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/main/resources/META-INF/schema/remote-datastore/remote-datastore.xsd
```diff
@@ -0,0 +1,101 @@
...
+        <xs:attribute name="port" type="PortNumber" use="required">
+            <xs:annotation>
+                <xs:documentation>
+                    <html:p>
+                        Defines the port on which the remote datastore is listening (1-65535).
+                    </html:p>
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+        <xs:attribute name="useSSL" type="xs:boolean" use="optional" default="true">
```
wdschei: Yep. All the way around. Updated.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/datastore/RemoteDatastoreServiceTest.groovy
```diff
@@ -0,0 +1,172 @@
...
+
+    def cleanupSpec() {
+        deproxy?.shutdown()
+        repose1?.stop()
+        repose2?.stop()
+        remote?.stop()
+    }
+
+    static def startRepose(String subName, int reposePort, int targetPort, int datastorePort, boolean client) {
+        def testProperties = new TestProperties(RemoteDatastoreServiceTest.canonicalName.replace('.', '/') + "/$subName")
```
wdschei: What would you use?

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/datastore/RemoteDatastoreServiceTest.groovy
```diff
@@ -0,0 +1,172 @@
...
+        testProperties.setReposePort(reposePort)
+        testProperties.setTargetPort(targetPort)
+        def reposeConfigProvider = new ReposeConfigurationProvider(testProperties)
+        def reposeValveLauncher = new ReposeValveLauncher(reposeConfigProvider, testProperties)
+        reposeValveLauncher.enableDebug()
+        def reposeLogSearch = new ReposeLogSearch(testProperties.getLogFile())
+        reposeLogSearch.cleanLog()
+        def params = testProperties.getDefaultTemplateParams()
+        def runDir = testProperties.runDirectory
+        params.put('repose.artifact.directory', "$runDir/artifacts")
```
wdschei: This was a left over from before the merge of the parallel testing branch. It isn't needed now that the `runDir` was exposed. Removed.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/datastore/RemoteDatastoreServiceTest.groovy
```diff
@@ -0,0 +1,172 @@
...
+        params.put('datastorePort', datastorePort)
+        reposeValveLauncher.configurationProvider.applyConfigs("common", params)
```
wdschei: I had it in there and it was pointed out that since this is the only test in this snowflake, then it probably wasn't needed. However, since you brought it up and I originally had it in there, then I think it should be there. Added back in.

repose-aggregator/components/services/datastore-service/datastore-service-api/src/main/java/org/openrepose/core/services/datastore/DatastoreService.java
dmnjohns: Hah, I'll take it. 👍

repose-aggregator/components/services/datastore-service/datastore-service-api/src/main/java/org/openrepose/core/services/datastore/DatastoreService.java
dmnjohns: Huh? I meant that the description test (`the newly created Datastore`) does not line up with the description text on the previous line (`indicates if SSL/TLS should be used`). Not a big deal at all, but you know me, I'll point out everything.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/datastore/RemoteDatastoreServiceTest.groovy
dmnjohns: A Java [Path](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html), probably. I basically never want to have to worry about the environment I'm running on, and `Path` lets me avoid that to a large extent. Of course, that would require a bit of rework because everything is a String right now.

dmnjohns: Oh, sorry, I'm not sure if I made that clear. There's nothing wrong with the way you are calling this constructor, I just had a small complain about the parameter that the constructor takes. That's on the other end, though, and not related to your changes.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/datastore/RemoteDatastoreServiceTest.groovy
dmnjohns: I mean, I think doing either would be fair. This was just another manifestation of my desire to be explicit.

But then, I also writes novels to explain even the simplest of things, so take my opinion with a grain of salt.

wdschei: LOL. I too think it is better styling to have it, so it's back.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/main/java/org/openrepose/core/services/datastore/impl/distributed/CacheRequest.java
wdschei: A comment was added for clarity. This was the original IntelliJ IDEA notification that made me look it up again:
```
Reports any usages of java.lang.StringBuffer and java.lang.StringBuilder which can be replaced with a single java.lang.String concatenation.
Using a String concatenation makes the code shorter and simpler.
This inspection only reports when the resulting concatenation is at least as efficient or more efficient than the original StringBuffer or StringBuilder code.
```

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
wdschei: Updated, but admonitions blocks aren't in the style guide yet. :smiley:

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
wdschei: Fixed.

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
wdschei: `Authentication` removed and `Rate Limiting` moved to the `rate-limiting.adoc`.

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
wdschei: **handle** it is.

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
wdschei: **Service** it is.

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
wdschei: `Currently` it is.

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
wdschei: Shorter links with human-friendly names added.

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
```diff
@@ -3,4 +3,293 @@
...
+By both validating the client credentials and encrypting communication with the client, data in the datastore is made more secure.
+
+Assuming all Repose nodes are configured identically, the most straight-forward way to make use of this security would be to use a single unique keystore as both the keystore and the truststore.
+This can be achieved by not explicitly configuring a separate truststore.
+Since each datastore node will have a copy of the keystore, each node will trust every other node.
+
+Client authentication in SSL/TLS can act as as alternate form of client validation, performing a task similar to that of an access control list.
+As such, the usage of client authentication may replace the need to configure the allowed-hosts section of the `dist-datastore.cfg.xml` file.
+
+Note that distributed datastore will use a connection pool to communicate across nodes.
```
wdschei: `Note` swapped with `The`.

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
```diff
@@ -3,4 +3,293 @@
...
+<5> The truststore used for validating the server this pool is connecting to.
+    This is typically set to the same path as the client authentication keystore.
+<6> The password for the client authentication truststore.
+    NOTE: This attribute is only used if the truststore-filename attribute is present.
+<7> Defines a list of hosts who has access to the distributed datastore API calls.
+    This does not add the host to the participating data storage nodes.
+<8> Setting the optional `allow-all` to `true` (defaults to `false`) will turn off host ACL checking.
+<9> Defines a host who has access to the distributed datastore API.
+<10> Defines the `port` on which all datastores in this `cluster` will listen.
+<11> Overrides the default `port`, defined for this `cluster` 's datasotores to listen, only for this particular `node`.
```
wdschei: Fixed.

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
```diff
@@ -3,4 +3,293 @@
...
+    AND the keystore-password and key-password attributes are no longer optional.
+    NOTE: IF the `keystore-filename` is defined,
+    THEN HTTPS will be used;
+    ELSE HTTP will be used.
+<3> The password for the client authentication keystore.
+<4> The password for the particular client authentication key in the keystore.
+<5> The truststore used for validating the server this pool is connecting to.
+    This is typically set to the same path as the client authentication keystore.
+<6> The password for the client authentication truststore.
+    NOTE: This attribute is only used if the truststore-filename attribute is present.
```
wdschei: Backticks and admonition inserted.

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
```diff
@@ -3,4 +3,293 @@
...
+        <port port="9999" cluster="repose"/>                <!--10-->
+        <port port="7777" cluster="repose" node="node2"/>   <!--11-->
+    </port-config>
+</distributed-datastore>
+----
+<1> HTTP Connection pool (ID) to use when communicating with other members of the distributed datastore.
+<2> IF this attribute is configured,
+    THEN it is assumed that it points to the Java keystore containing the client certificate to present for client authentication (e.g keystore.jks)
+    AND the keystore-password and key-password attributes are no longer optional.
+    NOTE: IF the `keystore-filename` is defined,
```
wdschei: Admonition inserted.

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
```diff
@@ -3,4 +3,293 @@
...
+    <port-config>
+        <port port="9999" cluster="repose"/>                <!--10-->
+        <port port="7777" cluster="repose" node="node2"/>   <!--11-->
+    </port-config>
+</distributed-datastore>
+----
+<1> HTTP Connection pool (ID) to use when communicating with other members of the distributed datastore.
+<2> IF this attribute is configured,
+    THEN it is assumed that it points to the Java keystore containing the client certificate to present for client authentication (e.g keystore.jks)
+    AND the keystore-password and key-password attributes are no longer optional.
```
wdschei: Yep. I missed these two. Updated.

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
dmnjohns: Oh man, you're right! I distinctly remember talking about this, and I was convinced I had added it to the style guide. Bad job, self.

wdschei: Why are you answering these???
You are supposed to be off having fun.

adrianjgeorge: No time like the present

Mario-Lopez: – Admiral Kathryn Janeway, alternate timeline

repose-aggregator/components/services/datastore-service/datastore-service-api/src/main/java/org/openrepose/core/services/datastore/DatastoreService.java
wdschei: Right. All of the `@param`'s line up. If there were multiple `@return`'s, which would be way bad and wrong, they would line up. The one's with multiple `@throw`'s should line up too.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/datastore/RemoteDatastoreServiceTest.groovy
wdschei: Right. I was just curious what your thoughts on changing the constructor were.

wdschei:
The one test failure is the flaky one that started acting up a while back:
 - https://repose.atlassian.net/browse/REP-5409

repose-aggregator/src/docs/asciidoc/release-notes.adoc
```diff
@@ -4,6 +4,8 @@ include::_includes/in-progress.adoc[]
 
 == Future Release
 
+* https://repose.atlassian.net/browse/REP-5234[REP-5234] - added the new <<services/datastores.adoc#_remote_datastore, Remote Datastore service>> which allows the Distributed Datastore service concept to work in a dynamic containerized environments like OpenShift.
```
dmnjohns: "... in a dynamic ... environment ..." or "... in dynamic ... environments ...". Mixed plurality here.

wdschei: Fixed.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/datastore/RemoteDatastoreServiceTest.groovy
dmnjohns: Oh, and yeah, not exploding the artifacts would be great (because really, it's unnecessary).

However, it appears that you are, in fact, using the `../repose_home/artifacts/` directory as I expected. The deployment directory is even set to `../respose_home` rather than `../test-homes/...`. I don't see why that wouldn't work, though. That is, I don't see what requires us to provide each Repose instance its own copy of the artifacts. All Repose is going to do is find the artifact files in the artifact directory, extract them to a deployment directory (named with a UUID generated by Repose, and so very likely unique between instances of Repose), and use the URL of that deployment directory to load classes.

```
~/dev/repose  REP-5234_RemoteDatastore ✔                                                                                                                             2d  ⍉
▶ cat repose-aggregator/tests/functional-tests/build/test-homes/features/services/datastore/RemoteDatastoreServiceTest/repose1/configs/container.cfg.xml
<repose-container xmlns='http://docs.openrepose.org/repose/container/v2.0'>
    <deployment-config>
        <deployment-directory auto-clean="true">
            /Users/dami6356/dev/repose/repose-aggregator/tests/functional-tests/build/repose_home
        </deployment-directory>
        <artifact-directory check-interval="1000">
            /Users/dami6356/dev/repose/repose-aggregator/tests/functional-tests/build/repose_home/artifacts
        </artifact-directory>
        <logging-configuration href="file:///Users/dami6356/dev/repose/repose-aggregator/tests/functional-tests/build/test-homes/features/services/datastore/RemoteDatastoreServiceTest/repose1/configs/log4j2-test.xml"/>
    </deployment-config>
</repose-container>
```

Since the test works, I'm thinking we don't need to do this copy at all. If we are going to do it, though, we need to fix it.

wdschei: Yep. The issue I was having appears to have been fixed with the exposing of the `repose.run.directory`. The copying of the artifacts has been removed.

PR #1731, opened 2017-05-02
REP-5412 Secure listening port for InfluxDB Graphite protocol

dmnjohns:


repose-aggregator/tests/performance-tests/roles/cloud_resource_create/tasks/create_network.yml
```diff
@@ -0,0 +1,8 @@
+- name: Create the Cloud Network for private communication between resources
+  local_action:
+    module: rax_network
+    credentials: ~/.raxpub
+    label: PerfNet
+    cidr: 192.168.3.0/24
+    region: IAD
+    state: present
```
dmnjohns: Truthfully, this network should always be present already. After all, static infrastructure like the Grafana and InfluxDB hosts already belong on it.

repose-aggregator/tests/performance-tests/roles/gatling/files/gatling.conf
```diff
@@ -12,10 +12,9 @@ gatling {
     }
   }
   data {
-    writers = [console, file, graphite] # The list of DataWriters to which Gatling write simulation data (currently supported : console, file, graphite, jdbc)
+    writers = [console, file, graphite]          # The list of DataWriters to which Gatling write simulation data (currently supported : console, file, graphite, jdbc)
     graphite {
-      host = "influxdb.openrepose.org"  # The host where the Carbon server is located
-      port = 13002                      # The port to which the Carbon server listens to (2003 is default for plaintext, 2004 is default for pickle)
+      host = "internal.influxdb.openrepose.org"  # The host where the a Graphite listener is located
```
dmnjohns: We talked about this a bit on Slack, but I created a public DNS entry to provide us with a static address to the InfluxDB host. Just pointing it out here.

Mario-Lopez: Drive by comment:
I would have expected: `influxdb.internal.openrepose.org`

InfluxDB shouldn't have a lot of DNS entries (only ever the private address), but we may add more internal entries.

dmnjohns: Fair enough. I didn't give it much thought, but that makes sense. I'll change it.

adrianjgeorge: How about influxdb.performance.openrepose.org? Since we've clearly delineated this network as being for performance stuff for now

Mario-Lopez: Even better since ideally we'll create another internal network for... other... stuff.

Mario-Lopez: Actually, could you make it influxdb.🗝.openrepose.org to make it really secure?

dmnjohns: But that isn't at least 12 characters long. Nor does it include alphabetical characters both uppercased and lowercased. Or a number. Therefore it is insecure?

repose-aggregator/tests/performance-tests/roles/cloud_resource_create/tasks/create_network.yml
adrianjgeorge: I'm assuming this is idempotent

repose-aggregator/tests/performance-tests/roles/cloud_resource_create/tasks/create_server_gatling.yml
```diff
@@ -6,7 +6,11 @@
     flavor: performance1-4
     image: "{{ cloud.server.image }}"
     files: "{{ cloud.server.files }}"
-    region: "{{ cloud.region }}"
+    region: IAD
+    networks:
+      - public
+      - private
```
adrianjgeorge: Is private the correct label for the service net?

repose-aggregator/tests/performance-tests/roles/cloud_resource_create/tasks/create_network.yml
dmnjohns: It sure is.

repose-aggregator/tests/performance-tests/roles/cloud_resource_create/tasks/create_server_gatling.yml
dmnjohns: Seemingly. I took `public` and `private` from the docs here:

>>> The network to attach to the instances. If specified, you must include ALL networks including the public and private interfaces. Can be id or label.

http://docs.ansible.com/ansible/rax_module.html

repose-aggregator/src/docs/resources/images/performance-test-framework.gliffy
```diff
@@ -0,0 +1 @@
...der":2,"name":"Interactions","active":false,"locked":false,"visible":true,"nodeIndex":18},{"guid":"8pbi2R4AZnwP","order":1,"name":"Hosts","active":true,"locked":false,"visible":true,"nodeIndex":94},{"guid":"TIAblKBtzODd","order":0,"name":"Base","active":false,"locked":false,"visible":true,"nodeIndex":26}],"shapeStyles":{},"lineStyles":{"global":{"fill":"none","stroke":"#980000","strokeWidth":2,"endArrow":2,"orthoMode":0,"startArrow":0}},"textStyles":{"global":{"face":"Arial","size":"12px","color":"#000000"}}},"metadata":{"title":"untitled","revision":0,"exportBorder":false,"loadPosition":"default","libraries":["com.gliffy.libraries.mru","com.gliffy.libraries.basic.basic_v1.default","com.gliffy.libraries.flowchart.flowchart_v1.default","com.gliffy.libraries.swimlanes.swimlanes_v1.default","com.gliffy.libraries.uml.uml_v2.class","com.gliffy.libraries.uml.uml_v2.sequence","com.gliffy.libraries.uml.uml_v2.activity","com.gliffy.libraries.erd.erd_v1.default","com.gliffy.libraries.ui.ui_v3.containers_content","com.gliffy.libraries.ui.ui_v3.forms_controls","com.gliffy.libraries.images"],"autosaveDisabled":false,"lastSerialized":1493933904951},"embeddedResources":{"index":0,"resources":[]}}
```
adrianjgeorge: This looks like straight json. Can gliffy still load it if you format it? I ask because I'd hope that they don't shift things around in the output too much, and it'd be much easier to see changes made later if that's true.

dmnjohns: Hm, fair question. Let me test that.

adrianjgeorge: I can't comment on the diagram directly, but are the load balancers and endpoints not also talking over the private network?

dmnjohns: Sure enough, Gliffy has no problem loading if it it's formatted. I can't speak to how much the output changes, but storing a formatted version makes sense to allow for better isolation of changes. I'll push that up.

adrianjgeorge: Not certain you saw the comment about the arrows since this comment got hidden.

dmnjohns: The arrows?

I missed the private network comment, if that's what you mean. So the load balancers and endpoints are not currently talking over the private network. I'm actually not sure why I added the Repose servers to the private network in the first place. I should probably remove that to avoid confusion.

I know Mario wants to do more work to get all of our performance infrastructure onto a private network (which would be awesome), but that was not accomplished as part of this story. I should also note that the performance test servers themselves are not really locked down -- neither basic authentication nor a firewall is set up. I suppose that is slightly mitigated by the fact that they are transient servers.

PR #1734, opened 2017-05-10
REP-5345 Attribute mapping validation on PUTs

dmnjohns:


repose-aggregator/components/filters/attribute-mapping-policy-validation-filter/build.gradle
```diff
@@ -0,0 +1,18 @@
+dependencies {
+    compile project(":repose-aggregator:commons:commons-utilities")
+    compile "org.scala-lang:scala-library"
+    compile "com.typesafe.scala-logging:scala-logging-slf4j_$scalaMajDotMin"
+    compile "com.rackspace.identity.components.attribute-mapper:mapper-core"
+    //todo: remove this once we can move the translation filter up to the same version of saxon
+    compile "net.sf.saxon:Saxon-EE:9.7.0-8"
+    compile "com.fasterxml.jackson.core:jackson-databind"
```
dmnjohns: Unfortunately, the Attribute Mapper library API forces us to deal with the Saxon, Jackson, and Java XML APIs as well. That is why these dependencies are here.

repose-aggregator/artifacts/extensions-filter-bundle/src/main/application/WEB-INF/web-fragment.xml
```diff
@@ -32,4 +32,8 @@
         <filter-name>saml-policy</filter-name>
         <filter-class>org.openrepose.filters.samlpolicy.SamlPolicyTranslationFilter</filter-class>
     </filter>
+    <filter>
+        <filter-name>attribute-mapping-policy-validation</filter-name>
```
dmnjohns: I have been told that this name is quite long. I am open to other suggestions, though I'm also fine leaving it as-is since it is only for one user.

repose-aggregator/components/filters/attribute-mapping-policy-validation-filter/src/main/scala/org/openrepose/filters/attributemapping/AttributeMappingPolicyValidationFilter.scala
```diff
@@ -0,0 +1,122 @@
...
+
+@Named
+class AttributeMappingPolicyValidationFilter extends Filter with LazyLogging {
+
+  import AttributeMappingPolicyValidationFilter._
+
+  // Create a transform factory using the default implementation and the classloader for this class
+  private val transformerFactory = TransformerFactory.newInstance(
+    "org.apache.xalan.xsltc.trax.TransformerFactoryImpl",
+    this.getClass.getClassLoader)
```
dmnjohns: This was done in the SAML filter, so I did it here, too. I don't want to have to deal with any class loader issues...

`org.apache.xalan.xsltc.trax.TransformerFactoryImpl` is actually the default `TransformerFactory` that you would get if you call `newInstance()`. This is just to get that factory to use the classloader for this class.

repose-aggregator/artifacts/extensions-filter-bundle/src/main/application/WEB-INF/web-fragment.xml
wdschei: I like it. It deters users from using it that shouldn't and the one that should only has to write it once.

repose-aggregator/components/filters/attribute-mapping-policy-validation-filter/src/main/scala/org/openrepose/filters/attributemapping/AttributeMappingPolicyValidationFilter.scala
```diff
@@ -0,0 +1,122 @@
...
+  override def destroy(): Unit = {}
+
+  override def doFilter(request: ServletRequest, response: ServletResponse, chain: FilterChain): Unit = {
+    val httpServletRequest = request.asInstanceOf[HttpServletRequest]
+    val httpServletResponse = response.asInstanceOf[HttpServletResponse]
+    val requestContentType = httpServletRequest.getContentType
+
+    // Buffer the request payload so that we can validate the policy, but forward the original payload
+    val requestPayloadBuffer = new BufferedServletInputStream(httpServletRequest.getInputStream)
+    requestPayloadBuffer.mark(Integer.MAX_VALUE)
```
dmnjohns: This allows us to reset the stream after it is validated. That means that we aren't doing any normalization -- the reason being that it is non-trivial to accomplish for JSON requests, we already validate the policy on application (in the SAML filter), and that was not part of the acceptance criteria for the story. I admit that normalizing is probably the right way to go, but I figured we can iterate on what is here.

repose-aggregator/components/filters/attribute-mapping-policy-validation-filter/src/main/scala/org/openrepose/filters/attributemapping/AttributeMappingPolicyValidationFilter.scala
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  def getPolicyAsXmlSource(contentType: String, policy: InputStream): Try[Source] = {
+    val contentTypeLowerCase = contentType.toLowerCase
+    val requestStreamSource = new StreamSource(policy)
+
+    if (contentTypeLowerCase.contains("json")) {
+      Try {
+        val outPolicyXml = new XdmDestination
+        AttributeMapper.policy2XML(requestStreamSource, outPolicyXml)
```
RackerWilliams: Why don't you push this logic into attribute mapper?  Really you just need a validate call.

This will shield you from changes to the attribute mapper overtime.  Additionally, I want to add better validation rules for JSON, this would prevent me from doing that work.

repose-aggregator/components/filters/attribute-mapping-policy-validation-filter/src/main/scala/org/openrepose/filters/attributemapping/AttributeMappingPolicyValidationFilter.scala
RackerWilliams: Not sure why you have to use this.  We don't use XSLTC anywhere in the code in attribute mapping.

dmnjohns: Oops. You're right, this isn't actually being used. It's vestigial code from when I was attempting to translate the `Source` coming out of `validatePolicy(...)` back into JSON.

When I was using this `TransformerFactory`, I simply copied the instantiation line we had in `SamlPolicyTranslationFilter`. I don't remember exactly what issue we were seeing there with the default class loader.

At any rate, I can remove this.

repose-aggregator/components/filters/attribute-mapping-policy-validation-filter/src/main/scala/org/openrepose/filters/attributemapping/AttributeMappingPolicyValidationFilter.scala
dmnjohns: My primary motivation was to save time. I wanted to get this story wrapped up by end-of-sprint, which is today. Pushing this logic into attribute mapper would require that I spend a bit of time figuring out more about that library (e.g., the testing framework). It would also require a release of that library.

Instead, I took the approach I would take with any third-party library -- I leveraged the functionality that currently exists. Of course, that does not prevent me (or us) from making a PR to that library in the future and adapting our code when we do. That is, I looked at this iteratively and chose to make a small first step, after which more steps could follow.

RackerWilliams: Okay makes sense.  Let's get this out there, for now.  I'll work on adding the JSON validation call and open up another story in Repose to integrate.

dmnjohns: Sounds great!

*wipes brow*

repose-aggregator/components/filters/attribute-mapping-policy-validation-filter/src/test/scala/org/openrepose/filters/attributemapping/AttributeMappingPolicyValidationFilterTest.scala
```diff
@@ -0,0 +1,241 @@
...
+
+        chain.getRequest shouldBe theSameInstanceAs(request)
+        chain.getResponse shouldBe theSameInstanceAs(response)
+      }
+    }
+  }
+}
+
+object AttributeMappingPolicyValidationFilterTest {
+  final val HttpMethods = Set(
```
wdschei: These are just the methods not supported by this filter. Maybe `UnsupportedHttpMethods` would be a better name.

repose-aggregator/src/docs/asciidoc/filters/attribute-mapping-policy-validation.adoc
```diff
@@ -0,0 +1,58 @@
+= Attribute Mapping Policy Validation Filter
+
+include::../_includes/in-progress.adoc[]
```
wdschei: Since this is a brand new filter, pointing them back to the old Wiki is probably useless.

repose-aggregator/src/docs/asciidoc/filters/attribute-mapping-policy-validation.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+This filter will validate attribute mapping policies in requests to the origin service.
+
+The origin service is expected to be an Identity service.
+This filter will only take action if the HTTP request method is `PUT`.
+
+== General filter information
+* *Name:* attribute-mapping-policy-validation
+* *Released:* TBD
```
wdschei: We probably want to add a task to the next release story to come back and update this.

repose-aggregator/components/filters/attribute-mapping-policy-validation-filter/src/test/scala/org/openrepose/filters/attributemapping/AttributeMappingPolicyValidationFilterTest.scala
dmnjohns: Oops. It used to be all HTTP methods, but I changed the tests a bit and that is what this became. I'll change it.

repose-aggregator/src/docs/asciidoc/filters/attribute-mapping-policy-validation.adoc
dmnjohns: Indeed. This is also no longer in progress. I'll remove it.

repose-aggregator/src/docs/asciidoc/filters/attribute-mapping-policy-validation.adoc
dmnjohns: Yeah, when I wrote that, I was wondering how we have handled this scenario up until now. Have we really just not released any new filters since moving to versioned docs?

Anyway, I agree. I don't think the next release story exists yet, though, so I'm not sure where to track that...

repose-aggregator/src/docs/asciidoc/filters/attribute-mapping-policy-validation.adoc
```diff
@@ -0,0 +1,56 @@
...
+None.
+
+=== Response Status Codes
+[cols="2", options="header,autowidth"]
+.Status Codes
+|===
+| Status Code
+| Reasons
+
+| `400`
```
wdschei: I would love to see us start using the full name of the Response status like this:

* _Bad Request_ (`400`)
* _Unsupported Media Type_ (`415`)

It is kind of two fold, it reinforces the intent of why we chose a particular code and it also educates the user a bit without forcing them to lookup every code.

Mario-Lopez: That seems really noisy to me. I would prefer the status codes be links with that info or maybe have tooltips (although a quick Google shows this isn't supported by Asciidoctor).

Mario-Lopez: That is to say, I would prefer a solution that keeps the default view concise enough that seasoned developers will not gloss over our documentation and that provides easy access to supplemental information for developers that may not understand a specific lingo.

dmnjohns: I don't particularly care either way. If we go with @wdschei's approach, I would propose the `400 Bad Request` format. If we go with @Mario-Lopez's approach, I assume we would link to RFC2616 (and the update status code RFC) like `[400](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.1)`? So... everyone fight, the winner decides!

However, if we are going to add more static information to the status code, I would want a way to do so automatically (like as a pre-processing stage to building the docs). I'm not sure that's feasible, though.

repose-aggregator/components/filters/attribute-mapping-policy-validation-filter/src/main/scala/org/openrepose/filters/attributemapping/AttributeMappingPolicyValidationFilter.scala
```diff
@@ -0,0 +1,146 @@
...
+
+import scala.util.{Failure, Success, Try}
+
+@Named
+class AttributeMappingPolicyValidationFilter extends Filter with LazyLogging {
+
+  import AttributeMappingPolicyValidationFilter._
+
+  // Magic
+  private val transformerFactory = new net.sf.saxon.TransformerFactoryImpl
```
dmnjohns: To everyone's horror, this is back.

I am being explicit about the class, because it really, really seems to matter. This is what got the XML `Source` to `InputStream` code working. It literally would not work with the Xalan `TransformerFactory` I was getting out of `javax.xml.transform.TransformerFactory.newInstance()`. The only reason I even tried this is because I noticed two things:

1. When using `javax.xml.transform.TransformerFactory.newInstance().newTransformer().transform(...)` (abbreviated into one line just for convenience here), I noticed that transform failed because the `Transformer` thought the `Source` was empty.
1. This is how Jorge was getting an Identity `Transformer` in the Attribute Mapper library.

So I figured there may be a difference in the `Transformer` being returned by the `TransformerFactory`. Even though from what I saw in the code, both `TransformerFactory`s should have been giving me an Identity `Transformer` by default. I did not want to try to learn everything about the Java XML APIs and the implementations thereof, so I did a quick test with the `TransformerFactory` replaced. Things worked, tests passed. It went in the bucket.

DON'T JUDGE ME.

repose-aggregator/components/filters/attribute-mapping-policy-validation-filter/src/main/scala/org/openrepose/filters/attributemapping/AttributeMappingPolicyValidationFilter.scala
```diff
@@ -0,0 +1,146 @@
...
+        // Wrap the policy bytes with a ServletInputStream
+        new ServletInputStreamWrapper(new ByteArrayInputStream(jsonOutput.toByteArray))
+      }
+    } else if (contentTypeLowerCase.contains("xml")) {
+      Try {
+        // Transform the normalized policy Source into bytes
+        val transformer = transformerFactory.newTransformer()
+        val xmlOutput = new ByteArrayOutputStream()
+        val xmlStreamResult = new StreamResult(xmlOutput)
+        transformer.transform(policy, xmlStreamResult)
```
dmnjohns: No, but seriously though. If anyone knows of a better way to do this, please, please let me know.

repose-aggregator/components/filters/attribute-mapping-policy-validation-filter/src/main/scala/org/openrepose/filters/attributemapping/AttributeMappingPolicyValidationFilter.scala
RackerWilliams: I'm very confused as to why you need a transformer in the first place. Will look this over tomorrow.

dmnjohns: The `Transformer` is being used to write the normalized XML policy `Source` returned from the `validatePolicy(...)` call to an `OutputStream`. The transformation itself is just the identity transformation. Once we have an `OutputStream` with the bytes from the `Source`, we can wrap the bytes from it in an `InputStream` that will be stuck in the `HttpServletRequest` and forwarded to the next filter in the filter chain.

In other words, I'm only using the `Transformer` to go from `Source` to `InputStream`.

repose-aggregator/src/docs/asciidoc/filters/attribute-mapping-policy-validation.adoc
wdschei: Given that Asciidoctor doesn't support tooltips for text at this time, only images, I'm good with the current format. I was just throwing the idea out there.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/attributemappingvalidation/AttributeMappingPolicyValidationFilterTest.groovy
```diff
@@ -0,0 +1,193 @@
...
+    @Unroll
+    def "#method should hit origin service #times times"() {
+        given:
+        Map headers = body ? ["content-type" : TEXT_PLAIN] : [:]
+
+        when:
+        MessageChain mc = deproxy.makeRequest(url: reposeEndpoint, method: method, headers: headers, requestBody: body)
+
+        then:
+        mc.handlings.size() == times
```
wdschei: Shouldn't we care that the body is not modified if it makes it to the origin?
```
if (times > 0 && StringUtils.isNotBlank(body)) {
    mc.handlings[0].request.body as String == body
}
```

repose-aggregator/src/docs/asciidoc/filters/attribute-mapping-policy-validation.adoc
dmnjohns: Always appreciated, sir.

repose-aggregator/components/filters/attribute-mapping-policy-validation-filter/src/main/scala/org/openrepose/filters/attributemapping/AttributeMappingPolicyValidationFilter.scala
RackerWilliams: You have to understand the internals of Saxon to know how to avoid using a transformer, but here's how you would do it...

````scala
  AttributeMapper.processor.writeXdmValue (new XdmNode(policy.asInstanceOf[NodeInfo]),  
   AttributeMapper.processor.newSerializer(xmlOutput))
````

You have to know that the source is of type ```NodeInfo``` which is only true if you're using Saxon.  The approach you took will pretty much always work though.

PR #1735, opened 2017-05-12
REP-5399 Add Documentation for new JMX Metrics names

dmnjohns:


repose-aggregator/src/docs/asciidoc/services/metrics.adoc
```diff
@@ -99,3 +114,63 @@ meter.mark(); // <3>
...
+| Power Filter Router
+| Meter
+| `org.openrepose.core.ResponseCode.<endpoint>.<response-code>`
+| Counts the number of responses with a `<response-code>` status code returned by `<endpoint>`.
+`<response-code>` is the HTTP status code in the response from `<endpoint>`.
+`<endpoint>` is the endpoint that the request was made to.
+
+| Power Filter Router
+| Meter
+| `org.openrepose.core.ResponseCode.ALL ENDPOINTS.<response-code>`
```
wdschei: Did we really put a space in `ALL ENDPOINTS`?

dmnjohns: We really did. However, thanks for calling this out. We actually case those words differently from how we case `ACROSS ALL`. This should say `All Endpoints`.

Because that is clear.

wdschei: Nothing like consistency. :stuck_out_tongue_winking_eye:

wdschei: While looking at the emoji list, I just found @adrianjgeorge 's favorite quote is now supported:
* /play butts

wdschei: GitHub links out to that page emoji page, but is doesn't actually embed them correctly it doesn't look like.

Mario-Lopez: It looks like that section is only for Campfire. No cool sounds for us. 😢

PR #1738, opened 2017-05-17
REP-5411 Modify Gatling Output Name

dmnjohns:


dmnjohns:
So as @Mario-Lopez pointed out yesterday, this does not give us separation between the languages used by the scripting filter performance test.

If we want to keep the current test structure, I suggest we achieve that separation by including the language in the scenario name. I can add that to this PR if that is the way we want to go.

dmnjohns:
Since the scripting language information is encoded in Ansible itself (as a test variable and directory) I had to make the change in Ansible (and thus, in the Gatling output name) rather tahn in the HTTP request name.

Since changing the `gatling.test.name` would have a number of complications, I just added a `gatling.test.id` that the scripting filter tests will use (although others could as well, if other tests end up in a similar situation). Then, the Gatling output name is set to the id if present, otherwise it is set to the name.

dmnjohns:
Retest this please.

dmnjohns:
I'm rerunning the tests that failed, but both passed locally.

Considering that my changes were exclusively to Ansible stuff, and Ansible isn't being run as part of the tests yet, I would say that this is probably good to go at any time.

wdschei:
If they pass locally, then I concur.

PR #1739, opened 2017-05-22
Updating the directory search for new output name

dmnjohns:


Mario-Lopez:
Should we also update the docs? Or do we plan on removing support for the Gatling HTML report in the near future anyway?

https://github.com/rackerlabs/repose/blame/master/repose-aggregator/src/docs/asciidoc/architecture/performance.adoc#L223

PR #1740, opened 2017-05-22
Removed Ruby and Lua support from Scripting filter docs (old versions)

Mario-Lopez:
Updated older versions HTML.


adrianjgeorge:
were there not examples of lua and ruby on those pages?

Mario-Lopez:
Nope. They were left as an exercise for the student.

I verified that all references to "ruby" and "lua" on those pages were removed. There wasn't much.

PR #1756, opened 2017-06-12
REP-5574 Dynamically determine list of performance tests to run

dmnjohns:
I used a new Jenkins plugin to provide the `findFiles` step. Please see:
https://github.com/rackerlabs/repose-infrastructure-ng/pull/67


repose-aggregator/tests/performance-tests/test_vars/filters/scripting.yml
```diff
@@ -1,3 +1,8 @@
+# TODO: Remove this file? Don't want it to get picked up by the automated performance tests, but
+# TODO: keeping this file gives us a sensibly named way to run generic scripting tests.
+#
+# TODO: YAML does not seem to support include or inherit. As such, there is more duplication than
+# TODO: necessary because each scripting language only changes the "script_lang" variable.
```
dmnjohns: Just pointing this out. Removing this file would allow us to remove the exception case in the determination logic. On the other hand, keeping the file makes some sort of sense if we will ever run the scripting filter test manually.  I lean toward just removing this file, and having to use the predefined language files even if we ever run a scripting test manually.

To the second point, I haven't been able to find a good solution. If anyone knows comes up with something, just let me know.

Mario-Lopez: I'm not sure I understand the case for keeping this file. If we delete it, we're talking about going from this:

```
ansible-playbook site.yml \
  -e '{perf_test: filters/scripting,
       script_lang: python}'
```

to this:

```
ansible-playbook site.yml \
  -e '{perf_test: filters/scripting/python}'
```

right?

I suppose someone might do something ridiculous like:

```
ansible-playbook site.yml \
  -e '{perf_test: filters/scripting/groovy,
       script_lang: python}'
```

Which we can't stop, but either way I'm fine with deleting this file.

dmnjohns: Correct on all points. So I guess the use-case would be if you wanted to throw a lot of custom stuff at the performance test framework. For example, say you wanted to ad-hoc testing Jaskell with the scripting filter. You could pass a `script_lang` of `jaskell` and custom `config_files` and without following the conventions of the performance testing framework, you could run a test using Jaskell. In that case, it would be more confusing to have to do what you did in your last example -- customize the `script_lang` for a test already named with a scripting language. Of course, that's a pretty advanced example, and honestly, that test should probably just conform to the conventions instead.

Which is all to say, I agree with you, so I will delete this file.

Mario-Lopez: I guess I do see the case a little bit more. In theory (after we're done-done with the perf tests), you should only have to create a new performance test if you create a new filter (or write some logic that should be performance tested).

By deleting this file, we also have to add "or add a new scripting language" to the list. I don't think it's that big of a deal, though. Still good with deleting this file (which has already been done).

dmnjohns:
I added a description that points to the infrastructure PR to add a Jenkins plugin that I used.

repose-aggregator/tests/performance-tests/Jenkinsfile
```diff
@@ -9,36 +9,33 @@ properties([
...
+        checkout scm
+        dir("repose-aggregator/tests/performance-tests/test_vars") {
+            def files = findFiles(glob: "**/*.yml")
+            for (int index = 0; index < files.size(); index++) {
+                perfTests.add(files[index].path.replace(".yml", ""))
+            }
+        }
+    }
+}
+
```
wdschei: I like this new dynamic way much better than the old static way.

repose-aggregator/tests/performance-tests/test_vars/filters/scripting/javascript.yml
```diff
@@ -0,0 +1,29 @@
+{
+  dir_prefix: "filters/scripting",
+  script_lang: "javascript",
+  cloud: {
+    naming_prefix: "perf-filters-scripting-{{ script_lang }}"
```
wdschei: Although it does lead to a little extra boiler plate around the test, but a required file will never be forgotten. Or at least caught fairly early as opposed to forgetting to add it to a list in some other file.

PR #1760, opened 2017-06-21
REP-5419 Ignored Tests

dmnjohns:
On-going effort to un-ignore, fix, or remove tests annotated with `@Ignore`.

I'm opening this PR just to track what work has been done. It isn't necessarily ready to be merged.


repose-aggregator/external/pjl-compressing-filter/src/test/java/org/openrepose/external/pjlcompression/LogTest.java
```diff
@@ -49,21 +48,6 @@
...
 public final class LogTest {
 
-    /**
-     * this test tests a commons logging implementation that we are *NOT* using
-     *
-     * @throws Exception
-     */
-    @Test
-    @Ignore
-    public void testJakartaImpl() throws Exception {
```
dmnjohns: Pretty sure this isn't necessary. We don't use the Jakarta logging for anything, and a search in IntelliJ doesn't show that class as existing on the classpath. Also, reflectively testing a constructor? No idea what the goal was there.

If I'm wrong, though, someone can correct me.

repose-aggregator/components/filters/rate-limiting-filter/src/test/java/org/openrepose/filters/ratelimiting/SchemaTest.java
```diff
@@ -87,19 +87,5 @@ public void shouldValidateAgainstLiveLimitsExample() throws Exception {
             assertThat("Expected element should have child elements",
                     jaxbUnmarshaller.unmarshal(sampleSource, Limits.class).getValue().getRates().getRate().size(), greaterThan(0));
         }
-
-        @Test
-        @Ignore
-        public void output() throws Exception {
```
dmnjohns: Also not necessary. This isn't actually a test (which is probably why it was ignored). It just prints an example as both XML and JSON. Considering it was brought in with the very first commit, I don't see any value here. You could argue that it tests that transformer, but that really isn't the job of this test. Furthermore, the transformer methods called here are not called anywhere else. So really, I should go delete those methods. Maybe that entire class? I'll try that and see if anything breaks.

dmnjohns: Everything still seemed to work. I would let it run through the full integration tests, of course.

adrianjgeorge:
These all seem legit, and pass tests. I lean towards merging them, any reason not to?

dmnjohns:
Seems legit to me.

PR #1767, opened 2017-06-27
REP-5645 Performance tests now with random quiet periods

adrianjgeorge:
Sets the quiet period to a random length between 15 seconds and 10 minutes. Note that we may still hit rate limiting, but it should be pretty unlikely at this point.

### DIFF
--- a/repose-aggregator/tests/performance-tests/Jenkinsfile
+++ b/repose-aggregator/tests/performance-tests/Jenkinsfile
@@ -27,6 +27,8 @@ stage("Performance Test") {
     def performanceJenkinsJob = "performance-test"
     def performanceTestParam = "perf_test"
     def extraOptionsParam = "extra_options"
+    def tenMinutesInSeconds = 600
+    def random = new Random()
 
     def jobsToRun = [:]
 
@@ -34,7 +36,9 @@ stage("Performance Test") {
         def perfTest = perfTests[index]
         jobsToRun[perfTest] = {
             retry(3) {
-                build(job: performanceJenkinsJob, parameters: [
+                build(job: performanceJenkinsJob,
+                      quietPeriod: random.nextInt(tenMinutesInSeconds),
+                      parameters: [
                         [$class: "StringParameterValue", name: performanceTestParam, value: perfTest],
                         [$class: "StringParameterValue", name: extraOptionsParam, value: params.extraOptions]])
             }


repose-aggregator/tests/performance-tests/Jenkinsfile
```diff
@@ -27,14 +27,17 @@ stage("Performance Test") {
...
 
     def jobsToRun = [:]
 
     for (int index = 0; index < perfTests.size(); index++) {
         def perfTest = perfTests[index]
         jobsToRun[perfTest] = {
             retry(3) {
-                build(job: performanceJenkinsJob, parameters: [
+                build(job: performanceJenkinsJob,
+                      quietPeriod: (random.nextInt() % 585) + 15,
```
dmnjohns: Can't `nextInt()` return a negative integer value? In which case, this could set the `quietPeriod` to a negative value. Not sure what the resultant behavior is, but it probably isn't what we want.

dmnjohns: Side note, I'm glad there is a parameter for the `quietPeriod`. I think we tried to configure the job itself, but I did not realize that was something we could set here. That works quite well.

dmnjohns: I would almost suggest giving the magic numbers named variables, but the behavior is pretty obvious, and it's not like we are likely to modify this file frequently. I would take it either way.

adrianjgeorge: That's true looking at the contract. it seems to just fire off immediately in that case, based on the behavior i saw earlier. do we actually need the 15 second minimum quiet period? I can trivially set it to anything within 10 minutes or i can keep this plus add an absolute value to it.

dmnjohns: I'm not sure that we do (need the minimum quiet period). Theoretically, it's possible, though very unlikely, that all jobs run at the same time regardless of whether or not we add a minimum quiet period. So what does the minimum quiet period buy us? Setting it to anything within 10 minutes seems legit to me, but if you've got a better idea, go for it.

The thing we should avoid, though, is a bunch of tests running at once because they are all given negative quiet periods. Well, I guess there's a 50% chance of any individual test getting a negative quiet period at the moment, but still.

adrianjgeorge: Yarp, passing a range to`nextInt` automatically ensures positive values.

adrianjgeorge: the script doesn't like constant, but is till pulled it out into a variable

PR #1773, opened 2017-07-19
REP-5737 IP User Schema bug

wdschei:


repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/ipuser/IpUserTest.groovy
```diff
@@ -89,4 +93,23 @@ class IpUserTest extends ReposeValveTest {
         !group.contains("local-lan-ip;q=0.6")
     }
 
+    def "should fail to start with bad config"() {
```
dmnjohns: I wonder if this should be broken out into its own test, for the sake of consistency.

wdschei: I was thinking about it, but this test suite is written in a very odd manner and doesn't follow our normal patterns anyway. That said, I don't have any heartache splitting this out for consistency. I guess this is as good a place as any to start making this thing right.

dmnjohns: Yeah, that's kind of where I'm at. My only concern is that, at some point in the future, we will check which filters do and do not have this bad config test, and the way we check may be to look for a certain test file. Or, more generally, forming more patterns makes searching much easier. So I vote for breaking it out, even though you're right.

PR #1774, opened 2017-07-24
REP-5748 Phone Home service NPE fix.

wdschei:


wdschei:
The functional test was added to an existing test that shouldn't have passed:
> https://github.com/rackerlabs/repose/blob/d8496b206995378171301a949f8d920365735d5f/repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/phonehomeservice/PhoneHomeServiceTest.groovy#L77

dmnjohns:
Ah, fair enough. I thought since the task was still open with 3 hours, we wanted to do something more comprehensive there. Maybe verify that the mock phone home service actually receives the message rather than relying on logs.

wdschei:
I didn't want to close it out in case we decided to go that route. I'm good either way. I updated the test and it failed with the NPE being logged. I updated the order in the service and then it passed.

dmnjohns:
Fine by me.

PR #1776, opened 2017-07-25
Updated the Jetty version.

wdschei:


wdschei:
I also tried the latest, v9.4.6.v20170531, and the latest 9.3.x, v9.3.20.v20170531, but these will require some rework to make them usable for our purposes.

wdschei:
This is a synopsis of the conversation in Slack:

Unfortunately we made the `Scripting` filter to clever and it uses the Scala `Try` construct which catches everything.
* https://github.com/rackerlabs/repose/blob/8.6.2.0/repose-aggregator/components/filters/scripting-filter/src/main/scala/org/openrepose/filters/scripting/ScriptingFilter.scala#L86-L88

```
scala> Try(throw new Throwable())
res0: scala.util.Try[Nothing] = Failure(java.lang.Throwable)
```

This prevents us from easily creating a test to prove if the Response Status Code is not set and an uncaught `Throwable` is thrown, then the Client is not returned an OK (200).

Just to see if the new Jetty version behaves as we want or maintains the same behavior, I commented out the the `sendError` line in the `Scripting` filter and ran a test locally. Unfortunately, it does behave the same undesired way and returned an OK (200) when a [`ClassNotFoundException`](http://docs.oracle.com/javase/8/docs/api/java/lang/ClassNotFoundException.html) was forced, but not caught.

wdschei:
It turns out this was actually an uncaught [Error](http://docs.oracle.com/javase/8/docs/api/java/lang/Error.html) that was causing the issue.

wdschei:
Rebuild this please.

wdschei:
Retest this please.

wdschei:
Retest this please.

dmnjohns:
Just to be clear, this is a version bump to Jetty, but not one that fixes the 200 on `Error`s?

Anyway, the changes themselves look good, but I am concerned that the PR job has been running for two and a half days...

wdschei:
Yes, this was intended to just bump the Jetty version to the latest 9.2.x release.
Yes, it concerns me too that it is having trouble with the CORS tests.
So, I think the end result of this is actually a new story to upgrade and fix all the things that it breaks since we know it doesn't fix the `OK` (`200`) on `Error`'s.

PR #1778, opened 2017-07-28
REP-5823 Multiple Tenant RegEx's in Keystone v2 filter

wdschei:


repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -340,14 +351,13 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
...
-                tokenTenant.startsWith(prefix) && tokenTenant.substring(prefix.length).equals(uriTenant)
-              )
-            }
+          val uriTenant = getTenantFromUri
+          val tokenTenants = validToken.defaultTenantId.toSet ++ validToken.tenantIds
+          val prefixes = Option(config.getTenantHandling.getValidateTenant.getStripTokenTenantPrefixes).map(_.split('/')).getOrElse(Array.empty[String])
+          tokenTenants find { tokenTenant =>
+            tokenTenant.equals(uriTenant) || prefixes.exists(prefix =>
+              tokenTenant.startsWith(prefix) && tokenTenant.substring(prefix.length).equals(uriTenant)
+            )
```
dmnjohns: I think this change goes away if you end up going with my suggestion and letting the lazy val do its thing.

repose-aggregator/components/filters/keystone-v2-filter/src/test/scala/org/openrepose/filters/keystonev2/KeystoneV2SchemaTest.scala
```diff
@@ -17,13 +17,12 @@
  * limitations under the License.
  * =_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_=_
  */
-
-package org.openrepose.filters.keystonev2.config
+package org.openrepose.filters.keystonev2
```
dmnjohns: Was this just moved to keep it consistent with other filters? I can't really think of a reason that this would make a difference, so I assume it's just consistency and/or preference.

repose-aggregator/components/filters/keystone-v2-filter/src/test/scala/org/openrepose/filters/keystonev2/KeystoneV2FilterTest.scala
```diff
@@ -2298,6 +2298,70 @@ with HttpDelegationManager {
...
+      """.stripMargin)
+
+    val filter = new KeystoneV2Filter(mockConfigurationService, mockAkkaServiceClientFactory, mock[AtomFeedService], mockDatastoreService)
+
+    filter.init(mockFilterConfig)
+    filter.KeystoneV2ConfigListener.configurationUpdated(configuration)
+    filter.SystemModelConfigListener.configurationUpdated(mockSystemModel)
+
+    val testUris = Seq("One", "Two", "Too")
+    testUris foreach { uri =>
```
dmnjohns: Is there any reason not to inline this variable? Especially since you are describing the values in the function passed to `foreach`.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/keystonev2/multiextractregex/MultiExtractRegexTest.groovy
```diff
@@ -0,0 +1,87 @@
...
+            method: 'GET',
+            headers: ['content-type': 'application/json', 'X-Auth-Token': fakeIdentityV2Service.client_token])
+
+        then: "everything with a valid Tenant ID that was extracted is sent to the origin service, otherwise it isn't"
+        mc.receivedResponse.code as Integer == serviceRespCode
+
+        if (serviceRespCode == SC_OK) {
+            assert mc.handlings.size() == 1
+            assert mc.handlings[0].request.headers.findAll("x-tenant-id").get(0).contains(fakeIdentityV2Service.client_tenantid)
+        } else assert mc.handlings.size() == 0
```
dmnjohns: Not a fan of branching assertions in tests. Not only does it make the expectation block more difficult to understand, but it means that the test name cannot be as descriptive since the test itself is covering more than one case. Would you be opposed to breaking the `serversBad` case into its own test? You do end up with some duplication that way, but I, personally, think that's preferred. Oh, and breaking that case out also lets you drop the `serviceRespCode` column and lets you unroll from a list (i.e., `uriSegment << ["serversOne", "serversTwo", "serversToo"]`) instead.

repose-aggregator/components/filters/keystone-v2-filter/src/test/scala/org/openrepose/filters/keystonev2/KeystoneV2SchemaTest.scala
wdschei: Yep. Consistency.

repose-aggregator/components/filters/keystone-v2-filter/src/test/scala/org/openrepose/filters/keystonev2/KeystoneV2FilterTest.scala
wdschei: Originally, I thought I would have more test sets, but now there is no reason. I'll update it.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/keystonev2/multiextractregex/MultiExtractRegexTest.groovy
wdschei: I can. I'm just not a fan of the duplication. However, I can see it as being more clear the next time I have to come back to it.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
wdschei: Yep. Updated.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/keystonev2/multiextractregex/MultiExtractRegexTest.groovy
```diff
@@ -0,0 +1,98 @@
...
+        MessageChain mc = deproxy.makeRequest(
+            url: "$reposeEndpoint/$uriSegment/${fakeIdentityV2Service.client_tenantid}",
+            method: 'GET',
+            headers: ['content-type': 'application/json', 'X-Auth-Token': fakeIdentityV2Service.client_token])
+
+        then: "everything with a valid Tenant ID that was extracted is sent to the origin service, otherwise it isn't"
+        mc.receivedResponse.code as Integer == SC_OK
+
+        assert mc.handlings.size() == 1
+        assert mc.handlings[0].request.headers.findAll("x-tenant-id").get(0).contains(fakeIdentityV2Service.client_tenantid)
```
dmnjohns: Since you split this apart, you can drop the `assert`, can't you? Since you dropped the conditional block, I mean.

wdschei: Yep. Removed.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -320,9 +330,10 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
       def getTenantScopedRoles(roles: Seq[Role]): Seq[Role] = {
         Option(config.getTenantHandling.getValidateTenant) match {
           case Some(validateTenant) if !validateTenant.isEnableLegacyRolesMode =>
+            val uriTenant = tenantFromUriOption.get
```
dmnjohns: This `.get` potentially changes the semantics of this method. Rather than always returning `true` or `false`, it may now throw a `NoSuchElementException` which will result in the filter returning a `500` response. Now, we should be protected because if `tenantFromUriOption` would be `None`, then an `UnparseableTenantException` should have already been thrown. The way it was before, that would not be a concern -- you would just end up with an empty `Seq`. Using `.get` also adds another line of code.

So, functionally, the behavior of the filter should not change. The behavior of this method changes slightly, but that likely won't be observed. Do we want to keep this change?

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -116,25 +116,17 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
...
-      }
+      lazy val tenantFromUriOption: Option[String] =
+        Option(config.getTenantHandling.getValidateTenant).flatMap({ validateTenantConfig =>
+          Option(validateTenantConfig.getUriExtractionRegex).flatMap({ uriExtractionRegexList =>
+            uriExtractionRegexList.asScala.toStream.map(_.r).flatMap({ uriExtractionRegex: Regex =>
+                          request.getRequestURI match {
+                            case uriExtractionRegex(tenantId, _*) => Option(tenantId)
+                            case _ => None
+                          }
+                        }).headOption
```
dmnjohns: Huh. I like the use of a `Stream`. I was trying to think of a way to use it yesterday, but my brain couldn't come up with it.

If you tack on `.getOrElse(throw UnparseableTenantException("Could not parse tenant from the URI"))` after `.headOption` you can keep the early throw and drop the `Option`. Alternatively, you can use `.orElse` to keep the `Option` type, but still throw the exception here on failure to parse. Either way lets you drop the `getTenantFromUri` method.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -115,16 +116,17 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
...
+          })
+        }).getOrElse(throw UnparseableTenantException("Could not parse tenant from the URI"))
```
dmnjohns: Whoa, did I just totally miss this line? Or did something change underneath me?

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
wdschei: Agreed.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
wdschei: Updated.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
dmnjohns: I do have to ask what's up with all of the `({` and `})`, though. In general, you shouldn't need to double up delimiters like that, and I think that holds true here, as well. Also, there is a bunch of whitespace before the pattern match, but it doesn't seem like that whitespace was used to make it line up with anything... so I'm wondering what that's about.

Personally, I would also remove the `Regex` type ascription in the `flatMap` since, considering the variable name, it doesn't really provide additional clarity or type safety. So I propose the following cleanup:

```
lazy val tenantFromUri: String =
  Option(config.getTenantHandling.getValidateTenant).flatMap(validateTenantConfig =>
    Option(validateTenantConfig.getUriExtractionRegex).flatMap(uriExtractionRegexList =>
      uriExtractionRegexList.asScala.toStream.map(_.r).flatMap(uriExtractionRegex =>
        request.getRequestURI match {
          case uriExtractionRegex(tenantId, _*) => Option(tenantId)
          case _ => None
        }
      ).headOption
    )
  ).getOrElse(throw UnparseableTenantException("Could not parse tenant from the URI"))
```

wdschei: Most of that was due to the fallout of fighting through the required syntax to make it all work and we knew the dot notation would definitely work. Then the innermost one actually did require the doubling of the delimiters. So the fallout was they all had them. I just pushed the removal of the doubling from the outer two and the whitespace on the match statement.

dmnjohns: Huh, so I just played with it... and this is super weird. If I remove both the extraneous `{}` and the type ascription, it compiles and passes tests just fine. If I only remove the `{}`, though, the compiler gets angry. It is not immediately obvious to me why that is the case.

I think I still stand by removing both the type ascription and the `{}`, but if you want to leave those, that's fine. Just let me know either way so I can merge this once tests pass!

wdschei: The extra curly's and the type were removed. That was a crazy syntactical nightmare.

PR #1780, opened 2017-08-01
REP-5617 Response Wrapper throw IllegalStateException on add/append after commit.

wdschei:


dmnjohns:
I think we may want to stick this named logger in the log4j2 config that we ship. We can leave it on by default, but I imagine this is a logger people will probably want to either turn off, or write to a different log.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -75,6 +76,7 @@ import scala.io.Source
   */
 class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMode: ResponseMode, bodyMode: ResponseMode, desiredOutputStream: ServletOutputStream)
   extends javax.servlet.http.HttpServletResponseWrapper(originalResponse) with HeaderInteractor {
+  private val addHeaderWarningLogger = LoggerFactory.getLogger(this.getClass.getName + "_addHeaderWarning")
```
dmnjohns: I wonder if we actually want to use `this.getClass.getName`. We could just use the static `classOf[HttpServletResponseWrapper].getName`. Obviously that means that extensions of this class will have the same named logger, but that's probably a good thing. It would be kind of annoying to have to manage multiple loggers to turn off or reroute these log messages. Also, it doesn't cost much, but `this.getClass` is slightly less performant than `classOf[HttpServletResponseWrapper]`.

repose-aggregator/src/docs/asciidoc/release-notes.adoc
```diff
@@ -17,6 +17,17 @@ As part of this correction, any configurations that were taking advantage of thi
...
 * https://repose.atlassian.net/browse/REP-5853[REP-5853] - Updated the <<filters/saml-policy.adoc#, SAML Policy Translation Filter>> and <<filters/attribute-mapping-policy-validation.adoc#,Attribute Mapping Policy Validation Filter>> to recover support for XML and JSON (which was removed in <<8.6.2.0 (2017-06-13)>>).
+* https://repose.atlassian.net/browse/REP-5617[REP-5617] - Updated the the internal HTTP Servlet Response Wrapper to log a WARNING when addHeader, addIntHeader, addDateHeader, or appendHeader is called after the response has been committed.
+
+[NOTE]
+====
+This message is logged to a separate logger and can be disabled by adding the following to the `log4j2.xml`:
+
+[source,xml]
+----
+<Logger name="org.openrepose.commons.utils.servlet.http.HttpServletResponseWrapper_addHeaderWarning" level="off"/>
```
dmnjohns: Ah, so this line reinforces my use of `classOf[HttpServletResponseWrapper]` rather than `this.getClass`.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
wdschei: Makes sense.

repose-aggregator/src/docs/asciidoc/release-notes.adoc
wdschei: Yep.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -223,6 +225,10 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
     addHeader(name, DateUtils.formatDate(new Date(timeSinceEpoch)))
 
   override def addHeader(name: String, value: String): Unit = {
```
wdschei: Yep.

wdschei: Maybe. Originally it was `add`'ing, then we included `append`'ing. If we also include set'ing, then we also have to include `replace`'ing.

I'm good with it and it is probably the right thing to do.

dmnjohns: Well, `replace` just calls through to `set`. But yeah, it seems to me like if the idea is that we are letting the user know that we might not have done what they wanted to do, then we should let them know in all cases.

wdschei: Should would add the warning to `setContentLength`, `setContentType`, & `setCharacterEncoding`?
Currently we check and just silently do nothing if it is committed.

dmnjohns: It seems like we may as well. We aren't actually changing behavior, just notifying of unexpected behavior. On a custom logger made just for that purpose, even. Is that ever a bad thing?

wdschei: I don't think so. I am going to change the name of the logger from `..._addHeader` though.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -75,6 +77,7 @@ import scala.io.Source
   */
 class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMode: ResponseMode, bodyMode: ResponseMode, desiredOutputStream: ServletOutputStream)
   extends javax.servlet.http.HttpServletResponseWrapper(originalResponse) with HeaderInteractor {
+  private val headerWarningLogger = LoggerFactory.getLogger(classOf[HttpServletResponseWrapper].getName + "_headerWarning")
```
dmnjohns: Should we change the name from `_headerWarning` to `.headerWarning`? Doing so makes `.` a consistent delimiter, and I can't come up with a reason why we need to use something different. I could just be missing it, though.

You could also use string interpolation here, but I don't think it particularly matters. That's probably a preference thing.

wdschei: I used the underscore as the dot typically indicates level. This is at the same level, but is a special logger of this same class.

I updated the String concatenation to String interpolation. It only adds one char in this case and it is more inline with the style of this class.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -222,7 +225,25 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
...
+    headerWarningLogger.warn(s"The following header $mayWill $notStill arrive at the client: $nameValue")
+  }
+
+  private def ifCommitted(name: String, value: String): Unit = {
+    if (isCommitted) {
+      logHeaderWarning(
+        "addHeader, addIntHeader, addDateHeader, setHeader, setIntHeader, setDateHeader, appendHeader, or replaceHeader",
+        May, Not, name, value)
+    }
+  }
```
dmnjohns: You created a few methods here for the sake of logging. I believe that was done to keep the code DRY, but personally, I don't believe that is worth the additional cognitive overhead and method calls on the stack. Additionally, this is a fairly rigid structure -- the log message must always be mostly the same when these methods are used.

If you don't want to repeat the log message everywhere, which is reasonable, I would propose, instead, that you use a string template for the log message in question. Doing so would let you keep your methods functional, and provide flexibility if we ever wanted to log something else when the response has already been committed. Something like:
`private def headerWarningMessage(method: String, headerName: String): String = s"Calls to $method after the response has been committed may be ignored -- the following header may not be modified: $headerName"`

I also wouldn't bother with the `ifCommitted` method since it only buys you one line per call site, and forces you to be vague in what method was called. Also, in `removeHeader` you still end up logging the header in a different way, which adds a little inconsistency (again, because of the rigidity).

In fact, I may also condense the log messages themselves into one message. Aren't the `info` and `warn` message saying nearly the same thing? If the header call is ignored, doesn't it directly follow that the header (the reason for the call) may/will not make it to the client?

So with all of that, I would probably end up with something like:
```
...

override def addHeader(name: String, value: String): Unit = {
  if (isCommitted) {
    headerWarningLogger.warn(headerWarningMessage("addHeader", name) + ": " + value)
  }

  ...
}

override def removeHeader(name: String, value: String): Unit = {
  ifMutable(headerMode) {
    if (isCommitted) {
      headerWarningLogger.warn(headerWarningMessage("removeHeader", name))
    }

    ...
  }
}

...
```

You might want to change a few words or parameters from what I have if you want to maintain the clarity that is currently being provided, but that should be easy enough to do using this structure.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: Well sure, it indicates logger level, that's true. Logger level does not need to directly correlate with the fully-qualified class name, though. That is, I would consider `headerWarning` to be a descendent of the class logger. After all, if I wanted to, say, turn off the class logger, I probably wouldn't want this logger to continue logging. Sure, it is distinct from the class logger, but it is still a logger pertaining to the class. The only other concern I can think of is that you will run into logger name collisions, but in what case would that happen? Other classes with loggers based on the fully-qualified class name should never collide following Java naming conventions. So then you are only concerned about loggers for this class, which you have complete control over (and thus, can provide collisions with). And even at a lower level, you can still fully customize this logger (although you may have to disable `additivity`, but that is always the case with nested loggers in Log4j2).

wdschei: Ok.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
wdschei: Updated.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -225,25 +225,15 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
...
 
-  private def logHeaderWarning(headerNames: String, mayWill: String, notStill: String, name: String, value: String): Unit = {
-    logHeaderWarning(headerNames, mayWill, notStill, s"$name: $value")
-  }
-
-  private def logHeaderWarning(headerNames: String, mayWill: String, notStill: String, nameValue: String): Unit = {
-    headerWarningLogger.info(s"Calls to $headerNames after the response has been committed $mayWill be ignored.")
-    headerWarningLogger.warn(s"The following header $mayWill $notStill arrive at the client: $nameValue")
-  }
+  private def headerWarningMessage(method: String, header: String): String = s"Calls to $method after the response has been committed may be ignored -- the following header may not be modified: $header"
```
dmnjohns: Oh, legit. Take a header string, then you can decide whether or not to pass the name and value or just value. I saw this before, but didn't think about it in my suggestions. Nice.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -506,3 +541,10 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
   }
 
 }
+
+object HttpServletResponseWrapper {
+  val May: String = "may"
+  val Will: String = "will"
+  val Not: String = "not"
+  val Still: String = "still"
+}
```
dmnjohns: With the recent changes, can these constants be removed?

wdschei: Removed.

repose-aggregator/artifacts/src/config/filters/log4j2.xml
```diff
@@ -29,6 +29,7 @@
         <Logger name="org.eclipse.jetty" level="off"/>
         <Logger name="org.springframework" level="warn"/>
         <Logger name="intrafilter-logging" level="info"/>
+        <Logger name="org.openrepose.commons.utils.servlet.http.HttpServletResponseWrapper_headerWarning" level="info"/>
```
dmnjohns: This should also be updated for the format change.

wdschei: Updated.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -28,6 +28,8 @@ import javax.servlet.{ServletOutputStream, ServletResponse}
 import javax.ws.rs.core.HttpHeaders
 
 import org.apache.http.client.utils.DateUtils
+import org.openrepose.commons.utils.servlet.http.HttpServletResponseWrapper._
```
dmnjohns: With the companion object gone, this breaks compilation.

wdschei: Removed.

PR #1788, opened 2017-08-15
REP-5315 Using metrics JMX object name convention for Spring managed beans

dmnjohns:


repose-aggregator/artifacts/cli-utils/src/main/java/org/openrepose/cli/command/datastore/local/ReposeJMXClient.java
```diff
@@ -36,6 +36,9 @@
 // close jmxc, but how on Earth would we test it since cli-utils doesn't seem to be used?  Suppressing this for now.
 public class ReposeJMXClient implements ReposeLocalCacheMBean {
 
+    private static final String hostname = NetUtilities.bestGuessHostname();
+    private static final String REPOSE_LOCAL_CACHE_OBJECT_NAME = hostname + ":001=\"org\",002=\"openrepose\",003=\"core\",004=\"services\",005=\"datastore\",006=\"distributed\",007=\"impl\",008=\"ehcache\",009=\"ReposeLocalCache\"";
+
```
dmnjohns: Super gross, I know, but semi-necessary because the hostname portion of the object name is dynamic.

Yes, we could do a query for the bean instead, but when I looked down that route, things got messier.

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/MBeanObjectNames.java
```diff
@@ -1,31 +0,0 @@
...
- * See the License for the specific language governing permissions and
- * limitations under the License.
- * =_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_=_
- */
-package org.openrepose.core;
-
-/**
- * Created by adrian on 11/11/16.
- */
-public class MBeanObjectNames {
```
dmnjohns: No longer necessary since the package and classname are already available to (and used by) the `ReposeJmxNamingStrategy`.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/core/services/datastore/distributed/impl/ehcache/ReposeLocalCache.java
```diff
@@ -35,8 +34,7 @@
 import java.security.NoSuchAlgorithmException;
 
 @Named("reposeLocalDatastore")
-@ManagedResource(objectName = MBeanObjectNames.REPOSE_LOCAL_CACHE,
-        description = "Repose local datastore MBean.")
+@ManagedResource(description = "Repose local datastore MBean.")
```
dmnjohns: No longer need the `objectName` since, again, it was just repeating the package and classname.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/core/services/reporting/metrics/MetricsJmxObjectNameFactory.java
```diff
@@ -23,12 +23,10 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.inject.Named;
 import javax.management.MalformedObjectNameException;
 import javax.management.ObjectName;
 import java.util.Hashtable;
 
-@Named
```
dmnjohns: This was not actually being wired in anywhere, so to prevent confusion, I just removed the annotation.

I actually considered leaving this a bean and wiring it in places, but we have some tests that don't like that approach. That is, this class is necessary for the `ReposeJmxNamingStrategy` bean to work. Right now, that bean instantiates this class itself so that it can always use an instance. If I wired this class in, it would break unit tests that do not scan the core classpath, but instead scan a tests classpath.

If we want this to be a bean, we can always make it so with some work in the future.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/core/spring/CoreSpringProvider.java
```diff
@@ -163,31 +164,31 @@ public void initializeCoreContext(String configRoot,
             // I'm not sure how well this will work, maybe it will fix the problem.
             GenericBeanDefinition mBeanServer = new GenericBeanDefinition();
             mBeanServer.setBeanClass(MBeanServerFactoryBean.class);
-            mBeanServer.setAutowireCandidate(true);
-            mBeanServer.setLazyInit(false); //Make sure that it's not lazy!
```
dmnjohns: These are the default values, so I just removed the redundant lines. We depend on other defaults implicitly, so I did not see a reason to be explicit about these.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/core/spring/CoreSpringProvider.java
```diff
@@ -163,31 +164,31 @@ public void initializeCoreContext(String configRoot,
...
             //Would have to give this an agent id?
             coreContext.registerBeanDefinition("reposeMBeanServer", mBeanServer);
 
+            GenericBeanDefinition jmxAttributeSource = new GenericBeanDefinition();
+            jmxAttributeSource.setBeanClass(AnnotationJmxAttributeSource.class);
+            coreContext.registerBeanDefinition("jmxAttributeSource", jmxAttributeSource);
+
+            GenericBeanDefinition jmxNamingStrategy = new GenericBeanDefinition();
+            jmxNamingStrategy.setBeanClass(ReposeJmxNamingStrategy.class);
+            jmxNamingStrategy.setAutowireMode(AbstractBeanDefinition.AUTOWIRE_CONSTRUCTOR);
```
dmnjohns: This is the primary component that we were missing before. We defined the naming strategy as a bean, but never tied it into the `MBeanExporter`. As a result, we were always using a different naming strategy that the `MBeanExporter` defaulted to. To tie it in, I programmatically define the bean (for the sake of consistency and to prevent the previously mentioned tests from breaking) and set it on the `MBeanExporter`.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/core/services/reporting/metrics/MetricsJmxObjectNameFactory.java
adrianjgeorge: I'm not following the test classpath problem? I'd lean towards keeping this a bean and wiring it in, can you elaborate what the problems are?

dmnjohns: So the problem with wiring this in is that this package is not scanned in certain unit tests. The `CoreSpringProvider` determines which package to scan based on a configuration file, as you can see here:

https://github.com/rackerlabs/repose/blob/master/repose-aggregator/core/repose-core/src/main/java/org/openrepose/core/spring/CoreSpringProvider.java#L49

Well, in some of our tests, we configure the `CoreSpringProvider` to only look in a test package rather than the standard package:
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/core/repose-core/src/test/resources/springConfiguration.conf#L1

As a result, the annotated `MetricsJmxObjectNameFactory` is never registered as a bean, as so it can't be found when Spring tries to wire it into the naming strategy and metrics service beans.

So we could make this an annotated bean, but it would involving changing the semantics, or at least the behavior, of those tests. I would have preferred to use this as an annotated bean, but didn't want to scope creep. If you prefer that, though, I can.

adrianjgeorge: Let's go ahead with the annotated bean, we are far enough ahead in this sprint that doing this right is worth the time. I'd make sure anything that needs it is looking for the interface and not this specific class, then in the couple of tests that look in a different package you can just stub out an implementation.

dmnjohns: So I went ahead and made this a bean, but there are a couple of things that I'm not super happy about.

For one, the interface this class implements comes from the metrics library. As a result, the metrics library becomes a direct dependency for core, where the `ReposeJmxNamingStrategy` lives. Of course, since the metrics service is currently a core service, we have already declared the metrics library as a core dependency... but this change couples things a bit more tightly. That may never matter, depending on what we do with the metrics service, but it still doesn't thrill me. Alternatively, I could wire in `MetricsJmxObjectNameFactory` explicitly instead of the `ObjectNameFactory` interface, which keeps the dependency on the service rather than the backing library. Considering the direction I went with the tests, would that be acceptable?

Two, I don't love the repeated code in the tests. That is, the classes which satisfy the bean dependency when we are scanning other packages. Of course, I don't really know of a way around that unless we modify the tests themselves or take a different approach.

It basically seems like we have three options here:

1. Don't make this class a bean. Either make it a singleton, or instantiate it when needed.
1. Make this class a bean that is defined with Java code and registered directly with the Spring context. This gets us around the package scanning in tests issue while keeping this a Spring bean without requiring us to stub implementations. Also, since the naming strategy bean itself is defined this way, there is some consistency there.
1. Make this class an annotated bean.

adrianjgeorge: So this opens a lot of questions about about how we want to gather metrics, reports metrics, and expose metrics. And who is responsible for what there. What do we want the metrics service to do? At the moment, I'm leaning towards metrics gathering being an intrinsic things repose does meaning that it's not the responsibility of the metrics service, but that it becomes a dependency more akin to log4j. This probably means we expose the registry as a bean that anything can wire in (assuming that the registry itself doesn't have the same kind of awareness that log4j/slf4j does). Then the metrics service is responsible for the other two needs reporting and exposure. In this case reporting is the tying in and dumping to graphite, and exposure is part of the non-existent admin api providing a restful interface into the metrics so people can potentially look in on repose live. I don't remember where we ended up landing with the metrics service when we hacked at it a bunch several months back, and what level of compatibility this has with that. I don't even know if this is a good way of looking at it. I think this way helps relieve some of the dependency weirdness you are pointing out here though? Did any of that make sense? Thoughts?

dmnjohns: I think that made sense, and that seems legit to me. Basically, we just need to register the `MetricRegistry` as a bean, which is pretty easy to do. In fact, I think the work we did a few months back actually makes that easier, since we stopped forcing all interactions with the registry to go through the service (now, you can get a reference to the registry from the service, which is similar to making it a bean, just a bit more work). 

Of course, the metrics service would have to acquire the `MetricRegistry` bean to be able to configure `Reporter`s to report the metrics to JMX and Graphite (and wherever else we may want to report metrics). At this point, I should note that anyone with a reference to the `MetricRegistry` will be able to create a reporter, which is pretty neat. I mean, that's already the case since the registry is exposed through the service, but still.

The only other thing to consider is the custom support we added for combining metrics. So the `SummingMeterFactory` and `MultiMeter`. These will still work regardless, and we can adjust the service to continue making it easy to access them, or we can move them into common utilities (since they are utilities for managing metrics in the registry rather than part of the service, really).

Oh, and there's the `MetricsJmxObjectNameFactory` that was the root of this conversation. Well, that's easy. Since the metrics library must be part of core, there is no problem using the interface. I would probably move this class to utilities as well. I was actually already considering doing that since this logic is now used for both Spring JMX beans and metrics JMX beans. I had hoped that there was a Java management interface for an `ObjectName` factory that I could use to make it a more standard utility, but unfortunately, there does not seem to be. So the interface was the only reason I kept it in the metrics service.

adrianjgeorge: To be clear I'm not saying to do these things now, but for this to be the direction we push things toward. Hopefully @Mario-Lopez and @wdschei chime in with their thoughts.

dmnjohns: Indeed.

As for the code here, I reverted the bean changes and did what made the most sense to me. I left the bean changes commit so that we have the history, and you can see both approaches. My new approach was to make the JMX `ObjectName` factory a common utility, since it really is our approach to naming any JMX bean at the moment, not just those reported as metrics. When I did that, I made the `ObjectName` creation method static, so there was no need to make a bean that had to be wired in places. To satisfy the contract for the metrics library, I kept `MetricsJmxObjectNameFactory` around as a very simple singleton wrapper to the common utility.

wdschei: I like the idea of exposing the registry as a bean so that anything that wants to report metrics can. I also like the ideas of making a facility for dumping to an external analysis tool as well providing an API for direct live queries. I also agree this is way out of scope of this task.

PR #1795, opened 2017-08-22
REP-5767 Adding support for JAVA_OPTS in Repose Docker images

dmnjohns:


repose-aggregator/artifacts/docker/src/docker/centos/Dockerfile
```diff
@@ -29,5 +29,9 @@ RUN touch /var/log/repose/current.log
 # "-p" flag with the Docker run command.
 EXPOSE 8080
 
+# This environment variable is used to set command-line options.
+# The user can manually set these options using the "-e" flag with the Docker run command.
+ENV JAVA_OPTS=
+
 # Start Repose.
-CMD ["java", "-jar", "/usr/share/repose/repose-valve.jar", "-c", "/etc/repose"]
+CMD java $JAVA_OPTS -jar /usr/share/repose/repose-valve.jar -c /etc/repose
```
dmnjohns: Unfortunately, the `exec` form of the `CMD` instruction will not substitute environment variables, so I switched over to the `shell` form.

repose-aggregator/artifacts/docker/src/docker/centos/Dockerfile
```diff
@@ -29,5 +29,9 @@ RUN touch /var/log/repose/current.log
 # "-p" flag with the Docker run command.
 EXPOSE 8080
 
+# This environment variable is used to set command-line options.
+# The user can manually set these options using the "-e" flag with the Docker run command.
+ENV JAVA_OPTS=
```
dmnjohns: This instruction is not strictly required. I added it to make it clear that we can handle a `JAVA_OPTS` environment variable, and in the future, we can provide a default value... but for now, it does nothing.

PR #1796, opened 2017-08-25
REP-5885 Don't return a 200 on failed requests when Errors are thrown

dmnjohns:


repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
```diff
@@ -423,8 +423,8 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
         } catch (URISyntaxException use) {
             LOG.debug("{}:{} -- Invalid URI requested: {}", clusterId, nodeId, wrappedRequest.getRequestURI(), use);
             wrappedResponse.sendError(HttpServletResponse.SC_BAD_REQUEST, "Error processing request");
-        } catch (Exception ex) {
-            LOG.error("{}:{} -- Exception encountered while processing filter chain.", clusterId, nodeId, ex);
+        } catch (Throwable t) {
```
dmnjohns: I'm wondering if we should handle `Exception`s and `Error`s separately, and rethrow `Error`s. That way we can still log the issue and set the status code, but we aren't claiming to have handled the `Error` in question by swallowing it. It feels a little sketchy to just swap `Exception` for `Throwable` in cases where the JVM runs out of memory or something even more strange.

Sonar may have a point on this one. While it makes sense for us to catch a `NoClassDefFoundError` (or maybe, more generally, a `LinkageError`) in certain cases, does it really make sense to catch `Throwable` here? What if the JVM hits  

https://sbforge.org/sonar/rules/show/squid:S1181?layout=false

adrianjgeorge: so rethrowing the error won't solve anything, because the finally block will still trigger

adrianjgeorge: or are you suggesting we catch it, log and set the response then rethrow it?

dmnjohns: Right, the ladder. Catch it, log it, set it, throw it. Bop it.

wdschei: I concur with the latter as well. Re-Throwing definitely makes sure the JVM is able to clean up after itself if it so chooses.

adrianjgeorge:
retest this please

dmnjohns:
retest this pelase

dmnjohns:
retest this please

adrianjgeorge:
Not mergeable at the moment.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/powerfilter/ExceptionAndErrorHandlingTest.groovy
```diff
@@ -0,0 +1,60 @@
...
+
+    def "should return a 500 response if an exception is thrown by a filter during processing"() {
+        when:
+        MessageChain messageChain = deproxy.makeRequest(
+            url: reposeEndpoint + "/exceptionanderror"
+        )
+
+        then:
+        // The PowerFilterChain catches the Exception and sets the response code to 500
+        messageChain.receivedResponse.code.toInteger() == 500
```
adrianjgeorge: Magic number, although we do that a lot. the code actually explicitly says we'll throw a 502.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/powerfilter/ExceptionAndErrorHandlingTest.groovy
```diff
@@ -0,0 +1,65 @@
...
+
+    def "should return a 500 response if an exception is thrown by a filter during processing"() {
+        when:
+        MessageChain messageChain = deproxy.makeRequest(
+            url: reposeEndpoint + "/exceptionanderror"
+        )
+
+        then:
+        // The PowerFilterChain catches the Exception and sets the response code to 500
+        messageChain.receivedResponse.code.toInteger() == SC_INTERNAL_SERVER_ERROR
```
adrianjgeorge: Code says it'll throw a bad gateway

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
```diff
@@ -424,8 +424,12 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
             LOG.debug("{}:{} -- Invalid URI requested: {}", clusterId, nodeId, wrappedRequest.getRequestURI(), use);
             wrappedResponse.sendError(HttpServletResponse.SC_BAD_REQUEST, "Error processing request");
         } catch (Exception ex) {
-            LOG.error("{}:{} -- Exception encountered while processing filter chain.", clusterId, nodeId, ex);
+            LOG.error("{}:{} -- Issue encountered while processing filter chain.", clusterId, nodeId, ex);
             wrappedResponse.sendError(HttpServletResponse.SC_BAD_GATEWAY, "Error processing request");
```
adrianjgeorge: right here.

dmnjohns: Oh, right, but no. The exception actually gets caught before this, in the `PowerFilterChain`. I added a comment in the test saying as much. Here:

https://github.com/rackerlabs/repose/blob/master/repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilterChain.java#L218-L222

So I guess this block is for `Exception`s thrown outside of a filter.

PR #1797, opened 2017-08-29
Updated the API Checker version from v2.3.0 to v2.4.0.

wdschei:


repose-aggregator/components/filters/saml-policy-translation-filter/src/test/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilterTest.scala
```diff
@@ -1088,7 +1088,7 @@ class SamlPolicyTranslationFilterTest extends FunSpec with BeforeAndAfterEach wi
       .compile(new StreamSource(new StringReader(brokenXslt)))
     val workingXslt =
       """<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
-        |                version="1.0">
+        |                version="3.1">
```
dmnjohns: IIRC, won't XSLT `1.0` still work, as long as Xalan is being used? Then Saxon supports 3.1, which is mostly backwards compatible with `2.0`?

Of course, for this test, there's no harm in bumping the version to `3.1`. I'm mostly wondering if we need to bump the XSLT version anywhere else.

PR #1799, opened 2017-09-11
REP-5765 Implement Scoverage

dmnjohns:
... and also some tweaks to JaCoCo.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -456,6 +456,7 @@ sonarqube {
         property "sonar.dynamicAnalysis", "reuseReports"
         property "sonar.jacoco.reportPath", "$rootDir/repose-aggregator/build/jacoco/jacocoTest.exec"
         property "sonar.jacoco.itReportPath", "$rootDir/repose-aggregator/build/jacoco/jacocoIntegrationTest.exec"
+        property "sonar.scoverage.reportPath", "$rootDir/repose-aggreagtor/build/scoverage-aggregate/scoverage.xml"
     }
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -261,11 +261,12 @@ subprojects {
 
     checkstyle {
         configFile = file("$rootDir/repose-aggregator/src/config/styles/checkstyle.xml")
-        sourceSets = sourceSets.matching {it.name != 'generated'}
+        sourceSets = sourceSets.matching {it.name != 'generated' && it.name != 'scoverage' && it.name != 'testScoverage'}
     }
 
     codenarc {
         configFile = file("$rootDir/repose-aggregator/src/config/styles/codenarc.xml")
+        sourceSets = sourceSets.matching {it.name != 'generated' && it.name != 'scoverage' && it.name != 'testScoverage'}
     }
 
     //todo: write a good one of these plugins, because this is garbage
@@ -286,7 +287,7 @@ subprojects {
         }
         exclude "**/*.cfg.xml"
         exclude "**/log4j2*.xml"
-        sourceSets = sourceSets.matching {it.name != 'generated'}
+        sourceSets = sourceSets.matching {it.name != 'generated' && it.name != 'scoverage' && it.name != 'testScoverage'}
     }
 
     //todo: for some reason jacoco freaks when we do clean and build together
@@ -382,10 +383,6 @@ subprojects {
         sign configurations.archives
     }
 
-    // todo: fix Scoverage running as dependency of check
-    //       it looks like it is being brought in by certain tasks (which are probably generated from source sets)
-    //       the following command removes Scoverage tasks:
-    //       gradlew build -x codenarcScoverage -x codenarcTestScoverage -x license -x checkstyleScoverage -x checkstyleTestScoverage -m
     // todo: fix jacocoTestCoverageVerification task skipping
     if (project.hasProperty("minCoverageRatio")) {
         def minCoverageRatio = (project.findProperty("minCoverageRatio") ?: "0.75") as double
--- a/build.gradle
+++ b/build.gradle
@@ -234,10 +234,6 @@ subprojects {
                 compileClasspath += generated.output
                 runtimeClasspath += generated.output
             }
-            scala {
-                compileClasspath += generated.output
-                runtimeClasspath += generated.output
-            }
             scoverage {
                 compileClasspath += generated.output
                 runtimeClasspath += generated.output
@@ -249,10 +245,6 @@ subprojects {
         }
         test {
             java {
-                compileClasspath += generated.output
-                runtimeClasspath += generated.output
-            }
-            scala {
                 compileClasspath += generated.output
                 runtimeClasspath += generated.output
             }
--- a/build.gradle
+++ b/build.gradle
@@ -311,7 +311,7 @@ subprojects {
         afterEvaluate {
             // Ignore Scala classes.
             classDirectories = files(classDirectories.files.collect { classDirectory ->
-                fileTree(dir: classDirectory, excludes: getScalaClassPatterns(sourceDirectories))
+                fileTree(dir: classDirectory, excludes: ['**/ObjectFactory.class'] + (getScalaClassPatterns(sourceDirectories) as List<String>))
             })
             sourceDirectories = sourceDirectories.filter({ dir -> dir.name != 'scala' })
         }
@@ -420,7 +420,7 @@ subprojects {
             afterEvaluate {
                 // Ignore Scala classes.
                 classDirectories = files(classDirectories.files.collect { classDirectory ->
-                    fileTree(dir: classDirectory, excludes: getScalaClassPatterns(sourceDirectories))
+                    fileTree(dir: classDirectory, excludes: ['**/ObjectFactory.class'] + (getScalaClassPatterns(sourceDirectories) as List<String>))
                 })
                 sourceDirectories = sourceDirectories.filter({ dir -> dir.name != 'scala' })
             }
--- a/build.gradle
+++ b/build.gradle
@@ -12,6 +12,7 @@ buildscript {
         classpath 'com.netflix.nebula:nebula-dependency-recommender:3.6.3'
         classpath 'com.netflix.nebula:gradle-extra-configurations-plugin:3.0.3'
         classpath 'gradle.plugin.org.openrepose:gradle-jaxb-plugin:2.4.1'
+        classpath "gradle.plugin.org.scoverage:gradle-scoverage:2.1.0"
         classpath 'com.github.jengelman.gradle.plugins:shadow:1.2.3'
         classpath 'org.github.ngbinh.scalastyle:gradle-scalastyle-plugin_2.11:0.8.2'
         classpath 'gradle.plugin.nl.javadude.gradle.plugins:license-gradle-plugin:0.12.1'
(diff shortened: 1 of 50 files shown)


dmnjohns:
I think this failed some builds because I did not upgrade Gradle in Puppet. Oops. Why don't the slaves use the Gradle wrapper again?

dmnjohns:
retest this please

dmnjohns:
Crap. Something else must be wrong.

dmnjohns:
retest this please

dmnjohns:
Boom! I fixed the exception numbers, but I have no idea what happened to fix the other build issues. 

I'd ship it.

build.gradle
```diff
@@ -227,6 +234,28 @@ subprojects {
                 compileClasspath += generated.output
                 runtimeClasspath += generated.output
             }
+            scala {
+                compileClasspath += generated.output
```
adrianjgeorge: Why was this necessary? Doesn't it aggregate them in?

build.gradle
```diff
@@ -227,6 +234,28 @@ subprojects {
...
+                runtimeClasspath += generated.output
+            }
+            testScoverage {
+                compileClasspath += generated.output
+                runtimeClasspath += generated.output
+            }
+        }
+        test {
+            java {
```
adrianjgeorge: These super shouldn't be necessary, because it takes all the main classpaths and adds them to test by default?

build.gradle
```diff
@@ -227,6 +234,28 @@ subprojects {
...
             }
+            scala {
+                compileClasspath += generated.output
+                runtimeClasspath += generated.output
+            }
+            scoverage {
+                compileClasspath += generated.output
+                runtimeClasspath += generated.output
+            }
+            testScoverage {
```
adrianjgeorge: if this is even needed shouldn't it be under test?

build.gradle
```diff
@@ -365,6 +405,39 @@ subprojects {
...
+            violationRules {
+                rule {
+                    limit {
+                        minimum = minCoverageRatio
+                    }
+                }
+            }
+            afterEvaluate {
+                // Ignore Scala classes.
+                classDirectories = files(classDirectories.files.collect { classDirectory ->
```
adrianjgeorge: This was needed twice?

build.gradle
dmnjohns: It does, which is why our tests were running prior to this change. However, adding this enabled IntelliJ to recognize generates sources in test code, which it has not been doing for me prior.

It would seem that generated classes are added to same output directory as Java classes, which is why they totally work in tests (i.e., are on the runtime classpath). However, IntelliJ does not know about the sources, so it highlights things in red.

For a little more context, I added some printlns so that the configuration looks like this:
```
        test {
            java {
                println("java compileClasspath: " + compileClasspath.files)
                println("java runtimeClasspath: " + runtimeClasspath.files)
                compileClasspath += generated.output
                runtimeClasspath += generated.output
                println("java compileClasspath: " + compileClasspath.files)
                println("java runtimeClasspath: " + runtimeClasspath.files)
            }
            ...
        }
```
and here is the output:
```
java compileClasspath: [/Users/dami6356/dev/repose/repose-aggregator/build/classes/main, /Users/dami6356/dev/repose/repose-aggregator/build/resources/main]
java runtimeClasspath: [/Users/dami6356/dev/repose/repose-aggregator/build/classes/test, /Users/dami6356/dev/repose/repose-aggregator/build/resources/test, /Users/dami6356/dev/repose/repose-aggregator/build/classes/main, /Users/dami6356/dev/repose/repose-aggregator/build/resources/main]
java compileClasspath: [/Users/dami6356/dev/repose/repose-aggregator/build/classes/main, /Users/dami6356/dev/repose/repose-aggregator/build/resources/main, /Users/dami6356/dev/repose/repose-aggregator/build/resources/generated]
java runtimeClasspath: [/Users/dami6356/dev/repose/repose-aggregator/build/classes/test, /Users/dami6356/dev/repose/repose-aggregator/build/resources/test, /Users/dami6356/dev/repose/repose-aggregator/build/classes/main, /Users/dami6356/dev/repose/repose-aggregator/build/resources/main, /Users/dami6356/dev/repose/repose-aggregator/build/resources/generated]
```

So as you can see, the generated resources directories are not on the classpath without this change... but we don't really care about those.

Which makes me wonder why this makes IntelliJ happy. They way IntelliJ utilizes Gradle must be odd, as adding directories we don't really care about to the `compileClasspath` seems to be helping IntelliJ find the generates sources. Instead, I think I should probably just add the `generatedDir` to the `srcDirs` for the configurations in the `test` source set. Does that make sense? I need to test it and see what happens, but that feels more right.

dmnjohns: Quick update -- I just testing setting the `srcDirs` and that also seems to satisfy IntelliJ. Plus, really, that's what I wanted to do in the first place.

dmnjohns: Hah. An unfortunate, but amusing, side-effect is that checkstyle will look at generated sources, which obviously fail the check.

dmnjohns: So `generated.output` is a `SourceSetOutput` that just holds references to a classes output directory and a resources output directory. Since the classes output directory is set to the main classes output directory, the practical effect of all of the lines doing something like `compileClasspath += generated.output` is to add the generated resources directory to the classpath. But the generated resources directory is always empty.

So then why did we add `compileClasspath += generated.output` and the like? Presumably to make IntelliJ happy, since we've established that generated classes live in the same directory as compiled code classes. And why does adding that make IntelliJ happy? That's the thing I can't answer.

build.gradle
dmnjohns: It totally does. Not sure why this was added. Let me try to remove it to see what happens. It everything works, I'll take it out, otherwise I'll report back with what broke.

dmnjohns: Seems to work.

repose-aggregator/components/filters/keystone-v2-filter/build.gradle
```diff
@@ -41,3 +41,16 @@ jaxb {
     }
 }
 
+jacocoTestCoverageVerification {
+    violationRules {
+        rules.forEach { rule ->
+            rule.limits.forEach { limit ->
+                if (limit.counter == 'INSTRUCTION' && limit.value == 'COVEREDRATIO') {
+                    // todo: set to the current coverage -- need more coverage, but we don't want the build to fail for the baseline
+                    limit.minimum = 0.46
```
adrianjgeorge: why is this one so low?

repose-aggregator/components/filters/saml-policy-translation-filter/build.gradle
```diff
@@ -32,3 +32,34 @@ dependencies {
     testCompile 'org.opensaml:opensaml-security-api'
     testCompile 'org.opensaml:opensaml-xmlsec-api'
 }
+
+license {
```
adrianjgeorge: what is this all about?

repose-aggregator/components/services/atom-feed-service/atom-feed-service-api/build.gradle
```diff
@@ -0,0 +1,12 @@
+jacocoTestCoverageVerification {
+    violationRules {
+        rules.forEach { rule ->
+            rule.limits.forEach { limit ->
+                if (limit.counter == 'INSTRUCTION' && limit.value == 'COVEREDRATIO') {
+                    // todo: set to the current coverage -- need more coverage, but we don't want the build to fail for the baseline
+                    limit.minimum = 0.0
```
adrianjgeorge: ???

adrianjgeorge:
Many of these don't make much sense to me as to why they are so low, it feels like mis-configuration.

build.gradle
dmnjohns: I swear I tried that and things broke, but I tried it again just now, and everything seems to be alright. So yes.

I guess putting it here caused the test source set to pull it in, but it's only really there.

build.gradle
adrianjgeorge: Man i super don't remember, but ti did take a lot of exploration and poking at things to get intellij a little happy.

dmnjohns: So we can remove the Scala `SourceDirectorySet` configuration since it extends the Java one anyway (as described in the comments on the other line), but I think we should keep the Java one. It certainly doesn't harm anything, and give us (or at least me) a nicer experience in IntelliJ.

build.gradle
dmnjohns: Unfortunately, yes. In one case, we are configuring the reporting task to ignore Scala sources. In the other case, we are configuring the verification task to ignore Scala sources. JaCoCo itself extends test tasks to generate the `.exec` files with all of the data. Both the report and the verification use those files to do their thing. The key there is that the verification task does not use the reports from the report task, so we have to ignore those classes in both places.

I mean, unless we don't care about reports. But it costs us nothing, so why not keep it configured?

repose-aggregator/components/filters/keystone-v2-filter/build.gradle
dmnjohns: I think it's almost all from missing generated source coverage. For this project, we configure XJC to add builder methods via the `-Xflient-api` flag:
```
jaxb {
    xjc {
        args = ["-Xdefault-value", "-Xfluent-api"]
    }
}
```

Well, we don't use all of the builder methods in our tests, so we don't cover those paths.

We may also be missing legitimate coverage of certain configurations, but I couldn't say without digging into the report more.

repose-aggregator/components/filters/saml-policy-translation-filter/build.gradle
dmnjohns: You know how Scoverage has to instrument and compile all the things? Well, to do so, it has defined its own set of tasks which mirror the tasks added by the Java plugin (e.g., `testScoverage`, `checkScoverage`). To coordinate all of that, Scoverage also defined its own source directory set, which you've already seen. Well, the output directories in that source directory set are not the same as the output directories for, say, the java source directory set. So instead of copying resources to `${project.buildDir}/resources/test/`, the Scoverage task will copy them to `${project.buildDir}/resources/testScoverage/`.

All of this is coordination to make sure things end up where the tests run during `testScoverage` can use them, and to make sure that the test itself knows where to look. Before this change, the test itself always tried to use a hardcoded path (which matched the java output directory, but not the scoverage output directory) as the `configRoot` (i.e., the place where some test resources live).

build.gradle
adrianjgeorge: you've created two sources of truth. I don't mind it being there for the reports, i just wish it were easier to have it written out once, in case we end up having to change it.

repose-aggregator/components/filters/keystone-v2-filter/build.gradle
adrianjgeorge: le sigh

repose-aggregator/components/filters/saml-policy-translation-filter/build.gradle
dmnjohns: To simplify this a bit, I could try and set the output resources directory for the Scoverage source directory set to the same as the one for the Java source directory set... I'm not sure what consequences there are to doing so, however.

build.gradle
dmnjohns: I recognize that, and did not really want to do it... which is why I extracted the bulk of this out into a method. However, that seems to be the way the plugin works, so I'm not sure how to get around that, especially since in one case configuration is conditional (i.e., executed based on the presence of the `minCoverageRatio` Project property).

repose-aggregator/components/services/atom-feed-service/atom-feed-service-api/build.gradle
dmnjohns: Ah. The API has some testable classes, but no tests in the project. All of the tests are in the impl project. So when run on its own, this project reports a coverage of 0%. That's not actually the case, though, since those classes are used in tests elsewhere.

I must have gotten a bit trigger-happy when going down the list to have missed that. However, the reason I hit this is probably because it failed a build, which means that the test coverage from the other project is not being counted here. I... am not entirely sure how to fix that. I could ignore api projects, but then we don't have numbers for things in API projects (e.g., Exceptions, enums). Or we could add tests to API projects, but... they would probably be really simple, and kind of pointless, tests. Ideally, we would be able to aggregate JaCoCo coverage metrics and act on the aggregation rather than each individual project, but support for doing so does not really exist in the JaCoCo plugin. We could probably come up with a way to do that (maybe like this: https://gist.github.com/aalmiray/e6f54aa4b3803be0bcac) but it's still not super pretty. Perhaps that is an enhancement we could save for a follow-up story?

PR #1800, opened 2017-09-14
REP-5727 Logging Trace ID Log Messages With New Logger

dmnjohns:


repose-aggregator/src/docs/asciidoc/release-notes.adoc
```diff
@@ -13,6 +13,16 @@
...
+* https://repose.atlassian.net/browse/REP-5727[REP-5727] - Extracted trace ID logging to its own named logger.
+** The `trace-id-logging` Logger in your Log4j2 configuration will determine the logging behavior for trace ID logging.
+If the `trace-id-logging` Logger has not been configured, it will inherit the `Root` logger's configuration.
++
+[IMPORTANT]
+====
+If Repose is already installed and configured, and the `Root` logger is not configured to log at the `trace` logging level, then upgrading Repose will drop some or all trace ID logging messages.
+Please add and configure the `trace-id-logging` Logger to regain those messages.
+====
+
```
dmnjohns: So the naive approach could change logging behavior underneath people that upgrade, but already have a `log4j2` configuration in place. It is somewhat likely that they will lose at least the trace level trace ID log messages. Does that have some implication with regards to the version of the release this work can be in?

repose-aggregator/src/docs/asciidoc/release-notes.adoc
```diff
@@ -13,6 +13,16 @@
 *** https://github.com/rackerlabs/api-checker/blob/api-checker-2.4.0/RELEASE.md[API Checker v2.4.0 release notes]
 ** Attribute Mapper: 1.3.0 → 1.3.1
 *** https://github.com/rackerlabs/attributeMapping/blob/attribute-mapper-1.3.1/RELEASE.md[Attribute Mapper v1.3.1 release notes]
+* https://repose.atlassian.net/browse/REP-5727[REP-5727] - Extracted trace ID logging to its own named logger.
+** The `trace-id-logging` Logger in your Log4j2 configuration will determine the logging behavior for trace ID logging.
+If the `trace-id-logging` Logger has not been configured, it will inherit the `Root` logger's configuration.
++
+[IMPORTANT]
```
dmnjohns: Important? Warning?

adrianjgeorge:
So it just occurred to me that we can achieve what we want without breaking compatibility in anyway. Instead of making this simply named logger, make it a sub logger of the of power filter.
`org.openrepose.powerfilter.PowerFilter.trace-id-logging`
Anyone who has tweaked logging of repose classes will keep their current behavior, but we still add the granularity to turn this off.

dmnjohns:
Man, that is a great idea that totally did not occur to me. Will do.

repose-aggregator/artifacts/src/config/filters/log4j2.xml
```diff
@@ -30,6 +30,7 @@
         <Logger name="org.springframework" level="warn"/>
         <Logger name="intrafilter-logging" level="info"/>
         <Logger name="org.openrepose.commons.utils.servlet.http.HttpServletResponseWrapper.headerWarning" level="warn"/>
+        <Logger name="trace-id-logging" level="trace"/>
```
dmnjohns: Looking at this again, should I set this to `info`? That is the current default for the `PowerFilter` logger, so I'm thinking probably so.

repose-aggregator/src/docs/asciidoc/release-notes.adoc
dmnjohns: This gets dropped with the child logger change.

adrianjgeorge:
rebase so the jenkins build works

dmnjohns:
Rebased.

PR #1804, opened 2017-09-25
REP-6050 Fixing the page styling

dmnjohns:
The icons relative position changed for some reason that I can't figure out (for I am certainly no CSS expert, much less a Compass expert), so I kind of cheated and adjusted their relational adjustment to make things work. I also fell back to the old slide spacing and just added exceptions for the information slides at the bottom of the page to make things look a little nicer.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -9750,7 +9750,7 @@ strong {
 
 /* line 11, ../scss/_custom.scss */
 .slide {
-  padding-top: 50px;
+  padding: 100px 0;
 }
 
 /* line 15, ../scss/_custom.scss */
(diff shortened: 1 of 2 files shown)


css/style.css
```diff
@@ -9750,7 +9750,7 @@ strong {
 
 /* line 11, ../scss/_custom.scss */
 .slide {
-  padding-top: 50px;
+  padding: 100px 0;
```
adrianjgeorge: lol, pretty sure this is how it use to be, and you changed it to this

dmnjohns: Yeah, 100%. But even after changing it back, the icons were still off, hence the minor adjustment to the icon class.

I initially changed this so that the text at the bottom wasn't so weirdly spaced out (with a ton of empty space). Of course, changing it changed all the slides, which led to the Narwhal imae being off. I guess I wasn't paying attention to the top of the page when I changed it initially. So to fix that, I went back to this standard spacing, then just made exceptions for the information text lower in the page. Really, I should have given that stuff it's own class... but...

PR #1806, opened 2017-09-26
REP-6098 Support different locations for default values in Attribute Mapping policies

dmnjohns:


repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlIdentityClient.scala
```diff
@@ -143,7 +143,7 @@ class SamlIdentityClient @Inject()(akkaServiceClientFactory: AkkaServiceClientFa
     * @param checkCache whether or not to use the HTTP request cache
     * @return the IDP ID if successful, or a failure if unsuccessful
     */
-  def getIdpId(issuer: String, token: String, traceId: Option[String], checkCache: Boolean): Try[String] = {
+  def getIdpId(issuer: String, token: String, traceId: Option[String], checkCache: Boolean): Try[ProviderInfo] = {
```
Mario-Lopez: The name of the method isn't really accurate anymore. Maybe `getIdp`?

Mario-Lopez: Since this was the only issue and does not impact the logic of the filter, we'll address this the next chance we get.

PR #1813, opened 2017-10-02
REP-5536 Added the new linkchecker plugin to the build.

wdschei:
Also updated the build to fail on ASCII Doctor `WARNING`'s.

### DIFF
--- a/repose-aggregator/build.gradle
+++ b/repose-aggregator/build.gradle
@@ -69,12 +69,11 @@ asciidoctor {
     }
 }
 
-tasks.linkchecker.dependsOn asciidoctor
 linkchecker {
     startFileName = "$buildDir/asciidoc/html5"
 }
 
-task('publishDocs', dependsOn: asciidoctor, group: 'release', description: 'Adds the new version of the docs to github pages') << {
+task('publishDocs', dependsOn: tasks.linkchecker, group: 'release', description: 'Adds the new version of the docs to github pages') << {
     def workingDir = project.file("${project.buildDir}/gh-pages")
 
     workingDir.deleteDir()
@@ -134,7 +133,8 @@ if (project.hasProperty("minCoverageRatio")) {
 }
 
 asciidoctor.dependsOn copyXsds
-build.dependsOn asciidoctor
+tasks.linkchecker.dependsOn asciidoctor
+build.dependsOn tasks.linkchecker
 publishDocs.dependsOn project.tasks.getByPath(':buildAll')
 project.tasks.getByPath(':release').dependsOn publishDocs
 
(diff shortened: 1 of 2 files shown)


wdschei:
Retest this please.

wdschei:
Retest this please.

repose-aggregator/build.gradle
```diff
@@ -50,9 +51,29 @@ asciidoctor {
...
+        }
+    }
+}
+
+linkchecker {
+    startFileName = "$buildDir/asciidoc/html5"
 }
 
-task('publishDocs', dependsOn: asciidoctor, group: 'release', description: 'Adds the new version of the docs to github pages') << {
+task('publishDocs', dependsOn: tasks.linkchecker, group: 'release', description: 'Adds the new version of the docs to github pages') << {
```
adrianjgeorge: I would leave this as it was, and make verify depend on the linkchecker

repose-aggregator/build.gradle
```diff
@@ -112,7 +133,8 @@ if (project.hasProperty("minCoverageRatio")) {
 }
 
 asciidoctor.dependsOn copyXsds
-build.dependsOn asciidoctor
+tasks.linkchecker.dependsOn asciidoctor
```
adrianjgeorge: see above

adrianjgeorge: although obviously this line is correct

repose-aggregator/build.gradle
wdschei: So instead of:
* `build.dependsOn tasks.linkchecker`

make it:
* `verify.dependsOn tasks.linkchecker`

at line 137?

wdschei: Or would it be:
* `test.dependsOn tasks.linkchecker`

adrianjgeorge: yep, it better represents what part of the lifecycle of the build it is. And build already depends on verify. 
And the publish task depends on our overall gate.
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/build.gradle#L116

dmnjohns: Gradle Java plugin tasks descriptions and lifecycle here:
https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_tasks

That has helped me a bunch. And, to me, it seems like you want:
`check.dependsOn tasks.linkchecker`

Is that what @adrianjgeorge meant? I don't think we have a `verify` task?

adrianjgeorge: You are correct that is what i meant. Where did i get verify? Maybe that's a maven goal?

wdschei:
Retest this please.

wdschei:
Retest this please.

PR #1815, opened 2017-10-04
REP-5765 Reporting Scoverage on a per-project basis

dmnjohns:
This is the appropriate way to do this, even though the Sonar Scoverage plugin may not work for us and our end solution may necessitate further changes.

So this is more correct than what we currently have, but doesn't solve our problem?


dmnjohns:
retest this please

PR #1817, opened 2017-10-04
REP-6135 Updated the Dockerfiles to support root group permissions.

wdschei:


dmnjohns:
I've manually tested the changes to our images in OpenShift, and I'm happy to say that our images run fine without any modification!

repose-aggregator/artifacts/docker/build.gradle
```diff
@@ -6,13 +6,9 @@ import groovy.json.JsonSlurper
...
-// This forces the usage of a more recent unix-socket-factory which makes Mac OS X happy
-// It also forces the usage of jackson dependencies which were not being brought in as expected
-// TODO: Remove unix-socket-factory after upgrading to gradle-docker-plugin 3.0.4
 configurations {
     dockerJava {
         resolutionStrategy {
-            force 'de.gesellix:unix-socket-factory:2016-04-06T22-21-19'
             force 'com.fasterxml.jackson.core:jackson-databind:2.6.4'
             force 'com.fasterxml.jackson.core:jackson-core:2.6.4'
             force 'com.fasterxml.jackson.core:jackson-annotations:2.6.4'
```
dmnjohns: Were these dependencies fixed by the upgrade? I would love not to have to keep this around.

If not, we should maybe keep the comment about what this is for around. If that comment isn't clear enough, our transitive dependency on Jackson was not resolving its dependencies correctly, so I forced them here to prevent version mismatch issues. That seems like an issue with either docker-java or jackson, so we may be able to remove our patch.

repose-aggregator/artifacts/docker/src/docker/centos/Dockerfile
```diff
@@ -14,16 +14,24 @@ RUN yum update -q -y && yum install -q -y wget java-1.8.0-openjdk-headless
 RUN wget --quiet -O /etc/yum.repos.d/openrepose.repo http://repo.openrepose.org/el/openrepose.repo
 RUN yum update -q -y && yum install -q -y repose-valve-$REPOSE_VERSION repose-filter-bundle-$REPOSE_VERSION repose-extensions-filter-bundle-$REPOSE_VERSION repose-experimental-filter-bundle-$REPOSE_VERSION
 
-# Expose /etc/repose as a volume for mounting.
-WORKDIR /etc/repose
-VOLUME /etc/repose
+ENV APP_ROOT=/etc/repose
+ENV APP_VARS=/var/repose
+ENV APP_LOGS=/var/log/repose
```
dmnjohns: I suspect you had keep this reference around because phone-home-messages are still being written to a file in this directory, and log4j2 seems to touch (i.e., create) the repose log file even though nothing is appending to is.

So in my eyes, we can either:
1. Keep this around to handle the auxiliary phone home logging and log4j2 file management.
1. Remove this and remove all file appenders from the log4j2 configuration in our Docker image.

repose-aggregator/artifacts/docker/README.adoc
```diff
@@ -0,0 +1,126 @@
...
+
+These are a few minor configuration items that may prevent some platforms from even executing the `docker run hello-world` test.
+They are not always well documented in the installation instructions, but are easily remedied.
+
+You will need to get a https://hub.docker.com[Docker Hub] account.
+Once you have that account, you will need to login using:
+
+- `docker login`
+
+Once you have confirmed your credentials, you will need to add them to the `~/.gradle/gradle.properties` file to allow them to work with the Repose build:
```
dmnjohns: Not technically true. You could also pass them as properties when executing Gradle. Pedantic, I know, but Gradle does its normal property lookup stuff.

repose-aggregator/artifacts/docker/README.adoc
```diff
@@ -0,0 +1,126 @@
...
+If you can't, then the steps below may get you up and going.
+Either way, you will need to perform them.
+
+== Minor updates
+
+These are a few minor configuration items that may prevent some platforms from even executing the `docker run hello-world` test.
+They are not always well documented in the installation instructions, but are easily remedied.
+
+You will need to get a https://hub.docker.com[Docker Hub] account.
+Once you have that account, you will need to login using:
```
dmnjohns: Not technically a requirement. Using `docker login` is a good way to test your credentials, for sure, but it isn't necessary as long as Gradle has your credentials.

repose-aggregator/artifacts/docker/build.gradle
wdschei: Do you remember what the Jackson stuff was causing to fail?

repose-aggregator/artifacts/docker/src/docker/centos/Dockerfile
wdschei: Yep, as far as the `/var/log/repose` reference goes.
I'm good commenting out both file appenders and removing this all together.
I'll do that.

repose-aggregator/artifacts/docker/build.gradle
dmnjohns: Not exactly, but without it, at least one of the Docker tasks would fail. So if you can remove this and run all of the tasks, we should be good.

It looks like the `docker-java` library uses Jackson to interact with the Docker API, so...

dmnjohns: Considering that `docker-java` is still on Jackson 2.6.4, and that they haven't added any new dependencies, I would guess this is still an issue:
https://github.com/docker-java/docker-java/blob/master/pom.xml#L93-L97

... but I could be wrong.

repose-aggregator/artifacts/docker/README.adoc
wdschei: Ok.

repose-aggregator/artifacts/docker/README.adoc
wdschei: "W"

repose-aggregator/artifacts/docker/build.gradle
wdschei: I can successfully execute both of the following with it removed:

* `:repose-aggregator:artifacts:docker:buildUbuntuImage`
* `:repose-aggregator:artifacts:docker:buildCentosImage`

dmnjohns: Noice.

dmnjohns:
With the phone-home change, commenting out the named appender means that the phone home message is not logged anywhere (since additivity is set to false). Is that what we want?

Also, the `/var/log/repose` directory still contains the `current.log` and `phone-home.log` files since they are still reference by appenders. They should never be written to, and thus, not giving them permissions is fine... but if we don't want those to be created at all, we'll have to remove or comment out the corresponding appenders.

wdschei:
These seem like reasonable behaviors inside a Docker container, but I'll leave it up for discussion.

dmnjohns:
Sure, they cause no harm. I don't particularly care either way, just pointing it out.

PR #1820, opened 2017-10-11
REP-6179 Convert Docs

wdschei:


repose-aggregator/src/docs/asciidoc/recipes/quick-start.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+image::http://dockeri.co/image/rackerlabs/repose[Repose Docker,link=https://hub.docker.com/r/rackerlabs/repose/]
+
+== Running a Docker Container
+After you have a running Docker Container Engine, you can run the latest *Repose* release with the following command:
+
+----
+docker run                  \ <1>
+   --detach                 \ <2>
+   --publish 8080:8080      \ <3>
```
dmnjohns: Using the long-form of options totally makes sense in examples. Not sure why I opted to use the short-form in other places. I like this.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
```diff
@@ -1,130 +1,147 @@
...
+   --publish 8080:8080                       \ <4>
+   --env JAVA_OPTS=-Xmx1024m                 \ <5>
+   --name my_repose                          \ <6>
+   rackerlabs/repose:latest                    <7>
 ----
 Let's break that command down and take a closer look at what it is doing:
 
 <1> This is the part of the command that tells Docker to create and start a container.
-<2> This flag will run the container in detached mode.
+<2> This option will run the container in detached mode.
```
dmnjohns: Not that it makes much difference, but why the change here?

repose-aggregator/src/docs/asciidoc/recipes/quick-start.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+To end the interactive session, issue the `exit` command.
+
+== Docker Container Logging
+The expectations of https://12factor.net/logs[Twelve-Factor App logging] is that all internal logging mechanisms within the container be disabled.
+The intent is that all traditional logging information should be streamed on STDOUT/STDERR.
+To access a containers log information simply use the following:
+
+----
+docker container logs --follow CONTAINER_ID
```
dmnjohns: I think you can drop the `container` bit from this command. `docker logs` is the shorthand, and seems to be the preferred way of accessing logs.

repose-aggregator/src/docs/asciidoc/recipes/quick-start.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+The output will contain lines similar to these:
+
+----
+< HTTP/1.1 301 Moved Permanently
+< Location: http://www.rackspace.com/
+----
+
+If you look at the container logs, then you will see a new entry under the `Repose ready` that contains `Tracing header: {...}`.
+This indicates that a new request was made to this instance and since you received the `301 Moved Permanently` response, then you know the default configured Respose is working as expected.
```
dmnjohns: `Respose` typo. Also, not bolded. Probably because of the typo. 😛

repose-aggregator/src/docs/asciidoc/recipes/quick-start.adoc
```diff
@@ -0,0 +1,166 @@
...
+If you connect to that same address using a regular web browser, you will automagically be redirected clear on through to the Rackspace homepage.
+
+== What's next?
+Out of the box a default configured *Repose* doesn't do much.
+This is because everybody needs it to do something specific to their needs.
+There are a lot of different <<../filters/index.adoc#,Filters>> available that can be used like building blocks to make a filter chain that does exactly what you need.
+One of the things that *Repose* is used for is <<preventing-xml-bomb.adoc#,Preventing XML bomb attacks>>.
+
+If you are ready to learn how to tweak the default configuration to do a little more or are feeling adventurous and want explore building your very own custom *Repose Docker* image,
+then head on over to the <<docker.adoc#,Docker>> recipe page.
```
dmnjohns: New line here? I thought we only put full sentences on new lines?

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
```diff
@@ -1,130 +1,147 @@
...
-After running that command, you should be able to see Repose running inside a Docker container!
-Use the following command to show all running containers:
+== Manually Building a Repose Docker Image
+The latest version of the Dockerfiles used to build the *Repose Docker* images published on Docker Hub can be found at:
 
-----
-docker ps
-----
+- https://raw.githubusercontent.com/rackerlabs/repose/master/repose-aggregator/artifacts/docker/src/docker/ubuntu/Dockerfile[Debian/Ubuntu]
+- https://raw.githubusercontent.com/rackerlabs/repose/master/repose-aggregator/artifacts/docker/src/docker/centos/Dockerfile[RedHat/Centos]
```
dmnjohns: Why not leverage the fact that our docs are versioned? You can link to the actual Dockerfiles in the repository, though if you link rather than include, you will probably have to copy those files into the docs project (like we do with schemas).

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
```diff
@@ -1,130 +1,147 @@
...
+- `remove`
 
-https://docs.docker.com/
+If you don't have the basics of these operations memorized, then that is fine.
+You can always look back at the <<quick-start.adoc#,Quick Start>> anytime you need to.
 
-== Docker Image
-All official, published Docker images for Repose, can be found on Docker Hub at:
+== Using a Repose Docker Image
+All official *Repose Docker* images are published on Docker Hub at:
```
dmnjohns: No idea what I was thinking with that line. This is better wordsmithing. 👍

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
```diff
@@ -1,130 +1,147 @@
 = Docker
 
-With the advent of container technology, Repose can be fully encapsulated and run as a service.
-By bundling the environment with the software itself, deploying Repose becomes a much quicker, simpler process.
-Docker boasts security through isolation, and when run on platforms with native support, it requires very little overhead.
-To make everything as easy and versatile as possible, the Repose team will release Docker images on Docker Hub alongside every release of Repose.
+With the advent of container technology, *Repose* can be fully encapsulated and run as a service.
```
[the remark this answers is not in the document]
dmnjohns: Looping back to this, I think my comment on the quick start page solves my issue.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
```diff
@@ -1,130 +1,147 @@
...
 
-== Docker
-To run Repose in Docker, there are just two basic steps:
+Even though the published *Repose Docker* image doesn't do a lot by default, this recipe will show you how to tap into some of that power hidden just below the surface.
 
-. Acquire a Repose Docker image.
-.. Download the Docker image. See: <<Docker Image>>
-.. Build the Docker image. See: <<Dockerfile>>
-. Run the Repose Docker image in a Docker container. See: <<Docker Container>>
```
dmnjohns: I have to say, I hate seeing some of this stuff go. I think what you have done here is write more of a recipe, which is absolutely appropriate, but I think there is still a place for more of this reference type material. I kind of want to stick it in an architecture page, but it's really more about Docker than it is Repose.

I wonder if the old page wouldn't work well as a README for the Docker project. I think it does a fair job detailing how to use Docker and build Repose with it in a structured way. Repose users probably don't want or need such information, but I think it does have value for developers.

Perhaps the team needs to have a discussion about user vs. developer documentation. The idea of putting developer documentation in the project as READMEs and user documentation on the site is starting to grow on me.

Thoughts?

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
```diff
@@ -1,130 +1,147 @@
...
+- `dockerhub.username=<USERNAME>`
+- `dockerhub.password=<PASSWORD>`
+
+Then try to build a quick *Repose Docker* image from the root of the *Repose* project.
 
 ----
-docker exec -it container_name_or_id /bin/bash
+./gradlew \
+    :repose-aggregator:artifacts:docker:buildUbuntuImageLocal \
+    -Prepose-version=8.7.0.2
```
dmnjohns: If we did keep a README around for the Docker project, I might put this in there. Will a Repose user ever need to build a Docker image using the Repose build system? This seems like developer information.

repose-aggregator/artifacts/docker/build.gradle
```diff
@@ -19,24 +19,29 @@ docker {
...
-        finalizedBy "remove${baseImage}Image"
+    task("build${baseImage}ImageLocal", type: DockerBuildImage) {
         inputDir = file("$projectDir/src/docker/${baseImage.toLowerCase()}")
         buildArgs = ['REPOSE_VERSION': reposeVersion]
     }
 
+    task("build${baseImage}Image") {
+        dependsOn "build${baseImage}ImageLocal"
+        finalizedBy "remove${baseImage}Image"
+        ext.imageId = tasks["build${baseImage}ImageLocal"].getImageId()
```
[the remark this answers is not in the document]
wdschei: I was wondering, but I figured you catch it if it was not needed.
I'll fix it.

repose-aggregator/artifacts/docker/build.gradle
```diff
@@ -19,24 +19,29 @@ docker {
...
+    }
+
     task("tag${baseImage}ImageVersion", type: DockerTagImage) {
         dependsOn "build${baseImage}Image"
         repository = 'rackerlabs/repose'
         tag = reposeVersion + tagSuffix
-        targetImageId { tasks["build${baseImage}Image"].getImageId() }
+        targetImageId { tasks["build${baseImage}ImageLocal"].getImageId() }
```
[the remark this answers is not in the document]
wdschei: :+1:

repose-aggregator/build.gradle
```diff
@@ -44,7 +44,8 @@ asciidoctor {
                'toc': 'left',
                'icons': 'font',
                'docinfo': 'shared',
-               'sectanchors': ''
+               'sectanchors': '',
+               'allow-uri-read': ''
```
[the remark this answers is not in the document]
wdschei: I didn't want to have to start copying all of the examples over too.
This seemed like an easy way to get there.

repose-aggregator/src/docs/asciidoc/architecture/container.adoc
```diff
@@ -173,13 +173,13 @@ For more details, see:
 * https://docs.oracle.com/javase/8/docs/technotes/guides/security/crypto/CryptoSpec.html#KeystoreImplementation
 
 == Proxy (Valve) Configuration of SSL/TLS Certificates
-Repose Valve is based on Jetty and uses its services for SSL/TLS.
+*Repose Valve* is based on Jetty and uses its services for SSL/TLS.
 
-To make the Repose Proxy Server deployment option work with origin SSL/TLS requests:
+To make the *Repose Proxy Server* deployment option work with origin SSL/TLS requests:
```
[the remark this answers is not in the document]
wdschei: That all sounds good.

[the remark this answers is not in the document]
wdschei: Updated.

repose-aggregator/src/docs/asciidoc/style-guide.adoc
```diff
@@ -55,4 +55,9 @@ Nesting in lists will correspond to the number of leading asterisks.
 ....
 
 == Citations
-* http://asciidoctor.org/docs/asciidoc-recommended-practices/[AsciiDoc Recommended Practices]
\ No newline at end of file
+* http://asciidoctor.org/docs/asciidoc-recommended-practices/[AsciiDoc Recommended Practices]
+
+== Repose is Bold
```
dmnjohns: Could/should this be generalized to all products? So *Docker* or *OpenShift* or other such things?

repose-aggregator/src/docs/asciidoc/recipes/getting-started.adoc
```diff
@@ -62,56 +62,60 @@ Following are some of the most common ones:
...
-
 After installing the WAR, there is additional setup to get everything working with Tomcat.
 See <<tomcat-setup.adoc#,Tomcat Setup>> for more details.
 
 ===== GlassFish Setup
-
 See <<glassfish-setup.adoc#,GlassFish Setup>> for more details.
 
 ===== Container Attributes known to affect Repose
-
```
dmnjohns: Should this be in the style guide? No blank lines after section headers?

repose-aggregator/src/docs/asciidoc/recipes/quick-start.adoc
```diff
@@ -0,0 +1,166 @@
+= Quick Start
+
+With the advent of container technology, *Repose* can be fully encapsulated and run as a service with minimal setup.
+
+By bundling the environment with the software itself, deploying *Repose* has become a much quicker and simpler process.
+Docker boasts security through isolation, and when run on platforms with native support, it requires very little overhead.
+To make everything as easy and versatile as possible, the *Repose* team now publishes Docker images on Docker Hub as part of every release.
```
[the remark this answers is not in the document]
wdschei: Sure. I like it.
Updated.

repose-aggregator/src/docs/asciidoc/recipes/quick-start.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+== Prerequisites
+Before you can use a *Repose Docker* image, you will need to install and have running a Docker Container Engine.
```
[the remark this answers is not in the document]
wdschei: I got the phrase from the [Docker Installation page](https://docs.docker.com/engine/installation/).
I not tied to it either way. I'm a Docker newbie that was just reading to get oriented.

repose-aggregator/src/docs/asciidoc/recipes/preventing-xml-bomb.adoc
```diff
@@ -1,3 +1,80 @@
...
+          </xsl>
+        </style>
+      </style-sheets>
+    </request-translation>
+  </request-translations>
+</translation>
+----
+
+== Limit methods
+Limit the amount of HTTP methods that are allowed by configuring the `http-methods` attribute.
```
dmnjohns: The amount of? Awkward wording. I would probably say the types of. Or maybe I would reword the whole section to Restrict methods.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
```diff
@@ -1,130 +1,147 @@
...
+This directory will also be treated as the Docker build context.
+====
+. If the Dockerfile was successfully built, a Docker image should be available within Docker.
+Run the following command to show all available images:
++
+----
+docker images
+----
+
+== Building a Repose Docker Image With the build file
```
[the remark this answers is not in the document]
dmnjohns: Oh, that's right. We are not very consistent about section header capitalization, are we? Need to get that in the style guide.

repose-aggregator/src/docs/asciidoc/recipes/quick-start.adoc
```diff
@@ -0,0 +1,166 @@
...
+The Docker Container Engine is available for almost all operating systems.
+The installation instructions for the Docker Container Engine are outside the scope of this Quick Start guide, but they can be found https://docs.docker.com/engine/installation/[here].
```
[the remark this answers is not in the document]
wdschei: Fair enough.
Updated.

repose-aggregator/src/docs/asciidoc/recipes/quick-start.adoc
wdschei: Cool. Updated.

repose-aggregator/src/docs/asciidoc/recipes/quick-start.adoc
wdschei: Yep. I found it somewhere else too.
Both fixed.

repose-aggregator/src/docs/asciidoc/recipes/quick-start.adoc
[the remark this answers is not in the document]
dmnjohns: Ah, I see. I suppose I can see calling that part of the Docker engine the container engine, but I stick by `Docker Engine` being the phrase which describes all the local Docker goodness (for the reasons in my last comment).

repose-aggregator/src/docs/asciidoc/recipes/quick-start.adoc
wdschei: I can change it, but I used this format in several places.
My rationale was that the sentence was very long and normal viewing would not let you see it.
It is at a logical break in the sentence structure and reinforces that these are `if`/`then` style sentences.

Thoughts?

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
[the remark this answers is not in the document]
wdschei: Totally done.

repose-aggregator/src/docs/asciidoc/services/index.adoc
```diff
@@ -15,11 +15,11 @@ Enables the local Repose node or an entire cluster to store various types of dat
...
 
 == <<metrics.adoc#, Metrics Service>>
-Enables the registration, collection, and reporting of metrics across Repose.
-Such metrics provide insight into the state of Repose and its components at any given time.
+Enables the registration, collection, and reporting of metrics across *Repose*.
+Such metrics provide insight into the state of *Repose* and its components at any given time.
 
 == <<phone-home.adoc#,Phone Home Service>>
-*IN PROGRESS*
+Sends usage data consisting of the filters and services in use along with contact information to the *Repose Team*.
```
dmnjohns: Since this is just a quick description of, and reference to, the service page, I probably wouldn't include the specifics of what data is sent. After all, that might change, and we probably don't want to have to update it here **and** on the service page.

repose-aggregator/src/docs/asciidoc/services/index.adoc
```diff
@@ -15,11 +15,11 @@ Enables the local Repose node or an entire cluster to store various types of dat
 The HTTP Connection Pool service is used for all outbound communication, and allows you to specify a number of things about those connections.
 
 == <<metrics.adoc#, Metrics Service>>
-Enables the registration, collection, and reporting of metrics across Repose.
-Such metrics provide insight into the state of Repose and its components at any given time.
+Enables the registration, collection, and reporting of metrics across *Repose*.
```
dmnjohns: Who wrote this glorious line?! So concise. So beautiful.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
wdschei: I do like the idea of putting `README`'s wherever they are needed.
It keeps the tribal knowledge from being lost and reinforces that others should be doing the same.

What all would you want to be in that document?

repose-aggregator/src/docs/asciidoc/services/index.adoc
```diff
@@ -3,10 +3,10 @@
 include::../_includes/in-progress.adoc[]
 
 == <<atom-feed-consumption.adoc#,Atom Feed Consumption Service>>
-*IN PROGRESS*
+Enables a *Repose* filter to read an Atom feed.
```
dmnjohns: Technically, another service could also use this service to consume an Atom feed.

Maybe something like:
> Consolidates the consumption of Atom feeds, and provides a simple interface to retrieve consumed data.

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
```diff
@@ -95,16 +110,16 @@ As mentioned above, this is achieved by configuring a keystore in the `dist-data
...
-Assuming all Repose nodes are configured identically, the most straight-forward way to make use of this security would be to use a single unique keystore as both the keystore and the truststore.
+Assuming all *Repose* nodes are configured identically, the most straight-forward way to make use of this security would be to use a single unique keystore as both the keystore and the truststore.
 This can be achieved by not explicitly configuring a separate truststore.
 Since each datastore node will have a copy of the keystore, each node will trust every other node.
 
 Client authentication in SSL/TLS can act as as alternate form of client validation, performing a task similar to that of an access control list.
 As such, the usage of client authentication may replace the need to configure the allowed-hosts section of the `dist-datastore.cfg.xml` file.
 
-The that distributed datastore will use a connection pool to communicate across nodes.
+The that distributed datastore will use the <<http-connection-pool.adoc#,HTTP Connection Pool service>> to communicate across nodes.
```
dmnjohns: > The that

Nope.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
```diff
@@ -1,130 +1,147 @@
...
+If you don't have the basics of these operations memorized, then that is fine.
+You can always look back at the <<quick-start.adoc#,Quick Start>> anytime you need to.
 
-== Docker Image
-All official, published Docker images for Repose, can be found on Docker Hub at:
+== Using a Repose Docker Image
+All official *Repose Docker* images are published on Docker Hub at:
 
-https://hub.docker.com/r/rackerlabs/repose/
+image::http://dockeri.co/image/rackerlabs/repose[Repose Docker,link=https://hub.docker.com/r/rackerlabs/repose/]
```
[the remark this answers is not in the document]
wdschei: I totally stole it from @dimtruck 's original docs.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
wdschei: The `man` pages call them `option`'s.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
wdschei: That is kind of where I was at. I didn't feel the need to `include::` it in, but at the same time wanted to make it as accessible as possible.
I didn't like this solution either, but couldn't think of a better one.

Do we have access to the version number in a way we could reference it during the doc build?

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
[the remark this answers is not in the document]
wdschei: Yes. Yes it is.
That must have been towards the end of one of my docs days.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
wdschei: Also, this block was just moved into another section.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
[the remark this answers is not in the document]
wdschei: Fixed.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
wdschei: I thought it was a logical next step to anyone that was building their own images to see how we build it. That way they can incorporate it into their own build system. Basically it was a way to say, "hey you shouldn't be building it by hand, so here's a better way."

repose-aggregator/src/docs/asciidoc/style-guide.adoc
wdschei: I don't know. I took it more as we were promoting us and our direct use of other technologies.

That said, I would say no. However, I'm open for discussion.

repose-aggregator/src/docs/asciidoc/recipes/getting-started.adoc
wdschei: Adding it now.

repose-aggregator/src/docs/asciidoc/services/index.adoc
wdschei: Makes sense.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
dmnjohns: Only if their build system is Gradle.

Honestly, I also wouldn't call this a better way. It's an automated way that fits in the build system of Repose, but it:
1. Relies on a Gradle plugin that has a history of giving us trouble.
1. Relies on a Gradle plugin that is pinned to a certain version of the Docker server API, and thus can get out-of-date. This is particularly annoying if you are trying to build an older Repose `Dockerfile` which puts you on a branch relying on an older version of the Gradle plugin.
1. Requires another tool (i.e., Gradle).
1. Reduces flexibility (e.g., you cannot tag an image when building it -- that's another step).

So yes, we benefit from using it because it is easy to incorporate in our release process, but if you aren't releasing Repose, I'm not sure why you would want to use it.

repose-aggregator/src/docs/asciidoc/services/index.adoc
wdschei: FIxed.

repose-aggregator/src/docs/asciidoc/recipes/rate-limiting.adoc
```diff
@@ -1,72 +1,124 @@
 = Rate Limiting
 
-include::../_includes/in-progress.adoc[]
+== Overview
+A common use case for *Repose* is rate limiting.
+It limits how many requests per some unit of time (e.g. 10 requests per minute) are allowed to be made.
+In our example, we'll be using the `X-PP-User` header to indicate who we are for rate limiting purposes.
```
dmnjohns: I might change this line to:
> In this recipe, we will be using the `X-PP-User` header to identity the user for rate limiting purposes.

Subtle change, but I think it helps. That could just be me, though.

repose-aggregator/src/docs/asciidoc/recipes/preventing-xml-bomb.adoc
wdschei: Updated.

repose-aggregator/src/docs/asciidoc/services/index.adoc
wdschei: :thinking:

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
wdschei: Yes. `The that`. Totally intentional.

repose-aggregator/src/docs/asciidoc/recipes/rate-limiting.adoc
```diff
@@ -1,72 +1,124 @@
...
 ----
-<?xml version="1.0" encoding="UTF-8"?>
-<distributed-datastore xmlns='http://docs.openrepose.org/repose/distributed-datastore/v1.0'>
-    <allowed-hosts allow-all="false"/>
-    <port-config>
-        <port port="9999" cluster="repose"/>
-        <port port="7777" cluster="repose" node="node2"/>
-    </port-config>
-</distributed-datastore>
+include::https://raw.githubusercontent.com/rackerlabs/repose/master/repose-aggregator/components/filters/rate-limiting-filter/src/main/resources/META-INF/schema/examples/rate-limiting.cfg.xml[]
```
dmnjohns: Another spot where we are relying on GitHub. In this case, you're also breaking versioning, since that configuration could differ between master and the current version. I understand that `include` effectively takes a snapshot of what is being included, but consider cutting a bug fix release, or some other release which puts us behind the current version.

Since you are using `include` anyway, why not just `include` the file in the repository directly? That should just be simple file resolution, so you shouldn't have to copy that file around or anything.

repose-aggregator/src/docs/asciidoc/services/datastores.adoc
wdschei: Fixed.

repose-aggregator/src/docs/asciidoc/recipes/rate-limiting.adoc
wdschei: Sure. That keeps it consistent with the wording.

repose-aggregator/src/docs/asciidoc/recipes/rate-limiting.adoc
```diff
@@ -1,72 +1,124 @@
...
-    <limit-group id="unlimited" groups="IP_Super" default="false"/>
-</rate-limiting>
+[source,shell]
+.console session
 ----
+# make the script executable
+chmod +x test_rate_limiting.sh
+
+# run the script
+./test_rate_limiting.sh
```
dmnjohns: Since you are assuming the working directory is the directory containing the `test_rate_limiting.sh` anyway (from the previous command), why not drop the leading `./`?

repose-aggregator/src/docs/asciidoc/recipes/rate-limiting.adoc
wdschei: Good point. Relative pathing hell, here I come.

repose-aggregator/src/docs/asciidoc/recipes/rate-limiting.adoc
wdschei: Unless they have the current working directory in their path (which is bad by the way), then you must include the `./` to make it go.

repose-aggregator/src/docs/asciidoc/recipes/rate-limiting.adoc
wdschei: This also obsoletes the `allow-uri-read` attribute.

repose-aggregator/src/docs/asciidoc/recipes/rate-limiting.adoc
dmnjohns: Oh derp. That's right. The shell resolves the pathing bits to give you an absolute path, doesn't it? So the script is not on the `PATH`, and thus, must be run from an absolute path, and the `./` is a shorthand for getting that. Not sure what I was thinking.

dmnjohns: So I guess the alternative would be:
`sh test_rate_limiting.sh`

... but that's not really any better. So I guess I don't care.

repose-aggregator/src/docs/asciidoc/recipes/quick-start.adoc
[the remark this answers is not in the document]
wdschei: Cool. I'll update them.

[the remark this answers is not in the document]
wdschei: Fixed.

repose-aggregator/src/docs/asciidoc/recipes/quick-start.adoc
wdschei: I thought they were better for the tutorial too.

repose-aggregator/src/docs/asciidoc/recipes/quick-start.adoc
dmnjohns: Fair arguments. Personally, I would prefer it on a single line for two reasons:
1. It will be rendered on a single line in the output. Therefore, I would like to see it closer to how the user sees it (although, I realize that the one-sentence-per-line rule already differentiates the source from the output).
1. Line wrapping can make it easy enough to read so that you don't have to worry about not seeing it. Most editors should support line wrapping.

In the end, I really only care that we are consistent. So if we want to line break on punctuation it's no big deal, as long as we always do so.

... but since this is a shared document, what do @adrianjgeorge and @Mario-Lopez think?

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
wdschei: I was referring to the building by hand and here is an automated way. The better being automated, not necessarily exactly our way.

wdschei: But if we do move this out to a README, then I'm good with that too.

repose-aggregator/src/docs/asciidoc/recipes/getting-started.adoc
wdschei: Added.

repose-aggregator/src/docs/asciidoc/services/index.adoc
wdschei: Updated.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
dmnjohns: Yeah, I went back and wrote this comment at the end. I know some of the stuff was migrated, but other stuff was removed (that is, a search for it only yielded one result, which was highlighted with red). Of course, some of the removed stuff may have just been changing to be worded in a different way, but some of it is actually gone.

To me, developer documentation would basically provide a quick reference to the concepts and tools used to do what we do with Docker.

So one of the major changes is the organization. This recipe groups information by the goal is helps to achieve. The old page grouped information by what Docker concept it fell under. The recipe approach is great for a user who just wants to achieve something, while the old page is great for a developer to quickly find a way to get something working. So if I, as a developer, wanted to know how to inspect the state of my container, having a section named `Inspection` makes it easy to find that information.

Which brings me to the `Inspecting` section. It is just plain gone. That has certainly been useful to me while working with Docker, so I believe it is probably useful to others, as well. 

Or the bit about log files. Yes, that section needs to be updated because mounting the log directory is not the recommended way of reading logs, but it's nice to have a quick reference on how to read the logs. I know that's on the quick start page, so that may just be my organizational complaint again... although, developer documentation could provide multiple options for reading logs (e.g., mounting the log directory as a volume vs using Docker log management) even though we would not want to tell users to mount a log directory. In this specific example, it may be easier just to mount the log directory if, say, we are debugging with a set of configs handed to us by some other team.

I also liked the `Docker` section that told you, at a high level, what is necessary to run Repose in Docker.

So yeah, different ways of presenting (sometimes different) information to different audiences. That's my argument for keeping most of the old page around (although perhaps with some needed updates) as developer documentation.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
dmnjohns: Fair enough. `option`s are the general term, right? And `flag`s are generally accepted to be boolean `option`s? That's how I've always used them, but being consistent with the `man` pages is totally legit.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
dmnjohns: Sure, the Asciidoctor plugin for Gradle sets the `project-version` attribute.

Source: http://asciidoctor.org/docs/asciidoctor-gradle-plugin/#options-attributes

I would still probably suggest just pulling the files you care about over just so that resources are stored and served consistently. That is, resources are served from the same servers and domain. Actually, I'm not sure if that's valid since we are hosting on GitHub anyway. I'm not sure if GitHub pages content is distributed the same was as Git repository, and while I would hope that GitHub pages do not go down every time GitHub.com does, I couldn't say for certain. Even so, pulling the files in insulates us from that concern if we do every switch hosts for some reason.

repose-aggregator/src/docs/asciidoc/recipes/preventing-xml-bomb.adoc
```diff
@@ -1,3 +1,80 @@
...
+          </xsl>
+        </style>
+      </style-sheets>
+    </request-translation>
+  </request-translations>
+</translation>
+----
+
+== Restricting Methods
+Control the of HTTP method verbs that are allowed by configuring the `http-methods` attribute.
```
dmnjohns: > Control the of HTTP

Looks like we have a vestigial word there.

repose-aggregator/src/docs/asciidoc/recipes/rate-limiting.adoc
dmnjohns: Oh no! I just learned about that, too!

Oh well.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
dmnjohns: I see. Yeah, I think this totally fits in a README since it's our system, I just can't come up with a reason why someone working with Repose would need to know how to do this. To be fair, the Docker build process is pretty automatic, you're just replacing a single Docker command with a single Gradle command. One, I think, is just clearer (Docker). Also, as far as I know, we haven't publicly documented other Gradle tasks.

repose-aggregator/src/docs/asciidoc/recipes/quick-start.adoc
wdschei: Since there was no input, I'll move it to one line for now.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
wdschei: Ok. So these were brought in as includes and then wrapped with the HTML to make them expandable.

Thoughts?

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
wdschei: This was added to the new developer docs in almost the same format.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
wdschei: This was moved into the new developer docs.

wdschei:
This also brought in the newest `gradle-linkchecker-plugin` and updated the `Dockerfile`'s to allow local logging on the container with some additional setup.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
```diff
@@ -76,24 +76,32 @@ Now if we wanted to, we could change the Docker options that define the environm
 If we wanted to forward random ports instead of explicitly declaring the port mapping, we could replace the `--publish 8080:8080` option with `--publish-all` instead.
 
 == Manually Building a Repose Docker Image
-The latest version of the Dockerfiles used to build the *Repose Docker* images published on Docker Hub can be found at:
+The Dockerfile's used to build v{project-version} of the *Repose Docker* images published on Docker Hub are:
 
-- https://raw.githubusercontent.com/rackerlabs/repose/master/repose-aggregator/artifacts/docker/src/docker/ubuntu/Dockerfile[Debian/Ubuntu]
-- https://raw.githubusercontent.com/rackerlabs/repose/master/repose-aggregator/artifacts/docker/src/docker/centos/Dockerfile[RedHat/Centos]
++++ <details><summary>Ubuntu</summary><div> +++
```
dmnjohns: Not safe or recommended. This line assumes that HTML will be the output format, which makes our documentation far less portable. 

To use HTML in our documentation, we should use [Conditional Preprocessor Directives](http://asciidoctor.org/docs/user-manual/#conditional-preprocessor-directives). It's pretty similar to what you have here, with one more step. Check out [the includable search document](https://raw.githubusercontent.com/rackerlabs/repose/master/repose-aggregator/src/docs/asciidoc/_includes/search.adoc) for an example.

dmnjohns: As a follow-up, that this is being used to make the included `Dockerfile`s hideable indicates that they will not display well as full includes. Did we actually want to switch from links to includes, or did we want to copy and serve the `Dockerfiles` from our site and just link to them here?

Copying files into our output isn't too hard, and you can see an example of it here:
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/build.gradle#L23-L36

Basically, have Gradle perform a literal file copy in a `Copy` task, then make the `asciidoctor` task depend on the `Copy` task.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
```diff
@@ -102,7 +110,7 @@ Let's break that command down and take a closer look at what it is doing:
     See our http://www.openrepose.org/versions/latest/release-notes.html[Release Notes] page for a list of available versions.
 <3> This option defines the tag to apply to the image.
     It is in `repository:tag` format.
-<4> The last parameter is the path to the *Repose* Dockerfile.
+<4> The last parameter is the directory that contains the Dockerfile.
```
dmnjohns: Heh, technically it's the path to the directory that contains the Dockerfile. The Docker man page just refers to it as path, though:
> Usage:	docker build [OPTIONS] PATH | URL | -`

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
```diff
@@ -76,24 +76,32 @@ Now if we wanted to, we could change the Docker options that define the environm
...
+[source,bash,subs="verbatim,attributes"]
 ----
-docker build                                             \ <1>
-   --build-arg REPOSE_VERSION=8.7.0.2                    \ <2>
-   --tag local/repose:v8.7.0.2-ubuntu                    \ <3>
-   ./repose-aggregator/artifacts/docker/src/docker/ubuntu/ <4>
+docker build                          \ <1>
+   --build-arg REPOSE_VERSION={project-version} \ <2>
+   --tag local/repose:v{project-version}-local  \ <3>
+   /dir/containing/dockerfile/          <4>
```
dmnjohns: Aw, the callouts alignment was lost with this change.

repose-aggregator/src/docs/asciidoc/developer/docker.adoc
```diff
@@ -0,0 +1,47 @@
+= Repose and Docker - Tips and Tricks
+
+This is a collection of information cobbled together from several sources, but mostly from sifting through the https://docs.docker.com/[Official Docker Documentation] and then applying it to *Repose*.
+
+The rest of this assumes you are already familiar with the contents of the <<../recipes/quick-start.adoc#,Quick Start>> and <<../recipes/docker.adoc#,Docker>> recipes.
+Obviously since you made it here, your hunger for *Repose Docker* knowledge was not satiated.
+
+== Overview
```
dmnjohns: Overview of what? The steps listed are specifically for running Repose in Docker, so I'm not a fan of just calling this section overview.

repose-aggregator/src/docs/asciidoc/recipes/index.adoc
```diff
@@ -26,17 +26,18 @@ Covers the typical Repose setup -- a Valve installation (i.e., standalone applic
 == <<war-installation.adoc#,WAR Installation>>
 Covers setting up Repose inside of a user-provided servlet container.
 
-== <<quick-setup.adoc#,Quick Setup>>
-*IN PROGRESS*
-If you want to stand up a Repose instance quickly, visit the Quick Setup page.
+== <<quick-start.adoc#,Quick Start>>
+If you want to stand up a Repose instance quickly, visit the Quick Start.
```
dmnjohns: Hm, looking at this again, the following seems a little redundant:
> visit the Quick Start.

This is just the index page, and the section header is already a link with the same works, so I'm not sure that bit is necessary. Just the description should suffice. I might suggest something like:
> Stand up a Repose instance quickly!
or
> Learn how to stand up Repose as quickly as possible!

dmnjohns: This comment goes for the other sections with lines reading:
`See ...`

repose-aggregator/src/docs/asciidoc/developer/docker.adoc
```diff
@@ -45,3 +46,49 @@ Since this method will not automatically tag the image, you will need to tag it
 docker tag IMAGE_ID local/repose:v{project-version}-ubuntu
 ----
 
+== Logging inside the Repose Docker Container
+The published images adhere to the expectations of 12Factor logging.
```
dmnjohns: I might include a snippet about `docker logs` here to show the standard way of accessing logs.

repose-aggregator/src/docs/asciidoc/recipes/preventing-xml-bomb.adoc
wdschei: Removed.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
wdschei: Updated.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
wdschei: Updated.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
wdschei: It will realign when the `-SNAPSHOT` is not present.

repose-aggregator/src/docs/asciidoc/developer/docker.adoc
wdschei: Updated.

repose-aggregator/src/docs/asciidoc/recipes/index.adoc
wdschei: Updated.

repose-aggregator/src/docs/asciidoc/developer/docker.adoc
wdschei: Updated.

repose-aggregator/src/docs/asciidoc/developer/docker.adoc
```diff
@@ -5,16 +5,16 @@ This is a collection of information cobbled together from several sources, but m
...
-.. Build the Docker image. See: <<../recipes/docker.adoc#_manually_building_a_repose_docker_image,Manually Building a Repose Docker Image>>
-.. Build the Docker image from the project. See <<Building a Repose Docker Image with the build file>>
-. Run the Repose Docker image in a Docker container. See: <<../recipes/quick-start.adoc#_running_a_docker_container,Running a Docker Container>>
+.. <<../recipes/quick-start.adoc#_repose_docker_images,Download a *Repose* Docker Image>>.
+.. <<../recipes/docker.adoc#_manually_building_a_repose_docker_image,Build a *Repose* Docker Image>>
+.. <<Build a Repose Docker Image with the build file>>
+. <<../recipes/quick-start.adoc#_running_a_docker_container,Run the Repose Docker image in a Docker container>>
 
-== Building a Repose Docker Image with the build file
+== Build a Repose Docker Image with the build file
```
dmnjohns: Why the change here? I honestly don't have a preference, and I don't know if one is more correct that the other, but now you have two distinct tenses in section headers (i.e., `Running` vs `Build`).

repose-aggregator/src/docs/asciidoc/developer/docker.adoc
```diff
@@ -91,4 +91,4 @@ docker run                                   \
     So in this case, the contents of `/var/log/repose` in the container will be available on the host's `/var/log/repose` directory.
     This allows us to `tail` or persist the logs even if the Docker container is restarted.
 
-If you didn't recognize it, this is almost the identical command provided in the <<../recipes/docker.adoc#_using_a_repose_docker_image,Using a Repose Docker Image>> section, but with the extra volume mount option.
+If you didn't recognize it, this is almost the identical command provided in the <<../recipes/docker.adoc#_using_a_repose_docker_image,Docker recipe's Using a Repose Docker Image>> section, but with the extra volume mount option.
```
dmnjohns: That became a long link. Might I suggest rephrasing to:
```
If you didn't recognize it, this is almost identical to the command provided in<<../recipes/docker.adoc#_using_a_repose_docker_image,Using a Repose Docker Image>>, but with the extra volume mount option.
```

repose-aggregator/src/docs/asciidoc/recipes/index.adoc
```diff
@@ -3,7 +3,7 @@
 include::../_includes/in-progress.adoc[]
 
 == <<quick-start.adoc#,Quick Start>>
-If you want to stand up a *Repose* instance quickly, visit the Quick Start.
+If you want to stand up a *Repose* instance quickly.
```
dmnjohns: Since this is the same page as the one linked to from the primary index, I would just use the same description.
> Learn how to stand up *Repose* as quickly as possible!

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
```diff
@@ -110,13 +101,9 @@ Let's break that command down and take a closer look at what it is doing:
...
 <3> This option defines the tag to apply to the image.
     It is in `repository:tag` format.
-<4> The last parameter is the directory that contains the Dockerfile.
-    The Dockerfile itself must exist as a child of the provided directory.
-+
-[NOTE]
-====
-This directory will also be treated as the Docker build context.
-====
+<4> The last parameter is the path of the `docker build` context.
```
dmnjohns: I can't say why, exactly,  but `path to the` feels more correct than `path of the` to me.

PR #1831, opened 2017-10-25
Prep for v8.7.1.0 release.

wdschei:


repose-aggregator/src/docs/asciidoc/release-notes.adoc
```diff
@@ -1,9 +1,10 @@
 = Release Notes
 
-== In Progress Work
-
+== 8.7.1.0 (2017-10-25)
 * https://repose.atlassian.net/browse/REP-6133[REP-6133] - Updated the published Docker images to turn off local logging by default to be more in line with the expectations of a https://12factor.net/logs[Twelve-Factor App].
 * https://repose.atlassian.net/browse/REP-6135[REP-6135] - Updated the published Docker images to support running the container using an arbitrarily assigned user ID as is expected by the https://docs.openshift.com/container-platform/3.6/creating_images/guidelines.html#openshift-container-platform-specific-guidelines[OpenShift Container Platform].
+* https://repose.atlassian.net/browse/REP-6179[REP-6179] - Converted more old Wiki Docs over to the new www.openrepose.org/versions/latest/[Versioned Docs].
+* https://repose.atlassian.net/browse/REP-6186[REP-6186] - Updated the automated Release Verification to force the use of Java 8 since some GNU/Linux distributions are already providing Java 9 by default.
```
dmnjohns: Do we actually care if this makes it in the release notes? I ask because it isn't user-facing, just internal process.

wdschei: I don't know. I felt like it was kind of user facing since it now documents that we only test against Java 8 even though Java 9 is available and should work.

wdschei: This also seems to beg the issue of when do we officially start supporting Java 9 and furthermore stop supporting Java 8.

wdschei: Here is the official Java EOL schedule:
* http://www.oracle.com/technetwork/java/eol-135779.html

dmnjohns: Good question on the Java 8 vs 9 thing, but I don't have an answer on that one.

As for stating that we only test against Java 8, the release notes don't seem like the best place for that information. I agree that we should tell the user what versions of Java are supported (which seems to be strangely lacking in our docs?), but I don't think doing so by making them parse the release notes is the right call. Consider when we move to Java 10. We just have another release note about the switch, but the user does not know all of the supported Java versions, only that the version currently being used to test is supported.

Anyway, I'll merge this if we're good with it here, but I don't really see the need.

dmnjohns: Actually, while I still think there's a better place for putting the supported Java versions, I'll go ahead and merge this since it is, at least, information for us that is tied to a specific release.

PR #1834, opened 2017-10-27
REP-6159 RegEx RBAC filter

wdschei:


repose-aggregator/components/filters/regex-rbac-filter/src/main/resources/META-INF/schema/config/regex-rbac.xsd
```diff
@@ -0,0 +1,145 @@
...
+    </xs:simpleType>
+
+    <xs:complexType name="ResourcesType">
+        <xs:annotation>
+            <xs:documentation>
+                <html:p>
+                    This is a list of resources, HTTP methods, and roles allowed to access them.
+
+                    The special HTTP methods ANY and ALL are shorthand for the four standard REST methods
+                    (GET, PUT, POST, DELETE). Other methods are supported, but they must be explicitly stated.
```
adrianjgeorge: Patch is not included in any/all?

wdschei: It wasn't for the Simple RBAC.

repose-aggregator/components/filters/regex-rbac-filter/src/main/resources/META-INF/schema/config/regex-rbac.xsd
```diff
@@ -0,0 +1,145 @@
...
+        </xs:annotation>
+
+        <xs:simpleContent>
+            <xs:extension base="ResourcesString">
+                <xs:attribute name="href" type="xs:anyURI" use="optional">
+                    <xs:annotation>
+                        <xs:documentation>
+                            <html:p>
+                                Specifies a location to an external file which contains the RegEx RBAC resources.
+                                If the message element has a value and the href attribute is configured,
```
adrianjgeorge: message element?

repose-aggregator/components/filters/regex-rbac-filter/src/main/resources/META-INF/schema/examples/regex-rbac.cfg.xml
```diff
@@ -0,0 +1,13 @@
...
+<regex-rbac xmlns="http://docs.openrepose.org/repose/regex-rbac/v1.0"
+             mask-rax-roles-403="false">
+    <resources>
+        /path/to/this   GET     role1,role2,role3,role4
+        /path/to/this   PUT     role1,role2,role3
+        /path/to/this   POST    role1,role2
+        /path/to/this   DELETE  role1
+        /path/to/that   GET,PUT ALL
+        /path/to/that   ALL     role1
+        /path/{to}/wild GET     role1
```
adrianjgeorge: can we be more obvious about the fact that these are regexs?

repose-aggregator/components/filters/regex-rbac-filter/src/main/resources/META-INF/schema/config/regex-rbac.xsd
wdschei: Copy/Paste. Should be `resources`.
Updated.

repose-aggregator/components/filters/regex-rbac-filter/src/main/resources/META-INF/schema/examples/regex-rbac.cfg.xml
wdschei: I have already updated that in my local copy.
Pushing now.

repose-aggregator/components/filters/regex-rbac-filter/src/main/resources/META-INF/schema/config/regex-rbac.xsd
wdschei: I can add it here if we will need it.

adrianjgeorge: That's super non-intuitive. why did we do this the first time?

wdschei: The original was intended to solve "simple" issues and worked under the assumption of only supporting basic CRUD operations.

wdschei: I can make `ANY`/`ALL` be truly any or all and basically just not check the method.

adrianjgeorge: I'd say go ahead and support them all on any/all. lets not make them weirdly counter intuitive

repose-aggregator/components/filters/regex-rbac-filter/src/main/resources/META-INF/schema/config/regex-rbac.xsd
```diff
@@ -0,0 +1,155 @@
...
+                <html:p>The root config type for the RegEx RBAC filter configuration file.</html:p>
+            </xs:documentation>
+        </xs:annotation>
+
+        <xs:sequence>
+            <xs:element name="delegating" type="DelegatingType" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="resources" type="ResourcesType" minOccurs="1" maxOccurs="1"/>
+        </xs:sequence>
+
+        <xs:attribute name="mask-rax-roles-403" type="xs:boolean" use="optional" default="false">
```
dmnjohns: I know naming this attribute this way makes the transition to the full `api-validator` filter easier, but I just wanted to point out that `rax-roles` is not defined in this context.

`mask-forbidden` or `mask-403` seem like more accurate and intuitive names here.

wdschei: We originally talked about keeping it the same name for consistency of the feature's behavior across all of the filters.

That said, I'm ok changing the name if we want to. This is more of an end user expectation kind of thing.

dmnjohns: I figured as much, just thought I'd chime in anyway.

In my mind, it's better to name the attribute to make sense in the context it is being used than to name the attribute for consistency across contexts. The reason being that a user is very unlikely to be using both this filter and either the `simple-rbac` or `api-validator` filter, so the consistency isn't buying us anything. Of course, it does buy us something when a user transitions from one filter to another, but I doubt many users would transition from the fuller-featured `simple-rbac` filter or `api-validator` filter to this filter, so it would only be helpful migrating from this filter up to one of those filters. So yes, there is value there, but considering how different this filter is from either of those filters (and thus, how much the user is already going to have to change), I would vote to keep this filter's configuration simpler by not including irrelevant information (i.e., `rax-roles`) in the attribute name. After all, simple is in the name of the filter. ;)

wdschei: Well simple is not in *this* filter's name, that's in the other one.

However, I see the point. I also think consistency in named behavior in the context of *Repose* in general and not just a specific configuration makes sense too. Especially since this is coming to light that this filter will probably be more limited in lifespan and that teams will have to migrate to the full WADL based solutions in the future where the name is used.

As these are just names, it will be easy to do a find/replace on it. I'll wait for the rest of the team to weigh in before making any changes.

dmnjohns: Gah. That's what I get for adding that line at the last second. My brain got mixed up because we kept tossing the name "super-simple-rbac" around. Anyway, I dropped that sentence to make you look silly.

repose-aggregator/artifacts/valve/src/test/scala/org/openrepose/valve/ReposeJettySSLTest.scala
```diff
@@ -145,7 +145,7 @@ class ReposeJettySSLTest extends FunSpec with Matchers with BeforeAndAfterAll {
 
   def selectiveRequest(protocols: Array[String] = null, ciphers: Array[String] = null): Unit = {
     //Protocol names come from here:
-    //http://docs.oracle.com/javase/7/docs/technotes/guides/security/StandardNames.html#SSLContext
+    //http://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html#SSLContext
```
adrianjgeorge: Did you check that link?

wdschei: Yep. I checked them all. There is one that was super weirdly changed, but it is the right one.

repose-aggregator/components/filters/regex-rbac-filter/src/main/scala/org/openrepose/filters/regexrbac/RegexRbacFilter.scala
```diff
@@ -0,0 +1,203 @@
...
+
+  override def doConfigurationUpdated(newConfigurationObject: RegexRbacConfig): Unit = {
+
+    def parseLine(line: String): Option[Resource] = {
+      val values = line.trim.split("\\s+")
+      values.length match {
+        case 3 =>
+          Some(Resource(
+            stringToRegexString(values(0)),
+            Try(values(1).split(',').toSet[String].map(_.trim) map (_.toUpperCase)).getOrElse(Set.empty),
```
dmnjohns: One of the maps on this line is called without using dot notation, which seems inconsistent with other usages.

wdschei: Yep. Completely agree with that.
Updated.

repose-aggregator/components/filters/regex-rbac-filter/src/main/resources/META-INF/schema/config/regex-rbac.xsd
```diff
@@ -0,0 +1,155 @@
...
+        <xs:simpleContent>
+            <xs:extension base="ResourcesString">
+                <xs:attribute name="href" type="xs:anyURI" use="optional">
+                    <xs:annotation>
+                        <xs:documentation>
+                            <html:p>
+                                Specifies a location to an external file which contains the RegEx RBAC resources.
+                                If the resources element has a value and the href attribute is configured,
+                                then the RegEx RBAC filter will use the contents of the referenced location for the
+                                endpoints configuration.
```
dmnjohns: It appears that the filter will actually prefer the inline resources to any externally referenced resources (which totally seems like the right call to me, but is not how I interpret this blurb).

wdschei: There is no preference, they are simply mutually exclusive ways of representing the resources.

dmnjohns: Derp. I didn't look for the assertion, just looked at the code. Fair enough. In that case, the whole "If the resources element has a value and the href attribute is configured..." bit here is probably unnecessary.

wdschei: Updated to make it more clear.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/regexrbac/RegexRbacAccessTest.groovy
```diff
@@ -0,0 +1,228 @@
...
+        "/all"                      | "role1, role2"
+        "/any"                      | "role1, role2"
+        "/all/any"                  | "role1"
+        "/any/all"                  | "role1"
+        "/roles/simple"             | "role1"
+        "/roles/simple"             | "role2"
+        "/roles/simple"             | "role1,role2"
+        "/roles/simple"             | "role1, role2"
+        "/roles/nbsp"               | "nb sp"
+        "/roles/nbsp/end"           | "nbsp "
```
dmnjohns: This test currently fails for me locally. The reason is that the space is at the end of the role. The configured role of `nbsp ` has the `&nbsp;` character translated into a normal space _after_ trimming occurs. The `HttpServletRequestWrapper`, however, trims the space around all header values. The result is that the translated, configured `nbsp ` role does not match the `nbsp` role pulled from the request.

So we can either:

1. Trim the space around configured roles _after_ replacing the `&nbsp;` character. We would just have to be clear that leading/trailing whitespace, including `&nbsp;` is ignored. Also, I'm not sure if a role can exist that starts or ends with a space in Identity. If it can, then we do not have a way to account for that at the moment.
1. Consider this test invalid and remove it.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/regexrbac/RegexRbacAccessTest.groovy
```diff
@@ -0,0 +1,228 @@
...
+        then: "Repose responds with a 403"
+        messageChain.handlings.size() == 0
+        messageChain.receivedResponse.code as Integer == SC_FORBIDDEN
+
+        where:
+        resource                    | roles
+        "/roles/simple"             | "forbidden-role"
+        "/roles/casing"             | "role1"
+        "/roles/nbsp"               | "nb sp"
+        "/roles/nbsp/end"           | "nbsp "
```
dmnjohns: You can't tell the different, but these tests use the `&nbsp;` character rather than the space character. I just wanted to point these tests out because they make it apparent that we do not support roles which include the `&nbsp;` character. When a role is configured with a `&nbsp;`, it will _always_ be translated into a normal space.

Again, I'm not sure if that's a valid role in Identity, but it is something to be aware of.

repose-aggregator/tests/functional-tests/src/integrationTest/configs/features/filters/regexrbac/common/regex-rbac.cfg.xml
```diff
@@ -0,0 +1,29 @@
...
+        /url/regex/[^/\d]+ ANY ANY
+        /method/head HEAD ANY
+        /method/get GET ANY
+        /method/put PUT ANY
+        /method/post POST ANY
+        /method/patch PATCH ANY
+        /method/delete DELETE ANY
+        /method/options OPTIONS ANY
+        /roles/simple ANY role1,role2
+        /roles/nbsp ANY role1,nb sp,role2
```
wdschei: This isn't a literal ASCII `nbsp` it is the Unicode NBSP char: `xA0`

wdschei: The conversion is here:
> https://github.com/rackerlabs/repose/blob/2adc0372d5af80e4d26773f67e3a06db799f7ff4/repose-aggregator/components/filters/regex-rbac-filter/src/main/scala/org/openrepose/filters/regexrbac/RegexRbacFilter.scala#L155

wdschei: Never mind. I jumped the gun on that one.

repose-aggregator/components/filters/regex-rbac-filter/src/test/scala/org/openrepose/filters/regexrbac/RegexRbacFilterTest.scala
```diff
@@ -0,0 +1,504 @@
...
+      filter.destroy()
+
+      Then("the filter should unregister with the ConfigurationService")
+      verify(mockConfigService).unsubscribeFrom(eql("another-name.cfg.xml"), same(filter))
+    }
+  }
+}
+
+object RegexRbacFilterTest {
+  private final val XRolesHeader = "X-Roles"
```
adrianjgeorge: This one should already be in a constants file somewhere.

repose-aggregator/components/filters/regex-rbac-filter/src/test/scala/org/openrepose/filters/regexrbac/RegexRbacFilterTest.scala
```diff
@@ -0,0 +1,504 @@
...
+  private final val XRelevantRolesHeader = "X-Relevant-Roles"
```
adrianjgeorge: Since this is a thing now, might as well make it a constant as well

repose-aggregator/components/filters/regex-rbac-filter/src/test/scala/org/openrepose/filters/regexrbac/RegexRbacSchemaTest.scala
```diff
@@ -0,0 +1,79 @@
...
+package org.openrepose.filters.regexrbac
+
+import org.junit.runner.RunWith
+import org.openrepose.commons.test.ConfigValidator
+import org.scalatest.{FunSpec, Matchers}
+import org.scalatest.junit.JUnitRunner
+import org.xml.sax.SAXParseException
+
+@RunWith(classOf[JUnitRunner])
+class RegexRbacSchemaTest extends FunSpec with Matchers {
```
adrianjgeorge: Thought we ended up with a base class for these? Did that not work out?

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/commons/config/manager/UpdateFailedException.java
```diff
@@ -20,6 +20,10 @@
 package org.openrepose.commons.config.manager;
 
 public class UpdateFailedException extends Exception {
+    public UpdateFailedException(String message) {
```
adrianjgeorge: Why was this added? I'm not seeing the usage (could be i'm losing it in everything else).

repose-aggregator/src/docs/asciidoc/filters/api-validator.adoc
```diff
@@ -153,7 +153,7 @@ This configuration expands the basic example in order to show off all of the fea
     Triggers off of `X-Roles` header. +
     *DEPRECATED:* Roles defined outside of the WADL will not be supported in *Repose* 9 and this attribute will not be available.
 <2> If the api-validator config `multi-match` is set to `true` then the default validator will be the first validator to process the incoming request.
-    If multi-match is set to `false` and if no validator is matched to the users' roles, then the filter will use the default validator. (Default: false) +
+    If multi-match is set to `false` and if no validator is matched to the users' roles, then the filter will use the default validator. (Default: `false`) +
```
adrianjgeorge: can you pull all these unrelated doc changes into a separate pr?

repose-aggregator/src/docs/asciidoc/filters/regex-rbac.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+[NOTE]
+====
+This header is not configurable.
+====
+
+=== Required Preceding Filters
+This filter has no dependencies on other filters and can be placed wherever it is needed in the filter chain.
+
+However, due to the nature of this filter, it is typically placed early in the filter chain immediately after any authentication filters.
```
adrianjgeorge: Maybe add `such as Keystone v2' with link.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/regexrbac/RegexRbacAccessTest.groovy
wdschei: Option was chosen, so this test should pass now.

repose-aggregator/src/docs/asciidoc/filters/regex-rbac.adoc
```diff
@@ -0,0 +1,176 @@
...
+    (Default: `regex-rbac`)
+<5> Specifies a location to an external file which contains the RegEx RBAC resources.
+    If the message element has a value and the `href` attribute is configured, the RegEx RBAC will use what is configured in the value.
+    If the file that the `href` attribute points to is modified, the RegEx RBAC will not reload the configuration.
+    So the new RBAC file should be placed in a new file name (e.g. Dated) and the regex-rbac.cfg.xml file updated to point to it in order to guarantee the changes are utilized.
+
+== Additional Information
+
+=== Delegation
+In some cases, you may want to delegate the validation of a request down the chain to either another filter or to the origin service.
```
adrianjgeorge: `In some cases (User access Events(link to probably non-existent recipe)), ...`

repose-aggregator/tests/performance-tests/roles/gatling/files/simulations/filters/regexrbac/RegexRbacFilterSimulation.scala
```diff
@@ -0,0 +1,120 @@
...
+import filters.regexrbac.RegexRbacFilterSimulation._
+
+import scala.concurrent.duration._
+
+/**
+  * RegEx RBAC filter performance simulation.
+  */
+class RegexRbacFilterSimulation extends Simulation {
+  // properties to configure the Gatling test
+  val conf = ConfigFactory.load("application.conf")
```
adrianjgeorge: Is this boilerplate? If so we should totally create our own subclass of `Simulation` to handle this crap.

repose-aggregator/tests/performance-tests/roles/gatling/files/simulations/filters/regexrbac/RegexRbacFilterSimulation.scala
```diff
@@ -0,0 +1,120 @@
...
+      exec(getFooBarBazSpace)
+    }
+    .inject(
+      nothingFor(warmUpDuration minutes),  // do nothing during warm up period
+      constantUsersPerSec(rampUpUsers) during(rampUpDuration seconds))
+    .throttle(
+      jumpToRps(throughput), holdFor((warmUpDuration + duration) minutes))
+
+  // run the scenarios
+  setUp(
```
adrianjgeorge: I suspect this is duplicate as well. we could just force definition of `warmup` and `mainScenario`

repose-aggregator/components/filters/regex-rbac-filter/src/test/scala/org/openrepose/filters/regexrbac/RegexRbacFilterTest.scala
wdschei: Updated.

repose-aggregator/components/filters/regex-rbac-filter/src/test/scala/org/openrepose/filters/regexrbac/RegexRbacFilterTest.scala
wdschei: Updated.

repose-aggregator/components/filters/regex-rbac-filter/src/test/scala/org/openrepose/filters/regexrbac/RegexRbacSchemaTest.scala
wdschei: Not that I'm aware of. We start with the same basic format, but then have specific tests for each schema.

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/commons/config/manager/UpdateFailedException.java
wdschei: It is used here:
> https://github.com/rackerlabs/repose/blob/688b55855bd5b4f0eff574858629ca34510c03d8/repose-aggregator/components/filters/regex-rbac-filter/src/main/scala/org/openrepose/filters/regexrbac/RegexRbacFilter.scala#L166

repose-aggregator/src/docs/asciidoc/filters/regex-rbac.adoc
wdschei: Updated.

repose-aggregator/src/docs/asciidoc/filters/regex-rbac.adoc
wdschei: Updated.

repose-aggregator/tests/performance-tests/roles/gatling/files/simulations/filters/regexrbac/RegexRbacFilterSimulation.scala
wdschei: It sure kind of looks like it. I will defer to the original creators of the performance test suite, @Mario-Lopez and @dmnjohns, as to why or if this is needed.

repose-aggregator/tests/performance-tests/roles/gatling/files/simulations/filters/regexrbac/RegexRbacFilterSimulation.scala
wdschei: Same as above.

repose-aggregator/src/docs/asciidoc/filters/api-validator.adoc
wdschei: These were all pulled out to #1838.

repose-aggregator/components/filters/regex-rbac-filter/src/test/scala/org/openrepose/filters/regexrbac/RegexRbacSchemaTest.scala
dmnjohns: `ConfigurationTest` is the class you are looking for. That would be the "base".

repose-aggregator/tests/performance-tests/roles/gatling/files/simulations/filters/regexrbac/RegexRbacFilterSimulation.scala
dmnjohns: It is boilerplate. I mean, you can use these values to configure your simulation, but I would view them as a reasonable suggestion. I've lost enough familiarity with Gatling that I could not say, without looking into Gatling again, whether or not creating an inheritable `Simulation` is a good idea. In general, doing so increases code reuse but (potentially) reduces flexibility. Not sure which one matters more for these tests.

repose-aggregator/components/filters/regex-rbac-filter/src/test/scala/org/openrepose/filters/regexrbac/RegexRbacSchemaTest.scala
wdschei: Updated.

repose-aggregator/tests/performance-tests/roles/gatling/files/simulations/filters/regexrbac/RegexRbacFilterSimulation.scala
adrianjgeorge: I don't currently see us changing these values much, and a lot of this boiler plate is loading up settings from elsewhere, we would probably just tweak those settings in the elsewhere and this code wouldn't change

wdschei: This is outside the current scope and has been opened as:

* https://repose.atlassian.net/browse/REP-6336

repose-aggregator/tests/performance-tests/roles/gatling/files/simulations/filters/regexrbac/RegexRbacFilterSimulation.scala
wdschei: This is outside the current scope and has been opened as:

* https://repose.atlassian.net/browse/REP-6336

repose-aggregator/components/filters/regex-rbac-filter/src/test/scala/org/openrepose/filters/regexrbac/RegexRbacSchemaTest.scala
wdschei: A story was opened to convert the all the others:

* https://repose.atlassian.net/browse/REP-6337

PR #1841, opened 2017-11-10
REP-6338 Versioned Doc updates to normalize the way configuration default values are presented.

wdschei:


repose-aggregator/src/docs/asciidoc/filters/valkyrie-authorization.adoc
```diff
@@ -117,7 +117,7 @@ For more information about specific response codes that *Repose* will receive fr
 ====
 
 == Account Admins
-If a user has the `account_admin` role in Valkyrie, when `enable-bypass-account-admin` is configured to be `true` (the default is `false`), the Valkyrie filter will pass the request along regardless of whether or not the device permission check fails.
+If a user has the `account_admin` role in Valkyrie, when `enable-bypass-account-admin` is configured to be `true` (Default: `false`), the Valkyrie filter will pass the request along regardless of whether or not the device permission check fails.
```
dmnjohns: I wonder if we want to keep the default noted on this line. I noticed it did not follow the new line pattern that all the others do, and that is because it is an inline aside, and so it can't really. That made me question what could be done about it, and I landed on removing it since leaving it in means that we have multiple sources of truth for the default value of that attribute. 

I'm actually fine leaving it as is, but have a slight preference toward removing it.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
```diff
@@ -55,7 +55,7 @@ Let's break that command down and take a closer look at what it is doing:
 <1> This is the part of the command that tells Docker to create and start a container.
 <2> This option will run the container in detached mode.
     In detached mode, the container will run in the background.
-    If we were to run in attached mode (the default) instead of detached mode, the terminal we used to execute this command would receive all of the output (i.e., STDIN, STDOUT, and STDERR) from the container.
+    If we were to run in the default attached mode instead of optional detached mode, the terminal we used to execute this command would receive all of the container output (e.g. STDOUT, and STDERR).
```
dmnjohns: I find "optional detached mode" to be slightly awkward phrasing. Really, including "optional" is redundant, so I would probably just drop that word.

Oh, and I think `i.e.` fits better here than `e.g.` since the enumerated values are exhaustive. That is, I would replace:
> (e.g. STDOUT, and STDERR).

with:

> (i.e., `STDOUT` and `STDERR`).

You may notice that I would also drop the comma before the "and" and include a comma after "i.e." because I believe that is more correct and also because it is consistent with the rest of this document. [On correctness](https://www.grammarly.com/blog/know-your-latin-i-e-vs-e-g/).

Oh, and I formatted STDOUT and STDERR as code just because that is what the Docker folks do in their docs.

repose-aggregator/src/docs/asciidoc/recipes/quick-start.adoc
```diff
@@ -54,7 +54,7 @@ Let's break that command down and take a closer look at what it is doing:
 <1> This is the part of the command that tells Docker to create and start a container.
 <2> This option will run the container in detached mode.
     In detached mode, the container will run in the background.
-    If we were to run in attached mode (the default) instead of detached mode, the terminal we used to execute this command would receive all of the container output (e.g. STDOUT, and STDERR).
+    If we were to run in the default attached mode instead of optional detached mode, the terminal we used to execute this command would receive all of the container output (e.g. STDOUT, and STDERR).
```
dmnjohns: I would also drop the optional here.

Really, I wonder if there's a way to include this content rather than duplicating it.

repose-aggregator/src/docs/asciidoc/services/http-connection-pool.adoc
```diff
@@ -57,49 +57,49 @@ The id/name of the pool corresponds to the connection pool id found in other ser
...
+     Default: `true`
 <13> If a Keep-Alive header is not present in the response, the value of keepalive.timeout is evaluated.
      If this value is 0, the connection will be kept alive indefinitely.
      If the value is greater than 0, the connection will be kept alive for the number of milliseconds specified.
-     Set to 1 to connect:close.
-     Default is 0.
-<14> The keystore filename either fully qualified or relative to the *Repose* configuration directory (/etc/repose is the default).
+     Set to 1 to connect:close. +
+     Default: `0`
+<14> The keystore filename either fully qualified or relative to the *Repose* configuration directory (Default: `/etc/repose`).
```
dmnjohns: I think this could have be put on its own line before the notes?

repose-aggregator/src/docs/asciidoc/services/http-connection-pool.adoc
```diff
@@ -57,49 +57,49 @@ The id/name of the pool corresponds to the connection pool id found in other ser
...
      Used for SSL/TLS Client Authentication. +
      NOTE: The `keystore-filename`, `keystore-password`, and `key-password` are all mutually required. If one exists, then they all must exist and client authentication is assumed to be required.
 <15> The password to access the keystore. +
      NOTE: The `keystore-filename`, `keystore-password`, and `key-password` are all mutually required. If one exists, then they all must exist and client authentication is assumed to be required.
 <16> The password for the particular key within the keystore. +
      NOTE: The `keystore-filename`, `keystore-password`, and `key-password` are all mutually required. If one exists, then they all must exist and client authentication is assumed to be required.
-<17> The truststore filename either fully qualified or relative to the *Repose* configuration directory (/etc/repose is the default).
+<17> The truststore filename either fully qualified or relative to the *Repose* configuration directory (Default: `/etc/repose`).
```
dmnjohns: This could also probably be on its own line.

repose-aggregator/src/docs/asciidoc/services/http-connection-pool.adoc
wdschei: These were intentionally left inline since their value is not the default of the field being defined, rather the default of the item referenced by the field being defined.

repose-aggregator/src/docs/asciidoc/services/http-connection-pool.adoc
wdschei: These were intentionally left inline since their value is not the default of the field being defined, rather the default of the item referenced by the field being defined.

repose-aggregator/src/docs/asciidoc/services/http-connection-pool.adoc
dmnjohns: Ah, I see. That's what I get for glancing.

In that case, this is similar to my other comment where I would considering removing it. Once again, this adds another source of truth, and it isn't strictly necessary to define the keystore filename attribute. Perhaps the best approach would be to drop the default, but link to the relevant portion of the system model document? Perhaps put the link on the words "configuration directory"?

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
Mario-Lopez: I also agree with using `i.e.` over `e.g.` and reluctantly acknowledge our standardization on using the following comma (i.e. using `i.e.,` 😉  ).

repose-aggregator/src/docs/asciidoc/recipes/quick-start.adoc
wdschei: I had looked at it, but it is only one callout. If anything, I would just remove the callout from the Docker recipe since we state the assumption is that the reader has already gone through the Quick Start.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
wdschei: Updated.

wdschei: This one slipped passed me. I updated this in a couple of other places.

repose-aggregator/src/docs/asciidoc/filters/valkyrie-authorization.adoc
wdschei: Removed since it is stated as the `Default` in the callout a few lines later.

repose-aggregator/src/docs/asciidoc/recipes/docker.adoc
wdschei: Worse than that, I changed this one to be wrong.

repose-aggregator/src/docs/asciidoc/services/http-connection-pool.adoc
wdschei: Linked to the Configuration doc.

repose-aggregator/src/docs/asciidoc/services/http-connection-pool.adoc
wdschei: Linked to the Configuration doc.

PR #1842, opened 2017-11-10
REP-6313 Adding ignoring specific roles to keystone v2

adrianjgeorge:


repose-aggregator/components/filters/keystone-v2-filter/src/main/resources/META-INF/schema/config/keystone-v2.xsd
```diff
@@ -49,6 +49,13 @@
             <xs:element name="require-service-endpoint" type="ServiceEndpointType" minOccurs="0" maxOccurs="1"/>
             <xs:element name="pre-authorized-roles" type="RolesList" minOccurs="0" maxOccurs="1"/>
         </xs:all>
+        <xs:attribute name="ignored-roles" type="xs:string" default="identity:tenant-access" use="optional">
+            <xs:annotation>
+                <xs:documentation>
+                    <html:p> A space separated list of roles that will be ignored when process the response from the identity provider.</html:p>
```
wdschei: This should be _"when process*ing* the response"_.

wdschei:
Config updates look good.

adrianjgeorge:
That test failure is weird, because i haven't actually changed any code yet.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneRequestHandler.scala
```diff
@@ -108,6 +108,7 @@ class KeystoneRequestHandler(identityServiceUri: String, akkaServiceClient: Akka
         val userId = (json \ "access" \ "user" \ "id").as[String]
         val roles = (json \ "access" \ "user" \ "roles").as[JsArray].value
           .map(jsRole => Role((jsRole \ "name").as[String], (jsRole \ "tenantId").asOpt[String]))
+          .filter(role => !ignoredRoles.contains(role.name))
```
dmnjohns: You might want to use `filterNot` here? I think it makes it clearer, but it's not like what you have is unclear at all.

`.filterNot(role => ignoredRoles.contains(role.name))`

repose-aggregator/components/filters/keystone-v2-filter/src/main/resources/META-INF/schema/config/keystone-v2.xsd
```diff
@@ -49,6 +49,13 @@
             <xs:element name="require-service-endpoint" type="ServiceEndpointType" minOccurs="0" maxOccurs="1"/>
             <xs:element name="pre-authorized-roles" type="RolesList" minOccurs="0" maxOccurs="1"/>
         </xs:all>
+        <xs:attribute name="ignored-roles" type="xs:string" default="identity:tenant-access" use="optional">
```
dmnjohns: Out of curiosity, since this is a space-separated list, why not use the XML Schema `list` type?

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -679,6 +679,8 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
       akkaServiceClient = akkaServiceClientFactory.newAkkaServiceClient(keystoneV2Config.getIdentityService.getConnectionPoolId)
       akkaServiceClientOld.foreach(_.destroy())
 
+      ignoredRoles = keystoneV2Config.getIgnoredRoles.split(' ').to[Set]
```
dmnjohns: This ties into my comment about using the XSD `list` type. I believe that type collapses whitespace when parsing, whereas this code does not. Thus, a user could end up with empty strings as ignored roles if a user was expecting the XSD `list` semantics.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneRequestHandler.scala
adrianjgeorge: I thought about it, but i find `filterNot` frankly more confusing than just not-ing my condition when it isn't what i want. Those methods always do the opposite of what i think they'll do.

repose-aggregator/components/filters/keystone-v2-filter/src/main/resources/META-INF/schema/config/keystone-v2.xsd
adrianjgeorge: mostly because that behaves super weirdly with defaults.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneRequestHandler.scala
dmnjohns: Fair enough.

repose-aggregator/components/filters/keystone-v2-filter/src/main/resources/META-INF/schema/config/keystone-v2.xsd
dmnjohns: Oh, yeah, I forgot about that. Freakin' JAXB.

PR #1844, opened 2017-11-14
REP-5616 Upgrading Jetty to the latest v9.2.x.

wdschei:


repose-aggregator/components/filters/cors-filter/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
```diff
@@ -229,14 +229,13 @@ class CorsFilter @Inject()(configurationService: ConfigurationService)
...
+      .setHost(normalizeHost(originUri.getHost))
       .build())
 
-  def normalizeHost(host: String): String =
-    if (InetAddresses.isInetAddress(host)) InetAddresses.forString(host).getHostAddress else host
-
-  def normalizeUriHost(host: String): String =
-    if (InetAddresses.isUriInetAddress(host)) InetAddresses.forUriString(host).getHostAddress else host
+  def normalizeHost(host: String): String = {
+    val hostReplaced = host.replaceFirst("""^\[?([0-9A-Fa-f:]+)\]?$""", "$1")
```
dmnjohns: I like the idea of consolidating the normalization methods, but I don't love using a regex replacement to remove the brackets so that you can achieve that. In this case, the regex actually changes the behavior. Consider the origin header value `[2001:db8:cafe::17`. This code would normalize and return it fine, but the old code would have returned the non-normalized value. I'm not sure whether or not that is acceptable behavior, but I am sure that we should be aware of the difference, which is made somewhat difficult since it requires grokking a regex. 

Also, this will replace the first byte of a IPv4 address with itself. No harm there, but it isn't really necessary, either.

Perhaps instead you could use the same logic as the `InetAddresses` class. It is less compact, but I think it is simpler and more precise:
```
if (hostAddr.startsWith("[") && hostAddr.endsWith("]")) {
      ipString = hostAddr.substring(1, hostAddr.length() - 1);
}
```

So here it might be:
```
val hostReplaced = if (hostAddr.startsWith("[") && hostAddr.endsWith("]")) hostAddr.subString(1, hostAddr.length - 1) else hostAddr
```

Although, even that is not quite the same behavior as before, since now `[255.255.255.255]` would be normalized and returned, whereas before the non-normalized value would have been returned.

So maybe just a straight combination:
```
def normalizeHost(host: String): String = {
  if (InetAddresses.isUriInetAddress(host)) {
    InetAddresses.toAddrString(InetAddresses.forUriString(host))
  } else if (InetAddresses.isInetAddress(host)) {
    InetAddresses.toAddrString(InetAddresses.forString(host))
  } else {
    host
  }
}
```

Of course, that has the effect of allowing brackets wherever `normalizeHost` is called, which were not allowed before... but that will be the case case if do any method combination here. So I think we should either take that last approach, or leave the two methods and just replace `getHostAddress` with `toAddrString`.

repose-aggregator/components/filters/cors-filter/src/test/scala/org/openrepose/filters/cors/CorsFilterTest.scala
```diff
@@ -708,7 +737,9 @@ class CorsFilterTest extends FunSpec with BeforeAndAfterEach with Matchers with
         servletRequest.setScheme("http")
         servletRequest.setServerName(null)
         servletRequest.setServerPort(-1)
-        forwardedHostHeaders foreach { servletRequest.addHeader(CommonHttpHeader.X_FORWARDED_HOST, _) }
+        forwardedHostHeaders foreach {
+          servletRequest.addHeader(CommonHttpHeader.X_FORWARDED_HOST, _)
+        }
```
dmnjohns: If you wanted, you could keep this on one line like:
```
forwardedHostHeaders.foreach(servletRequest.addHeader(CommonHttpHeader.X_FORWARDED_HOST, _))
```

repose-aggregator/components/filters/cors-filter/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
wdschei: Sure. Updating that logic for removal of the leading and trailing brackets is easy enough. The reals issue is that as you pointed out the behavior is different. The old behavior was incorrect. In turn it would fail and simply pass the original `String` back as you have stated. The desired behavior is to extract the host, normalize it, and return the normalized value.

repose-aggregator/components/filters/cors-filter/src/test/scala/org/openrepose/filters/cors/CorsFilterTest.scala
wdschei: Updated.

PR #1845, opened 2017-11-14
REP-5616 Upgrading Jetty to the latest version.

wdschei:


repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
```diff
@@ -126,37 +126,63 @@ class ReposeJettyServer(val nodeContext: AbstractApplicationContext,
...
+        // https://github.com/eclipse/jetty.project/commit/0a1b0b2bc69ea7e7f5f44992f47a84f926cdeebb
+        cf.addExcludeCipherSuites(
+          List(
+            "SSL_RSA_WITH_DES_CBC_SHA",
+            "SSL_DHE_RSA_WITH_DES_CBC_SHA",
+            "SSL_DHE_DSS_WITH_DES_CBC_SHA",
+            "SSL_RSA_EXPORT_WITH_RC4_40_MD5",
+            "SSL_RSA_EXPORT_WITH_DES40_CBC_SHA",
+            "SSL_DHE_RSA_EXPORT_WITH_DES40_CBC_SHA",
+            "SSL_DHE_DSS_EXPORT_WITH_DES40_CBC_SHA"): _*)
```
dmnjohns: Were these defaults present in Jetty 9.2.0 (currently in-use by the `master` branch), or did they come in on of the other upgrades? If the latter, do we actually need to add them back in?

dmnjohns: Looking at the Jetty 9.2.0 source I've pulled down locally, I do not see the defaults referenced in your link. I wonder if that's why we were using `add` instead of `set`? They were functionally equivalent in Jetty 9.2.0?

repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
```diff
@@ -126,37 +126,63 @@ class ReposeJettyServer(val nodeContext: AbstractApplicationContext,
...
         Option(ssl.getIncludedProtocols).foreach { ip =>
           if (ip.getProtocol.nonEmpty) {
             cf.setIncludeProtocols(ip.getProtocol.toList: _*)
           }
-
         }
-        Option(ssl.getExcludedCiphers).foreach { xc =>
-          cf.addExcludeCipherSuites(xc.getCipher.toList: _*)
+        Option(ssl.getExcludedProtocols).foreach { ep =>
+          if (ep.getProtocol.nonEmpty) {
```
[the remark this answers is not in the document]
dmnjohns: So to sum up this comment and my others, I think we should:

1. Always set the includes/excludes (not add).
    1. This should be backwards compatible, and thus, just a small implementation change.
1. Allow empty sequences for includes/excludes (that the user has to explicitly declare to clear the defaults).
    1. This should be a new feature, since previously the user was not allowed to declare an empty sequence.

repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
wdschei: This is the commit they changed in:
* https://github.com/eclipse/jetty.project/commit/0a1b0b2bc69ea7e7f5f44992f47a84f926cdeebb

repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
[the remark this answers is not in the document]
wdschei: 1. Updated.
1. Current behavior on `included`'s is to overwrite (`set`), but current behavior on `excluded`'s is to append (`add`) to any defaults. This is what actually caused the issue with the `ReposeJettySSLTest` and led me down this path. So for backwards compatibility, `excluded`'s must remain `add`'s.
1. Any user defined `excluded-ciphers` are still being added. The difference here is to override the new Jetty defaults with the old ones. That is why the `TODO`'s to change these behaviors in Repose v9.x.

repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
dmnjohns: Sure, I saw that in your comment. So it looks like the oldest tag that commit is in is a Jetty 9.3.7 tag. Which means the behavior from Jetty 9.2.0 (our current version) to Jetty 9.3.7 changed. To be more exact, Jetty 9.3.0 seems to have introduced the default excluded list, and Jetty 9.3.7 modified it. However, I believe we want to maintain current behavior (from Jetty 9.2.0) which means no default exclusions.

repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
```diff
@@ -126,37 +126,51 @@ class ReposeJettyServer(val nodeContext: AbstractApplicationContext,
...
-        }
-        Option(ssl.getExcludedCiphers).foreach { xc =>
-          cf.addExcludeCipherSuites(xc.getCipher.toList: _*)
-        }
-
-        Option(ssl.getIncludedCiphers).foreach { ic =>
-          if (ic.getCipher.nonEmpty) {
-            cf.setIncludeCipherSuites(ic.getCipher.toList: _*)
+        Option(ssl.getIncludedProtocols) foreach { ip => cf.setIncludeProtocols(ip.getProtocol.toList: _*) }
+        Option(ssl.getExcludedProtocols) foreach { ep => cf.addExcludeProtocols(ep.getProtocol.toList: _*) }
```
dmnjohns: I think we can set this rather than add, right? In 9.2.0, there are no default includes or excludes as far as I can tell, so it should be safe to do the following:
```
cf.setIncludeProtocols()
cf.addExcludeProtocols()
cf.setIncludeCipherSuites()
cf.setExcludeCipherSuites()
Option(ssl.getIncludedProtocols) foreach { ip => cf.setIncludeProtocols(ip.getProtocol.toList: _*) }
Option(ssl.getIncludedProtocols) foreach { ip => cf.setExcludeProtocols(ep.getProtocol.toList: _*) }
Option(ssl.getIncludedCiphers) foreach { ic => cf.setIncludeCipherSuites(ic.getCipher.toList: _*) }
Option(ssl. getExcludedCiphers) foreach { ic => cf.setExcludeCipherSuites(ec.getCipher.toList: _*) }
```

That gives us 9.2.0 behavior in 9.3.0 and onward.

In the future, we may want to consider using the Jetty defaults by default or even making our own defaults if we want to decouple with Jetty.

repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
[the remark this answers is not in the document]
dmnjohns: In response to 2, I think your issue was with the defaults being updated. If you wipe the defaults and then set excludes, you should be okay. And from what I can tell, you only really need to wipe the cipher exclusion defaults, but I would wipe them all for consistency (both with Jetty 9.2.0 and with each other).

repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
```diff
@@ -126,37 +126,27 @@ class ReposeJettyServer(val nodeContext: AbstractApplicationContext,
...
-
-        Option(ssl.getIncludedCiphers).foreach { ic =>
-          if (ic.getCipher.nonEmpty) {
-            cf.setIncludeCipherSuites(ic.getCipher.toList: _*)
-          }
-        }
+        Option(ssl.getIncludedProtocols) foreach { ip => cf.setIncludeProtocols(ip.getProtocol.toList: _*) }
+        Option(ssl.getExcludedProtocols) foreach { ep => cf.setExcludeProtocols(ep.getProtocol.toList: _*) }
+        Option(ssl.getIncludedCiphers) foreach { ic => cf.setIncludeCipherSuites(ic.getCipher.toList: _*) }
+        Option(ssl.getExcludedCiphers).foreach { ec => cf.setExcludeCipherSuites(ec.getCipher.toList: _*) }
```
dmnjohns: Won't you still need to clear the default excluded ciphers to maintain the 9.2.0 behavior? It only comes into play when the user does not configure any excluded ciphers, but still.

dmnjohns: Also, I would probably just clear all of the includes/excludes for consistency.

Then in Repose 9.0 we can decide what we want the default includes/excludes to be.

wdschei: Updated.

repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
wdschei: Added.

repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
wdschei: Removed.

repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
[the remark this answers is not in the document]
wdschei: Restructured.

repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
```diff
@@ -126,37 +126,32 @@ class ReposeJettyServer(val nodeContext: AbstractApplicationContext,
...
-        }
+        // Clear Jetty defaults that were added in later v9.2.x releases.
+        cf.setIncludeProtocols()
+        cf.setExcludeProtocols()
+        cf.setIncludeCipherSuites()
+        cf.setExcludeCipherSuites()
+        Option(ssl.getIncludedProtocols) foreach { ip => cf.setIncludeProtocols(ip.getProtocol.toList: _*) }
+        Option(ssl.getExcludedProtocols) foreach { ep => cf.setExcludeProtocols(ep.getProtocol.toList: _*) }
+        Option(ssl.getIncludedCiphers) foreach { ic => cf.setIncludeCipherSuites(ic.getCipher.toList: _*) }
+        Option(ssl.getExcludedCiphers).foreach { ec => cf.setExcludeCipherSuites(ec.getCipher.toList: _*) }
```
dmnjohns: Woo! I like it.

So two things to track from this:

1. We should write a story to allow empty include/exclude sequences in the container configuration.
1. For Repose 9, we should decide if we want to change the default include/exclude protocols/ciphers. That may mean defaulting to the same defaults as Jetty, defaulting to whatever is supported by the environment, or defaulting to our own set of defaults.

wdschei: Concur.

wdschei:
Now to figure out what the 41'ish functional integration test failures are all about.

repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
dmnjohns: Wait, why is there a rando `.foreach` on the last set here?

dmnjohns: Oh, also, I think you could use the following if it makes things clearer:
```
Option(ssl.getExcludedCiphers).map(_.getCipher.toList).foreach(cf.setExcludeCipherSuites)
```

repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/ReposeJettyServer.scala
```diff
@@ -126,37 +126,32 @@ class ReposeJettyServer(val nodeContext: AbstractApplicationContext,
...
-        }
+        // Clear Jetty defaults that were added in later v9.2.x releases.
+        cf.setIncludeProtocols()
+        cf.setExcludeProtocols()
+        cf.setIncludeCipherSuites()
+        cf.setExcludeCipherSuites()
+        Option(ssl.getIncludedProtocols) foreach { ip => cf.setIncludeProtocols(ip.getProtocol.toList: _*) }
+        Option(ssl.getExcludedProtocols) foreach { ep => cf.setExcludeProtocols(ep.getProtocol.toList: _*) }
+        Option(ssl.getIncludedCiphers) foreach { ic => cf.setIncludeCipherSuites(ic.getCipher.toList: _*) }
+        Option(ssl.getExcludedCiphers) foreach { ec => cf.setExcludeCipherSuites(ec.getCipher.toList: _*) }
```
dmnjohns: I wrote this comment at a bad time so it may have been hidden. I'll repeat it to make sure it is seen.

If it make it clearer, I believe you could replace these lines with lines like:
```
Option(ssl.getExcludedCiphers).map(_.getCipher.toList).foreach(cf.setExcludeCipherSuites)
```

Or to be even more explicit:
```
Option(ssl.getExcludedCiphers).map(_.getCipher).map(_.toList).foreach(cf.setExcludeCipherSuites)
```

wdschei: Updated to the more clear/less explicit version.

wdschei: Ok. That caused time mismatches. I reverted for now.

dmnjohns: Ah, I was afraid of that. Darn. I don't know if there's a good way to handle the typing weirdness for varargs. Oh well.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
```diff
@@ -129,13 +130,13 @@ class HttpServletRequestWrapper(originalRequest: HttpServletRequest, val inputSt
...
     getHeadersScala(headerName).foldLeft(List.empty[String])((list, s) => list ++ s.split(","))
-      .map(_.trim)
+      .map(_.trim).filter(StringUtils.isNotBlank(_))
 
   def getHeadersScala(headerName: String): List[String] = {
     if (removedHeaders.contains(headerName)) {
       List.empty[String]
     } else {
-      headerMap.getOrElse(headerName, super.getHeaders(headerName).asScala.toList)
+      headerMap.getOrElse(headerName, super.getHeaders(headerName).asScala.toList).filter(StringUtils.isNotBlank(_))
```
dmnjohns: I believe you should be able to replace:
`.filter(StringUtils.isNotBlank(_))`
with:
`.filter(StringUtils.isNotBlank)`

Which cuts out a few unnecessary characters.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/connectionframework/ConnectionFrameworkTest.groovy
```diff
@@ -62,7 +63,7 @@ class ConnectionFrameworkTest extends ReposeValveTest {
         messageChain = deproxy.makeRequest(url: reposeEndpoint + "/", headers: headers, addDefaultHeaders: false)
 
         then:
-        messageChain.handlings[0].request.headers.getFirstValue("accept") == null
+        StringUtils.isBlank(messageChain.handlings[0].request.headers.getFirstValue("accept"))
```
dmnjohns: Nothin' wrong here, just calling attention to it since it demonstrates the Jetty behavior change.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/proxy/RequestQueryParamTest.groovy
```diff
@@ -78,48 +78,49 @@ class RequestQueryParamTest extends ReposeValveTest {
...
         messageChain.receivedResponse.code == "400"
 
         where:
-        uriSuffixGiven                      | method
-        "/path/to/resource?key=value value" | "GET"
+        uriSuffixGiven << [
+            "/path/to/resource?key=value value",
+            "/path/to/resource?key=value@%",
+            "/path/to/resource?key=va%lu@e"
+        ]
```
dmnjohns: ... and here's the other behavior change.

wdschei: Yep. Between Jetty v9.3.17.v20170317/v9.4.3.v20170317 and v9.3.18.v20170406/v9.4.4.v20170414 the handling of malformed query parameters was changed. Now if it isn't legal, then it doesn't make it in the front door anymore.

 - https://github.com/eclipse/jetty.project/compare/jetty-9.3.17.v20170317...jetty-9.3.18.v20170406
 - https://github.com/eclipse/jetty.project/commit/dff8fb6b90952a92e308d7687bfaf597393876f4

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/connectionframework/ConnectionFrameworkTest.groovy
wdschei: Yep. Jetty v9.2.x supports HTTP/1.1 ([RFC 2616](https://tools.ietf.org/html/rfc2616)), javax.websocket, and SPDY v3. Jetty v9.3.x supports HTTP/1.1 ([RFC 7230](https://tools.ietf.org/html/rfc7230)), HTTP/2 (RFC 7540), WebSocket (RFC 6455, JSR 356), and FastCGI. One of the differences between HTTP/1.1 as defined by RFC 2616 and RFC 7230 is the way headers are defined and handled.

Specifically empty header values are now returned instead of a null. That is there is no longer a differentiation between an empty header and a header with an empty string as the value.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/powerfilter/RequestSizeTest.groovy
```diff
@@ -43,7 +43,7 @@ class RequestSizeTest extends ReposeValveTest {
         deproxy.addEndpoint(properties.targetPort)
     }
 
-    @Unroll("request with header size of #headerSize should respond with 413")
+    @Unroll("request with header size of #headerSize should respond with 431")
```
wdschei: This is the last behavior change. [RFC 2616](https://tools.ietf.org/html/rfc2616) was also updated by [RFC 6585](https://tools.ietf.org/html/rfc6585) to add additional HTTP Status Codes. The one in particular that is now in use by Jetty v9.3.x+ is `431` which indicates that the server is unwilling to process the request because its header fields are too large.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletRequestWrapper.scala
wdschei: Updated.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/httpconnectionpool/HttpClientStateTest.groovy
```diff
@@ -139,13 +136,12 @@ class HttpClientStateTest extends ReposeValveTest {
             def value = session.getAttribute("KEY")
             println("Session ID: ${session.getId()} HEADER COUNT: ${req.getHeader("x-count-thingy")} KEY:${value} cookies: ${cookies}")
             session.setAttribute("KEY", "VALUE")
-            def cookie = new Cookie("TESTCOOKIE", "TEH VALUE")
+            def cookie = new Cookie("TESTCOOKIE", "THE-VALUE")
```
wdschei: Between jetty [v9.4.2.v20170220 and v9.4.3.v20170317](https://github.com/eclipse/jetty.project/compare/jetty-9.4.2.v20170220...jetty-9.4.3.v20170317) the Cookie value syntax check was updated. The Cookie values are now properly checked against [RFC6265](https://tools.ietf.org/html/rfc6265#section-4.1.1) and may not contain the US-ASCII space character (0x20) which the failing test used.
 - https://github.com/eclipse/jetty.project/commit/528c1f7745fbd20de06d91263c794cd4035c8991

wdschei:
With all the rebase'ing and that last update, this should all be working.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/proxy/RequestQueryParamTest.groovy
dmnjohns: Also, if Jetty is validating the query string now, can we remove the following line?
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java#L401

wdschei: Is that line only intended to vet the query parameters?

dmnjohns: It is not, which is a fair point.

... but I wonder if Jetty is stricter about URIs in general now. If Jetty validates the entire URI, then this line would no longer be necessary. Should be easy enough to remove this line and run the tests to see what happens, if we are interesting. That can even be done by Jenkins on a separate branch.

wdschei:
Retest this please.

PR #1846, opened 2017-11-15
Removing unreachable statement

dmnjohns:
This was a warning in the Gradle build, so fixing it removes the warning.


adrianjgeorge:
That was a weird one, sure it didn't have coverage. Was it maybe meant to be scoped at a different level?

dmnjohns:
Judging by how similar it is to the previous case, I don't think so... We also handle the default case in all cases, so I'm not too worried about it, either.

PR #1848, opened 2017-11-29
Updated the CodeNarc configuration to use a specific version.

wdschei:


adrianjgeorge:
i'm going to let it run to make sure it works.

PR #1849, opened 2017-11-29
Added the OWASP Dependency Check plugin to the subprojects.

wdschei:


wdschei:
I use the script below that I put in my `<REPOSE>/local` directory to gather up all the individual result files so they can be easily accessed:

```bash
#!/usr/bin/env bash
SCRIPT_DIR=$( cd "$( dirname "$0" )" && pwd )
REPOSE_DIR=${SCRIPT_DIR}/..
REPORTS_DIR=${REPOSE_DIR}/build/reports
cd ${REPOSE_DIR}
mkdir -p ${REPORTS_DIR}

# Add the header
cat > ${REPORTS_DIR}/dependency-check-report-index.html << EOM
<!DOCTYPE html>
<html>
    <head>
        <title>Dependency-Check Reports Index</title>
        <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
    </head>
    <body>
      <h2>Dependency-Check Reports Index</h2>
      <ul>
EOM

REPORT_FILES=( $(find . -name dependency-check-report.html | sort) )
for file in ${REPORT_FILES[*]} ; do
    echo "          <li><a href=\"../../$file\">$file</a></li>" >> ${REPORTS_DIR}/dependency-check-report-index.html
done

cat >> ${REPORTS_DIR}/dependency-check-report-index.html << EOM
      </ul>
    </body>
</html>
EOM
```

wdschei:
Here is the plugins homepage:
> https://jeremylong.github.io/DependencyCheck/dependency-check-gradle/index.html

And the task is:
> `gradle dependencyCheckAnalyze`

wdschei:
Another way to get a good list of all the findings is to use:

`./gradlew dependencyCheckAnalyze | grep 'CVE-20' | sort -u`

adrianjgeorge:
retest this please

versions.properties
```diff
@@ -1,12 +1,12 @@
-jacksonVersion=2.8.9
+jacksonVersion=2.9.8
```
dmnjohns: This is the cause of the current test failures when api-checker is being used. api-checker has a dependency on Jackson that requires Jackson >= 2.8.0 and < 2.9.0.

Updating api-checker to use the newer version of Jackson right now would put a blocker on this work, and I'm not certain how much effort it would  take to do so. It seems like we may want to upgrade to `2.8.10` and stay on that version for the time being. Of course, there  are known security vulnerabilities with that version, so I'll look into those and try to figure out how they affect us.

dmnjohns: Of course, we could also upgrade the version of Jackson we use directly and let the `extensions-filter-bundle` continue using the older working version for the time being.

wdschei: I like isolating our exposure as much as possible if we can't eliminate it completely.
Then we add a story to figure out how to move it in the external dependencies.

wdschei: I like isolating our exposure as much as possible if we can't eliminate it completely.
Then we add a story to figure out how to move it in the external dependencies.

dmnjohns: Gah. Unfortunately, that might not actually work. I can force the dependency back to the version that works with api-checker which will put the expected version of the Jackson JAR in the  `extensions-filter-bundle` EAR, however we still encounter a failure at runtime. 

My theory is that the thread classloader is being used rather than the EAR classloader. This would be because we do not load the Jackson class until we validate a request.

There should be a way to force class loading to occur at the appropriate time, but such a solution quickly pushes us into the realm where complexity outweighs the benefit.

dmnjohns:
Dang it. I messed up my push which deleted this branch and closed this PR. See #2029 for the new PR.

PR #1853, opened 2017-12-14
REP-5401 Environment Variable Substitution in Configuration Files

dmnjohns:


dmnjohns:
Down to a single test failure after that last commit. It looks like there is a classpath collision for `org.objectweb.asm.tree.ClassNode`. JTwig has a dependency on Parboiled which has an ASM dependency. GlassFish also has an ASM dependency. The versions clearly do not match, and so we see something like:
> Caused by: java.lang.NoSuchMethodError: org.objectweb.asm.tree.ClassNode.<init>(I)V

repose-aggregator/src/docs/asciidoc/architecture/configuration.adoc
```diff
@@ -25,3 +25,78 @@ The configuration files are checked for updates every 15 seconds.
...
+
+[IMPORTANT]
+====
+When substituting an environment variable, *Repose* strictly requires that the named environment variable exists.
+If not, *Repose* will throw an exception when processing the containing configuration file.
+====
+
+[IMPORTANT]
+====
+Updating an environment variable will *not* cause Repose to reload any configuration which references that environment variable.
```
adrianjgeorge: something we should potentially look at fixing in the future.

dmnjohns: Sure, I think we can do that. It should just be a matter of polling environment variables and checking if the map has changed.

adrianjgeorge:
Missed a license somewhere

dmnjohns:
Bleh. Sure enough. It's the `glassfish-web.xml` file, which is a XML test resource, which is probably why I didn't catch it. Now I am wondering why that did not seem to be required when this file was in the `artifacts:web-application` project. Especially since those files actually make it into artifacts, while this file does not (where it lives now).

adrianjgeorge:
i think we have an exclusion setup in the other module

adrianjgeorge:
it appears you broke some other stuff

dmnjohns:
retest this please

dmnjohns:
Great. So one flaky test with fixed, the other two failed once and are now passing, and now a fourth has popped up.

Not cool, tests. Not cool.

dmnjohns:
retest this please

dmnjohns:
retest this please

repose-aggregator/src/docs/asciidoc/architecture/configuration.adoc
adrianjgeorge: Well this turns out to not be a thing

dmnjohns: Whoops! But we still want to leave that warning in place, right?

PR #1854, opened 2017-12-15
REP-6390 Cutting off common functionality in preparation for splitting Keystone v2 Filter

adrianjgeorge:


adrianjgeorge:
retest this please

Mario-Lopez:
retest this please

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  def handleFailures(authResult: Try[Unit.type]): Option[Reject] = {
+    authResult match {
+      case Failure(e: InvalidTenantException) => Option(Reject(SC_UNAUTHORIZED, Some(e.getMessage)))
+      case Failure(e: UnauthorizedEndpointException) => Option(Reject(SC_FORBIDDEN, Some(e.getMessage)))
+      case Failure(e: UnparseableTenantException) => Option(Reject(SC_UNAUTHORIZED, Some(e.getMessage)))
+      case _ => None
+    }
+  }
```
dmnjohns: Rather than making this a method, you could make this a function. It saves you the parameter and the match statement.

```
val handleFailures: (Try[Unit.type] => Option[Reject]) = {
  case Failure(e: InvalidTenantException) => Option(Reject(SC_UNAUTHORIZED, Some(e.getMessage)))
  case Failure(e: UnauthorizedEndpointException) => Option(Reject(SC_FORBIDDEN, Some(e.getMessage)))
  case Failure(e: UnparseableTenantException) => Option(Reject(SC_UNAUTHORIZED, Some(e.getMessage)))
  case _ => None
}
```

If you'd like to be a bit clearer than that, you could also give the function a type:
```
type FailureHandler = (Try[Unit.type] => Option[Reject])
val failureHandler: FailureHandler = {
  case Failure(e: InvalidTenantException) => Option(Reject(SC_UNAUTHORIZED, Some(e.getMessage)))
  case Failure(e: UnauthorizedEndpointException) => Option(Reject(SC_FORBIDDEN, Some(e.getMessage)))
  case Failure(e: UnparseableTenantException) => Option(Reject(SC_UNAUTHORIZED, Some(e.getMessage)))
  case _ => None
  }
```

It doesn't make much of a difference, but I think it's kind of neat.

repose-aggregator/commons/commons-test/src/main/scala/org/openrepose/commons/test/ConfigValidator.scala
```diff
@@ -56,7 +56,7 @@ object ConfigValidator {
       classOf[ConfigValidator].getClassLoader)
     factory.setFeature("http://apache.org/xml/features/validation/cta-full-xpath-checking", true)
     new ConfigValidator(factory
-      .newSchema(schemas.map(_.openStream).map(new StreamSource(_)).toArray[Source])
+      .newSchema(schemas.map(u => new StreamSource(u.toString)).toArray[Source])
```
dmnjohns: I suppose it's somewhat of preference thing, but I prefer the way this was being done where each transformation is in its own `map` call. So:
```
.newSchema(schemas.map(_.toString).map(new StreamSource(_))).toArray[Source])
```

I think that ends up being one character longer, but for whatever reason, it seems clearer and cleaner to me. Not that it is hard to understand as-is, by any means.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -55,518 +54,373 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
...
+    token flatMap { authToken =>
+      validateToken(authToken) flatMap { validToken =>
+        val authResult = KeystoneV2Authorization.doAuthorization(config, request, validToken, endpoints)
+
+        addTokenHeaders(authResult.scopedToken, authResult.matchedTenant)
+        authResult match {
+          case AuthorizationPassed(scopedToken, _) =>
+            addCatalogHeader(endpoints)
+            addGroupsHeader(getGroups(token.get, scopedToken))
+          case AuthorizationFailed(_, _, exception) => Failure(exception)
```
dmnjohns: I think you can nest the endpoints a bit more, simplifying this block a small amount:

```
    getAuthToken flatMap { authToken =>
      validateToken(authToken) flatMap { validToken =>
        lazy val endpoints = getEndpoints(authToken) // Prevents making call if its not needed
        val authResult = KeystoneV2Authorization.doAuthorization(config, request, validToken, endpoints)

        addTokenHeaders(authResult.scopedToken, authResult.matchedTenant)
        authResult match {
          case AuthorizationPassed(scopedToken, _) =>
            addCatalogHeader(endpoints)
            addGroupsHeader(getGroups(token.get, scopedToken))
          case AuthorizationFailed(_, _, exception) => Failure(exception)
        }
      }
    }
```

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -55,518 +54,373 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
...
+    response.commitToResponse()
+  }
 
-              case None =>
-                logger.debug(s"Rejecting with status $statusCode")
+  override def doAuth(request: HttpServletRequestWrapper): Try[Unit.type] = {
+    /**
+      * STATIC REFERENCE TO CONFIG
+      */
+    val config = configuration
```
dmnjohns: Hm, keeping this line here kind of breaks the idea of the static reference. The current configuration object is already captured in the `doWork` method of the abstract filter. Capturing it again here using the global variable means that it could have changed between when it was captured in `doWork` and here. I suppose the thing to do here, if we do want to make sure that the configuration doesn't change under us, is to pass the config reference from `doWork` into `doAuth`.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/AbstractKeystoneV2Filter.scala
```diff
@@ -0,0 +1,147 @@
...
+      */
+    val config = configuration
+
+    /**
+      * DECLARE COMMON VALUES
+      */
+    lazy val request = new HttpServletRequestWrapper(servletRequest)
+    lazy val response = new HttpServletResponseWrapper(servletResponse, MUTABLE, PASSTHROUGH)
+
+    def isWhitelisted(requestUri: String): Boolean = {
```
dmnjohns: You know, I'm not really sure why I nested these methods in the first place. I think I've come to regret that. Not that it's in scope of this story.

However, if we are going to leave these nested, I personally prefer them to be placed below the actual work being done by the containing method. To me, it makes it easier to see what this method is meant to be doing. Thought I admit it is strange to invoke a method before it has been defined.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
dmnjohns: Actually, the return type should probably be:
```
type FailureHandler = PartialFuction[Try[Unit.type], Option[Reject]]
```

That way you can combine partial functions to handle failure handling the way you are doing now, as a chain of sorts.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -55,518 +54,373 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
...
+      authResult match {
+        case Failure(e: MissingAuthTokenException) => Option(Reject(SC_UNAUTHORIZED, Some(e.getMessage)))
+        case Failure(e: NotFoundException) => Option(Reject(SC_UNAUTHORIZED, Some(e.getMessage)))
+        case Failure(e: UnparseableTenantException) => Option(Reject(SC_UNAUTHORIZED, Some(e.getMessage)))
+        case Failure(e: IdentityCommunicationException) => Option(Reject(SC_BAD_GATEWAY, Some(e.getMessage)))
+        case Failure(e: OverLimitException) =>
+          if (isSelfValidating) {
+            Option(Reject(e.statusCode, Some(e.getMessage), Map(HttpHeaders.RETRY_AFTER -> e.retryAfter)))
+          } else {
+            Option(Reject(SC_SERVICE_UNAVAILABLE, Some(e.getMessage), Map(HttpHeaders.RETRY_AFTER -> e.retryAfter)))
```
dmnjohns: So here, if we did the `PartialFunction` thing, it would look like:
```
KeystoneV2Authorization.handleFailures orElse {
  case Failure(e: MissingAuthTokenException) => ...
  ...
}
```

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/AbstractKeystoneV2Filter.scala
dmnjohns: Oh. I guess I nested them so that I didn't have to keep passing the static `config` reference around.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
```diff
@@ -0,0 +1,152 @@
...
+import org.openrepose.filters.keystonev2.AbstractKeystoneV2Filter.Reject
+import org.openrepose.filters.keystonev2.KeystoneV2Common.{Endpoint, EndpointsData, Role, ValidToken}
+import org.openrepose.filters.keystonev2.config._
+
+import scala.collection.JavaConverters._
+import scala.util.{Failure, Success, Try}
+
+object KeystoneV2Authorization extends LazyLogging {
+
+  def doAuthorization(config: KeystoneV2Config, request: HttpServletRequestWrapper, validToken: ValidToken, endpoints: => Try[EndpointsData]): AuthorizationInfo = {
```
dmnjohns: I don't love the new case class that is basically a `Try` with more information on failure. That makes me think that this method should have its concerns separated somehow, but I don't think that is trivial to do. Furthermore, doing so would probably mean more code in both the authentication and authorization filters to do the same thing. At least, until we pull the authorization code out of the authentication filter. 

Perhaps this is just something to look at when we do finally yank the authorization code out of the authentication filter. After all, things become a bit simpler once we aren't worrying about scoping the token stuff that we write to headers.That is, the add everything then remove stuff the user isn't authorized to see approach that we get by default by separating the authentication and authorization filters seems cleaner and clearer than the current approach where we trying to figure out what to add before adding it to headers.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -55,518 +54,373 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
...
-                }
-              case Failure(e) if e.getCause.isInstanceOf[AkkaServiceClientException] && e.getCause.getCause.isInstanceOf[TimeoutException] =>
-                Reject(SC_GATEWAY_TIMEOUT, Some(s"Call timed out: ${e.getMessage}"))
-              case Failure(_: AdminTokenUnauthorizedException) if isSelfValidating =>
-                Reject(SC_UNAUTHORIZED, Some("Token unauthorized"))
-              case Failure(e) => Reject(SC_INTERNAL_SERVER_ERROR, Some(e.getMessage))
-            }
-          }
+  override def doWork(servletRequest: HttpServletRequest, servletResponse: HttpServletResponse, chain: FilterChain): Unit = {
+    val response = new HttpServletResponseWrapper(servletResponse, MUTABLE, PASSTHROUGH)
```
dmnjohns: I don't love having to wrap the response here to then wrap it again when `super.doWork` is called. I get that we do that so that we can keep the super code common yet still `uncommit` and fix the things that need to be fixed, but still, bleh. 

I guess the alternative would be to add a hook in the abstract filter to allow processing of the response before it is committed, but then the wrapper response would have to be passed if we wanted to avoid re-wrapping. I'm not sure if that's actually any better.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -55,518 +54,373 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
...
-                addIdentityStatusHeader(confirmed = false)
-                delegationHeaders foreach { case (key, values) =>
-                  values foreach { value =>
-                    request.addHeader(key, value)
-                  }
-                }
+    super.doWork(servletRequest, response, chain)
 
-                chain.doFilter(request, response)
+    val keystoneAuthenticateHeader = s"Keystone uri=${configuration.getIdentityService.getUri}"
```
dmnjohns: Seems like we may as well move this inside of the conditional block.

repose-aggregator/core/repose-core-api/src/main/scala/org/openrepose/core/filter/AbstractConfiguredFilter.scala
```diff
@@ -113,7 +113,14 @@ abstract class AbstractConfiguredFilter[T: ClassTag](val configurationService: C
...
     */
-  override def isInitialized: Boolean = initialized
+  override def isInitialized: Boolean = listenerInitialized
+
+  /**
+    * Indicates that the filter has done all initialization and is ready to serve.
+    *
+    * @return
+    */
+  def filterInitialized: Boolean = isInitialized
```
dmnjohns: Hm, so a way to differentiate between the listener in the abstract filter being initialized and the filter itself being initialized. Makes sense.

I don't love having to leave `isInitialized` public, though, since that might cause some confusion. I almost want to suggest having the abstract filter contain an `UpdateListener` rather than **be** an `UpdateListener`. That seems cleaner to me, I think...

repose-aggregator/commons/commons-test/src/main/scala/org/openrepose/commons/test/ConfigValidator.scala
adrianjgeorge: I actually find it less clear with an added layer, besides physically creating a new object you are also mentally creating another object that has to be tracked.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
adrianjgeorge: Yeah, that's better. I think the current state was from a different set of changes that required it, but as things moved around more it could collapse again.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
adrianjgeorge: I was never a fan of this being here. I was just preserving what it use to do. I have no qualms about just cutting it entirely.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/AbstractKeystoneV2Filter.scala
adrianjgeorge: Weirdly it wouldn't let me put them below any more, irc. There were a couple of places where it things use to be defined below their usage and it didn't care, but now it does. I didn't bother trying to fight it i just made it happy.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
adrianjgeorge: Yeah, I wasn't super excited by it either, but due to how everything was currently wired together I didn't really see another way out, without massive rewrites, and even then i suspect we still would have ended with something similar.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
adrianjgeorge: Yep, another 'fix' i wasn't super happy with, but had to live with.

repose-aggregator/core/repose-core-api/src/main/scala/org/openrepose/core/filter/AbstractConfiguredFilter.scala
adrianjgeorge: I thought the same, then i started to think out the implications of that change and it got pretty gross. I don't remember it all right now though.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
adrianjgeorge: yep, it was a holdover of things getting moved around.

repose-aggregator/commons/commons-test/src/main/scala/org/openrepose/commons/test/ConfigValidator.scala
dmnjohns: Fair enough.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
dmnjohns: I'd say go for it, then. Using the global variable holding the configuration simplifies everything a little bit, and it means that methods can be de-nested (un-nested? whatever) without having to pass the configuration around all the time. That makes things easier to read, I think, and easier to unit test.

Besides, few if any of our other filters keep a static reference, and I can't think of a time where the config changing during processing has been an issue.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/AbstractKeystoneV2Filter.scala
dmnjohns: Oh, huh, weird. I was going to say, I know a bunch of languages either don't support calling methods/functions before they are defined, or they make you forward declare them, but I thought it would "just work" with Scala (mostly because it seemingly does work with some of the other nested things).

Oh well.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
adrianjgeorge: as written your suggestion doesn't work

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/AbstractKeystoneV2Filter.scala
```diff
@@ -0,0 +1,146 @@
...
+
+  val handleFailures: PartialFunction[Try[Unit.type], KeystoneV2Result]
+
+  def doAuth(request: HttpServletRequestWrapper): Try[Unit.type]
+
+  def doWork(servletRequest: HttpServletRequest, servletResponse: HttpServletResponse, chain: FilterChain): Unit = {
+    /**
+      * STATIC REFERENCE TO CONFIG
+      */
+    val config = configuration
```
dmnjohns: Since we're using the `configuration` member object in the concrete implementation, I don't see much value in keeping this around here, either.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/AbstractKeystoneV2Filter.scala
```diff
@@ -0,0 +1,146 @@
...
+
+    val filterResult =
+      if (isWhitelisted(request.getRequestURI)) {
+        Pass
+      } else {
+        val authResult = doAuth(request)
+        handleFailures.applyOrElse(authResult, (t: Try[Unit.type]) => t match {
+          case Success(_) => Pass
+          case Failure(e) => Reject(SC_INTERNAL_SERVER_ERROR, Some(e.getMessage))
+        })
```
dmnjohns: This could also be written as:

```
        handleFailures.applyOrElse(authResult, (_: Try[Unit.type]) match {
          case Success(_) => Pass
          case Failure(e) => Reject(SC_INTERNAL_SERVER_ERROR, Some(e.getMessage))
        })
```

Unfortunately, the argument types of an anonymous function must be fully known, so that's as simple as I could get it.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/AbstractKeystoneV2Filter.scala
dmnjohns: Oh, and if we do get rid of these references, do we want to pop all of the nested methods out? I personally vote yes.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -55,519 +54,363 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
...
+      * DECLARE COMMON VALUES
+      */
+    lazy val traceId = Option(request.getHeader(CommonHttpHeader.TRACE_GUID)).filter(_ => sendTraceHeader)
+    lazy val requestHandler = new KeystoneRequestHandler(configuration.getIdentityService.getUri, akkaServiceClient, traceId)
 
-                chain.doFilter(request, response)
+    /**
+      * DEFINING FUNCTIONS IN SCOPE
+      */
+    def isWhitelisted(requestUri: String): Boolean = {
```
dmnjohns: Hm, this shouldn't be here anymore, should it? It looks like it was moved to the abstract filter.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -55,519 +54,363 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
...
-            }
-        }
-
-        // Construct and add authenticatedBy, if available
-        token.authenticatedBy foreach {
-          _ foreach {
-            request.addHeader(OpenStackServiceHeader.AUTHENTICATED_BY, _)
-          }
-        }
+    def addIdentityStatusHeader(confirmed: Boolean): Unit = {
```
dmnjohns: This was also moved to the abstract filter.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -55,519 +54,363 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
...
+      */
+    getAuthToken flatMap { authToken =>
+      validateToken(authToken) flatMap { validToken =>
+        lazy val endpoints = getEndpoints(authToken) // Prevents making call if its not needed
+        val authResult = KeystoneV2Authorization.doAuthorization(configuration, request, validToken, endpoints)
+
+        addTokenHeaders(authResult.scopedToken, authResult.matchedTenant)
+        authResult match {
+          case AuthorizationPassed(scopedToken, _) =>
+            addCatalogHeader(endpoints)
```
dmnjohns: Hm, we are losing the return value from this call. Since `endpoints` is lazy, we might end up making the endpoints call here, so we need to capture the failure if that call fails. I suppose we could go back to nesting these calls to fix that.

```
addCatalogHeader(endpoints) flatMap { _ =>
  addGroupsHeader(getGroups(authToken, scopedToken))
}
```

adrianjgeorge: Lol, this was your suggested change

dmnjohns: I think they were side-by-side before my suggestion, though admittedly, I did not catch this when I made that suggestion.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/AbstractKeystoneV2Filter.scala
adrianjgeorge: I get the desire, but it's adding work onto an already long running story.

repose-aggregator/components/filters/keystone-v2-filter/src/test/scala/org/openrepose/filters/keystonev2/KeystoneV2AuthorizationTest.scala
```diff
@@ -0,0 +1,178 @@
...
+
+    it("should fail when there wasn't a tenant match") {
+      authorizeTenant(true, None).failed.get shouldBe a [InvalidTenantException]
+    }
+
+    describe("authorizeEndpoints") {
+      val config = new ServiceEndpointType().withName("banana").withRegion("ord").withPublicUrl("http://foo.com")
+
+      it("should succeed if unconfigured") {
+        authorizeEndpoints(null, false, Success(EndpointsData("", Vector.empty))).success
```
dmnjohns: Somehow I have never known about the `success` and `failure` wrappers. This is extremely useful.

adrianjgeorge: it also gives you a nice way of unpacking them for other comparisons. It's just in most of our cases we get a success of unit.

adrianjgeorge:
retest this please

PR #1856, opened 2018-01-09
REP-6400 Create Auth-Z filter (split Keystone v2 filter)

dmnjohns:


repose-aggregator/components/filters/keystone-v2-filter/src/main/resources/META-INF/schema/config/keystone-v2-common.xsd
```diff
@@ -129,6 +132,17 @@
             </xs:element>
         </xs:sequence>
 
+        <xs:attribute name="header-extraction-name" type="xs:string" use="optional">
```
dmnjohns: I wasn't sure if we wanted to default this attribute or not. Right now, I left it with no default since it's purely optional -- the user can choose to extract the tenant ID either from the URI, a header, or both. If/when we do drop support for URI extraction in this filter, we will probably want a default.

repose-aggregator/components/filters/keystone-v2-filter/src/test/scala/org/openrepose/filters/keystonev2/config/KeystoneV2SchemaTest.scala
```diff
@@ -88,5 +88,51 @@ class KeystoneV2SchemaTest extends ConfigurationTest {
...
+      val config = """<keystone-v2 xmlns="http://docs.openrepose.org/repose/keystone-v2/v1.0">
+                     |    <identity-service uri="https://some.identity.com"/>
+                     |    <tenant-handling>
+                     |        <validate-tenant header-extraction-name="x-expected-tenant"/>
+                     |    </tenant-handling>
+                     |</keystone-v2>""".stripMargin
+      validator.validateConfigString(config)
+    }
+
+    it("should successfully validate config if both a tenant URI extraction regex nor a tenant header extraction name are provided") {
```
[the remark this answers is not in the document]
dmnjohns: Fixed.

repose-aggregator/components/filters/keystone-v2-filter/src/main/resources/META-INF/schema/config/keystone-v2-common.xsd
```diff
@@ -100,7 +103,9 @@
                     </xs:documentation>
                 </xs:annotation>
                 <xs:complexType>
+                    <!-- TODO: Does the authorization filter need a default-tenant-quality and roles-tenant-quality? -->
```
dmnjohns: It seems to me like those two attributes should not mean anything in the authorization filter. Currently they do nothing.

This is kind of painful to tease out, but I think we could do it. It's easy enough to remove these from this type and extend this type in the authentication filter, but then to require the new type means we probably have to restrict the root config type. So then we end up with a restriction and extension of the root config type in order to straighten things out.

repose-aggregator/components/filters/keystone-v2-filter/src/main/resources/META-INF/schema/config/keystone-v2-common.xsd
```diff
@@ -100,7 +103,9 @@
...
                     <xs:attribute name="default-tenant-quality" type="QualityType" default="0.9" use="optional"/>
+                    <!-- TODO: Should this be renamed to matching-tenant-quality? Not backwards compatible. -->
```
dmnjohns: This is probably just something we want to do eventually. I'm not sure we want to do anything about this right now, though.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneRequestHandler.scala
```diff
@@ -83,6 +83,7 @@ class KeystoneRequestHandler(identityServiceUri: String, akkaServiceClient: Akka
       case Success(serviceClientResponse) =>
         serviceClientResponse.getStatus match {
           case statusCode if statusCode >= 200 && statusCode < 300 =>
+            // TODO: Handle character encoding set in the content-type header rather than relying on the default system encoding
```
dmnjohns: So yeah... we definitely are not explicit about character encoding when parsing the response from Identity. We just assume the JVM default encoding. We also don't send the `accept-charset` header in our request, but could do so if we wanted to express a preference for a certain encoding.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
```diff
@@ -32,18 +32,15 @@ import scala.util.{Failure, Success, Try}
...
-            }
-          ).headOption
-        )).getOrElse(throw UnparseableTenantException("Could not parse tenant from the URI"))
+  val handleFailures: PartialFunction[Try[Unit.type], KeystoneV2Result] = {
+    case Failure(e: InvalidTenantException) => Reject(SC_UNAUTHORIZED, Some(e.getMessage))
+    case Failure(e: UnauthorizedEndpointException) => Reject(SC_FORBIDDEN, Some(e.getMessage))
+    case Failure(e: UnparseableTenantException) => Reject(SC_UNAUTHORIZED, Some(e.getMessage))
+  }
 
+  // TODO: Take tenants to roles mapping instead of token to calculate scoped roles.
```
dmnjohns: Doing this is kind of necessary to drop the requirement of a token request attribute for the authorization filter (without refactoring this a bunch, but since we're doing the map anyway, why not use it). Turns out, though, that we only need the tenant-to-role mapping and endpoints to do all of our authorization stuff.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
```diff
@@ -59,13 +56,24 @@ object KeystoneV2Authorization extends LazyLogging {
     }
   }
 
-  val handleFailures: PartialFunction[Try[Unit.type], KeystoneV2Result] = {
-    case Failure(e: InvalidTenantException) => Reject(SC_UNAUTHORIZED, Some(e.getMessage))
-    case Failure(e: UnauthorizedEndpointException) => Reject(SC_FORBIDDEN, Some(e.getMessage))
-    case Failure(e: UnparseableTenantException) => Reject(SC_UNAUTHORIZED, Some(e.getMessage))
+  // TODO: Switch to getRequestTenants (plural) and include all tenants extracted. This is a behavior change.
```
dmnjohns: So as written, this code will look for a tenant in a header as well as in the URI, giving preference to the value in a header. However, it only returns a single tenant ID. This is consistent with current behavior given multiple URI extraction regexes, but is going to change with the support for multi-tenants story. I just wanted to make a note of that.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2AuthorizationFilter.scala
```diff
@@ -0,0 +1,110 @@
...
+      case Failure(e@(_: MissingTokenException |
+                      _: MissingEndpointsException |
+                      _: InvalidTokenException |
+                      _: InvalidEndpointsException)) =>
+        Reject(SC_INTERNAL_SERVER_ERROR, Some(e.getMessage))
+    }
+  }
+
+  override def doAuth(request: HttpServletRequestWrapper): Try[Unit.type] = {
+    // TODO: Stop using request attributes. Stop using the token. Just use the tenant to roles map, tenant ids, and roles headers.
```
dmnjohns: This ties back to a previous comment I made, but it's straightforward enough. The one thing I will call out is that, as written, everything is being transported as request attributes right now. I think we all agree that using the headers we are or will be populating for transport is a better idea.

I used request attributes when writing this because we do not yet populate the tenant-to-role map header, so I still needed the token. Since I was already putting the token in a request attribute, I decided to also throw endpoints in a request attribute as kind of a shim. The token attribute cannot be removed until the next story is worked, but the endpoints attribute can be dropped at any time, we will just have to decode and parse the catalog header. As-is, I think we have a MVP, but...

Dropping the endpoints attribute is probably what I'll try to work through today.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Common.scala
```diff
@@ -61,4 +67,13 @@ object KeystoneV2Common {
 
   case class EndpointsData(json: String, vector: Vector[Endpoint])
 
+  object Endpoint {
+    implicit val endpointReads: Reads[Endpoint] = (
```
dmnjohns: This is pretty cool. I had originally intended to stick this here just so I could use it in both filters, and to import it into the same scope as the JSON processing calls. However, after some quick research on implicit resolution, I learned that implicits in companion objects to associated types are in scope. So by sticking this here, I don't even have to import it anywhere. Yay!

Here's the Scala documentation about implicits resolution:
https://docs.scala-lang.org/tutorials/FAQ/finding-implicits.html#where-do-implicits-come-from

repose-aggregator/components/filters/keystone-v2-filter/src/main/resources/META-INF/schema/config/keystone-v2-common.xsd
adrianjgeorge: I think originally the thinking was that authorization was going to long term be writer of these headers so the configuration made sense here, to end up in both. We've teased out a little more how we are actually going to implement, so it makes less sense now.

repose-aggregator/components/filters/keystone-v2-filter/src/main/resources/META-INF/schema/config/keystone-v2-common.xsd
adrianjgeorge: default tenant quality and matching tenant quality are two different concerns

repose-aggregator/components/filters/keystone-v2-filter/src/main/resources/META-INF/schema/config/keystone-v2-common.xsd
adrianjgeorge: I don't think i like this as an attribute still. I series of elements much like the uri regex make more sense to me. The other way is they do a header translation and/or merge header filter to put a bunch of values together, which while very much the toolbox approach feels a little klugey.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2AuthorizationFilter.scala
```diff
@@ -0,0 +1,131 @@
...
+
+  def scopeRolesHeader(request: HttpServletRequestWrapper, scopedToken: ValidToken): Unit = {
+    request.removeHeader(ROLES)
+    scopedToken.roles.map(_.name).foreach(request.appendHeader(ROLES, _))
+  }
+}
+
+object KeystoneV2AuthorizationFilter {
+
+  case class MissingTokenException(message: String, cause: Throwable = null) extends Exception(message, cause)
```
adrianjgeorge: I'm not a huge fan of all these extending straight exception. We should make a base class that they extend.

repose-aggregator/src/docs/asciidoc/filters/keystone-v2-authorization.adoc
```diff
@@ -0,0 +1,204 @@
...
+This filter does not require any request headers.
+
+=== Required Preceding Filters
+This filter requires that the <<keystone-v2.adoc#, Keystone v2 Filter>> precede it to populate necessary user data.
+
+=== Request Headers Created
+The following headers are created:
+
+* `X-Tenant-Id` - The Tenant ID's for the user.
+  This header will only be added if this filter is configured to validate tenants.
```
adrianjgeorge: Didn't we just say yesterday it'll almost always do at least one?

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/keystonev2/authorization/AuthorizationFilterDelegationTest.groovy
```diff
@@ -0,0 +1,157 @@
...
+
+        and: "Delegation header is present"
+        def delegatedHeader = mc.handlings[0].request.headers.findAll("X-Delegated").toString()
+        delegatedHeader.contains("status_code=403")
+        delegatedHeader.contains("component=keystone-v2")
+        delegatedHeader.contains("message=User did not have the required endpoint")
+        delegatedHeader.contains("q=0.8")
+    }
+
+    @Unroll
```
adrianjgeorge: Why bother when you only provided a single set of values?

adrianjgeorge: nevermind, i read the where wrong

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
```diff
@@ -59,19 +56,30 @@ object KeystoneV2Authorization extends LazyLogging {
...
+  def getRequestTenant(config: ValidateTenantType, request: HttpServletRequestWrapper): String = {
+    Option(config).flatMap(validateTenantConfig =>
+      Option(validateTenantConfig.getHeaderExtractionName).flatMap(headerName =>
+        request.getSplittableHeaderScala(headerName).headOption
```
adrianjgeorge: Pretty sure these are still going to have qualities dangling off them.

repose-aggregator/components/filters/keystone-v2-filter/src/main/resources/META-INF/schema/config/keystone-v2-common.xsd
dmnjohns: Oh, the TODO is actually above the line for the `uri-tenant-quality` which is now being used not just for the URI tenant but for the matching tenant. This line was referring to the `uri-tenant-quality` attribute, not the `default-tenant-quality` attribute. GitHub just chooses to give you the commented on line and the line above it, but not the line below...

repose-aggregator/src/docs/asciidoc/filters/keystone-v2-authorization.adoc
dmnjohns: The authentication filter will, but this filter only messes with the tenant header if tenant authorization is performed. What would it do otherwise? Theoretically, the authentication filter has already written the header with appropriate values.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2AuthorizationFilter.scala
dmnjohns: Fair enough, I'll make the base class.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
dmnjohns: Dang it. Totally forgot about that. Will fix.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/keystonev2/authorization/AuthorizationFilterTest.groovy
```diff
@@ -0,0 +1,200 @@
...
+
+        and: "It should have never made it to the origin service"
+        mc.handlings.size() == 0
+    }
+
+    @Unroll
+    def "User sends the prefixed tenant (#tenantPrefix) in the URI should succeed and be extracted"() {
+        when: "User sends a request through repose"
+        MessageChain mc = deproxy.makeRequest(
+            url: reposeEndpoint + "/v2/extract/" + tenantPrefix + fakeIdentityV2Service.client_tenantid + "/ss",
```
dmnjohns: It took me a weirdly long time to figure out why this test was failing, but I think it's this test, as written, is invalid. Prefixes are stripped from tenants in the access response, not from the tenant we extract from the request.

Here are the current tests proving that for the KeystoneV2 filter:
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/keystonev2/tenantvalidation/StripTenantPrefixTest.groovy#L62-L96

The fix would be to set one of the tenant IDs in the Identity response to the prefixed value, and drop the prefix in the URI. Easy enough to do with:
```
fakeIdentityV2Service.client_tenantid = tenantPrefix + fakeIdentityV2Service.client_tenantid
```

wdschei: That makes sense. I guess I need more remedial Keystone v2 training. :smile:

dmnjohns:
Retest this please.

repose-aggregator/components/filters/keystone-v2-filter/src/main/resources/META-INF/schema/examples/keystone-v2-authorization.cfg.xml
```diff
@@ -0,0 +1,44 @@
...
+
+    <white-list>
+        <uri-regex>/application\.wadl$</uri-regex>
+    </white-list>
+
+    <tenant-handling send-all-tenant-ids="true">
+        <validate-tenant>
+            <uri-extraction-regex>.*/servers/([-|\w]+)/?.*</uri-extraction-regex>
+            <uri-extraction-regex>.*/clients/([-|\w]+)/?.*</uri-extraction-regex>
+            <header-extraction-name>x-expected-tenant</header-extraction-name>
```
adrianjgeorge: This order isn't required right?

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneRequestHandler.scala
```diff
@@ -262,20 +258,20 @@ object KeystoneRequestHandler {
     }
   }
 
-  trait IdentityException
+  abstract class IdentityException(message: String, cause: Throwable) extends Exception(message, cause)
```
adrianjgeorge: man this was weird before

repose-aggregator/components/filters/keystone-v2-filter/src/main/resources/META-INF/schema/examples/keystone-v2-authorization.cfg.xml
dmnjohns: Nope. Any order should work.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneRequestHandler.scala
dmnjohns: Right? I don't remember why this was a trait... I can kind of see that for classifying exceptions but still allowing them to call different constructors on `Exception`, but we always call the same constructor here, and make the base exception a class seems more natural.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
adrianjgeorge: Pretty sure this comment is out of date

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
```diff
@@ -59,19 +56,31 @@ object KeystoneV2Authorization extends LazyLogging {
...
-  val handleFailures: PartialFunction[Try[Unit.type], KeystoneV2Result] = {
-    case Failure(e: InvalidTenantException) => Reject(SC_UNAUTHORIZED, Some(e.getMessage))
-    case Failure(e: UnauthorizedEndpointException) => Reject(SC_FORBIDDEN, Some(e.getMessage))
-    case Failure(e: UnparseableTenantException) => Reject(SC_UNAUTHORIZED, Some(e.getMessage))
+  // TODO: Switch to getRequestTenants (plural) and include all tenants extracted. This is a behavior change.
+  def getRequestTenant(config: ValidateTenantType, request: HttpServletRequestWrapper): String = {
+    Option(config).flatMap(validateTenantConfig =>
+      (validateTenantConfig.getUriExtractionRegexAndHeaderExtractionName.asScala.toStream map {
+        case headerExtraction: HeaderExtractionType =>
+          request.getPreferredSplittableHeaders(headerExtraction.getValue).asScala.headOption
```
adrianjgeorge: this seems super broke. you can only possibly match on the first header of a given quality

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
dmnjohns: Not in this code, which does not yet support the multi-tenant use-case. The next story handles it, and removes the TODO.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
dmnjohns: That is exactly what it does. You'll also notice that I turn the list of extractions into a `Stream`, and then terminate the stream with a `find`. The big idea is that the first tenant that matches will be returned.

This is the single-tenant approach consistent with current behavior, but it will change with the next story and PR.

adrianjgeorge: Yeah, i was thinking of the other header of tenants, not the ones to match. derp.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
adrianjgeorge: I meant your pr comment, not the code comment.

adrianjgeorge:
Every single example and test shows the tenant checks as urls then headers, I'd like to see at least one that has both intermixed to prove that it works.

dmnjohns:
Legit. I will add that test.

PR #1858, opened 2018-01-18
Opentracing

dimtruck:
PR is ready to be fully reviewed:

This PR introduces OpenTracing service that wraps http://opentracing.io implementation in Repose.

The configuration is not required.  If it's omitted, the service will not start and will not be used.

Bare minimum configuration is to specify:

* `tracer` - type of Tracer implementation.  Currently only `jaeger` is supported
* `name` - service name.  This is unique to your service.  Best practice is to name it for the service that it's in front of and add `-repose` suffix.  E.g. if it's in front of Nova, name the service `nova-repose`.

There are two ways of communicating with a tracer:

* UDP - sends data to agent, which then buffers and forwards data to collector
* HTTP - sends data directly to collector **(default if not specified)**

The service won't start if repose cannot connect to the tracer right away when using UDP so make sure your tracer agent is up and running prior to starting repose.  The service will start and **silently fail to send traces** if using HTTP.  Functionality to track failed traces is out of scope for this PR (done via DropWizard).

For UDP sender, you must configure `agent-port` and `agent-host`, which point to the tracer agent.

For HTTP sender, you must configure `collector-endpoint`, which points to collector instance.

For HTTP sender, you can optionally configure basic authentication with `username` and `password` OR bearer token authentication with `token`.

The service also supports sampling configuration.  Sampling, in this context, means reporting and tracking a request in the tracer.  Sampling is on and will always report by default.

The following configurations are supported:

* `const` - `1` is to always report **(default if not specified)**.  `0` is to never report (or sample)
* `probability` - Ranges from `0.0` to `1.0`.  `0.0` will never report and `1.0` will always report.  Default is `0.001`, which will report one out of every 1000 requests.
* `rate-limited` - Double that corresponds to number of requests per second sampled.  

Repose will trace both origin service and 3rd party (e.g. keystone) traffic.


repose-aggregator/commons/commons-utilities/build.gradle
```diff
@@ -11,6 +11,7 @@ dependencies {
     compile "com.fasterxml.jackson.core:jackson-databind"
     compile "org.apache.httpcomponents:httpclient"
     compile "org.jvnet.jaxb2_commons:jaxb2-basics-runtime"
+    compile "io.opentracing:opentracing-api:0.30.0"
```
Mario-Lopez: Need to add the library to `dependencies.txt`.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneRequestHandler.scala
```diff
@@ -69,6 +69,8 @@ class KeystoneRequestHandler(identityServiceUri: String, akkaServiceClient: Akka
       )
     )
 
+    // TODO: add tracing info here and add it to trace guid
```
dimtruck: Discussion point:  Currently, there are 3 trace points that have been configured: 

* `PowerFilter`
* `RequestProxyServiceImpl`
* `AkkaServiceClientImpl`

`PowerFilter` is a wrapper around all of Repose and the trace point there (although I'd like the entry point to be sooner since there are a number of places we can fail before we emit a trace).

`RequestProxyServiceImpl` also is a fine trace point because it just starts a new span and injects the child headers.

`AkkaServiceClientImpl` I go back and forth.  On one hand, it's good that it's a single point of entry for all filters that need to talk to 3rd party services (Identity, etc.); however, it's not as simple to wrap it in a span since it's an async request that ends before we ever get a response back.  For that, it's better to do it per filter, and it'd be easier to name each child span (e.g. `keystone-filter-span`); however, then we duplicate code and we'd need to remember to add a span for any new filter (including 3rd party filters that may use Repose's akka stuff).

Help!

dimtruck:
@adrianjgeorge it's ready!!!!

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneRequestHandler.scala
adrianjgeorge: the right spot is to put it down in the http client. that covers everything the akka client does as well as when we go to the origin service, it might even cover the dd (i'd have to look at the code for that one).

repose-aggregator/components/filters/herp-filter/src/main/scala/org/openrepose/filters/herp/HerpFilter.scala
```diff
@@ -51,6 +52,7 @@ import scala.util.matching.Regex
 
 @Named
 class HerpFilter @Inject()(configurationService: ConfigurationService,
+                           openTracingService: OpenTracingService,
```
adrianjgeorge: Why? You've added a bunch of deps to this filter, but don't seem to use them at all?

repose-aggregator/components/services/akka-http-client-service/akka-http-client-service-impl/src/main/java/org/openrepose/core/services/serviceclient/akka/impl/AkkaServiceClientImpl.java
```diff
@@ -103,10 +113,39 @@ public ServiceClientResponse get(String hashKey, String uri, Map<String, String>
...
     public ServiceClientResponse get(String hashKey, String uri, Map<String, String> headers, boolean checkCache) throws AkkaServiceClientException {
-        AuthGetRequest authGetRequest = new AuthGetRequest(hashKey, uri, headers);
+
+
         try {
             Timeout timeout = new Timeout(socketTimeout + CONNECTION_TIMEOUT_BUFFER_MILLIS, TimeUnit.MILLISECONDS);
-            Future<ServiceClientResponse> future = getFuture(authGetRequest, timeout, checkCache);
+            Future<ServiceClientResponse> future;
+
+            if (openTracingService.isEnabled()) {
```
adrianjgeorge: Is the open tracing api not written in a way where it's safe for us always to call it regardless of how the back end is configured? Kind of log4j/slf4j where we can always safely make our calls to loggers regardless of what level or output, if any, they are configured for.

repose-aggregator/components/services/akka-http-client-service/akka-http-client-service-impl/src/main/java/org/openrepose/core/services/serviceclient/akka/impl/AkkaServiceClientImpl.java
```diff
@@ -103,10 +113,39 @@ public ServiceClientResponse get(String hashKey, String uri, Map<String, String>
...
+                // get active span
+                ActiveSpan parentSpan = openTracingService.getGlobalTracer().activeSpan();
```
adrianjgeorge: Can there only be one active span at a time? Or is it one per thread?

adrianjgeorge: Or are some of these your constructs?

repose-aggregator/components/services/akka-http-client-service/akka-http-client-service-impl/src/main/java/org/openrepose/core/services/serviceclient/akka/impl/AkkaServiceClientImpl.java
```diff
@@ -103,10 +113,39 @@ public ServiceClientResponse get(String hashKey, String uri, Map<String, String>
...
+
+                try (ActiveSpan span = openTracingService.getGlobalTracer().buildSpan(uri)
+                    .asChildOf(parentSpan)
+                    .withTag(Tags.SPAN_KIND.getKey(), Tags.SPAN_KIND_CLIENT)
+                    .withTag(Tags.HTTP_URL.getKey(), uri)
+                    .withTag(Tags.HTTP_METHOD.getKey(), "GET")
+                    .startActive()) {
+
+                    // this needs to be better ... however, it looks like headers are an immutable map
+                    Map<String, String> copiedHeaderMap = new HashMap<>(headers.size() + 1);
```
adrianjgeorge: Why are you manually setting the size?

adrianjgeorge: or for that matter doing this operation in two lines?

repose-aggregator/components/services/akka-http-client-service/akka-http-client-service-impl/src/main/java/org/openrepose/core/services/serviceclient/akka/impl/AkkaServiceClientImpl.java
```diff
@@ -103,10 +113,39 @@ public ServiceClientResponse get(String hashKey, String uri, Map<String, String>
...
+                    copiedHeaderMap.putAll(headers);
+
+                    // adds the trace guid (tracer specific)
+                    openTracingService.getGlobalTracer().inject(
+                        span.context(), Format.Builtin.TEXT_MAP,
+                        new TraceGUIDInjector(copiedHeaderMap, CommonHttpHeader.TRACE_GUID));
```
adrianjgeorge: Side effects on a passed in parameter is bad business. Why are you passing the whole map and modifying it? Do you use anything other than the one value? If not I'd ay pass the value you care about and and have it return the new value.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneRequestHandler.scala
dimtruck: addressed.

dimtruck:
@adrianjgeorge this is ready for review again.

repose-aggregator/components/services/http-client-service/http-client-service-impl/src/main/java/org/openrepose/core/services/httpclient/impl/HttpConnectionPoolProvider.java
```diff
@@ -58,8 +62,8 @@
     private HttpConnectionPoolProvider() {
     }
 
-    public static HttpClient genClient(String configRoot, PoolType poolConf) {
-
+    public static HttpClient genClient(String configRoot, PoolType poolConf, OpenTracingService openTracingService) {
+        // to enable some of the new features available in Apache HTTP Client 4.3, updating to builder pattern
```
adrianjgeorge: pretty sure you didn't do this?

repose-aggregator/components/services/opentracing-service/opentracing-service-api/build.gradle
```diff
@@ -0,0 +1,19 @@
...
+}
+
+
+jacocoTestCoverageVerification {
+    violationRules {
+        rules.forEach { rule ->
+            rule.limits.forEach { limit ->
+                if (limit.counter == 'INSTRUCTION' && limit.value == 'COVEREDRATIO') {
+                    // todo: set to the current coverage -- need more coverage, but we don't want the build to fail for the baseline
+                    limit.minimum = 0.22
```
adrianjgeorge: yeah, this number is pretty low

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/out/production/resources/META-INF/schema/config/opentracing.xsd
```diff
@@ -0,0 +1,135 @@
...
+            <xs:annotation>
+                <xs:documentation>
+                    <html:p>
+                        Service name to use.  Will identify this service (e.g. identity-repose)
+                    </html:p>
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+
+        <xs:attribute name="tracer.host" type="xs:string" default="localhost">
```
adrianjgeorge: this name is not consistent to our standards, should be `tracer-host`

repose-aggregator/components/services/opentracing-service/opentracing-service-api/build.gradle
dimtruck: copy/pasta :shame:

repose-aggregator/components/services/http-client-service/http-client-service-impl/src/main/java/org/openrepose/core/services/httpclient/impl/HttpConnectionPoolProvider.java
dimtruck: :D ... i did ... then cursed myself and reverted ... except for the tiny comment ... will remove :)

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/out/production/resources/META-INF/schema/config/bindings.xjb
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
adrianjgeorge: Where is this? It seems like it's in a super weird non-standard spot

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/out/production/resources/META-INF/schema/config/opentracing.xsd
```diff
@@ -0,0 +1,135 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
adrianjgeorge: As is this one. Besides being in the wrong directory, they are on the wrong project, they should be in the api.

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/out/production/resources/META-INF/schema/config/opentracing.xsd
```diff
@@ -0,0 +1,135 @@
...
+            <xs:annotation>
+                <xs:documentation>
+                    <html:p>
+                        Tracer host (reporter host for zipkin and jaeger).  Default set to localhost.
+                    </html:p>
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+
+        <xs:attribute name="tracer.port" type="PositiveInt" default="9411">
```
adrianjgeorge: Same thing with the name here

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/out/production/resources/META-INF/schema/config/opentracing.xsd
```diff
@@ -0,0 +1,135 @@
...
+            <xs:annotation>
+                <xs:documentation>
+                    <html:p>
+                        Tracer port (reporter port for zipkin and jaeger).  Default set to 9411.
+                    </html:p>
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+
+        <xs:attribute name="flush.interval.ms" type="PositiveInt" default="1000">
```
adrianjgeorge: And again

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/out/production/resources/META-INF/schema/config/opentracing.xsd
```diff
@@ -0,0 +1,135 @@
...
+            <xs:annotation>
+                <xs:documentation>
+                    <html:p>
+                        Flush interval in milliseconds.  How often to send span data to tracer.  Defaults to 1 second.
+                    </html:p>
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+
+        <xs:attribute name="max.buffer.size" type="PositiveInt" default="10000">
```
adrianjgeorge: Last time?

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/out/production/resources/META-INF/schema/config/opentracing.xsd
dimtruck: I honestly had no idea where this needed to live.  moving!

dimtruck:
@adrianjgeorge take 20.  All comments addressed (includes the ones that aren't hidden due to directory move to the correct module)

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/src/main/java/org/openrepose/core/services/opentracing/OpenTracingServiceImpl.java
```diff
@@ -0,0 +1,269 @@
...
+        this.configurationService = configurationService;
+        this.healthCheckServiceProxy = healthCheckService.register();
+        this.configurationListener = new ConfigurationListener();
+    }
+
+    @PostConstruct
+    public void init() {
+        LOG.debug("Initializing OpenTracingService");
+
+        healthCheckServiceProxy.reportIssue(
```
adrianjgeorge: Health Check Service isn't really doing what we want it to, so we don't use it much. Feel free to not do all this.

dimtruck: If I don't have it, will it not be able to start my service?  I've noticed that the healthcheck checks if anything that's registered is broken and then won't start repose...

dimtruck:
retest this please

dimtruck:
Retest this please.

dimtruck:
retest this please

dimtruck:
retest this please

dimtruck:
retest this please

dimtruck:
retest this please

.gitignore
```diff
@@ -28,6 +28,7 @@ classes/
 # Gradle bits from: https://github.com/github/gitignore/blob/master/Gradle.gitignore
 .gradle
 build/
+out/
```
adrianjgeorge: Where did this directory come from?

wdschei: That is created by IntelliJ IDEA instead of it using the `build/` stated by the `build.gradle`. That is one of the issues I've been having too.

dimtruck: THE WORST.  Especially when I committed it before and it added a billion files.

build.gradle
```diff
@@ -184,7 +184,7 @@ subprojects {
         doLast {
             file(outputDir).mkdirs()
 
-            //Find all XSDs in the src/main/resources/META-INF.schema
+            //Find all XSDs in the src/main/resources/schema
```
adrianjgeorge: This isn't correct

repose-aggregator/components/filters/cors-filter/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
```diff
@@ -211,7 +211,7 @@ class CorsFilter @Inject()(configurationService: ConfigurationService)
     val maybeForwardedHostUri = Try(maybeForwardedHost
       .map(forwardedHost => new URIBuilder(s"${request.getScheme}://$forwardedHost"))
       .map(uri => uri.setPort(normalizePort(uri.getPort, uri.getScheme)).setHost(normalizeHost(uri.getHost)).build()))
-    lazy val hostUri = new URIBuilder().setScheme(request.getScheme).setHost(normalizeHost(request.getServerName))
+    lazy val hostUri = new URIBuilder("").setScheme(request.getScheme).setHost(normalizeHost(request.getServerName))
```
adrianjgeorge: What did you bring in that forced this change? I know it was talked about at one point but i don't recall.

repose-aggregator/components/services/http-client-service/http-client-service-impl/src/test/groovy/org/openrepose/core/services/httpclient/impl/HttpConnectionPoolProviderTest.groovy
```diff
@@ -35,12 +35,16 @@ import org.junit.Test
...
 
 import javax.servlet.ServletException
 import javax.servlet.http.HttpServletRequest
 import javax.servlet.http.HttpServletResponse
 import java.nio.charset.Charset
 
+import static org.mockito.Mockito.mock
+
+
 class HttpConnectionPoolProviderTest {
```
adrianjgeorge: You didn't add test to prove that you add the interceptors at the appropriate time.

repose-aggregator/components/services/opentracing-service/opentracing-service-api/src/main/resources/META-INF/schema/config/bindings.xjb
```diff
@@ -0,0 +1,31 @@
...
+  -->
+
+
+<bindings xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://java.sun.com/xml/ns/jaxb"
+          version="2.0"
+          xsi:schemaLocation="http://java.sun.com/xml/ns/jaxb http://java.sun.com/xml/ns/jaxb/bindingschema_2_0.xsd"
+          schemaLocation="opentracing.xsd">
+
+    <schemaBindings>
+        <package name="org.openrepose.core.service.opentracing.config"/>
```
adrianjgeorge: Are you certain that this is a core service and not a node service? I'm not pushing for a change, I'm looking for your reasoning.

repose-aggregator/components/services/opentracing-service/opentracing-service-api/src/main/resources/META-INF/schema/config/opentracing.xsd
```diff
@@ -0,0 +1,315 @@
...
+  -->
+
+
+<xs:schema xmlns:html="http://www.w3.org/1999/xhtml"
+           xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           xmlns="http://docs.openrepose.org/repose/opentracing/v1.0"
+           elementFormDefault="qualified"
+           targetNamespace="http://docs.openrepose.org/repose/opentracing/v1.0">
+
+    <xs:element name="opentracing" type="OpenTracingConfig"/>
```
adrianjgeorge: Maybe hyphenated?

repose-aggregator/components/services/opentracing-service/opentracing-service-api/src/main/resources/META-INF/schema/examples/opentracing.cfg.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<opentracing xmlns="http://docs.openrepose.org/repose/opentracing/v1.0"
+             name="test-repose"
+             tracer="jaeger"
+             collector-endpoint="http://localhost:8081">
+    <jaeger-sampling-config sample-type="rate-limited">
```
adrianjgeorge: I have feedback on this config, but i want to give it more consideration before i make recommendations.

repose-aggregator/components/services/opentracing-service/opentracing-service-api/src/test/scala/org/openrepose/core/services/opentracing/config/OpenTracingSchemaTest.scala
```diff
@@ -0,0 +1,108 @@
...
+import org.openrepose.core.service.opentracing.config.ObjectFactory
+import org.xml.sax.SAXParseException
+
+class OpenTracingSchemaTest extends ConfigurationTest {
+  override val schema: URL = getClass.getResource("/META-INF/schema/config/opentracing.xsd")
+  override val exampleConfig: URL = getClass.getResource("/META-INF/schema/examples/opentracing.cfg.xml")
+  override val jaxbContextPath: String = classOf[ObjectFactory].getPackage.getName
+
+  describe("schema validation") {
+    it("should successfully validate the config with http sender and sampling set to rate limited") {
```
adrianjgeorge: Why did you feel these cases needed testing? Typically we just do the example which you get for free by extending our test, and any assertions we put in the xsd.

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+    compile "io.opentracing:opentracing-util:0.31.0"
+    compile "com.uber.jaeger:jaeger-core:0.24.0"
+    compile "com.uber.jaeger:jaeger-apachehttpclient:0.24.0"
+
+    compileOnly "javax.servlet:javax.servlet-api"
+    compileOnly "javax.inject:javax.inject"
+    compileOnly "javax.ws.rs:javax.ws.rs-api"
+    compileOnly "com.google.code.findbugs:jsr305" //i hate this hack
+
+    testCompile project(":repose-aggregator:components:services:opentracing-service:opentracing-service-api")
```
adrianjgeorge: Why is this here it's already a `compile` dependency

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+    compile "org.springframework:spring-beans"
+    compile "io.opentracing:opentracing-api:0.31.0"
+    compile "io.opentracing:opentracing-util:0.31.0"
+    compile "com.uber.jaeger:jaeger-core:0.24.0"
+    compile "com.uber.jaeger:jaeger-apachehttpclient:0.24.0"
+
+    compileOnly "javax.servlet:javax.servlet-api"
+    compileOnly "javax.inject:javax.inject"
+    compileOnly "javax.ws.rs:javax.ws.rs-api"
+    compileOnly "com.google.code.findbugs:jsr305" //i hate this hack
```
adrianjgeorge: Did you need this or did ti come from where ever you copy-pasta'd

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+
+    testCompile project(":repose-aggregator:components:services:opentracing-service:opentracing-service-api")
+    testCompile "junit:junit:4.12"
+    testCompile "org.hamcrest:hamcrest-all"
+    testCompile "org.mockito:mockito-all"
+    testCompile "org.codehaus.groovy:groovy-all"
+    testCompile "org.eclipse.jetty:jetty-server"
```
adrianjgeorge: Why did you need jetty?

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+    compileOnly "javax.servlet:javax.servlet-api"
+    compileOnly "javax.inject:javax.inject"
+    compileOnly "javax.ws.rs:javax.ws.rs-api"
+    compileOnly "com.google.code.findbugs:jsr305" //i hate this hack
+
+    testCompile project(":repose-aggregator:components:services:opentracing-service:opentracing-service-api")
+    testCompile "junit:junit:4.12"
+    testCompile "org.hamcrest:hamcrest-all"
+    testCompile "org.mockito:mockito-all"
+    testCompile "org.codehaus.groovy:groovy-all"
```
adrianjgeorge: Seems like you moved to scala test, why is this and spock still present?

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/build.gradle
```diff
@@ -0,0 +1,33 @@
+dependencies {
+    compile project(":repose-aggregator:commons:commons-utilities")
+    compile project(":repose-aggregator:core:repose-core-api")
+    compile project(":repose-aggregator:components:services:opentracing-service:opentracing-service-api")
+    compile "org.slf4j:slf4j-api"
+    compile "org.slf4j:jcl-over-slf4j"
+    compile "org.springframework:spring-beans"
+    compile "io.opentracing:opentracing-api:0.31.0"
```
adrianjgeorge: Shouldn't be specifying versions in any of your build files

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+    testCompile "org.codehaus.groovy:groovy-all"
+    testCompile "org.eclipse.jetty:jetty-server"
+    testCompile "xerces:xerces-xsd11"
+    testCompile "org.spockframework:spock-core"
+    testCompile "org.springframework:spring-test:4.1.4.RELEASE"
+    testCompile "org.scalatest:scalatest_$scalaMajDotMin"
+    testCompile "io.opentracing:opentracing-mock:0.31.0"
+    testCompile "org.apache.logging.log4j:log4j-slf4j-impl"
+    testCompile "org.apache.logging.log4j:log4j-core"
+    testCompile group: 'org.apache.logging.log4j', name: 'log4j-core', classifier: 'tests'
```
adrianjgeorge: Are you checking for log messages in some of your tests?

.gitignore
adrianjgeorge: Reasonable addition then

repose-aggregator/components/filters/cors-filter/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
dimtruck: jaeger-apachehttpclient (https://github.com/jaegertracing/jaeger-client-java/blob/v0.24.0/jaeger-apachehttpclient/build.gradle#L5-L7)

build.gradle
dimtruck: EEK! sorry about that.  Will update.

repose-aggregator/components/services/opentracing-service/opentracing-service-api/src/main/resources/META-INF/schema/config/bindings.xjb
dimtruck: I... don't know what that means.  What's a different between a node service and a core service and where should one belong over the other?

repose-aggregator/components/services/opentracing-service/opentracing-service-api/src/main/resources/META-INF/schema/config/opentracing.xsd
dimtruck: it's not hyphenated in their docs so I kept it the same.

repose-aggregator/components/services/opentracing-service/opentracing-service-api/src/test/scala/org/openrepose/core/services/opentracing/config/OpenTracingSchemaTest.scala
dimtruck: So this was originally added to get coverage on schema tests (similarly to AtomHopperSchemaTest).  Definitely not something that's required.

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/build.gradle
dimtruck: For some reason, my tests could not find the API.  I'll remove it and re-run.

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/build.gradle
dimtruck: copy/pasta.  Will remove.

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/build.gradle
dimtruck: good catch. Removing.

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/build.gradle
dimtruck: yup!  good catch.  I specified it already in the properties so I'll remove all this.

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/build.gradle
dimtruck: I removed those tests but not the dependency.  Cleaning up.

repose-aggregator/components/filters/cors-filter/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
dimtruck: Mainly this in 4.5.3 - https://github.com/apache/httpcomponents-client/blob/4.5.3/httpclient/src/main/java/org/apache/http/client/utils/URIBuilder.java#L493-L497 vs 4.4.1 - https://github.com/apache/httpcomponents-client/blob/4.4.1/httpclient/src/main/java/org/apache/http/client/utils/URIBuilder.java#L496-L499

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/src/main/scala/org/openrepose/core/services/opentracing/TracerExtractor.scala
```diff
@@ -0,0 +1,105 @@
...
+          }) {
+            valuesList += valuesIterator.nextElement
+          }
+          headersResult += (headerName -> valuesList.toList)
+        }
+      }
+
+      headersResult.toMap
+
+    }
```
dmnjohns: You made this way more complex than it needs to be. Since you're taking a wrapped request anyway, use the convenience methods it provides. Also, leverage the power of Scala! 
```
  def servletHeadersToMultiMap(httpServletRequest: HttpServletRequestWrapper): Map[String, List[String]] = {
    Option(httpServletRequest).map(
      _.getHeaderNamesScala
        .map(headerName => headerName -> httpServletRequest.getHeadersScala(headerName))
        .toMap
    ).getOrElse(Map.empty)
  }
```

dmnjohns: And really, if you do that, you don't even need the nested/local method.

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/src/main/scala/org/openrepose/core/services/opentracing/TracerExtractor.scala
```diff
@@ -0,0 +1,105 @@
...
+import java.util
+import java.util.AbstractMap.SimpleImmutableEntry
+
+import com.typesafe.scalalogging.slf4j.LazyLogging
+import io.opentracing.propagation.TextMap
+import org.openrepose.commons.utils.servlet.http.HttpServletRequestWrapper
+
+import scala.collection.mutable.ListBuffer
+
+class TracerExtractor() extends TextMap with LazyLogging {
```
dmnjohns: Why have a default constructor that you never use?

Just make this class signature:
```
class TracerExtractor(httpServletRequestWrapper: HttpServletRequestWrapper) extends TextMap with LazyLogging {
```

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/src/main/scala/org/openrepose/core/services/opentracing/TracerExtractor.scala
```diff
@@ -0,0 +1,105 @@
...
+  var headers: Map[String, List[String]] = _
```
dmnjohns: This variable initializes to `null`. One again, I'm not sure why. It's also a variable that you never reassign, so why make it a variable at all? I would compress this and your constructor into something like:
```
val headers: Map[String, List[String]] = Option(httpServletRequestWrapper).map(request =>
    request.getHeaderNamesScala
      .map(headerName => headerName -> request.getHeadersScala(headerName))
      .toMap
  ).getOrElse(Map.empty)
```

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/src/main/scala/org/openrepose/core/services/opentracing/TracerExtractor.scala
```diff
@@ -0,0 +1,105 @@
...
+      }
+
+      headersResult.toMap
+
+    }
+  }
+
+  override def put(key: String, value: String): Unit = ???
+
+  override def iterator() = new MultivaluedMapFlatIterator[String, String](this.headers.toSet)
```
dmnjohns: You can drop the `Iterator` class and just transform the headers map instead. I'll just rewrite the rest of this class and paste the whole thing here in a bit.

repose-aggregator/components/services/opentracing-service/opentracing-service-impl/src/main/scala/org/openrepose/core/services/opentracing/TracerExtractor.scala
```diff
@@ -0,0 +1,105 @@
...
+        listIterator = mapEntry._2.iterator
+        println(s"listIterator $listIterator")
+      }
+      println(s"listIterator 2 ${listIterator.hasNext}")
+      if (listIterator.hasNext) new SimpleImmutableEntry[K, V](mapEntry._1, listIterator.next)
+      else new SimpleImmutableEntry[K, V](mapEntry._1, null)
+    }
+
+  }
+}
```
dmnjohns: Here's my rewrite of your class!
```
class TracerExtractor(httpServletRequestWrapper: HttpServletRequestWrapper) extends TextMap with LazyLogging {

  val headers: Map[String, List[String]] = Option(httpServletRequestWrapper).map(request =>
    request.getHeaderNamesScala
      .map(headerName => headerName -> request.getHeadersScala(headerName))
      .toMap
  ).getOrElse(Map.empty)

  override def put(key: String, value: String): Unit = ???

  override def iterator(): util.Iterator[util.Map.Entry[String, String]] = {
    headers.toIterator
      .flatMap(tupled((key, values) => values.map(key -> _)))
      .map(tupled((key, value) => new util.AbstractMap.SimpleEntry(key, value)))
      .asJava
      .asInstanceOf[util.Iterator[util.Map.Entry[String, String]]]
  }
}
```

dmnjohns: As a side note, that `.asInstanceOf` line was annoying. Despite `SimpleEntry` being an implementation of the `Map.Entry` interface, the typing is not correct because:
> Note: java.util.AbstractMap.SimpleEntry[String,String] <: java.util.Map.Entry[String,String], but Java-defined trait Iterator is invariant in type E.
> You may wish to investigate a wildcard type such as `_ <: java.util.Map.Entry[String,String]`. (SLS 3.2.10)

PR #1860, opened 2018-02-05
REP-6447 Support multi-tenant in Keystone/Auth-Z

dmnjohns:


wdschei:
Retest this please.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
```diff
@@ -38,30 +38,44 @@ object KeystoneV2Authorization extends LazyLogging {
...
 
-  // TODO: Switch to getRequestTenants (plural) and include all tenants extracted. This is a behavior change.
-  def getRequestTenant(config: ValidateTenantType, request: HttpServletRequestWrapper): String = {
-    Option(config).flatMap(validateTenantConfig =>
-      (validateTenantConfig.getUriExtractionRegexAndHeaderExtractionName.asScala.toStream map {
+  // TODO: Rather than throwing an exception, this method should return a Try.
+  // NOTE: Throwing an exception currently works because this method is always called in a [[Try]] stack.
+  @throws[UnparsableTenantException]
+  def getRequestTenants(config: ValidateTenantType, request: HttpServletRequestWrapper): Set[String] = {
+    Option(config).map(validateTenantConfig =>
```
adrianjgeorge: This option is unnecessary, because you are calling this method on a map of an option of the same config.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
```diff
@@ -71,83 +85,64 @@ object KeystoneV2Authorization extends LazyLogging {
         case _ =>
           logger.error("An unexpected tenant extraction type was encountered")
           None
-      }).find(_.nonEmpty).flatten
-    ).getOrElse(throw UnparsableTenantException("Could not parse tenant from the URI and/or the configured header"))
+      })
```
adrianjgeorge: Mixing lists and optionals is pretty gross

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
```diff
@@ -71,83 +85,64 @@ object KeystoneV2Authorization extends LazyLogging {
...
-          role.tenantId.forall(roleTenantId =>
-            roleTenantId.equals(tenantToMatch)))
-      case _ =>
-        roles
+  // NOTE: Replaces the tenant and roles scoping methods. For the roles scoping, also accounts for tenant prefixes, which was not previously the case.
+  def getScopedTenantToRolesMap(config: ValidateTenantType, tenantsToMatch: Set[String], tenantToRolesMap: TenantToRolesMap): TenantToRolesMap = {
+    val prefixes = Option(config.getStripTokenTenantPrefixes).map(_.split('/')).getOrElse(Array.empty[String])
+    tenantToRolesMap filterKeys { tenant =>
+      tenant == DomainRoleTenantKey || tenantsToMatch.exists(tenantToMatch =>
+        tenant == tenantToMatch || prefixes.exists(prefix =>
```
adrianjgeorge: we do this twice, extract to function?

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
```diff
@@ -71,83 +85,64 @@ object KeystoneV2Authorization extends LazyLogging {
...
-              name = Option(configuredEndpoint.getName),
-              endpointType = Option(configuredEndpoint.getType),
-              region = Option(configuredEndpoint.getRegion)
-            )
-
-          if (userIsPreAuthed || endpoints.vector.exists(_.meetsRequirement(requiredEndpoint))) Success(Unit)
-          else Failure(UnauthorizedEndpointException("User did not have the required endpoint"))
-        }
-      case None => Success(Unit)
+      if (endpoints.vector.exists(_.meetsRequirement(requiredEndpointModel))) Success(Unit)
```
adrianjgeorge: Took me a minute to figure out why a vector was involved. It's serialization.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -238,54 +239,50 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
...
-        case (None, Some(uri)) =>
-          Some(PreferredTenant(uri, uriTenantQuality))
-        case (Some(default), None) =>
-          Some(PreferredTenant(default, defaultTenantQuality))
-        case (None, None) =>
-          if (roleTenants.nonEmpty) Some(PreferredTenant(roleTenants.head, rolesTenantQuality))
-          else None
-      }
+      val preferredTenants =
+        if (matchedTenants.nonEmpty) {
```
adrianjgeorge: I feel like there's a match in here somewhere, but all i can come up with still has these if's

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -238,54 +239,50 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
...
 
       if (sendAllTenants && sendQuality) {
-        val priorityTenants = (defaultTenant, matchedUriTenant) match {
-          case (Some(default), Some(uri)) => Vector(s"$default;q=$defaultTenantQuality", s"$uri;q=$uriTenantQuality")
-          case (Some(default), None) => Vector(s"$default;q=$defaultTenantQuality")
-          case (None, Some(uri)) => Vector(s"$uri;q=$uriTenantQuality")
-          case (None, None) => Vector.empty[String]
-        }
-        priorityTenants ++ roleTenants.map(tid => s"$tid;q=$rolesTenantQuality")
+        val matched = matchedTenants.map(matched => s"$matched;q=$matchedTenantQuality")
```
adrianjgeorge: Ohh, this is gross. Why are we making the header values by hand?

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
dmnjohns: I mean, you could do a match, but you would probably end up using guards... which are just ifs anyway. If/elseif/else branches are actually the simpler construct, and one (of a few) that pattern matching uses behind-the-scenes, so if it makes sense, I tend to prefer if/elseif/else.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
dmnjohns: Totally is.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
dmnjohns: Hah. Yes, I was conflicted about that one. It works because they are both `TraversableOnce`. I actually think an `Option` is a better type for what will only ever be zero or one values, but then it isn't strongly consistent with the collection of zero to many values I need for the header. The thing is, the semantics differ depending on what type of extraction we are performing. 

I can totally switch the `Option` to a `List` if you prefer the consistency.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
dmnjohns: Yarp.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
dmnjohns: So... I thought about this about, but maybe you can come up with a better solution. The problem is, we are doing similar things, but not quite the same.

In one case, we want to scope the map down to just the relevant tenants and roles. So we have to iterate over all of the entries in the tenant-to-roles map to validate the tenant. We also want to keep prefixed tenants that match. Oh, and we also want to keep our special "tenant" for domain roles.

In the other case, we only need to verify that all of the tenants extracted from the request match something. We can't just do a size check between the request tenants and the scoped tenants because of prefixes, so I ended up iterating over the extracted tenants instead. There are two small optimizations here, though. First, we only pass the scoped tenants, so we know each user tenant matches some request tenant. Second, we stop as soon as any user tenant matches a request tenant and move on to the next request tenant.

I suppose I could write more code to track which request tenants matched when scoping the tenant-to-roles map, but then I conflate scoping and authorization, which I didn't want to do because it complicates things a little and the return type gets gross.

Or did you just mean extract the prefixed equality check? I can totally do that.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
dmnjohns: Dangit. Yeah, I changed the way we were handling the config so that every method didn't have to wrap it in an `Option`, but apparently I missed this. Will fix.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/keystonev2/authorizationstandalone/AuthorizationFilterMultiTenantTest.groovy
```diff
@@ -0,0 +1,369 @@
...
+        mapRolesDecode == [tenantId as String, "role3", "repose/domain/roles", "role4"] as Set
+    }
+
+    def "User with Multiple Tenants and invalid URI extraction should receive a 401 UNAUTHORIZED response"() {
+        given: "a configured tenant"
+        def tenantId = Math.abs(random.nextInt())
+
+        when: "User sends a request through repose"
+        MessageChain mc = deproxy.makeRequest(
+            url: """$reposeEndpoint/extract/${Math.abs(random.nextInt())}/ss""",
```
adrianjgeorge: There's a small chance that this test will randomly fail. If `nextInt` gives the same value twice in a row this test fails. Maybe just add one to the first value you got to guarantee there's never a collision?

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/keystonev2/authorizationstandalone/AuthorizationFilterMultiTenantTest.groovy
```diff
@@ -0,0 +1,369 @@
...
+        when: "User sends a request through repose"
+        MessageChain mc = deproxy.makeRequest(
+            url: """$reposeEndpoint/ss""",
+            method: "GET",
+            headers: [
+                "X-Catalog": xCatalog,
+                "X-Map-Roles": base64Encode("""{"tenant1":["role1","role2","role3"],"$tenantId":["role3"],"repose/domain/roles":["role4"]}"""),
+                "X-Tenant-Id": """$tenantId;q=1,tenant1;q=0.5""",
+                "X-Roles": "role1,role2,role3,role4",
+                "X-Expected-Tenant": Math.abs(random.nextInt())])
```
adrianjgeorge: Same problem

adrianjgeorge:
Couple of comments and todos still lingering, that might need cleanup.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
adrianjgeorge: do the semantics differ? they could be doing multiple capture groups now

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
adrianjgeorge: Yeah, that's what i was talking about. the differences between line 97-98 and line 114-115 is a parenthesis at the end.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
adrianjgeorge: That's what i was saying, every match i could come up with just had the if guards, and actually became less readable. I just feel like there is another version of this that somehow doesn't do that, but i can't find it. This is fine as written.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
dmnjohns: Oh, are we supporting multiple capture groups in a single regex now? I was not aware of that, and did not add support for it. If that's the case, then sure, I'll add support and use a `List` here.

adrianjgeorge: I don't see why we shouldn't? It's entirely possible that the multiple tenants we get will all be in the url.

dmnjohns: No, that's totally legit. Since any group can be a non-capturing group, and since we are now supporting multiple tenants, there is no reason not to treat each group as a tenant.

I suppose I just didn't remember talking about that or capturing it in the story, and I didn't really think about it when I was writing this code that kept some of the old behavior despite the fact that we are changing our approach.

I will add this.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
dmnjohns: Done.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
dmnjohns: The answer is that it's what we have been doing since the old auth filters.

However, I'm totally alright with changing it if you would like. I mean, I would like to if there are no objections.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
```diff
@@ -238,54 +240,53 @@ class KeystoneV2Filter @Inject()(configurationService: ConfigurationService,
...
-      val preferredTenant = (defaultTenant, matchedUriTenant) match {
-        case (Some(default), Some(uri)) =>
-          val quality = if (default.equals(uri)) {
-            math.max(defaultTenantQuality, uriTenantQuality)
-          } else {
-            uriTenantQuality
+      val preferredTenants =
+        if (matchedTenants.nonEmpty) {
+          matchedTenants map {
+            case matchedTenant if defaultTenant.exists(matchedTenant.equals) =>
```
dmnjohns: This is the carryover of the default tenant quality override. I just wanted to comment to say that I don't love that we give special treatment to this case. It's not like we always use the highest applicable quality, just when the default tenant ID matches. This leads to more complex behavior which may not always be obvious to a user.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
```diff
@@ -38,116 +38,118 @@ object KeystoneV2Authorization extends LazyLogging {
...
+  // NOTE: Throwing an exception currently works because this method is always called in a [[Try]] stack.
+  @throws[UnparsableTenantException]
+  def getRequestTenants(config: ValidateTenantType, request: HttpServletRequestWrapper): Set[String] = {
+    config.getUriExtractionRegexAndHeaderExtractionName.asScala.toSet[ExtractionType] flatMap {
+      case headerExtraction: HeaderExtractionType =>
+        request.getSplittableHeaderScala(headerExtraction.getValue)
+      case uriExtraction: UriExtractionType =>
+        val uriRegex = uriExtraction.getValue.r
+        request.getRequestURI match {
+          case uriRegex(all @ _*) => all
```
dmnjohns: I want to point out the change here. Instead of only extracting the value from the first capture group, this will now extract all values from all capturing capture groups, and all extracted values must be validated. I am slightly concerned that this is not backwards compatible with current configurations.

Of course, you can make a capturing group non-capturing using the prefix `?:` like so: `(?:hello)`.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
adrianjgeorge: Man, I thought we were using the highest quality that matched. You should flag this behavior for removal in 9.x. It's possible this behavior is needed by someone though. It's also possible it's just a hold over from a different world.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
adrianjgeorge: So that's a valid concern. I'd say revert the behavior back to what it use to be, except the uri extractor filter also only works on the first capture group. We seem to assume that things will be the first capture group in several places. I think I lean towards reverting this behavior back to what it use to be. Deprecating the uri extraction config here if we haven't already, and adding a fast follow story to make uri extract to header filter work on every capture group. We also have a potential backwards compatibility problem there as well, but we can finagle the config if we need to easier there.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala
dmnjohns: Will do.

repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Authorization.scala
dmnjohns: Will do.

dmnjohns:
Whelp. The last run failed, but the run before that passed:
https://jenkins.openrepose.org/job/repose-gradle-pull_request/405/

So everything up to the last commit should be good, and the last commit is just documentation (that failed due to the Scoverage issue).

dmnjohns:
Retest this please.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/string/Base64Helper.scala
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package org.openrepose.commons.utils.string
+
+import java.util.Base64
+import java.nio.charset.StandardCharsets.UTF_8
+
+object Base64Helper {
+  def base64EncodeUtf8(encodeString: String): String = {
+    new String(Base64.getEncoder.encodeToString(encodeString.getBytes(UTF_8)))
```
dmnjohns: `encodeToString` means you don't need to wrap it in `new String(...)`.

wdschei: You would be correct.
Updated.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/string/Base64Helper.scala
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package org.openrepose.commons.utils.string
+
+import java.util.Base64
+import java.nio.charset.StandardCharsets.UTF_8
+
+object Base64Helper {
+  def base64EncodeUtf8(encodeString: String): String = {
+    Base64.getEncoder.encodeToString(encodeString.getBytes(UTF_8))
```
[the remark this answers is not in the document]
dmnjohns: To summarize that, the only thing I would change here is not using the `encodeToString` method and instead doing it the hard way:
```
new String(Base64.getEncoder.encode(encodeString.getBytes(UTF_8)), StandardCharsets.ISO_8859_1)
```

Although, that is _precisely_ what the deprecated `encodeToString` does, so really, we don't _have_ to do anything here. Still, using deprecated methods when we don't need to isn't my favorite.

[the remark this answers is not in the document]
wdschei: The `encodeToString()` method isn't itself deprecated, they just have the `@SuppressWarnings("deprecation")` turned on since they haven't fixed the underlying implementation yet.

PR #1862, opened 2018-02-12
Updated API-Checker from v2.5.1 to v2.6.0.

wdschei:


repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/apivalidator/ApiValidatorMultiTenantTest.groovy
```diff
@@ -0,0 +1,475 @@
...
+    def "Should succeed when appropriate headers are set"() {
+        when:
+        MessageChain mc = deproxy.makeRequest(
+            url: "$reposeEndpoint/v1/resource/other",
+            method: "GET",
+            headers: [
+                "X-Tenant"   : tenants,
+                "X-Other"    : "other",
+                (ROLES)      : "foo,bar",
+                "X-Map-Roles": mapHeaderValue])
```
dmnjohns: This should probably use the constant like `ROLES`. So `OpenStackServiceHeader.TENANT_ROLES_MAP`.

wdschei: I looked for it, but I didn't find it.
Now it makes sense. It is on your branch that isn't on master yet.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/apivalidator/ApiValidatorMultiTenantTest.groovy
```diff
@@ -0,0 +1,475 @@
...
+                "X-Tenant"   : "1,5",
+                "X-Other"    : "other",
+                (ROLES)      : "foo,bar",
+                "X-Map-Roles": mapHeaderValue])
+
+        then:
+        mc.getReceivedResponse().getCode() as Integer == SC_OK
+        mc.handlings.size() == 1
+        List<String> relevantRolesValues = mc.handlings[0].request.headers.findAll(RELEVANT_ROLES)
+        relevantRolesValues.contains("a:admin/{X-TENANT}")
```
dmnjohns: Huh? Relevant roles contain the literal `{X-TENANT}`? Shouldn't that be replaced with the parameterized request tenant?

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/apivalidator/ApiValidatorMultiTenantTest.groovy
```diff
@@ -0,0 +1,475 @@
...
+    }
+
+    @Unroll("Should succeed on #method #url when appropriate headers are set (1, 5)")
+    def "Should succeed when appropriate headers are set (1, 5)"() {
+        when:
+        MessageChain mc = deproxy.makeRequest(
+            url: "$reposeEndpoint$url",
+            method: method,
+            headers: [
+                "X-Tenant"   : "1,5",
```
dmnjohns: So this test shows that the required role must be present on tenant 1 OR 5, right? Rather than AND. Since we pass this filter, but only tenant 5 has role `a:creator` when we `POST` to `/v1/resource/other`.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/apivalidator/ApiValidatorMultiTenantTest.groovy
```diff
@@ -0,0 +1,475 @@
...
+                "X-Other"    : "other",
+                (ROLES)      : "foo,bar",
+                "X-Map-Roles": mapHeaderValue])
```
dmnjohns: This seems kind of odd because the tenants in `X-Tenant` and the roles in `X-Roles` do not match what is in `X-Map-Roles`. It doesn't necessarily matter -- in fact, it probably makes testing easier since it shows where `api-checker` is getting the data it uses from -- but it's not the expected input to `api-checker`, I don't believe. It's similar to the case where we are not authorizing the tenant nor are we sending all tenants, but still not quite the same since we have multiple values of `X-Tenant`.

I think it's fine, but I wanted to point it out and maybe think about it some more.

repose-aggregator/tests/functional-tests/src/integrationTest/configs/features/filters/apivalidator/multitenant/headerAllAtMethodTenantExplicit.wadl
```diff
@@ -0,0 +1,161 @@
...
+                           required="true"
+                           type="xsd:int"
+                           repeating="true"
+                           rax:isTenant="true"/>
+                    <param name="X-TENANT"
+                           style="header"
+                           required="true"
+                           type="xsd:boolean"
+                           repeating="true"
+                           rax:isTenant="true"/>
```
dmnjohns: It may come with the documentation, but what's this boolean header param about?

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/apivalidator/ApiValidatorMultiTenantTest.groovy
dmnjohns: I'm told this is what `api-checker` does. Weird. It seems like if we had the actual tenant we would be able to scope down the request tenants even more in the `TenantCullingFilter`. That is, having the exact relevant role(s)/tenant(s) would be nice data to have.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/apivalidator/ApiValidatorMultiTenantTest.groovy
```diff
@@ -0,0 +1,475 @@
...
+
+        then:
+        mc.getReceivedResponse().getCode() as Integer == SC_OK
+        mc.handlings.size() == 1
+        List<String> relevantRolesValues = mc.handlings[0].request.headers.findAll(RELEVANT_ROLES)
+        relevantRolesValues.isEmpty()
+    }
+
+    @Unroll("Should fail on #method #url when appropriate headers are set but there's no tenant access (#tenants)")
+    def "Should fail on when appropriate headers are set but there's no tenant access"() {
```
dmnjohns: "should fail on when" should probably just be "should fail when". Not that it would ever matter...

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/apivalidator/ApiValidatorMultiTenantTest.groovy
RackerWilliams: @dmnjohns  :  No, it is not replaced with the tenant value -- it is the name of the parameter.  Consider that it may actually be multiple values.  `TenantCullingFilter` shouldn't scope based on the tenant, it should scope based on roles in the relevant roles header. In fact there are use cases that would break if you scoped down to the tenant first -- TenantCulling should just cull based on roles.

repose-aggregator/tests/functional-tests/src/integrationTest/configs/features/filters/apivalidator/multitenant/headerAllAtMethodTenantExplicit.wadl
RackerWilliams: boolean header means that the validator will only allow the value of X-TENANT to be true or false.  The other param only allows ints.  So this is saying the values of X-TENANT should be either ints or boolean values -- will fail if that's not the case.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/apivalidator/ApiValidatorMultiTenantTest.groovy
RackerWilliams: ... also `TenantCullingFilter` should ignore anything after `/` in the `X-RELEVANT-ROLES` header.

repose-aggregator/tests/functional-tests/src/integrationTest/configs/features/filters/apivalidator/multitenant/headerAllAtMethodTenantExplicit.wadl
dmnjohns: Ah, I see. So it's just a way of validating the content. For the test that uses thing config, we only seem to be providing int tenant values, so is the boolean type parameter necessary? I would think not?

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/apivalidator/ApiValidatorMultiTenantTest.groovy
dmnjohns: Right, that is exactly how the `TenantCullingFilter` works right now.

I suppose what I was saying is that while the `X-Tenant` param may be multi-valued, when we get to the point that we have relevant roles, we should know exactly what role(s) on what tenant(s) allowed the request through, right? So we should be able to forward that information along for whatever purpose, such as additional culling. So say I have a wadl with:
```
<method name="POST" rax:roles="a:creator/{X-TENANT}">
               <request>
                   <param name="X-TENANT"
                          style="header"
                          required="true"
                          type="xsd:int"
                          repeating="true"
                          rax:isTenant="true"/>
```

That snippet essentially says that to `POST`, a user must have the role `a:creator` on at least one of the tenants we parse from the `X-Tenant` header, right? Well, if that header is `X-Tenant: tenant1, tenant2, tenant3`, and if the user has `a:creator` on `tenant1` and `tenant2` but not on `tenant3`, then rather than saying relevant roles are `a:creator/{X-TENANT}` could we not say they are `a:creator/tenant1, a:creator/tenant2`? It's just further specification. I don't actually know that such specification would serve a purpose, but...

So in the case with the `TenantCullingFilter`, it wouldn't be scoping based on tenant per se, it would be scoping based on roles knowing specifically what roles were relevant rather than vaguely what roles were relevant. Using the example above, then, we could strip `tenant3` out of the request.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/string/Base64Helper.scala
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package org.openrepose.commons.utils.string
+
+import java.util.Base64
+import java.nio.charset.StandardCharsets.UTF_8
+
+object Base64Helper {
+  def base64EncodeUtf8(encodeString: String): String = {
+    new String(Base64.getEncoder.encodeToString(encodeString.getBytes(UTF_8)))
```
dmnjohns: I assume you'll drop the commit that brings this in and rebase if the other PR gets merged first, in which case this will be a collision. Or, if we keep this one, then you don't need the `new String` again.

wdschei: This was `cherry-pick`'d so should not collide and should be fine either way. The other branch already has the new commit to update this.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/apivalidator/ApiValidatorMultiTenantTest.groovy
wdschei: Yeah, I was just removing the variable parts of the `Unroll` name for the `def` name.

wdschei:
These API Validator tests are derived from this [API Checker test suite](https://github.com/rackerlabs/api-checker/blob/api-checker-2.6.0/core/src/test/scala/com/rackspace/com/papi/components/checker/ValidatorWADLRaxRolesHeaderMultiValueTenantSuite.scala).

The main difference is that only one WADL and one Configuration was used since all the permutations are already over there. This is intended to provide a sanity check to ensure no regression issues in the future.

repose-aggregator/tests/functional-tests/src/integrationTest/configs/features/filters/apivalidator/multitenant/headerAllAtMethodTenantExplicit.wadl
RackerWilliams: I would think not.  I think this is derived from tests on my side and there I did a sanity check on bool vs int.

wdschei: Yes. It is based on this [API Checker test suite](https://github.com/rackerlabs/api-checker/blob/api-checker-2.6.0/core/src/test/scala/com/rackspace/com/papi/components/checker/ValidatorWADLRaxRolesHeaderMultiValueTenantSuite.scala).

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/string/Base64Helper.scala
dmnjohns: Derp. Fair enough. For some reason, I was thinking that commit was modified rather than being two separate commits.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/apivalidator/ApiValidatorMultiTenantTest.groovy
dmnjohns: I thought as much. Not a big deal, but eh.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/string/Base64Helper.scala
wdschei: I thought about it, but I knew I had `cherry-pick`'d it and didn't want to deal with the conflicts unless I had to.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/apivalidator/ApiValidatorMultiTenantTest.groovy
RackerWilliams: Nope, actually in the example above if the header on the request is `X-TENANT: 1, 2, 3` (note `tenant1` doesn't validate against int).  And the user only has `a:creator` on `1` and `2` but not on `3`. Then the request would fail, since one of the tenants is not allowed the post.  The difference between `a:creator` on  Relevant Role header and `a:creator/{X-TENANT}` is simply stated that we made the tenant check.  If you want to allow access based on *any* tenant having the role then you should say `rax:roles="a:creator"` makes sense?

repose-aggregator/tests/functional-tests/src/integrationTest/configs/features/filters/apivalidator/multitenant/headerAllAtMethodTenantExplicit.wadl
dmnjohns: I see. So I'm always a fan of scoping tests to just what we care about because it makes them easier to grok and maintain. Plus it helps isolate the functionality under test. I kind of want to suggest we drop that second parameter for this test suite, even though it is copied from the `api-checker` suite, because we unlike the `api-checker` suite, we don't need it.

But I won't complain too much about that.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/apivalidator/ApiValidatorMultiTenantTest.groovy
dmnjohns: Fair enough, I should have used ints. So it's an *AND* rather than an *OR* for the tenanted role check. That makes way more sense to me from a multi-tenant perspective. I think my confusion arose from the fact that all of the other role requirements seem to be *OR*s, right? That is, `rax:roles="role1 role2"` would require `role1` or `role2`? And same with ancestral role requirements (like `rax:roles="role1"` on `/` and `rax:roles="role2"` on `/foo`)?

Anyway, in that case, we're good. Effectively, `relevantRole/{X-Tenant}` (or whatever the parameter happens to be named) denotes that `relevantRole` was relevant for *ALL* tenants taken from the `X-Tenant` parameter.

RackerWilliams: It's an `OR` for roles and an `AND` for tenants.   Or in other words any one of these roles on all of the tenants.  Soon, we need to do an `AND` for tenants -- I'm not sure what the syntax for that will look like...

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/apivalidator/ApiValidatorMultiTenantTest.groovy
dmnjohns: Nope. It's an `AND`. Both tenants `1` and `5` have the `a:admin` role, so we're good.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/apivalidator/ApiValidatorMultiTenantTest.groovy
wdschei: This header wasn't added in it's own commit, so I can't just `cherry-pick` it.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/apivalidator/ApiValidatorMultiTenantTest.groovy
wdschei: This was stupid naming anyway. I don't know why I did it. They are now fixed.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/apivalidator/ApiValidatorMultiTenantTest.groovy
dmnjohns: Nope. You would have to rebase, I guess. Or I could break that constant out into its own PR...

But since tests just passed on the branch, it should be good to go after one more quick glance form Adrian. So maybe just wait for that? Or don't use the constant, but I'm really not a fan of that.

wdschei: Yep. Sounds good. Hopefully it will be in shortly.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/apivalidator/ApiValidatorMultiTenantTest.groovy
```diff
@@ -124,9 +124,9 @@ class ApiValidatorMultiTenantTest extends ReposeValveTest {
             url: "$reposeEndpoint/v1/resource/other",
             method: "GET",
             headers: [
-                "X-Tenant"   : "1,5",
-                "X-Other"    : "other",
-                "X-Map-Roles": mapHeaderValue])
+                (TENANT_ID)       : "1,5",
+                (HDR_OTHER)       : VAL_OTHER,
+                (TENANT_ROLES_MAP): mapHeaderValue])
```
dmnjohns: Just to show the difference here, you could do something like:
```
headers: ([(TENANT_ID): "1,5"] + standardHeaders - [(ROLES): "foo,bar"]))
```

Although , I'm kind of annoyed that I can't find a way to remove the `ROLES` header by key rather than by entry. Well, there's `remove`, but that mutates the map and returns the removed value, which means it can't be done in a single expression.

PR #1868, opened 2018-02-20
REP-6645 WADL headers should be case insensitive.

wdschei:


dmnjohns:
I'm going to let the tests pass since I don't think we are in a rush to merge this.

As a side note, in IntelliJ CE, you register file extensions to file types. In `Preferences`, navigate to `Editor` > `File Types` and select `XML` from `Recognized File Types`. Then, under `Registered Patterns`, click the `+` symbol and add `*.wadl` to the list. It's not full WADL support (that I assume you get in the paid version of IntelliJ), but it does give you XML support for WADLs.

I don't mind appending `.xml` to the WADL, but I do think `.wadl` is the standard file extension. This is what I found about it in the specification:
https://www.w3.org/Submission/wadl/#x3-330005

wdschei:
Cool. Older versions of Community wouldn't let you add `*.wadl` to the list of XML file extensions. Glad to know this works now. We should add this to the list of things to clean up at some time.

PR #1869, opened 2018-02-20
REP-6550 Valkyrie Filter updates to support qualities on tenant ids

adrianjgeorge:


dmnjohns:
Looks good! I'll review the functional tests once available.

repose-aggregator/components/filters/valkyrie-authorization-filter/src/main/scala/org/openrepose/filters/valkyrieauthorization/ValkyrieAuthorizationFilter.scala
```diff
@@ -338,7 +318,7 @@ class ValkyrieAuthorizationFilter @Inject()(configurationService: ConfigurationS
         case _ =>
           Try(akkaServiceClient.get(cacheKey(callType, transformedTenant, contactId),
             valkyrieServer.getUri + uri,
-            authToken.map(token => Map("X-Auth-Token" -> token)).getOrElse(Map.empty) ++ requestTracingHeader)
+            authToken.map(token => Map(AUTH_TOKEN -> token)).getOrElse(Map.empty) ++ requestTracingHeader)
```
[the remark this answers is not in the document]
adrianjgeorge: I didn't actually remove all the `"X-Auth-Token"` references, their old custom stuff used a header by the same name, in the instances where we are dealing with their auth scheme i left it, i only switched it where the code was referring to a keystone token.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/valkyrie/ValkyrieAuthorizationQualityTest.groovy
```diff
@@ -0,0 +1,176 @@
...
+
+        deproxy = new Deproxy()
+        deproxy.addEndpoint(port: properties.targetPort)
+        deproxy.addEndpoint(port: properties.valkyriePort, name: "valkyrie service", defaultHandler: mockValkyrie.getHandler())
+    }
+
+    def setup() {
+        mockValkyrie.resetCounts()
+        mockValkyrie.resetHandlers()
+        mockValkyrie.resetParameters()
```
dmnjohns: Could you set the `contact_id` to some new, unique value here instead of in the tests? It seems like it's just being used to avoid cache collisions anyway, since the mock will, by default, return a body only containing permissions for the `contact_id` it pulls from the URL.

dmnjohns: Ideally, to keep things isolated, the `contact_id` would not be the same as the `tenant_id`. Not that it really matters, it just lead to me asking why that was always the case, so I'm thinking more isolation might make it more obvious that those values do not have to match.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/valkyrie/ValkyrieAuthorizationQualityTest.groovy
```diff
@@ -0,0 +1,176 @@
...
+        where:
+        [a, b] << [["PUT", "POST", "DELETE", "PATCH"],
+                   [";q=0.98", ""]
+        ].combinations()
+
+        method = a
+        quality = b
+
+    }
+}
```
dmnjohns: I would love to see a test that "fails" (passes, but with `SC_FORBIDDEN`) because the wrong tenant ID is used (as opposed to "failing" because the permission does not grant authorization). That would probably involve writing a replacement Valkyrie handler that would return a the failure template on one tenant ID, and the success template on another tenant ID. Then you can combine templates and qualities and make sure that you get back the right template (indirectly, via inspection of the response) for each request.

adrianjgeorge: why would that happen? all we do is get device permissions and make decisions on that?

dmnjohns: Right, so if the wrong tenant ID value was selected from the header, we might get an error response back from Valkyrie? Because the tenant we were requesting permissions for is not in Valkyrie or whatever the case may be. Basically, I just want a case capturing what happens when the wrong tenant ID is selected.

adrianjgeorge: pretty sure there are already existing tests for that, from before this change.

dmnjohns:
Needs a rebase to fix the release notes conflict. Or I can do it when merging.

dmnjohns:
Retest this please

dmnjohns:
While a failure was reported, it was a failure in reporting and not in the build or tests themselves. As such, I am going to merge this now.

PR #1870, opened 2018-02-21
REP-6448 Multi-Tenant RBAC

wdschei:


repose-aggregator/tests/functional-tests/src/integrationTest/configs/features/filters/simplerbac/multitenant/simple-rbac.cfg.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<simple-rbac xmlns="http://docs.openrepose.org/repose/simple-rbac/v1.0"
+             roles-header-name="X-Roles"
+             tenants-header-name="X-Tenant-Id"
```
dmnjohns: Does this need to be updated to case-match the roles below? I think I saw this being straight injected into the WADL, so I assume so.

repose-aggregator/tests/performance-tests/roles/gatling/files/simulations/filters/simplerbac/SimpleRbacFilterSimulation.scala
```diff
@@ -0,0 +1,235 @@
...
+
+  var feederGet: Array[Map[String, Any]] =
+    (0 until ScenariosGet.length).toArray map (i => {
+      val (path, roles, respcode) = ScenariosGet(i)
+      Map(
+        "getPath" -> path,
+        "getRoles" -> roles,
+        "getRespCode" -> respcode
+      )
+    })
```
dmnjohns:
```
ScenariosGet.toArray map { scenario =>
  val (path, roles, respcode) = scenario
  Map(
    "getPath" -> path,
    "getRoles" -> roles,
    "getRespCode" -> respcode
  )
}
```

Also, if a `Feeder` needs to be an `Array`, why not create them as such rather than creating them as `Seq`s and wrapping them as `Array`s?

repose-aggregator/tests/functional-tests/src/integrationTest/configs/features/filters/simplerbac/multitenant/simple-rbac.cfg.xml
wdschei: Good catch. Updating now.

repose-aggregator/tests/performance-tests/roles/gatling/files/simulations/filters/apivalidator/ApiValidatorFilterSimulation.scala
```diff
@@ -0,0 +1,232 @@
...
+  val percentSuccessfulRequest = conf.getInt(s"$confRoot.expectations.percent_successful_requests")
+
+  // this value is provided through a Java property on the command line when Gatling is run
+  val baseUrl = conf.getString("test.base_url")
+
+  val httpConf = http.baseURL(s"http://$baseUrl")
+
+  var feederGet: Array[Map[String, Any]] =
+    ScenariosGet map { scenario =>
+      val (path, roles, respcode) = scenario
```
dmnjohns: Heh. I just realized that this is silly. May as well use the power of pattern matching.
```
ScenariosGet map { case (path, roles, respcode) =>
  Map(
    ...
  )
}
```

repose-aggregator/tests/performance-tests/roles/gatling/files/simulations/filters/simplerbac/SimpleRbacFilterSimulation.scala
wdschei: True. Updated.

repose-aggregator/tests/performance-tests/roles/gatling/files/simulations/filters/apivalidator/ApiValidatorFilterSimulation.scala
wdschei: Sure. Updated.

repose-aggregator/components/filters/simple-rbac-filter/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
```diff
@@ -225,12 +225,34 @@ class SimpleRbacFilter @Inject()(configurationService: ConfigurationService,
...
+          case Some(tenantsHeader) =>
+            s""">
+               |        <request>
+               |          <param name="$tenantsHeader"
+               |                 style="header"
+               |                 required="true"
+               |                 type="xsd:int"
+               |                 repeating="true"
+               |                 rax:isTenant="true"/>
+               |          <param name="$tenantsHeader"
```
adrianjgeorge: I'm confused, this seems to be a repeat other than changing the type to boolean?

repose-aggregator/components/filters/simple-rbac-filter/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
```diff
@@ -263,10 +285,11 @@ class SimpleRbacFilter @Inject()(configurationService: ConfigurationService,
...
         }.mkString("\n")
-        val footer = s"""  </resources>
+        val footer = s"""
+                        |  </resources>
                         |</application>""".stripMargin
 
-        Some(s"$header\n$resources\n$footer")
+        Some(s"$header$resources$footer")
     }
   }
```
adrianjgeorge: I thought we were just going to add this once at the top, it looks like you are adding it basically everywhere?

repose-aggregator/components/filters/simple-rbac-filter/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: Yes. It says the header can contain either an `int` or a `boolean` value. This is actually what caused the error to appear in the API-Checker tests. It hit a corner case where the case of the `name` attribute was different and caused weird things to happen. There are now tests in API-Checker and this proves as long as they are the same case that it still works.

repose-aggregator/components/filters/simple-rbac-filter/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: You have to add it everywhere because of WADL.

repose-aggregator/src/docs/asciidoc/recipes/role-based-access-control.adoc
```diff
@@ -206,6 +206,88 @@ Please refer to the <<../filters/api-validator.adoc#,API Validator filter>> docu
...
+ 2+| Forbidden           2+^| 403            4+| A requested resource or method requires a specific X-Roles header and that header is not found.
+ 2+| Method Not Allowed  2+^| 405            4+| The URI is valid, but the method is not appropriate for the URI.
+|===
+
+[NOTE]
+====
+The case of the actual header does not matter, but the case of the `rax:roles` tenant must match the case of the `param` element's `name` attribute.
+====
+
+The difference between the regular RAX-Roles and the Tenanted RAX-Roles is subtle from a configuration standpoint, but has
```
adrianjgeorge: but has what?

wdschei: D'oh.

repose-aggregator/components/filters/simple-rbac-filter/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
adrianjgeorge: when would it have a boolean?

repose-aggregator/components/filters/simple-rbac-filter/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
adrianjgeorge: to be clear i was talking about the new param, if it is needed everywhere...le sigh

repose-aggregator/components/filters/simple-rbac-filter/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: It isn't that it should, it is that it is legal by the WADL standard so it proves we don't blow up.

repose-aggregator/components/filters/simple-rbac-filter/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: Yep. And, yep.

wdschei: The issue it that a `param` can be a direct sub of a `resource`, but a `request` can't be. A `request` must be a sub of a `method`.

repose-aggregator/components/filters/simple-rbac-filter/src/main/scala/org/openrepose/filters/simplerbac/SimpleRbacFilter.scala
wdschei: We are validating that API-Checker will handle a valid WADL. It is valid in a WADL to define that a particular header will have a value of either type x or type y. The `rax:isTenant` extension just enforces this capability. I could have used a non-standard type like Positive Integers and/or UUID's. The `int`'s and `boolean`'s happen to be standard types that I didn't have to define.

wdschei: On top of that, that is the corner case scenario that caused the bug in API-Checker to surface.

adrianjgeorge: this is the prod code that generates the wadl we feed to api checker right? why would we ever say a boolean is a valid value in our wadl. although now you bring up an excellent point that int is too restrictive. we can totally have tenant id's with other values, and in fact do have a bunch `hybrid:...` and there will be more like that in the future.

wdschei: Crap. I had the wrong context in my head.

wdschei: You're correct. This should only need the one. Do we make it a `String`?

wdschei:
The failure is expected due to an assumption during the writing of this test. The casing issue has been corrected in the (REP-6658_ParamCaseFix)[https://repose.atlassian.net/browse/REP-6658] branch and it passes there.

PR #1871, opened 2018-02-23
REP-6658 Updated the API Validator to account for Param case fixes in API-Checker.

wdschei:


wdschei:
The final of this will require the next API-Checker release.

Yes arbitrary casing is supported, but the `param` and `rax:roles` casing must match.

I'll update to show this.

wdschei:
When I went back to look at updating the case so that each resource had a different case with the `param` element's `name` attribute and the `rax:roles` embedded value being the same case, there is a snag.

The current test has an overarching `admin` role that has to be the same case as each of the sub-resources.

* https://github.com/rackerlabs/repose/blob/2e8eb684d50fef3e1acc9aff4bf34996aeb84ff4/repose-aggregator/tests/functional-tests/src/integrationTest/configs/features/filters/apivalidator/multitenant/headerAllAtMethodTenantExplicit.wadl.xml#L6

This in turn makes all of them need to be the same case.

We could update this test to not account for that `admin` role, but then we loose that portion of the test too. So we could make another test that basically strips out the `admin` role and acts accordingly. Thoughts?

dmnjohns:
Ah, I see. We probably want to continue verifying that hierarchical role access works when we are dealing with multiple tenants. So I would propose either making another test as you suggested, or just creating new resources on this test that do not fall into the existing hierarchy, and thus, are able to have deviant casing. I think we get the coverage we are looking for either way, so whatever seems clearer to you.

PR #1872, opened 2018-02-23
REP-6588 Fixing commitToResponse Issue with the Catalina Response and Response Wrapper Improvements

dmnjohns:


repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -75,7 +76,8 @@ import scala.io.Source
   *                            depending on mode selection of the body
   */
 class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMode: ResponseMode, bodyMode: ResponseMode, desiredOutputStream: ServletOutputStream)
-  extends javax.servlet.http.HttpServletResponseWrapper(originalResponse) with HeaderInteractor {
+  extends javax.servlet.http.HttpServletResponseWrapper(originalResponse) with HeaderInteractor with LazyLogging {
```
adrianjgeorge: I think we've settled into putting each of these on their own line with the `with`s being further tabbed in.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -85,6 +87,7 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
     case ResponseMode.MUTABLE => new MutableServletOutputStream(desiredOutputStream)
   }
 
+  private var statusCode: Int = originalResponse.getStatus
```
adrianjgeorge: Oh god, i understand why, but i suspect we are going to get some super weird behavior out of this.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: For whatever reason, I never got in the habit of doing that, but it makes sense. Oh, and the Scala style guide agrees:
https://docs.scala-lang.org/style/declarations.html#classes

Will update.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: I hope not. It does kind of depend on the container, but between copying the wrapped response's behavior and always defaulting to our own code, I thought this was the better idea. That is, if no one has called either `setStatus` or `sendError`, we probably shouldn't modify the response code, but we should report it accurately. So this does that.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -105,21 +113,31 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
...
+    logger.trace(s"setStatus called with status code: $i")
+    setStatusCommon(i)(originalResponse.setStatus(i))
   }
 
   override def setStatus(i: Int, s: String): Unit = {
+    logger.trace(s"setStatus called with status code: $i, message: $s")
+    setStatusCommon(i, s)(originalResponse.setStatus(i, s))
+  }
+
+  private def setStatusCommon(i: Int, s: String = null)(setStatusFunc: => Unit): Unit = {
```
adrianjgeorge: Don't really like this method name. I'm assuming you can't just further overload it.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -161,40 +184,74 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
 
     // If we are not in a mutable mode, immediately send error to the wrapped response.
     if (headerMode != ResponseMode.MUTABLE && bodyMode != ResponseMode.MUTABLE) {
-      originalResponse.sendError(i, s)
+      sendErrorFunc
     }
   }
 
-  def isError: Boolean = sentError
+  def isError: Boolean = {
```
adrianjgeorge: I don't know that there is a lot of value in adding logging into what is effectively getters and setters.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
```diff
@@ -161,40 +184,74 @@ class HttpServletResponseWrapper(originalResponse: HttpServletResponse, headerMo
...
+  }
 
-  override def isCommitted: Boolean = super.isCommitted || committed
+  override def isCommitted: Boolean = {
+    logger.trace("isCommitted called")
+    super.isCommitted || committed
+  }
 
-  override def getResponse: ServletResponse = throw new UnsupportedOperationException("getResponse is not supported")
+  override def getResponse: ServletResponse = {
```
adrianjgeorge: What are these methods? I'm assuming they are part of the contract, but what are they supposed to do?

repose-aggregator/core/repose-core/src/test/scala/org/openrepose/core/spring/NextVersionTimebombTest.scala
```diff
@@ -77,6 +77,8 @@ class NextVersionTimebombTest extends FunSpec with Matchers with TestFilterBundl
        *     Before doing so, verify that this behavior is not useful.
        *
        * 11. Remove the population of X-Auth-Token-Key from Keystone v2 Filter.
+       *
+       * 12. Allow Jetty (in a Valve deployment) to handle errors when sendError is called on the response.
```
adrianjgeorge: I'm not certain we want to do that actually. I think we may want to change the behavior of an empty body, but I don't think we ever want valve to give an html response.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/wrappers/response/GlassfishResponseWrapperTest.groovy
```diff
@@ -0,0 +1,100 @@
...
+        glassfishLauncher.start()
+        glassfishLauncher.waitForDesiredResponseCodeFromUrl(reposeEndpoint, 200)
+    }
+
+    def cleanupSpec() {
+        deproxy?.shutdown()
+        glassfishLauncher?.stop()
+    }
+
+    def "a filter writes a header and response body then calls sendError on the response"() {
```
adrianjgeorge: I'm assuming that this test failed before the change?

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/wrappers/response/GlassfishResponseWrapperTest.groovy
```diff
@@ -0,0 +1,100 @@
...
+
+        and: "the response body should be the container-server error page"
+        (messageChain.receivedResponse.body as String).length() != 0
+        messageChain.receivedResponse.body as String != ERROR_BODY
+
+        and: "the header should be written"
+        messageChain.receivedResponse.headers.findAll(ERROR_HEADER_NAME).size() == 1
+    }
+
+    def getReposeEndpoint() {
```
adrianjgeorge: why?

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/wrappers/response/TomcatResponseWrapperTest.groovy
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    def "a filter writes a header and response body then calls sendError on the response"() {
+        when: "a request is made to Repose"
+        def messageChain = deproxy.makeRequest(reposeEndpoint + "/scripting")
+
+        then: "the response code should be the error code, not a 500"
+        messageChain.receivedResponse.code as Integer == ERROR_CODE
+
+        // TODO: For some reason, Tomcat is not sending any reason phrase in this case
```
adrianjgeorge: I don't know that it is strictly part of the spec, so maybe they chose to throw it away?

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/wrappers/response/ResponseWrapperTest.groovy
```diff
@@ -274,6 +274,38 @@ class ResponseWrapperTest extends ReposeValveTest {
...
+        when: "any request is made"
+        MessageChain messageChain = deproxy.makeRequest(url: reposeEndpoint, method: "GET", headers: headers)
+
+        then: "the origin service does not receive the request"
+        messageChain.handlings.isEmpty()
+
+        and: "the response code is correctly set"
+        messageChain.receivedResponse.code as Integer == responseCode
+
+        and: "the response received by the user should be empty correct content type"
```
wdschei: `correct` => `without a`

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: These methods are inherited from the `HttpServletResponseWrapper` class. They allow the user of the wrapper to get and set the underlying response. Obviously, that does not work out very well when we are keeping state in the wrapper. I imagine that is the reason we disabled the methods here.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/wrappers/response/GlassfishResponseWrapperTest.groovy
wdschei: This is weird and not used.

dmnjohns: Oh, it's totally used, it's just used via Groovy magic. Instead of calling `getRequestEndpoint`, this is called everywhere you see `reposeEndpoint` as a variable.

As for why, it was just to be consistent with the `ReposeValveTest` which does the same thing. Even though I am not extending `ReposeValveTest` (because it looks like container support started to get added to that, but was never fully added... and I'm not sure why it was started to begin with), I wanted the test to be familiar. I suspect this was originally added to `ReposeValveTest` as a minor convenience.

https://github.com/rackerlabs/repose/blob/master/repose-aggregator/tests/functional-test-framework/src/main/groovy/org/openrepose/framework/test/ReposeValveTest.groovy#L150-L173

dmnjohns: That all being said, I'm fine with inlining this method if desired.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/wrappers/response/TomcatResponseWrapperTest.groovy
dmnjohns: Well that would explain not writing _our_ reason phrase, but it seems very odd to me that _no_ reason phrase is written.

Apparently, however, that may be allowable by HTTP specification. The relevant ABNF is:
```
Status-Line = HTTP-Version SP Status-Code SP Reason-Phrase CRLF
Status-Code    =
            "100"  ; Section 10.1.1: Continue
          | "101"  ; Section 10.1.2: Switching Protocols
          | "200"  ; Section 10.2.1: OK
          | "201"  ; Section 10.2.2: Created
          | "202"  ; Section 10.2.3: Accepted
          | "203"  ; Section 10.2.4: Non-Authoritative Information
          | "204"  ; Section 10.2.5: No Content
          | "205"  ; Section 10.2.6: Reset Content
          | "206"  ; Section 10.2.7: Partial Content
          | "300"  ; Section 10.3.1: Multiple Choices
          | "301"  ; Section 10.3.2: Moved Permanently
          | "302"  ; Section 10.3.3: Found
          | "303"  ; Section 10.3.4: See Other
          | "304"  ; Section 10.3.5: Not Modified
          | "305"  ; Section 10.3.6: Use Proxy
          | "307"  ; Section 10.3.8: Temporary Redirect
          | "400"  ; Section 10.4.1: Bad Request
          | "401"  ; Section 10.4.2: Unauthorized
          | "402"  ; Section 10.4.3: Payment Required
          | "403"  ; Section 10.4.4: Forbidden
          | "404"  ; Section 10.4.5: Not Found
          | "405"  ; Section 10.4.6: Method Not Allowed
          | "406"  ; Section 10.4.7: Not Acceptable
          | "407"  ; Section 10.4.8: Proxy Authentication Required
          | "408"  ; Section 10.4.9: Request Time-out
          | "409"  ; Section 10.4.10: Conflict
          | "410"  ; Section 10.4.11: Gone
          | "411"  ; Section 10.4.12: Length Required
          | "412"  ; Section 10.4.13: Precondition Failed
          | "413"  ; Section 10.4.14: Request Entity Too Large
          | "414"  ; Section 10.4.15: Request-URI Too Large
          | "415"  ; Section 10.4.16: Unsupported Media Type
          | "416"  ; Section 10.4.17: Requested range not satisfiable
          | "417"  ; Section 10.4.18: Expectation Failed
          | "500"  ; Section 10.5.1: Internal Server Error
          | "501"  ; Section 10.5.2: Not Implemented
          | "502"  ; Section 10.5.3: Bad Gateway
          | "503"  ; Section 10.5.4: Service Unavailable
          | "504"  ; Section 10.5.5: Gateway Time-out
          | "505"  ; Section 10.5.6: HTTP Version not supported
          | extension-code
      extension-code = 3DIGIT
      Reason-Phrase  = *<TEXT, excluding CR, LF>
```

dmnjohns: Well anyway, since this is known behavior, I'll update this test to reflect that and drop the TODO, at least.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/wrappers/response/GlassfishResponseWrapperTest.groovy
wdschei: Oh yes. The Groovy magic. I'm down with the inlining as well.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/wrappers/response/GlassfishResponseWrapperTest.groovy
dmnjohns: Correct.

repose-aggregator/core/repose-core/src/test/scala/org/openrepose/core/spring/NextVersionTimebombTest.scala
dmnjohns: Fair enough, I'll drop this note.

dmnjohns:
As far as the logging goes, the trace logging on every method is similar to what we did in the Keystone v2 filter. My thinking was that the trace logging provides an audit for interactions with the response, which can come in handy if we get into a bad state, but we don't get a stacktrace or anything including the call that put us in a bad state. That is, the trace logging might provide more insight for cases where everything technically works, but we aren't seeing the behavior we would expect to see.

Most of the behavior in the wrapper and its streams is actually pretty simple and straightforward. Things start to get really complicated when you look at the interactions between the wrapper and all of the components that interact with it. That is, the majority of our methods are one-line or close to one-line implementations, and the complexity comes from emergent behavior.

With that said, do you still want me to remove the per-method trace logging? I can, of course, still add more logging to capture everywhere the response would be committed, and the few places where there is complexity. Otherwise, I'll add that logging and remove the traces. Either way, I will add logging to the streams.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: I could overload `setStatus` further, I suppose. Although I probably am not allowed to use currying if I do, because I expect that makes this method indistinguishable from the other `setStatus` methods. (Just tested, and that's totally true)

Interestingly, IntelliJ complains when I try to overload `setStatus` here for some reason. It wants me to use the `override` modifier, but this method clearly does not override any other method. And, in fact, the code only compiles *without* the `override` modifier.

So, would you be happier with this?:
```
override def setStatus(i: Int): Unit = {
    logger.trace(s"setStatus called with status code: $i")
    setStatus(i, setStatusFunc = originalResponse.setStatus(i))
  }

  override def setStatus(i: Int, s: String): Unit = {
    logger.trace(s"setStatus called with status code: $i, message: $s")
    setStatus(i, s, originalResponse.setStatus(i, s))
  }

  private def setStatus(i: Int, s: String = null, setStatusFunc: => Unit): Unit = {
    if (isCommitted) {
      throw new IllegalStateException("Cannot call setStatus after the response has been committed")
    }

    statusCode = i
    reason = Option(s)

    if (headerMode != ResponseMode.MUTABLE && bodyMode != ResponseMode.MUTABLE) {
      setStatusFunc
    }
  }
```

Or perhaps with something slightly less obscure:
```
override def setStatus(i: Int): Unit = {
    logger.trace(s"setStatus called with status code: $i")
    setStatus(originalResponse.setStatus(i), i)
  }

  override def setStatus(i: Int, s: String): Unit = {
    logger.trace(s"setStatus called with status code: $i, message: $s")
    setStatus(originalResponse.setStatus(i, s), i, s)
  }

  private def setStatus(setStatusFunc: => Unit, i: Int, s: String = null): Unit = {
    if (isCommitted) {
      throw new IllegalStateException("Cannot call setStatus after the response has been committed")
    }

    statusCode = i
    reason = Option(s)

    if (headerMode != ResponseMode.MUTABLE && bodyMode != ResponseMode.MUTABLE) {
      setStatusFunc
    }
  }
```

Both of those also work, but personally, I find what I currently have to be the clearest and most flexible expression of this method. I could also just change the name to something else.

adrianjgeorge: Yeah, i'm mostly ok with the implicit usage, i just don't like the name. maybe `doSetStatus` that's a sort of standard for this type of thing.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/wrappers/response/GlassfishResponseWrapperTest.groovy
adrianjgeorge: yeah, inlining makes things clearer here

adrianjgeorge:
Having logging on every method still feels a bit excessive to me, although I'm not that concerned. Having individual log messages on every variant of `addHeader` seems pointless to me, especially given that they all just chain to the one call. I can't think of a single instance where i care which one was called, but there may be value in knowing that it was called at all in rare cases. I think the place we most care about the logging is everything that deals with commiting and the commited  state of the response, those are the places where things get weird, most of the other methods are super straight forward.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/servlet/http/HttpServletResponseWrapper.scala
dmnjohns: Sure, I'll update the method names to follow that, then.

Oh, and just to be clear, no implicits are being used here, just multiple parameter lists. I know this is the pattern followed when implicits are used, but I wanted to call that out so that no one thinks there is magic happening in this case.

dmnjohns:
With all of that in mind, I took a new approach to logging here. Hopefully I was consistent, because I was kind of developing the approach as I went:

- [Trace] Log any time state is modified. This will help track down emergent behavior. For call chains, logging was only added at the end of the chain.
- [Debug] Log any time the underlying response is modified. This goes with the previous bullet, but for non-local state.
- [Error] Log any time an error case is hit. This will help isolate issues that would otherwise be suppressed (e.g., by dropping an exception).

adrianjgeorge:
Seems like your taking of the responses status code made a million tests fail.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/proxy/MisbehavingOriginTest.groovy
```diff
@@ -51,15 +51,15 @@ class MisbehavingOriginTest extends ReposeValveTest {
...
         given: "something is going to do bad things"
 
         when: "Request goes through repose"
         MessageChain mc = deproxy.makeRequest(url: reposeEndpoint)
 
-        then: "repose should return a 500 Internal Server Error"
+        then: "repose should return a 503 Service Unavailable"
         mc.receivedResponse != null
-        mc.receivedResponse.code == "500"
+        mc.receivedResponse.code == "503"
```
dmnjohns: We explicitly handle the `IOException` that is thrown as a result of the server closing the socket unexpectedly by setting the response code to `503`. I'm not sure why we were seeing `500` before, but is my expectation looking at the code. Although, I do wonder if this test should wait to close the server socket until the end of the test rather than immediately after writing some bytes on the wire. That would probably change the behavior in Repose... but that is neither here nor there.

dmnjohns: Looking at this again, it appears that the `500` is expected when an `IOException` isn't being thrown. My theory is that we hit that case if Apache HttpClient processes the garbage response before the test closes the connection, otherwise the closed connection leads to an `IOException` which results in a `503` instead of a `500`.

dmnjohns: Whelp. Waiting to close the socket leads to Apache HttpClient throwing a `SocketTimeoutException` instead of a `ConnectionResetException`. Both are `IOException`s, though, and should result in a `503`. So much for the theory that HttpClient was handling the response leading to `500`s instead. So I'm still not sure where the `500` or `200` are coming from.

dmnjohns: So yeah. When run on Jenkins, the log shows that a `org.apache.http.client.ClientProtocolException` is thrown, leading to the expected `500` response code. When run locally, a `SocketException` (Connection Reset) is thrown leading to the expected `503`. That is when checking out the exact same commit, cleaning with both Git and Gradle, and running the same single test.

At this point, I suspect it may be a JDK thing. I'm running `Java 1.8.0_161 Hotspot` while the Jenkins server is running `Java 1.8.0_131 OpenJDK`.

I'm just going to try and get it to pass on Jenkins.

dmnjohns:
retest this please

dmnjohns:
retest this please

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/proxy/MisbehavingOriginTest.groovy
dmnjohns: The `200` is from the `PowerFilterChain` catching a `URISyntaxException` that occurs sporadically, setting the status code, but not committing the change. Committing the status code change should fix the issue, but I haven't been able to figure out why a `URISyntaxException` ever occurs. For some reason, Repose determines the destination `URI` to be `//` which is not a valid `URI`, however looking at this test and the determination logic, `//` should never be the destination `URI`.

PR #1876, opened 2018-03-06
REP-6654 Instrumenting the Phone Home Service for OpenTracing

dmnjohns:


repose-aggregator/components/services/phone-home-service/src/main/scala/org/openrepose/core/services/phonehome/PhoneHomeService.scala
```diff
@@ -154,14 +156,19 @@ class PhoneHomeService @Inject()(@Value(ReposeSpringProperties.CORE.REPOSE_VERSI
...
       updateMessage
     }
 
     def sendUpdateMessage(collectionUri: String, message: String): Unit = {
       logger.trace("sendUpdateMessage method called")
 
+      val scope = tracer.buildSpan(TracingOperationName)
+        .withTag(Tags.HTTP_URL.getKey, collectionUri)
+        .ignoreActiveSpan()
+        .startActive(true)
```
dmnjohns: I will note that the much (but not all) of the work captured by this span will also be captured by the HttpClient Interceptor span. This span will always be a root span. While ideally this span would also be a parent of the HttpClient span, I do not believe that is guaranteed since the `AkkaServiceClient` handles requests in `Future`s, which are likely to be computed on a different thread making it impossible to access the active span through the `ThreadLocalScopeManager` (although a reference to the parent span can be passed directly, if desired).

I am actually a bit worried about the relationship between spans created in the HttpClient Interceptor and spans created outside of it. Because `ThreadLocal` and `Future`s (really, the `ExecutionContext` of `Future`s). I wonder if we want to make those spans ignore the active span and optionally set a parent span that is passed to the client? Of course, that would require further modification of our take on Jaeger's `TracingRequestInterceptor`.

If we did do such a thing, I kind of question the value of the "phone-home" span. The boundary crossing will be captured by the HttpClient Interceptor span anyway. I guess the operation name might be clearer here than in the Interceptor since it has to be more generic, but even so.

dmnjohns:
This probably needs to be reworked a little bit. I don't think we need to set HTTP tags on the new span here because they will be set by the HttpClient child span anyway. We will also have to pass the `Scope` to the HttpClient once we do the work there to accept a `Scope` rather than relying on the `ThreadLocalScopeManager` which has issues when using `Future`s.

dmnjohns:
To follow-up on my last comment, the solution to manage span relationships in the `HttpClient` does not require any changes to this PR, so we should be good-to-go.

repose-aggregator/components/services/phone-home-service/src/main/scala/org/openrepose/core/services/phonehome/PhoneHomeService.scala
```diff
@@ -186,7 +188,9 @@ class PhoneHomeService @Inject()(@Value(ReposeSpringProperties.CORE.REPOSE_VERSI
           ))
         }
       } catch {
-        case e: Exception => logger.error("Could not send an update to the collection service", e)
+        case e: Exception =>
+          logger.error("Could not send an update to the collection service", e)
+          scope.span.setTag(Tags.ERROR.getKey, true)
```
dmnjohns: I could replace this with:
```
tracer.activeSpan.setTag(...)
```

That would allow us to remove the scope parameter to this method. I kind of liked being explicit, but I would be happy to change this if someone prefers the other way.

wdschei: I like the explicitness of passing the `Scope` in too.

wdschei:
The only failures are the four known issues and the OpenTracing tests that aren't complete.

PR #1877, opened 2018-03-06
Added the new ReposeTracer based on the OpenTracing GlobalTracer.

wdschei:


repose-aggregator/core/repose-core/src/test/scala/org/openrepose/core/opentracing/ReposeTracerTest.scala
```diff
@@ -0,0 +1,99 @@
...
+    it("should allow the registration a Tracer") {
+      ReposeTracer.register(testTracer)
+      ReposeTracer.get should equal(testTracer)
+      ReposeTracer.isRegistered shouldBe true
+    }
+
+    it("should allow the registration a different Tracer") {
+      val testTracerToo = new TestTracer
+      ReposeTracer.register(testTracerToo)
+      testTracer shouldNot equal(testTracerToo)
```
adrianjgeorge: Not sure what the point in this is? testing that two objects you newed up aren't the same? I think you are trying to make sure that you can swap it, but this assertion doesn't do that, and the test itself doesn't either without depending on prior tests already running.

repose-aggregator/core/repose-core/src/test/scala/org/openrepose/core/opentracing/ReposeTracerTest.scala
```diff
@@ -0,0 +1,99 @@
...
+      ReposeTracer.get should equal(testTracerToo)
+      ReposeTracer.isRegistered shouldBe true
+    }
+
+    it("should convert to a readable string") {
+      ReposeTracer.toString().startsWith("ReposeTracer {")
+    }
+  }
+
+  class TestTracer extends Tracer {
```
adrianjgeorge: They probably already have a test tracer you can use.

repose-aggregator/core/repose-core/src/test/scala/org/openrepose/core/opentracing/ReposeTracerTest.scala
wdschei: That specific line is to ensure the two Tracers are in fact not the same. The next line is what ensures the new Tracer was registered. I'll add in another line to make sure the first Tracer was added.

repose-aggregator/core/repose-core/src/test/scala/org/openrepose/core/opentracing/ReposeTracerTest.scala
wdschei: I could extend theirs. I just needed to ensure the hash would be different to prevent the equality from short circuiting things.

adrianjgeorge: does their's give back the same hash? weird.

wdschei: It didn't. I removed this homemade one and moved to their `MockTracer`.

PR #1878, opened 2018-03-06
REP-6654 PowerFilter updates

wdschei:


repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
```diff
@@ -385,42 +387,25 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
+        LOG.debug("Got the span context from request: {}", context);
 
-            LOG.debug("Start a new span {}", scope.span());
+        Tracer.SpanBuilder spanBuilder = tracer.buildSpan(String.format("%s %s", wrappedRequest.getMethod(), wrappedRequest.getRequestURI()));
+        if (context != null) {
+            spanBuilder = spanBuilder.asChildOf(context);
         }
+        Span activeSpan = spanBuilder.withTag(Tags.SPAN_KIND.getKey(), Tags.SPAN_KIND_CLIENT).start();
+
+        Scope scope = tracer.scopeManager().activate(activeSpan, false);
```
dmnjohns: Since you active the span right after creating it, and `activeSpan` is not referenced later, why not just use `startActive(...)` rather than `start()`?

dmnjohns: Also, since every time the `Scope` is `finished()` it is also `closed()`, why not call `startActive(true)` rather than `startActive(false)`? That is, auto-finish the `Scope` on `close()`, and then no need to call `finish()`.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
```diff
@@ -469,35 +453,19 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
+            closeScope(scope, HttpServletResponse.SC_BAD_GATEWAY);
             wrappedResponse.sendError(HttpServletResponse.SC_BAD_GATEWAY, "Error processing request");
         } catch (Error e) {
             LOG.error("{}:{} -- Error encountered while processing filter chain.", clusterId, nodeId, e);
-            if (scope != null) {
-                scope.span().setTag(Tags.HTTP_STATUS.getKey(), HttpServletResponse.SC_BAD_GATEWAY);
-                scope.span().finish();
-                scope.close();
-            }
+            closeScope(scope, HttpServletResponse.SC_BAD_GATEWAY);
```
dmnjohns: Do we need all these `closeScope` calls? Can we just handle it in the `finally` block? You should be able to get the status code even if the response is "committed", and we could repeat ourselves less. Also, repeating the status code in the `closeScope` and `sendError` opens the path for the status to mismatch at some point. Why do that rather than always using the actual status set on the response?

dmnjohns: Actually, we can probably follow the recommendation of the Java implementation of OpenTracing, minus the try-with-resource construct, in the README here:
https://github.com/opentracing/opentracing-java

That is, I propose:
```
io.opentracing.Tracer tracer = ...;
...
Scope scope = tracer.buildSpan(...).startActive(true);
// asChildOf stuff
try {
    // Do things.
} catch(Exception ex) {
    Tags.ERROR.set(scope.span(), true);
} finally {
    Tags.HTTP_STATUS.set(scope.span(), wrappedResponse.getStatus());
    scope.close();
}
```

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
```diff
@@ -531,6 +495,12 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
         MDC.clear();
     }
 
+    private static void closeScope(Scope scope, Number value) {
+        scope.span().setTag(Tags.HTTP_STATUS.getKey(), value);
```
dmnjohns: To be consistent, maybe use the `Tag` to set the tag?
```
Tags.HTTP_STATUS.set(scope.span(), value);
```

dmnjohns: Maybe consistent isn't the right word, because it begs the question consistent with what... if anything, the way it is currently is consistent with the builder's call to `withTag(...)`. So I'll just throw that out as an option, but I won't go so far as to recommend it. Though, overall, we probably should be consistent.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
```diff
@@ -24,6 +24,8 @@
 import io.opentracing.Span;
 import io.opentracing.SpanContext;
 import io.opentracing.Tracer;
+import io.opentracing.noop.NoopSpan;
+import io.opentracing.noop.NoopSpanContext;
```
dmnjohns: I'm not seeing any reference to these imports. Are they vestigial?

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
```diff
@@ -385,42 +387,25 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
-
-            scope = openTracingService.get().getGlobalTracer().scopeManager().activate(activeSpan, false);
+        LOG.trace("Let's see if there are any OpenTracing spans passed-in");
+        SpanContext context = null;
+        try {
+            context = tracer.extract(Format.Builtin.HTTP_HEADERS, new TracerExtractor(wrappedRequest));
+        } catch (RuntimeException re) {
+            LOG.error("Incoming tracer could not be parsed.");
+            LOG.error("We're going to start a new root span here; even though this is most likely part of a larger span.");
+            LOG.error("Check out following exception for more details:", re);
```
dmnjohns: I suspect making three separate `error` log calls means that these may not show up in order in the log (since other messages may squeeze between them). I worry that might be confusing for the user.

On the other hand, I totally get wanting these messages on separate lines.

On the other other hand, you can't actually guarantee that these messages will be on separate lines since that is up to the log format, right? And speaking of the log format, logging these messages separately will likely produce a bunch of redundant data in the logs (things like the thread ID, timestamp, etc.).

My gut feeling is that these should be collapsed into a single call, but I'm going to ask @adrianjgeorge what he thinks, and if we do collapse these, how we want to write it so that it is both clear in the code and the log.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
```diff
@@ -385,42 +387,25 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
+        }
+        LOG.debug("Got the span context from request: {}", context);
```
dmnjohns: `context` could still be null. I feel like that makes this log line a bit confusing. Maybe throw this in the `try` after the context is extracted?

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
wdschei: Yep. These were left over from the `NoOp` optimizations I removed. They're gone now.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
wdschei: I updated the verbiage a bit, but I want to log if the context was `null`.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
wdschei: I'm all for being consistent, even if it is just within the project.
Let's make this an action item for standup.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
wdschei: These were originally one big concatenated `String` and it was just ugly. I agree that there will be some redundant data and the possibility of interspersed messages, but the messages would all be linear as far as a given trace. I'm definitely open for suggestions on how we want to handle this large amount of data.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
wdschei: Yep. I agree. This was the first pass at cleaning up the prototype code that had those three lines copy/pasted all over the place. Again, consistency even within the project is the best practice.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
wdschei: Sure. I definitely don't have the familiarity with this API yet and that is basically what was in the original prototype cleaned up a bit. All that is now in place.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
wdschei: Something like below would give the same result as the original prototype, but is a bit more readable:
```java
            LOG.error(String.format("%s %s %s",
                "Incoming tracer could not be parsed.",
                "We're going to start a new root span here; even though this is most likely part of a larger span.",
                "Check out following exception for more details:"
            ), re);
```
Even the following is still consistent with other locations of logging exceptions/stacktraces and has the added benefit of only doing the formatting if necessary:
```java
            LOG.error("%s %s",
                "Incoming tracer could not be parsed.",
                "We're going to start a new root span here; even though this is most likely part of a larger span.");
            LOG.error("Check out following exception for more details:", re);
```

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
```diff
@@ -390,9 +389,11 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
         try {
             context = tracer.extract(Format.Builtin.HTTP_HEADERS, new TracerExtractor(wrappedRequest));
         } catch (RuntimeException re) {
-            LOG.error("Incoming tracer could not be parsed.");
-            LOG.error("We're going to start a new root span here; even though this is most likely part of a larger span.");
-            LOG.error("Check out following exception for more details:", re);
+            LOG.error("%s %s %s",
```
dmnjohns: Huh, does the `%s` pattern work here, or do you have to use `{}`?

wdschei: Yep. Fixed.

repose-aggregator/core/repose-core/build.gradle
```diff
@@ -52,6 +52,7 @@ dependencies {
     compile "io.opentracing:opentracing-util"
     compile "com.uber.jaeger:jaeger-core"
 
+    compileOnly "javax.inject:javax.inject"
```
dmnjohns: It looks like there currently a `compile` dependency on this that I believe should be removed.

PR #1879, opened 2018-03-07
REP-6654  Instrumenting the Atom Feed Service for OpenTracing

dmnjohns:


wdschei:
The only failures are the four known issues and the OpenTracing tests that aren't complete.

PR #1880, opened 2018-03-08
REP-6654 dist datastore

wdschei:
This also includes a non-functional formatting of the `PowerFilter` that removed a few more unnecessary imports.

### DIFF
--- a/repose-aggregator/components/services/datastore-service/datastore-service-impl/build.gradle
+++ b/repose-aggregator/components/services/datastore-service/datastore-service-impl/build.gradle
@@ -25,7 +25,11 @@ dependencies {
     testCompile "org.mockito:mockito-all"
     testCompile "org.codehaus.groovy:groovy-all"
     testCompile "org.hamcrest:hamcrest-all"
+    testCompile "org.springframework:spring-test"
+    testCompile "io.opentracing:opentracing-mock"
+    testCompile "org.apache.logging.log4j:log4j-core"
     testCompile "org.apache.logging.log4j:log4j-slf4j-impl"
+    testCompile "org.slf4j:jcl-over-slf4j"
 }
 
 jacocoTestCoverageVerification {
--- a/repose-aggregator/components/services/datastore-service/datastore-service-impl/build.gradle
+++ b/repose-aggregator/components/services/datastore-service/datastore-service-impl/build.gradle
@@ -9,6 +9,7 @@ dependencies {
     compile "org.eclipse.jetty:jetty-servlet"
     compile "org.springframework:spring-beans"
     compile "net.sf.ehcache:ehcache-core"
+    compile "io.opentracing:opentracing-api"
     // TODO: Add this dependency. Beware! Doing so pulls in a transitive "ehcache" artifact
     //       which shares some classes with "ehcache-core" above. The classpath conflict
     //       should be resolved before this dependency is uncommented.
@@ -27,7 +28,7 @@ dependencies {
     testCompile "org.hamcrest:hamcrest-all"
     testCompile "org.springframework:spring-test"
     testCompile "io.opentracing:opentracing-mock"
-    testCompile "org.apache.logging.log4j:log4j-core"
+    testCompile "org.apache.logging.log4j:log4j-slf4j-impl"
     testCompile "org.slf4j:jcl-over-slf4j"
 }
 
--- a/repose-aggregator/components/services/datastore-service/datastore-service-impl/build.gradle
+++ b/repose-aggregator/components/services/datastore-service/datastore-service-impl/build.gradle
@@ -28,7 +28,6 @@ dependencies {
     testCompile "org.springframework:spring-test"
     testCompile "io.opentracing:opentracing-mock"
     testCompile "org.apache.logging.log4j:log4j-core"
-    testCompile "org.apache.logging.log4j:log4j-slf4j-impl"
     testCompile "org.slf4j:jcl-over-slf4j"
 }
 
--- a/repose-aggregator/components/services/datastore-service/datastore-service-impl/src/main/java/org/openrepose/nodeservice/distributed/servlet/DistributedDatastoreServlet.java
+++ b/repose-aggregator/components/services/datastore-service/datastore-service-impl/src/main/java/org/openrepose/nodeservice/distributed/servlet/DistributedDatastoreServlet.java
@@ -175,7 +175,6 @@ public class DistributedDatastoreServlet extends HttpServlet {
                     break;
                 default:
                     resp.sendError(SC_INTERNAL_SERVER_ERROR);
-                    break;
             }
         } catch (IOException ioe) {
             LOG.error(ioe.getMessage(), ioe);
@@ -189,22 +188,18 @@ public class DistributedDatastoreServlet extends HttpServlet {
         // This servlet is only allowed to communicate with configured Repose instances,
         // so there is no chance of exposing sensitive information.
         // So it is safe to suppress warning squid:S1989
-        if (CacheRequest.isCacheRequestValid(req)) {
-            try {
-                final CacheRequest cachePut = CacheRequest.marshallCacheRequestWithPayload(req);
-                localDatastore.put(cachePut.getCacheKey(), objectSerializer.readObject(cachePut.getPayload()), cachePut.getTtlInSeconds(), TimeUnit.SECONDS);
-                resp.setStatus(SC_ACCEPTED);
-            } catch (IOException ioe) {
-                LOG.error(ioe.getMessage(), ioe);
-                throw new DatastoreOperationException("Failed to write payload.", ioe);
-            } catch (ClassNotFoundException cnfe) {
-                LOG.error(cnfe.getMessage(), cnfe);
-                throw new DatastoreOperationException("Failed to deserialize a message. Couldn't find a matching class.", cnfe);
-            } catch (MalformedCacheRequestException mcre) {
-                handleputMalformedCacheRequestException(mcre, resp);
-            }
-        } else {
-            resp.setStatus(SC_NOT_FOUND);
+        try {
+            final CacheRequest cachePut = CacheRequest.marshallCacheRequestWithPayload(req);
+            localDatastore.put(cachePut.getCacheKey(), objectSerializer.readObject(cachePut.getPayload()), cachePut.getTtlInSeconds(), TimeUnit.SECONDS);
+            resp.setStatus(SC_ACCEPTED);
+        } catch (IOException ioe) {
+            LOG.error(ioe.getMessage(), ioe);
+            throw new DatastoreOperationException("Failed to write payload.", ioe);
+        } catch (ClassNotFoundException cnfe) {
+            LOG.error(cnfe.getMessage(), cnfe);
+            throw new DatastoreOperationException("Failed to deserialize a message. Couldn't find a matching class.", cnfe);
+        } catch (MalformedCacheRequestException mcre) {
+            handleMalformedCacheRequestException(mcre, resp);
         }
     }
 
@@ -214,24 +209,19 @@ public class DistributedDatastoreServlet extends HttpServlet {
         // This servlet is only allowed to communicate with configured Repose instances,
         // so there is no chance of exposing sensitive information.
         // So it is safe to suppress warning squid:S1989
-        if (CacheRequest.isCacheRequestValid(req)) {
-            try {
-                final CacheRequest cacheDelete = CacheRequest.marshallCacheRequest(req);
-                localDatastore.remove(cacheDelete.getCacheKey());
-                resp.setStatus(SC_NO_CONTENT);
-            } catch (MalformedCacheRequestException e) {
-                LOG.trace("Malformed cache request on Delete", e);
-                switch (e.getMessage()) {
-                    case UNEXPECTED_REMOTE_BEHAVIOR:
-                        resp.setStatus(SC_BAD_REQUEST);
-                        break;
-                    default:
-                        resp.setStatus(SC_NO_CONTENT);
-                        break;
-                }
+        try {
+            final CacheRequest cacheDelete = CacheRequest.marshallCacheRequest(req);
+            localDatastore.remove(cacheDelete.getCacheKey());
+            resp.setStatus(SC_NO_CONTENT);
+        } catch (MalformedCacheRequestException e) {
+            LOG.trace("Malformed cache request on Delete", e);
+            switch (e.getMessage()) {
+                case UNEXPECTED_REMOTE_BEHAVIOR:
+                    resp.setStatus(SC_BAD_REQUEST);
+                    break;
+                default:
+                    resp.setStatus(SC_NO_CONTENT);
             }
-        } else {
-            resp.setStatus(SC_NOT_FOUND);
         }
     }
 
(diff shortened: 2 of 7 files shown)


wdschei:
427 of the failing tests are the new OpenTracing tests that are just not ready yet. The remaining 5 I'm looking at now.

wdschei:
4 of the 5 unexpected failing tests are because:

* `org.springframework.beans.factory.NoSuchBeanDefinitionException: No bean named 'powerFilter' is defined`

And the error is on the main [`REP-6654_OpenTracing`](https://github.com/rackerlabs/repose/tree/REP-6654_OpenTracing) branch.

The last test is passing locally and hasn't failed on other branches, so it may be a new flaky test.

repose-aggregator/components/services/datastore-service/datastore-service-impl/build.gradle
```diff
@@ -9,6 +9,8 @@ dependencies {
     compile "org.eclipse.jetty:jetty-servlet"
     compile "org.springframework:spring-beans"
     compile "net.sf.ehcache:ehcache-core"
+    compile "io.opentracing:opentracing-api"
+    compile "io.opentracing:opentracing-util"
```
dmnjohns: I glanced and didn't see anything, but is this dependency actually being used?

wdschei: Good call. These were moved into the `commons-utilities` project.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/main/java/org/openrepose/nodeservice/distributed/servlet/DistributedDatastoreServlet.java
```diff
@@ -127,7 +139,12 @@ protected void doTrace(HttpServletRequest req, HttpServletResponse resp) throws
         // This servlet is only allowed to communicate with configured Repose instances,
         // so there is no chance of exposing sensitive information.
         // So it is safe to suppress warning squid:S1989
-        resp.sendError(SC_METHOD_NOT_ALLOWED);
+        final Scope scope = startSpan(req, tracer, LOG);
```
dmnjohns: The `service` method delegates to the individual HTTP verb methods. So my first question is, do we really want to create a new span after we just created a span in `service`? Especially considering that the operation name will already include the method and URI. If we do, still want new spans, then this still doesn't work because:
1. It does not set the span created in `service` as the parent span.
2. It will attempt to create a new span with the same operation name (i.e., method and URI) as the span created in `service`.

Personally, I would be happy to drop the spans created in each individual verb processor method.

wdschei: True. Updated.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/main/java/org/openrepose/nodeservice/distributed/servlet/DistributedDatastoreServlet.java
```diff
@@ -106,18 +113,23 @@ public void init(ServletConfig config) throws ServletException {
     }
 
     @Override
-    protected void service(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
-        if (isRequestValid(request, response)) {
-            String traceGUID = TracingHeaderHelper.getTraceGuid(request.getHeader(CommonHttpHeader.TRACE_GUID));
-            MDC.put(TracingKey.TRACING_KEY, traceGUID);
-            LOG.trace("SERVICING DISTDATASTORE REQUEST");
+    protected void service(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
+        final Scope scope = startSpan(req, tracer, LOG);
```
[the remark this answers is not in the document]
wdschei: Yep. Updated.

repose-aggregator/components/services/datastore-service/datastore-service-impl/build.gradle
```diff
@@ -9,8 +9,6 @@ dependencies {
     compile "org.eclipse.jetty:jetty-servlet"
     compile "org.springframework:spring-beans"
     compile "net.sf.ehcache:ehcache-core"
-    compile "io.opentracing:opentracing-api"
```
dmnjohns: Oh wait. You probably still want this dependency. We don't like relying on transitive dependencies, and you do reference the `Tracer` class from this dependency directly.

I mostly just meant that we probably didn't need the `opentracing-util` dependency here since we aren't using the `GlobalTracer` or anything.

dmnjohns:
I also just noticed that there aren't any unit tests for `DistributedDatastoreServlet`. Have you looked into what is required to add some? I would love to see a sanity/regression test for creating a span here, but if there's a good reason that we never created unit tests for that class, then I can let it stand.

dmnjohns:
retest this please

wdschei:
That should answer the mail for the missing tests.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/test/scala/org/openrepose/nodeservice/distributed/servlet/DistributedDatastoreServletTest.scala
```diff
@@ -0,0 +1,93 @@
...
+    val badRequest = List("PUT", "PATCH").map(method => (method, SC_BAD_REQUEST)) // No data or Cache Key sent
+    val notAllowed = List("POST", "TRACE").map(method => (method, SC_METHOD_NOT_ALLOWED)) // Methods Not Allowed
+    val okMethods = List("OPTIONS").map(method => (method, SC_OK)) // This is handled by the parent HTTP Servlet
+    val notImplemented = List("CUSTOM", "BOGUS").map(method => (method, SC_NOT_IMPLEMENTED)) // Not Implemented
+
+    (notFound ++ badRequest ++ notAllowed ++ okMethods ++ notImplemented) foreach { case (httpMethod, result) =>
+      it(s"should build an OpenTracing span for HTTP method $httpMethod") {
+        servletRequest.setMethod(httpMethod)
+        distributedDatastoreServlet.service(servletRequest, servletResponse)
+        verify(tracerSpy).buildSpan(s"$httpMethod $CACHE_URI_PATH")
```
dmnjohns: I'm not sure you actually need to use a `spy` here. Wouldn't `MockTracer.finishedSpans()` give you the same information since `service` is a synchronous call? So in this case, couldn't you replace this line with:
```
tracer.finishedSpans.size shouldEqual 1
tracer.finishedSpans.get(0).operationName() shouldEqual s"$httpMethod $CACHE_URI_PATH"
```

Of course, I have nothing against spies, I just found it interesting that a `spy` was being used in conjunction with a `MockTracer` since they are serving a similar purpose.

repose-aggregator/components/services/datastore-service/datastore-service-impl/build.gradle
```diff
@@ -25,7 +25,11 @@ dependencies {
     testCompile "org.mockito:mockito-all"
     testCompile "org.codehaus.groovy:groovy-all"
     testCompile "org.hamcrest:hamcrest-all"
+    testCompile "org.springframework:spring-test"
+    testCompile "io.opentracing:opentracing-mock"
+    testCompile "org.apache.logging.log4j:log4j-core"
     testCompile "org.apache.logging.log4j:log4j-slf4j-impl"
+    testCompile "org.slf4j:jcl-over-slf4j"
```
dmnjohns: The two logging dependencies here look like they were added in anticipation of tests that verify log messages, but I don't see any such tests. For now, should we just drop these dependencies?

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/test/scala/org/openrepose/nodeservice/distributed/servlet/DistributedDatastoreServletTest.scala
```diff
@@ -0,0 +1,93 @@
...
+
+
+@RunWith(classOf[JUnitRunner])
+class DistributedDatastoreServletTest extends FunSpec with BeforeAndAfterEach with Matchers with MockitoSugar {
+
+  private val mockDatastoreService = mock[DatastoreService]
+  private val mockDatastore = mock[Datastore]
+  Mockito.when(mockDatastoreService.getDefaultDatastore).thenReturn(mockDatastore)
+  private val distributedDatastoreConfiguration = mock[DistributedDatastoreConfiguration]
+  val tracerSpy = Mockito.spy(new MockTracer)
```
dmnjohns: Personally, I am a fan of re-creating mocks for each test to ensure that there is no test bleed. The pattern I have started following is to declare all of my variables as `var`s then assign mocks to those variables in the `beforeEach` method. For example:
```
var datastoreService: DatastoreService = _
var datastore: Datastore = _

override def beforeEach(): Unit = {
  datastoreService = mock[DatastoreService]
  datastore = mock[Datastore]

  when(mockDatastoreService.getDefaultDatastore).thenReturn(mockDatastore)
}
```

You could, of course, use `val`s and `Mockito.reset(...)` instead of `var`s and `new`s, but I would rather rely on the language doing the right thing than the `Mockito` library. Although, there is a slight overhead to doing it my way.

I would also recreate the class under test for each test for the same reason (unless I am testing a singleton or static methods on a class).

Interestingly, you have already done this with the request and response.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/test/scala/org/openrepose/nodeservice/distributed/servlet/DistributedDatastoreServletTest.scala
```diff
@@ -0,0 +1,93 @@
...
+    servletResponse = new MockHttpServletResponse
+    tracerSpy.reset()
+  }
+
+  describe("Distributed Datastore calls without a Cache Key") {
+    val notFound = List("GET", "HEAD").map(method => (method, SC_NOT_FOUND)) // Head calls Get and no Cache Key sent
+    val badRequest = List("PUT", "PATCH").map(method => (method, SC_BAD_REQUEST)) // No data or Cache Key sent
+    val notAllowed = List("POST", "TRACE").map(method => (method, SC_METHOD_NOT_ALLOWED)) // Methods Not Allowed
+    val okMethods = List("OPTIONS").map(method => (method, SC_OK)) // This is handled by the parent HTTP Servlet
+    val notImplemented = List("CUSTOM", "BOGUS").map(method => (method, SC_NOT_IMPLEMENTED)) // Not Implemented
```
dmnjohns: This looks an awful lot like a map to me. You could either go for the straightforward method-to-response-code map like:
```
val methodToResponseCode = Map(
  "GET" -> SC_NOT_FOUND,
  "HEAD" -> SC_NOT_FOUND,
  "PUT" -> SC_BAD_REQUEST,
  ...)

methodToResponseCode foreach { case (httpMethod, responseCode) =>
  ...
}
```
or if you want to maintain your groupings and perhaps the comments:
```
val responseCodeToMethods = Map(
  SC_NOT_FOUND -> Set("GET", "HEAD"), // Head calls Get and no Cache Key sent
  SC_BAD_REQUEST -> Set("PUT", "PATCH") // No data or Cache Key sent
  ...)

responseCodeToMethods foreach { case (responseCode, methods) =>
  methods foreach { method =>
    ...
  }
}
```

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/test/scala/org/openrepose/nodeservice/distributed/servlet/DistributedDatastoreServletTest.scala
wdschei: Yep. I hadn't fully explored the `MockTracer` API and knew the `spy` would work.

Updated.

repose-aggregator/components/services/datastore-service/datastore-service-impl/build.gradle
wdschei: The `org.apache.logging.log4j:log4j-slf4j-impl` was a premature add, but the `org.slf4j:jcl-over-slf4j` is needed because the `DistributedDatastoreServlet` immediately constructs a `Logger` and without the dependency throws a `ClassNotFoundException`.

The first one has been removed.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/test/scala/org/openrepose/nodeservice/distributed/servlet/DistributedDatastoreServletTest.scala
wdschei: This is one of those we either trust what Mockito/OpenTracing provides or we don't. If we do, then I don't see an issue using their provided API. If we don't, then let's find something else.

That said, I typically use the pattern of if it is intended to be a long lived entity (e.g. a `Servlet`), then I like to test it under those conditions so we have a better chance of possibly finding the odd ball scenarios that cause a failure. As always it is never a guarantee, but it is more accurate testing for it's intended lifecycle.

That is also the intent behind the creation of the request and response every time. These are short lived one off entities and this is in line with their lifecycle.

In the end, I did remove the `Spy`, but left the `reset` on the `MockTracer` and the`DistributedDatastoreServlet` remains long lived as this is their intended lifecycle.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/test/scala/org/openrepose/nodeservice/distributed/servlet/DistributedDatastoreServletTest.scala
wdschei: You are correct, it could be a `Map`. I just chose to break it out into a `Seq` of tuples. Converting it to a single `Map` with redundant codes explicitly coded just seemed, well redundant. Converting it to a `Map` of `Seq`'s seems overly complex and causes more code in the end with the need of the extra `foreach`.

I swapped the `List` for the more generic `Seq` and moved to the less verbose underscore notation for the `map` method. I did leave the `okMethods` creation in the more verbose form for consistency even though it could have simply been defined as a single tuple.

dmnjohns: To be fair, I believe the `Map` of `Seq`s is still less code (since you lose all the `.map(...)` calls), but it does add another scope level. To get around that, you could do something like:
```
val responseCodeToMethods = Map(
  SC_NOT_FOUND -> Set("GET", "HEAD"), // Head calls Get and no Cache Key sent
  SC_BAD_REQUEST -> Set("PUT", "PATCH") // No data or Cache Key sent
  ...)

responseCodeToMethods.toSeq.flatMap({case (key, values) => values.map(key -> _)}) foreach { case (responseCode, method) =>
  ...
}
```

A `Map` still seems more natural to me, but whether your data structure is a `Seq` of tuples or a `Map`, you have to do some transformation. For the `Seq`, first off, you have to declare many independendent `Seq`s and combine them. For each `Seq`, you have to `.map(...)` all of the values into a tuple. For the `Map`, you have to (at least, to drop the nested `.foreach(...)`) break the one-key-to-value constraint. Or you accept the additional nested scope.

I guess I don't care too much, it just seemed to me like you should call a map a map.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/test/scala/org/openrepose/nodeservice/distributed/servlet/DistributedDatastoreServletTest.scala
dmnjohns: Your intended lifecycle point is an interesting one. I almost want to agree, but as I think about it more, I don't think I do. The problem I have is that while your tests may match the lifecycle of the object under tests, they also allow state/behavior to leak between tests. That makes individual tests less reliable since leftover state from a different test has the potential to change the behavior in the current test. That means that something like changing the ordering of tests might change the results, which does not seem good. Additionally, it becomes more difficult to reason about tests since once again you may have to account for leakage. If every test is self-contained, reasoning about the test is much more straightforward.

Now, a counter argument would be to say that tests failing because of test bleed is a good thing because it tells us when there is bleed. However, as noted above, that bleed may not be trivial to track down even with the tests. It also may not be an issue in the production code, but instead, an issue in the test code. And, of course, there is the fact that these are unit tests. Unit tests are only meant to test a unit of work, and in my mind, the best way to do so is to isolate the unit in question so that I can identify/specify the exact causality for some behavior.

As for the `Mockito.reset(...)` vs `var`/`new` thing, sure, either way is fine. It's less that I don't trust `Mockito.reset(...)` and more than there's no need to trust `Mockito.reset(...)`. To me, `Mockito.reset(...)` seems like a performance optimization to `var`/`new`, but as it turns out, the incredibly minor performance gain we get from it does not really bother me when it comes to tests. `var`/`new` is a pattern than works for non-mocked objects as well, so there's a consistency argument there. Ultimately, though, unless you're using `reset(...)` within a test (which is a test smell!), I don't overly care.

Just my 2c. As it stands, these tests should meet what we want to get out of them, so if you still want to leave them as-is, that's fine.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/test/scala/org/openrepose/nodeservice/distributed/servlet/DistributedDatastoreServletTest.scala
wdschei: The `Map` is a `Map`.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/test/scala/org/openrepose/nodeservice/distributed/servlet/DistributedDatastoreServletTest.scala
```diff
@@ -0,0 +1,97 @@
...
+      "GET"     -> SC_NOT_FOUND,
+      "HEAD"    -> SC_NOT_FOUND,   // Head calls Get
+      "PUT"     -> SC_BAD_REQUEST, // No data sent
+      "PATCH"   -> SC_BAD_REQUEST, // No data sent
+      "POST"    -> SC_METHOD_NOT_ALLOWED,
+      "TRACE"   -> SC_METHOD_NOT_ALLOWED,
+      "CUSTOM"  -> SC_NOT_IMPLEMENTED,
+      "BOGUS"   -> SC_NOT_IMPLEMENTED,
+      "OPTIONS" -> SC_OK) // This is handled by the parent HTTP Servlet
+      // @formatter:on
```
dmnjohns: Holy carp, how did I just learn about this magic?! I need more `@formatter` in my life.

repose-aggregator/components/services/datastore-service/datastore-service-impl/src/test/scala/org/openrepose/nodeservice/distributed/servlet/DistributedDatastoreServletTest.scala
wdschei: I too agree, the extremely minimal cost of the `val`/`reset()` vs `var`/`new` shouldn't be a concern in these tests.

I've been thinking some more on this and your point about the state/behavior leakage is a fair one. As they were written, these tests were more in line with this projects use of Functional Integration Tests. That said, I've updated these to be more static and repeatable by converting them to `var`'s instead of `val`'s thereby constraining each execution to minimize unintended leakage.

wdschei:
Down to only the 4 expected failures and all of the OpenTracing failures.

PR #1882, opened 2018-03-09
REP-6654 Service Changes

adrianjgeorge:


adrianjgeorge:
I forgot to go back and look over all the schema docs, i fixed ones i knew were wrong a couple of times, but it would probably still be good for someone to look at them all.

repose-aggregator/components/services/open-tracing-service/open-tracing-service-api/src/main/resources/META-INF/schema/config/open-tracing.xsd
```diff
@@ -0,0 +1,313 @@
...
+                            <html:p>
+                                Optional token.
+                            </html:p>
+                        </xs:documentation>
+                    </xs:annotation>
+                </xs:attribute>
+                <xs:assert vc:minVersion="1.1"
+                           test="(@username and @password and not(@token)) or (not(@username) and not(@password) and @token) or (not(@username) and not(@password) and not(@token))"
+                           xerces:message="Username/password must be specified or a token must be specified."
+                           saxon:message="Username/password must be specified or a token must be specified."
```
wdschei: The third case of none being specified is not in the message.

dmnjohns: Fair point. Since this assertion is just preventing the error case, how do you feel about this message instead?:

> Username/password and a token cannot both be specified.

I actually kind of want to replace the assertion test with a conditional that is a bit easier for me to grok, but since that might just be me, I'd like to get your opinion on that:
```
                           test="if (@username or @password) then @username and @password and not(@token)
                                 else if (@token) then not(@username) and not(@password)
                                 else true()"
                           xerces:message="Username/password and a token cannot both be specified."
                           saxon:message="Username/password and a token cannot both be specified."/>
```

The main difference is in the approach taken. In the current assertion, all valid states are enumerated. In the proposed assertion, only exclusionary states are enumerated with a default pass for all other states. In practice, that only really allows us to replace the no-auth state with the catch-all state, so no big difference there. It's just something about the structure that makes things clearer in my mind. Thoughts?

repose-aggregator/components/services/open-tracing-service/open-tracing-service-api/src/main/resources/META-INF/schema/config/opentracing.xsd
```diff
@@ -1,315 +0,0 @@
...
-                <xs:documentation>
-                    <html:p>
-                        Tracer agent port.  Defaults to 6831.  Used with udp protocol.  Ignored otherwise.
-                    </html:p>
-                </xs:documentation>
-            </xs:annotation>
-        </xs:attribute>
-
-
-        <xs:attribute name="collector-endpoint" type="xs:string" default="http://localhost:14268">
```
dmnjohns: It seems like we lost this after the update? It looks like the agent host/port was conflated with the collector host/port.

wdschei: Is that the real intent or an oversight?

dimtruck: Collector =! agent .  Agent would be a thing living alongside Repose that you can push to via UDP while Collector is something remote you can push to via TCP.

Out of the box, we should support both (right now our implementation of Jaeger does not support agents but that should change soon ™️  as using agents is best practice)

dmnjohns: A great question! At first I thought it was an oversight, but now I'm thinking it was intentional. My initial thinking was that you would always have an agent _and_ collector, and you would only even "talk" to the agent, but it appears that you only ever talk to one depending on the transport protocol being used. So it's an either/or thing. Either you configure a host/port for an agent if you're using UDP (thrift) as the transport, or you configure a host/port for the collector if you're using HTTP as the transport.
See: https://jaeger.readthedocs.io/en/latest/architecture/#agent

In either case, you only have to specify one endpoint to push data to, so I suspect that is why Adrian collapsed them. Plus, this is all Jaeger specific, so it was moved into the Jaeger configuration as well.

repose-aggregator/components/services/open-tracing-service/open-tracing-service-api/src/main/resources/META-INF/schema/config/open-tracing.xsd
```diff
@@ -0,0 +1,337 @@
...
+                    <xs:choice id="jaegerConnection">
+                        <xs:element name="connection-udp" type="JaegerConnectionUdp"/>
+                        <xs:element name="connection-http" type="JaegerConnectionHttp"/>
+                    </xs:choice>
+
+                    <xs:choice id="jaegerSampling">
+                        <xs:element name="sampling-constant" type="JaegerSamplingConstant"/>
+                        <xs:element name="sampling-probabilistic" type="JaegerSamplingProbabilistic"/>
+                        <xs:element name="sampling-rate-limiting" type="JaegerSamplingRateLimiting"/>
+                    </xs:choice>
```
dmnjohns: I tried three separate approached to clean this up a bit:
1. `xs:choice` with bindings. This is what you see here and in the `bindings.xjb` file. Basically, allows the user to choice one of the enumerated `element`s. The bindings magic just binds the resultant type in the generated object to the `JaegerSampling` parent type so that we have a more natural expression of polymorphism in the JVM world.
1. XML polymorphism. Look through the commit history for a commit about using raw XML types to see the implementation. Rather than giving the user a choice, force the user to always supply an `element` of the parent type (e.g., `JaegerSampling`). This cleans up both the XML (by not requiring us to enumerate the types we've already defined) and the generate Java object. However, it does require the user to be explicit about the type of their `element` by using the `xsi:type` attribute.
Honestly, this felt like the most natural approach to me, but the user experience is debatably worse than the other choices.
1. `substitutionGroup`.  This seemed like the natural XML way to handle this case, which is why I gave it a shot. I have the implementation stashed locally because JAXB does not play well with substitution groups. That is, JAXB will treat the substituted `element` as a `JAXBElement` and return that rather than a generated type, which makes manipulation of the generated object quite cumbersome. Substitution groups allow the user to choose an `element` by name rather choosing a type (similar to the `choice` approach). It also does not require binding magic. It does, however, still require enumeration of already defined types, and that enumeration takes the form of additional elements at the root level of the schema.
Here's a partial implementation for the sake of completeness:
```
<xs:complexType name="JaegerTracerConfig">
  ...
  <xs:sequence>
      <xs:element ref="connection"/>
  </xs:sequence>
  ...
</xs:complexType

<xs:element name="connection" type="JaegerConnection" abstract="true"/>
<xs:element name="connection-http" type="JaegerConnectionHttp" substitutionGroup="connection"/>
    <xs:element name="connection-udp" type="JaegerConnectionUdp" substitutionGroup="connection"/>
```

So that's the story. I am leaving this comment so that we can discuss which approach is best.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/core/opentracing/ReposeTracer.scala
```diff
@@ -28,14 +28,14 @@ import javax.annotation.PostConstruct
...
   *
   * The main differences are this Tracer:
   *  - immediately registers with the OpenTracing GlobalTracer
   *  - allows for registration of a new Tracer at any time.
   */
 @Named
-class ReposeTracer extends Tracer with LazyLogging {
+class ReposeTracer extends DelegatingTracer with LazyLogging {
 
   private var tracer: Tracer = NoopTracerFactory.create()
```
dmnjohns: I know this isn't part of this PR, but there was some discussion about replacing the default `NoopTracer` with a more fully-featured `Tracer` that simply does not report trace data (i.e., is configured with a `NoopReporter`).

Doing so, however, would probably mean keeping the configured header name in the system model so that it could be configured for use by this default `Tracer` as well. Not a big deal, just something to note. The service would have to listen for system model updates again, but that also opens the way to making this service a node service and handling it in the system model services list if we want.

Tangentially, I was under the impression that the `enabled` attributed on the service was dropped since a constant sampling could be configured to off. However, wouldn't that still lead to trace data being reported, just not exposed (perhaps even aggregated) via the tracing server? That makes me think we should either add the `enabled` attribute back, or if we switch to a node service, watch the system model to check if this service is listed.

Tangentially, does the Open Tracing Service actually needs to be a core service anymore? The `ReposeTracer` is always available without the service, the service just acts as a way to configure it. Making the service a node service would allow per-cluster configuration, and

repose-aggregator/components/services/open-tracing-service/open-tracing-service-api/src/main/resources/META-INF/schema/config/open-tracing.xsd
```diff
@@ -0,0 +1,349 @@
...
+                    <html:p>
+                        The name of the service that is reporting traces.
+                        This value will be used to identify the service and should therefore be unique
+                        (e.g., identity-repose).
+                    </html:p>
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+
+        <xs:attribute name="span-context-key" type="xs:string" default="TBD">
```
dmnjohns: This still needs to be dealt with. As @dimtruck pointed out, there is a RFC meant to standardize on the header names:
https://w3c.github.io/distributed-tracing/report-trace-context.html

Of course, we have to wait for that to have its status bumped up, and for Jaeger (or other tracers) to support it. Especially since the proposal (at this time) calls for multiple headers being added to the request.

For now, Dimitry suggested defaulting to whatever the Jaeger library does, which seems fine, but I'm not sure how we would do that. Using a default here means that we always have some value to set when configuring the `Tracer`. We could allow this attribute to be null, but then if we want to add a default in the future doing so will probably be backwards incompatible. Or we could plan to never default this attribute. If the user does not configure it, we let the tracing library do its thing, otherwise, we write a single header with the configured header name. Once again, that will probably not suffice once that RFC status changes.

repose-aggregator/components/services/open-tracing-service/open-tracing-service-api/src/main/resources/META-INF/schema/config/open-tracing.xsd
wdschei: I like Option 1. I understand that Option 2 is the more XML way to do things, but the typical Repose user/maintainer doesn't want to deal with XML any more than they have too. Option 3 leads us to a tooling issue and since JAXB is how we do things now and into the foreseeable future, I have to vote against it simply for clarity.

My 2&#162; are Option 1.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/core/opentracing/ReposeTracer.scala
adrianjgeorge: It doesn't work as a node service, because they'd try to replace each others tracer. We could maybe make it work by moving the `ReposeTracer` down to a lower level, but then we run into problems with the http client.

adrianjgeorge:
I'm starting to lean towards not letting the header being configured, and just embrace the fact that we are using jaeger for now. It means we'll have a breaking change when the rfc is actually finalized, but I'm tired of trying to fight this. Thoughts?

wdschei:
My concern there is that we are then waiting on a solution to become available. If someone wanted to make there system compatible immediately and use a non-Jaeger backend, then our custom thing would allow it.

dmnjohns:
I can see both points. The one thing I will note with maintaining the header name configuration is that dropping support for a custom header in the future will be backwards incompatible (i.e., a major version change). That is, I don't love putting something in now that we will probably want to yank out later. But I also don't love having to wait on the solution to become available, as Bill noted.

So I'm torn.

wdschei:
I kind of like releasing this feature as a new _incubating_ state. Then we can say hey, things have changed and we are changing too without having to make it a major version change.

repose-aggregator/components/services/open-tracing-service/src/test/scala/org/openrepose/core/services/opentracing/OpenTracingServiceImplTest.scala
```diff
@@ -0,0 +1,313 @@
...
+        isA(classOf[UpdateListener[OpenTracingConfig]]))
+    }
+  }
+
+  describe("itInitialized") {
+    it("should return false if the service has not been configured") {
+      openTracingService.isInitialized shouldBe false
+    }
+
+    it("should return false if the service has been configured") {
```
wdschei: Typo. `false` => `true`

repose-aggregator/components/services/open-tracing-service/src/test/scala/org/openrepose/core/services/opentracing/OpenTracingServiceImplTest.scala
```diff
@@ -0,0 +1,313 @@
...
+          .withProbability(probability))
+
+      val samplerConfig = openTracingService.getJaegerSamplerConfiguration(config)
+
+      samplerConfig.getType shouldBe ProbabilisticSampler.TYPE
+      samplerConfig.getParam shouldEqual probability
+    }
+  }
+
+  describe("getJaegerSenderConfiguration") {
```
wdschei: Why are we testing the XJB Bindings/XJC generated code here?

wdschei: Wouldn't they make more sense in the Schema test?

repose-aggregator/components/services/open-tracing-service/src/test/scala/org/openrepose/core/services/opentracing/OpenTracingServiceImplTest.scala
```diff
@@ -0,0 +1,313 @@
...
+    it("should unsubscribe an opentracing configuration listener") {
+      openTracingService.destroy()
+
+      verify(configurationService).unsubscribeFrom(
+        isEq("open-tracing.cfg.xml"),
+        isA(classOf[UpdateListener[OpenTracingConfig]]))
+    }
+  }
+
+  describe("itInitialized") {
```
dmnjohns: Typo. `itInitialized` => `isInitialized`

repose-aggregator/components/services/open-tracing-service/src/test/scala/org/openrepose/core/services/opentracing/OpenTracingServiceImplTest.scala
dmnjohns: This is actually testing ever-so-slightly more than just the JAXB generated objects. It's testing the translation from our configuration object (which is a JAXB object) to a Jaeger sampler configuration. These tests are just to verify that configuration is being applied in the right places.

PR #1891, opened 2018-03-26
REP-6654 Open Tracing

adrianjgeorge:
I'm opening this up even the functional tests are being fixed still, and a couple of features are still inbound. I'll be making notes here as to things I see that still need fixing or work, especially if i don't have the time to get them myself. We should all review this one since we all had our fingers in the pie.


repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/opentracing/ScopeHelper.scala
```diff
@@ -0,0 +1,59 @@
...
+import io.opentracing.propagation.Format
+import io.opentracing.tag.Tags
+import io.opentracing.{Scope, SpanContext, Tracer}
+import javax.servlet.http.{HttpServletRequest, HttpServletResponse}
+
+import org.slf4j.Logger
+
+import scala.util.{Failure, Success, Try}
+
+object ScopeHelper {
```
adrianjgeorge: This needs a test

repose-aggregator/artifacts/valve/build.gradle
```diff
@@ -23,6 +23,7 @@ dependencies {
     compile "org.eclipse.jetty:jetty-servlet"
     compile "org.eclipse.jetty:jetty-server"
     compile "org.eclipse.jetty:jetty-util"
+    compile "javax.inject:javax.inject"
```
dmnjohns: So as with our filters, we get this dependency transitively by depending on `repose-core`. Do we want to make this a `compileOnly` dependency?

I still like calling out the dependency here, but I think it was stuck at the `compile` level when we were under the impression that the container (either Valve or the container running the WAR) would provide this dependency. Since servlet containers will not necessarily provide this dependency (as it is part of JEE, but not the servlet specification), we always pull it in to core. To keep things simple, since we know core will always be available in an environment running Repose, we add it as a `compileOnly` (i.e., `provided`) dependency everywhere else it is needed.

repose-aggregator/components/filters/cors-filter/src/main/scala/org/openrepose/filters/cors/CorsFilter.scala
```diff
@@ -211,7 +211,7 @@ class CorsFilter @Inject()(configurationService: ConfigurationService)
     val maybeForwardedHostUri = Try(maybeForwardedHost
       .map(forwardedHost => new URIBuilder(s"${request.getScheme}://$forwardedHost"))
       .map(uri => uri.setPort(normalizePort(uri.getPort, uri.getScheme)).setHost(normalizeHost(uri.getHost)).build()))
-    lazy val hostUri = new URIBuilder().setScheme(request.getScheme).setHost(normalizeHost(request.getServerName))
+    lazy val hostUri = new URIBuilder("").setScheme(request.getScheme).setHost(normalizeHost(request.getServerName))
```
dmnjohns: What's this change about? It seems like switching from the empty constructor to the string constructor with an empty string would only change the path of the URI that gets built, but I haven't looked into it too much. If we want to change the path, though, I would suggest calling `.setPath` on the builder instead. Of course, I'm not sure why we are touching this class as part of this PR to begin with...

adrianjgeorge: This was a weird problem @dimtruck ran into, and this ended up being the fix. Don't remember what it was related too though.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/opentracing/OpenTracingServiceConstOffHttpTest.groovy
```diff
@@ -0,0 +1,215 @@
...
+        messageChain.handlings.get(0).request.headers.getCountByName(TRACING_HEADER) == 2
+
+
+        and: "request should have tracer header pass through as well as a new header added"
+        def newTraceId
+        def validateCount = 0
+        def validateUniqueList = []
+        messageChain.handlings.get(0).request.headers.each {
+            if (it.name == TRACING_HEADER) {
+                if (!validateUniqueList.contains(it.value)) validateUniqueList << it.value
```
adrianjgeorge: not a deal breaker but `{`/`}`s would be preferable here. I'm surprised this passed codenarc.

adrianjgeorge:
Feels like the placement of the performance tests doesn't follow our standard? is that right or am i mis-remembering?

adrianjgeorge:
Also several of the functional tests (like all of them) have single line ifs without curlys.

repose-aggregator/artifacts/valve/build.gradle
adrianjgeorge: seems legit to me

dmnjohns:
Performance test placement seems to follow our standard, with the one difference being that a new top-level directory was added for OpenTracing. So what I'm calling top-level directories are now:
* `filters`
* `opentracing`
* `usecases`

adrianjgeorge:
Changes look good.

Mario-Lopez:
Everyone has approved everyone's code, so I will merge it.

PR #1892, opened 2018-03-26
Fix Opentracing functional tests

dimtruck:


repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/opentracing/OpenTracingServiceDisabledHttpTest.groovy
```diff
@@ -1,152 +0,0 @@
...
- */
-package features.services.opentracing
-
-import org.openrepose.framework.test.ReposeValveTest
-import org.openrepose.framework.test.mocks.MockTracerAgent
-import org.openrepose.framework.test.mocks.MockTracerCollector
-import org.rackspace.deproxy.Deproxy
-import spock.lang.Unroll
-
-class OpenTracingServiceDisabledHttpTest extends ReposeValveTest {
```
dmnjohns: Can we keep this test? Instead of using the `enabled` attribute, you can just not include an `open-tracing.cfg.xml` file. The assertions should be good, though.

dimtruck: There's already a test where there's no config.

dimtruck: https://github.com/rackerlabs/repose/pull/1892/files#diff-22becc3deae6d86316e46362feff7c24

PR #1894, opened 2018-03-28
REP-6654 Tags

adrianjgeorge:


repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/opentracing/ReposeTags.scala
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_=_
+ */
+package org.openrepose.commons.utils.opentracing
+
+object ReposeTags {
+  val ReposeVersion = "repose.version"
```
dmnjohns: For the sake of convention, do we want to make this a `final val`? Since it's in an object, it's not like the object can be extended (and therefore, the value cannot be overridden), so that change would just be for the sake of consistency.

Since it's public, do we want to annotate it with its type (See: https://docs.scala-lang.org/style/types.html#inference)?

PR #1899, opened 2018-04-02
Adding the release version to the live OpenTracing service page

dmnjohns:


wdschei:
The failure was the `Scoverage Report` and nothing to do with what was actually changed.

PR #1902, opened 2018-04-02
REP-6674 Switching from LazyLogging to StrictLogging

dmnjohns:
The one thing that I want to call out here is that I intentionally did not switch the following classes to `StrictLogging`:

- `MutableServletOutputStream`
- `PassthroughServletOutputStream`
- `ReadOnlyServletOutputStream`

The reason being that we create those classes quite frequently, and each class will create a reference to a logger for itself, but those loggers will not log anything in most cases.

Although, thinking about it, we may just want to create a logger in a companion object for those classes (and still use `StrictLogging`?).

I'd love thoughts on the subject if anyone has them!

### DIFF
--- a/repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/spring/ValveRunner.scala
+++ b/repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/spring/ValveRunner.scala
@@ -26,7 +26,7 @@ import java.util.concurrent.atomic.AtomicReference
 import javax.inject.{Inject, Named}
 import javax.management.{InstanceNotFoundException, ObjectName}
 
-import com.typesafe.scalalogging.slf4j.LazyLogging
+import com.typesafe.scalalogging.slf4j.StrictLogging
 import org.openrepose.commons.config.manager.UpdateListener
 import org.openrepose.core.container.config.ContainerConfiguration
 import org.openrepose.core.services.config.ConfigurationService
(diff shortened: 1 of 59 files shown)


wdschei:
Now that you mention it, I'm curious as to why we haven't always defined the loggers for Scala classes in their companion objects. This seems like the most correct place to put them based on how they are created and used. This would also be in line with the Java way of creating them as `private static final`'s.

It feels like a bit of scope creep, but it does seem more correct.

dmnjohns:
After looking into it a bit more, while companion objects may be the ideal solution in most cases, I'm not sure we actually care. Remember that the `LoggerFactory` will only create one instance of a logger for each name (i.e., the class name when using the Scala logging traits). So mixing in the trait does have a real cost of one logger reference per instance, but that's pretty minimal. Especially considering that we could not use the Scala Logging traits, and would have to create a companion object for a bunch of classes.

For a little more discussion on the topic, see:
https://github.com/lightbend/scala-logging/issues/10

So I think in the end, just switching everything to `StrictLogging` is probably the right call. It's the most straightforward approach with acceptable overhead in the worst case, and optimal overhead in the best.

Thoughts?

adrianjgeorge:
Yeah, i super don't feel the need to add them to the companion object or add a companion object where it doesn't exist. A good percentage of our objects are singleton (no gain) or already contain a large size (minimal impact), so I see no need to change from what we currently do.

adrianjgeorge:
retest this please

PR #1906, opened 2018-04-04
REP-6778 Fixed jacoco recording to not 'dirty' the test outputs

adrianjgeorge:
This will let the incremental build work correctly again, sonar looks like it's good to go. I can't find the integration test coverage anymore, so maybe we got rid of that when we added scoverage?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -338,12 +338,6 @@ subprojects {
         }
     }
 
-    test {
-        jacoco {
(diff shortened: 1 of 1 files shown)


wdschei:
The failure for one job was the flaky `MisbehavingOriginTest` that is being taken care of elsewhere and the other failed for `jacocoTestCoverageVerification` that has only failed here one time.

PR #1907, opened 2018-04-04
REP-6793 Build Flakiness Improvements

dmnjohns:


repose-aggregator/core/repose-core/src/test/java/org/openrepose/core/services/reporting/destinations/DestinationInfoLogicTest.java
```diff
@@ -117,16 +116,36 @@ public void shouldReturnZeroIfNoResponseTime() {
...
+
+            for (int i = 0; i < totalResponses; i++) {
                 destinationInfoLogic.incrementResponseCount();
             }
 
+            // By sleeping for a certain amount of time that is significantly longer than the expected runtime of this
+            // test, we can minimize the potential difference between the expected and actual throughput.
+            // That is, as wait time increases while runtime is relatively constant (or, at least, with minimal variance),
+            // our expected and actual throughput converge.
+            Thread.sleep(sleepTime);
```
dmnjohns: This solution actually solves two problems:

1. Precision. By waiting, we don't have to worry about calculating `Infinity` due to `System.currentTimeMillis()` returning the same time (i.e., the test running within a single millisecond).
1. Variance. By waiting, we decrease the effect of test runtime on our calculations.

That being said, I hate all of this. If I had the time and inclination, I would either remove this functionality (if not necessary/desired), or create a time service that could be injected for more exact testing here. Or maybe something else that I haven't thought of.

repose-aggregator/core/repose-core/src/test/java/org/openrepose/core/services/reporting/destinations/DestinationInfoLogicTest.java
```diff
@@ -69,9 +68,9 @@ public void shouldIncrementStatusCodeCount() {
 
         @Test
         public void shouldAccumulateResponseTime() {
-            destinationInfoLogic.accumulateResponseTime(1000l);
+            destinationInfoLogic.accumulateResponseTime(1000L);
```
adrianjgeorge: good call switching to big L

PR #1910, opened 2018-04-06
REP-6795 new URI Redaction Service

wdschei:


adrianjgeorge:
Config looks good.

dmnjohns:
I thought of another test case we may want to have here that I did not yet see captured. What if there are nested capture groups? I can't think of a good real world example for that use-case, but we should probably be clear about the expected behavior anyway. An example configured regex would be:
```
\w+/(\w+:(\d+))/\w+
```

Also, are we accounting for the group index changes after we perform a redaction? That is, what if the redaction token is a different length than the text being replaced? Does that mess up the next replacement?

adrianjgeorge:
Lol, i was thinking about nested groups this morning. My take away was to specifically document not to use them, because things'll get weird.

wdschei:
The short answer to the question about do we account for the group index changes after we perform a redaction is, yes. That was specifically handled in this commit: https://github.com/rackerlabs/repose/pull/1910/commits/3f823eb8602184105ddad3388ca17916ed8b9318

wdschei:
I think documenting nested capture groups not being supported is a great idea.

repose-aggregator/components/services/uri-redaction-service/uri-redaction-service-impl/src/main/scala/org/openrepose/core/services/uriredaction/UriRedactionServiceImpl.scala
```diff
@@ -0,0 +1,90 @@
...
+ * =_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_=_
+ */
+package org.openrepose.core.services.uriredaction
+
+import com.typesafe.scalalogging.slf4j.StrictLogging
+import javax.annotation.{PostConstruct, PreDestroy}
+import javax.inject.{Inject, Named}
+import org.openrepose.commons.config.manager.UpdateListener
+import org.openrepose.core.services.config.ConfigurationService
+import org.openrepose.core.services.uriredaction.UriRedactionServiceImpl._
```
dmnjohns: I think we normally stick this in the class itself. I would suggest we do that here for the sake of consistency.

As for whether or not we should do that everywhere, I haven't really formulated a strong opinion. I don't think the Scala style guide is explicit about it, but it does import the companion into the class in the example:
https://docs.scala-lang.org/tour/singleton-objects.html#companion-objects

repose-aggregator/commons/commons-utilities/src/test/scala/org/openrepose/commons/utils/opentracing/ScopeHelperTest.scala
```diff
@@ -20,38 +20,50 @@
...
 import org.scalatest.junit.JUnitRunner
 import org.scalatest.mock.MockitoSugar
-import org.scalatest.{FunSpec, Matchers}
+import org.scalatest.{BeforeAndAfterEach, FunSpec, Matchers}
 import org.slf4j.Logger
 import org.springframework.mock.web.{MockHttpServletRequest, MockHttpServletResponse}
 
 @RunWith(classOf[JUnitRunner])
-class ScopeHelperTest extends FunSpec with MockitoSugar with Matchers {
+class ScopeHelperTest extends FunSpec with MockitoSugar with Matchers with BeforeAndAfterEach {
```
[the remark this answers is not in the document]
wdschei: This was basically already being done, so I just left it alone and added the [`when` clause](https://github.com/rackerlabs/repose/blob/5292e5d425914df30a35aa9cfa68f2fa896eeb0a/repose-aggregator/components/services/datastore-service/datastore-service-impl/src/test/scala/org/openrepose/nodeservice/distributed/servlet/DistributedDatastoreServletTest.scala#L93) to the [test](https://github.com/rackerlabs/repose/blob/5292e5d425914df30a35aa9cfa68f2fa896eeb0a/repose-aggregator/components/services/datastore-service/datastore-service-impl/src/test/scala/org/openrepose/nodeservice/distributed/servlet/DistributedDatastoreServletTest.scala#L93-L97).

repose-aggregator/commons/commons-utilities/src/test/scala/org/openrepose/commons/utils/opentracing/httpclient/ReposeTracingRequestInterceptorTest.scala
```diff
@@ -28,72 +28,69 @@ import org.mockito.Matchers.any
...
 import org.openrepose.commons.utils.opentracing.ReposeTags
+import org.openrepose.core.services.uriredaction.UriRedactionService
 import org.scalatest.junit.JUnitRunner
 import org.scalatest.mock.MockitoSugar
-import org.scalatest.{BeforeAndAfter, FunSpec, Matchers}
+import org.scalatest.{BeforeAndAfterEach, FunSpec, Matchers}
 
 @RunWith(classOf[JUnitRunner])
-class ReposeTracingRequestInterceptorTest extends FunSpec with Matchers with MockitoSugar with BeforeAndAfter {
+class ReposeTracingRequestInterceptorTest extends FunSpec with Matchers with MockitoSugar with BeforeAndAfterEach {
```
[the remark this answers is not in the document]
wdschei: I can speckle the test everywhere, but is it needed? Wanted?

repose-aggregator/tests/performance-tests/roles/repose/files/config/services/uriredaction/uri-redaction.cfg.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<uri-redaction xmlns="http://docs.openrepose.org/repose/uri-redaction-service/v1.0">
+    <!-- Redact the third path segment of a four segment URI hitting the v1 API. -->
+    <redact>^/v1/[^/]+/([^/]+)/[^/]+.*</redact>
+    <!-- Redact the third path segment of a four segment URI hitting the v2 API's admin endpoint. -->
+    <redact>^/v2/admin/([^/]+)/[^/]+.*</redact>
```
dmnjohns: I know Mario wanted a bunch of regexes here to really gauge the effect of applying all regexes. I would actually love to see a config where multiple redactions apply to each request, and multiple redactions do not.

repose-aggregator/tests/performance-tests/roles/gatling/files/simulations/services/uriredaction/UriRedactionServiceSimulation.scala
```diff
@@ -0,0 +1,87 @@
...
+    mainScenario
+  ).assertions(
+    global.responseTime.percentile3.lte(percentile3ResponseTimeUpperBound),
+    global.successfulRequests.percent.gte(percentSuccessfulRequest)
+  ).protocols(httpConf)
+
+  def getRequest: HttpRequestBuilder = {
+    http(session => session.scenario)
+      .get("/v2/admin/redactMe/required")
+      .check(status.is(200))
```
dmnjohns: Do we also want a sanity check to prove that the redaction was processed as expected? I know functional tests should prove that, but here it would just be to make sure that the performance test is set up correctly.

repose-aggregator/tests/performance-tests/test_vars/services/uriredaction.yml
```diff
@@ -0,0 +1,29 @@
...
+  dir_prefix: "services/uriredaction",
+  repose: {
+    config: {
+      system_model: {
+        services: [
+          # todo: when it becomes a service, add "uri-redaction"
+        ]
+      }
+    },
+    config_files: {
```
dmnjohns: I haven't looked through the Ansible tasks in a while, but will Repose run without specifying the system model and container configuration templates here? I seem to remember us being explicit when applying configuration.

repose-aggregator/tests/performance-tests/roles/repose/files/config/services/uriredaction/uri-redaction.cfg.xml
wdschei: Unless we run set of redactions as their own scenario, then they would all be intermixed in the run and the numbers would meld together. So do we want multiple scenarios? Multiple Simulations?

repose-aggregator/commons/commons-utilities/src/test/scala/org/openrepose/commons/utils/opentracing/ScopeHelperTest.scala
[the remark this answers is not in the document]
dmnjohns: Huh? I don't see it being done anywhere. I mean, I see where the service would be called, but not test assertion that the service _is_ being called. That is, if I drop the call to the service in `ScopeHelper`, `ScopeHelperTest` still passes, which doesn't seem right.

repose-aggregator/commons/commons-utilities/src/test/scala/org/openrepose/commons/utils/opentracing/httpclient/ReposeTracingRequestInterceptorTest.scala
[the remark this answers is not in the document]
dmnjohns: I think so. It's a small test that proves that we are performing redactions when, where, and as expected.

repose-aggregator/tests/performance-tests/test_vars/services/uriredaction.yml
wdschei: After looking through and confirming with @Mario-Lopez , the Repose perf tests are set up to have these [defaults](https://github.com/rackerlabs/repose/blob/master/repose-aggregator/tests/performance-tests/roles/repose/defaults/main.yml#L53-L57) as long as there are no specific `templates` defined.

repose-aggregator/tests/performance-tests/roles/repose/files/config/services/uriredaction/uri-redaction.cfg.xml
dmnjohns: Right. Personally, I'm fine with just one number right now. In my mind, we just want to capture something close to a worst-case scenario so that we can say with reasonable certainty that the service will be at least so fast in most use-cases.

Although, I'm also not against the idea of running multiple scenarios... but I certainly wouldn't require it for this story.

repose-aggregator/tests/performance-tests/test_vars/services/uriredaction.yml
dmnjohns: Ah, I see. Terrific!

dmnjohns: Man, I posted that comment a moment too soon.

I was just running the performance test to verify it, and it failed with the following:
```
failed: [perf-services-uriredaction-repose-test01] (item=system-model.cfg.xml) => {
    "changed": false,
    "item": "system-model.cfg.xml",
    "msg": "AnsibleUndefinedVariable: 'dict object' has no attribute 'filters'"
}
```

So you are right that the system model and container configs will be defaulted, but templating of the system model is currently broken if no filters are defined.

I suggest adding a default for filters the way that we have a default for services here:
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/tests/performance-tests/roles/repose/defaults/main.yml#L20

repose-aggregator/commons/commons-utilities/src/test/scala/org/openrepose/commons/utils/opentracing/ScopeHelperTest.scala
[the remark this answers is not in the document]
wdschei: It isn't in the `ScopeHelperTest`, it was in the `DistributedDatastoreServletTest` which does use the `ScopeHelper`. But I can speckle this test everywhere.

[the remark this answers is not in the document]
wdschei: The `ScopeHelperTest` has been updated to specifically check that the URI Redaction service is called as expected.

repose-aggregator/commons/commons-utilities/src/test/scala/org/openrepose/commons/utils/opentracing/httpclient/ReposeTracingRequestInterceptorTest.scala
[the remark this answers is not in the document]
wdschei: The `ReposeTracingRequestInterceptorTest` has been updated to specifically check that the URI Redaction service is called as expected.

repose-aggregator/components/services/uri-redaction-service/uri-redaction-service-impl/src/main/scala/org/openrepose/core/services/uriredaction/UriRedactionServiceImpl.scala
wdschei: This `import` was moved inside the class definition.

repose-aggregator/tests/performance-tests/roles/gatling/files/simulations/services/uriredaction/UriRedactionServiceSimulation.scala
wdschei: How would you check that? It would be in the communication with the `OpenTracing` service.

repose-aggregator/tests/performance-tests/test_vars/services/uriredaction.yml
wdschei: Empty list of `filters` added to the default system model.

repose-aggregator/tests/performance-tests/roles/gatling/files/simulations/services/uriredaction/UriRedactionServiceSimulation.scala
```diff
@@ -0,0 +1,87 @@
...
+
+package services.uriredaction
+
+import com.typesafe.config._
+import io.gatling.core.Predef._
+import io.gatling.http.Predef._
+import io.gatling.http.request.builder.HttpRequestBuilder
+
+import scala.concurrent.duration._
+import scala.util.Random
```
dmnjohns: Unused import.

repose-aggregator/tests/performance-tests/roles/gatling/files/simulations/services/uriredaction/UriRedactionServiceSimulation.scala
dmnjohns: Fair point. Not sure what I was thinking there...

dmnjohns:
I ran the performance test again after the recent change, and it still failed. This time it seems to be because the NodeJS dependencies are not being installed (which is an issue we were having with the OpenTracing performance test a week or two ago). I don't think I ever figured out why that was happening.

repose-aggregator/tests/performance-tests/roles/gatling/files/simulations/services/uriredaction/UriRedactionServiceSimulation.scala
wdschei: Removed.

dmnjohns:
I just noticed that the URL is leaked in one more place -- the `http.url` tag. That tag is set by Jaeger's implementation of a `HttpRequestInterceptor` which we extend with our own `ReposeTracingRequestInterceptor`. Therefore, all requests going through Apache HTTPClient are affected.

We should not leak the raw URL in that tag, either. Therefore, I propose that we add the following line to the `onSpanStarted(...)` method of our `ReposeTracingRequestInterceptor` class:
```
clientSpan.setTag(Tags.HTTP_URL.toString, uriRedactionService.redact(httpRequest.getRequestLine.getUri))
```

repose-aggregator/commons/commons-utilities/src/test/scala/org/openrepose/commons/utils/opentracing/httpclient/ReposeTracingRequestInterceptorTest.scala
```diff
@@ -19,81 +19,146 @@
...
+      httpRequestInterceptor.onSpanStarted(span, httpRequest, httpContext)
 
-      jaegerRequestInterceptor.process(httpRequest, httpContext)
+      verify(span).setTag(VIA, "1234")
+      verify(uriRedactionService).redact(requestLine.getUri)
+      verify(span).setTag(HTTP_URL.toString, redactedPath)
+    }
+  }
+
+  describe("testGetOperationName") {
```
dmnjohns: Ah. When I made my comment about testing `getOperationName` and `onSpanStarted`, I guess I didn't realize that they are protected methods. So you were probably right that they don't strictly need to be tested, and thus, we may not strictly need the anonymous subclass that grants access to them. Also, as you pointed out, testing them leads to some duplication.

dmnjohns:
Merge conflict with the release notes should be fixed.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/uriredaction/UriRedactionServiceKeystoneTest.groovy
```diff
@@ -0,0 +1,207 @@
...
+
+        when: "User passes a request through repose with valid token"
+        MessageChain messageChain = deproxy.makeRequest(url: reposeEndpoint + "/servers/test", method: 'GET',
+            headers: ['content-type': 'application/json', 'X-Auth-Token': fakeIdentityV2Service.client_token])
+
+        then: "The request should have reached the origin service"
+        messageChain.handlings.size() == 1
+
+        and: "keystone request contains tracing header"
+        messageChain.orphanedHandlings.each {
```
dmnjohns: Rather than sticking an `if` in the `each`, you could simplify this slightly by filtering using `findall` first:
```
// first, check to make sure it's not the tracer request (since that's done asynchronously and may count in orphaned handlings)
messageChain.orphanedHandlings.findAll { it.request.path != "/?format=jaeger.thrift" }.each {
  assert it.request.headers.contains(TRACING_HEADER)
  def traceId = URLDecoder.decode(it.request.headers.getFirstValue(TRACING_HEADER), "UTF-8")
  spanList << traceId
}
```

repose-aggregator/commons/commons-utilities/src/test/scala/org/openrepose/commons/utils/opentracing/httpclient/ReposeTracingRequestInterceptorTest.scala
wdschei: I'm good either way. The subclass is only in the test, so I'm not that concerned with it. Since it is only two methods and they have very specific behavior, I'm not opposed to testing them. If anything, then removing the dupes in favor of just making sure the two tested protected methods are called by the one public.

repose-aggregator/src/docs/asciidoc/services/uri-redaction.adoc
```diff
@@ -0,0 +1,109 @@
+= URI Redaction service
+
+The URI Redaction service enables *Repose* to redact sensitive information contained in a URI.
+
+This is extremely useful for removing user tokens from API calls prior to creating OpenTracing spans.
+The URI Redaction service can be used by any service or filter, even custom ones, that need to redact any information from a URI during processing.
+
+== Configuration
+* *Default Configuration:* uri-redaction.cfg.xml
+* *Released:* v8..?.?.?
```
dmnjohns: Since we think the next release is `8.8.4.0` (and in sprint), do we want to go ahead and populate that here?

repose-aggregator/src/docs/asciidoc/services/uri-redaction.adoc
```diff
@@ -0,0 +1,109 @@
...
+The configuration for this service is extremely simple.
+It consists of a root element and a repeating list of a single element.
+The values of the repeating `redact` elements are regular expressions (RegEx's) that need to adhere to the https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html[Java Regular Expression] syntax.
+Each of the RegEx's will be processed in order from top to bottom.
+The first RegEx will be passed the original URI.
+Then the result passed to the next RegEx and so on through the list.
+Each capture group will be replaced with the literal `XXXXX`.
+If multiple capture groups are present in a single RegEx, then all of them are redacted.
+
+[NOTE]
```
dmnjohns: Maybe a `WARNING` instead of a `NOTE`?

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/uriredaction/UriRedactionServiceKeystoneTest.groovy
```diff
@@ -0,0 +1,207 @@
...
+            assert logLines.size() == 1
+        }
+
+        and: "OpenTracingService has logged that span was sent to tracer"
+        def traceId = URLDecoder.decode(messageChain.handlings.get(0).request.headers.getFirstValue(TRACING_HEADER), "UTF-8")
+        def logLines = reposeLogSearch.searchByString("Span reported: $traceId")
+        logLines.size() == 1
+
+        and: "The sent trace doesn't have the un-redacted token in it"
+        assertUntilTrue({ assert fakeTracer.batches.collect({ it.spans}).flatten().collect({ it.operationName }).any({ it.contains('/XXXXX/bar/XXXXX') }) })
```
dmnjohns: Missing a space after `it.spans`.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/uriredaction/UriRedactionServiceKeystoneTest.groovy
```diff
@@ -0,0 +1,207 @@
...
+ */
+package features.services.uriredaction
+
+import org.openrepose.framework.test.ReposeValveTest
+import org.openrepose.framework.test.mocks.MockIdentityV2Service
+import org.openrepose.framework.test.mocks.MockTracerCollector
+import org.rackspace.deproxy.Deproxy
+import org.rackspace.deproxy.MessageChain
+
+class UriRedactionServiceKeystoneTest extends ReposeValveTest {
```
dmnjohns: I would love a test case capturing the behavior when capture text appears more than once in a URI. So like, using the already defined regex `/[a-zA-Z]+/specific/(path)`, if the URI is `/path/specific/(path)`, I would expect the redacted URI to be `/path/specific/XXXXX` and not `/XXXXX/specific/XXXXX`.

repose-aggregator/commons/commons-utilities/src/test/scala/org/openrepose/commons/utils/opentracing/httpclient/ReposeTracingRequestInterceptorTest.scala
dmnjohns: I guess I'm also good with it either way.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/uriredaction/UriRedactionServiceKeystoneTest.groovy
wdschei: Added this during the `Unroll` flattening.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/uriredaction/UriRedactionServiceKeystoneTest.groovy
wdschei: I ran the formatter on the file and then decided to flatten this into a `Unroll` for readability.

repose-aggregator/src/docs/asciidoc/services/uri-redaction.adoc
wdschei: Ok. Updated.

repose-aggregator/src/docs/asciidoc/services/uri-redaction.adoc
wdschei: Sure. Updated.

wdschei:
Retest this please.

PR #1913, opened 2018-04-16
WIP: REP-6839 Updated Attribute Mapper from v2.2.0 to v2.2.1.

wdschei:


dmnjohns:
These jobs "failed" because I killed them to work on upgrading Jenkins. I will kick them again once that's all settled.

wdschei:
This was named incorrectly and therefore would be referenced improperly in Jira.
It has been moved to #1914

PR #1915, opened 2018-04-23
REP-6878 Regex RBAC Perf Test Fix

dmnjohns:
…in environments

### DIFF
--- a/repose-aggregator/tests/performance-tests/roles/gatling/files/simulations/filters/regexrbac/RegexRbacFilterSimulation.scala
+++ b/repose-aggregator/tests/performance-tests/roles/gatling/files/simulations/filters/regexrbac/RegexRbacFilterSimulation.scala
@@ -110,7 +110,7 @@ class RegexRbacFilterSimulation extends Simulation {
   def getFooBarBazSpace: HttpRequestBuilder = http(session => session.scenario)
     .get("/foo/bar/baz")
     .header(HttpHeaderNames.Host, "localhost")
-    .header("X-Roles", "role with space")
+    .header("X-Roles", "role5")
     .check(status.is(StatusCodeOk))
 }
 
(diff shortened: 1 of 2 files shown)


adrianjgeorge:
retest this please

repose-aggregator/tests/performance-tests/roles/gatling/files/simulations/filters/regexrbac/RegexRbacFilterSimulation.scala
```diff
@@ -110,7 +110,7 @@ class RegexRbacFilterSimulation extends Simulation {
   def getFooBarBazSpace: HttpRequestBuilder = http(session => session.scenario)
     .get("/foo/bar/baz")
     .header(HttpHeaderNames.Host, "localhost")
-    .header("X-Roles", "role with space")
```
adrianjgeorge: this seems awfully intentional to be removing. What's your reasoning for dropping it?

dmnjohns: I believe it was intentional, but it is what was causing this performance test to bomb. For whatever reason, keeping the role with non-breaking spaces in the configuration broke parsing of the configuration, causing Repose to always return 5xx responses, causing the test to fail. Since this is a performance test rather than a unit/system/functional test, I was less concerned about testing the functionality of the filter than the performance. I should note that we do have functional tests verifying the behavior around non-breaking spaces. To best keep our old numbers for this test relevant, I tried to make a minimal change, which was just substituting out the role that broke things with a different role.

dmnjohns: Oh, I should add that I ran Repose with the unmodified configuration file locally and did not see the configuration parsing issue that I was seeing on the nightly performance test job.

PR #1916, opened 2018-04-23
Prep for v8.8.4.0 release.

wdschei:


wdschei:
Since this was coming in so quick, I went ahead and put in v8.8.4.0 after discussions with @dmnjohns during the code review. REP-6850 is not needed since the version didn't change.

PR #1924, opened 2018-05-02
REP-6882 (REP-5334) Collect JMX Metrics for Performance Tests

dmnjohns:
This PR requires the companion infrastructure PR at:
https://github.com/rackerlabs/repose-infrastructure-ng/pull/119

### DIFF
--- a/repose-aggregator/src/docs/asciidoc/architecture/performance.adoc
+++ b/repose-aggregator/src/docs/asciidoc/architecture/performance.adoc
@@ -404,7 +404,7 @@ ansible-playbook start_test.yml \
   -e '{perf_test: filters/saml,
        repose: {
          systemd_opts: {
-           java_opts: "-Xmx2g -Xms2g -XX:+UseG1GC -XX:+UseStringDeduplication"}}}' \
+           java_opts: ["-Xmx2g", "-Xms2g", "-XX:+UseG1GC", "-XX:+UseStringDeduplication"]}}}' \
   -vvvv
 ----
 
--- a/repose-aggregator/tests/performance-tests/group_vars/all.yml
+++ b/repose-aggregator/tests/performance-tests/group_vars/all.yml
@@ -1,3 +1,5 @@
+# TODO: Add variable(s) for jmxtrans URL/version
+# TODO: Add variable(s) for jmxtrans configuration (e.g., InfluxDB connection information)
 {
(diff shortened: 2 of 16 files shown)


adrianjgeorge:
why are we supporting deb and rpm stuff?

repose-aggregator/src/docs/asciidoc/architecture/performance.adoc
```diff
@@ -404,7 +404,7 @@ ansible-playbook start_test.yml \
   -e '{perf_test: filters/saml,
        repose: {
          systemd_opts: {
-           java_opts: "-Xmx2g -Xms2g -XX:+UseG1GC -XX:+UseStringDeduplication"}}}' \
+           java_opts: ["-Xmx2g", "-Xms2g", "-XX:+UseG1GC", "-XX:+UseStringDeduplication"]}}}' \
```
wdschei: Why the move here to an array?

dmnjohns: Oh man, that change was from one of prototype commits, so I can't say for certain... but thinking about it, I suspect it was for three reasons:

1. I found an array easier to read when dealing with a bunch of options than a string.
1. I found an array easier to manipulate (in this case, add elements to) than a string. This probably came up when I was trying to figure out where to put the JMX options. I probably, at one point, took whatever `java_opts` were present at the time and added the new JMX options to that array. Sure, I could have used string concatenation, but then it would be possible to end up with a space at the start of the `java_opts` string if no `java_opts` were previously defined, which isn't a big deal, I just don't like the imprecision of it. Of course, that isn't how things are being done at this point, but I probably left it because of reason 1.
1. Arrays are probably easier to use for control flow, if necessary. For example, the Ansible `with_items` control just works on an array.

On the other hand, I realize that using a string rather than an array is less verbose and arguable simpler.

So I guess, in summary, a string provide less verbosity while an array provides more flexibility. Although, for maximum verbosity and flexibility, a dictionary may be the best way to go. That would allow additionally filtering possibilities, and thus, is probably the most flexible.

For this change, at this time, I don't overly care one way or the other. Let me know if you'd like me to change this.

wdschei: I'm good with it. I was just curious what the thought process was.

PR #1925, opened 2018-05-02
REP-6826 Update Jaeger version.

wdschei:
Since the latest version of Jaeger doesn't directly support the Apache HTTP Client anymore, I borrowed the old classes and tailored them for Repose until we can update our use of the Apache HTTP Client to use the non-deprecated builder API.


wdschei:
Retest this please.

wdschei:
Retest this please.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/opentracing/httpclient/ReposeTracingRequestInterceptor.scala
```diff
@@ -19,47 +19,77 @@
...
-    s"${httpRequest.getRequestLine.getMethod} ${uriRedactionService.redact(httpRequest.getRequestLine.getUri)}"
+      Option(httpRequest.getFirstHeader(REQUEST_ID))
+        .map(_.getValue)
+        .foreach(clientSpan.setTag(REQUEST_ID, _))
+      Option(httpRequest.getFirstHeader(VIA))
+        .map(_.getValue)
+        .foreach(clientSpan.setTag(VIA, _))
+      clientSpan.setTag(ReposeVersion, reposeVersion)
+
+      httpContext.setAttribute("io.opentracing.Span", clientSpan)
```
dmnjohns: Extract this to a constant somewhere so that it is always correct between the request and response interceptors?

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/opentracing/httpclient/ReposeTracingRequestInterceptor.scala
```diff
@@ -19,47 +19,77 @@
...
-    super.process(httpRequest, httpContext)
+    try {
+      val currentActiveSpan = tracer.activeSpan
+      val requestLine = httpRequest.getRequestLine
+      val redactedUri = uriRedactionService.redact(requestLine.getUri)
 
-    spanInjectionInterceptor.process(httpRequest, httpContext)
-  }
+      val clientSpanBuilder = tracer.buildSpan(s"${requestLine.getMethod} $redactedUri")
+      if (currentActiveSpan != null) clientSpanBuilder.asChildOf(currentActiveSpan)
```
dmnjohns:
```
Option(currentActiveSpan).foreach(clientSpanBuilder.asChildOf)
```

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/opentracing/httpclient/ReposeTracingRequestInterceptor.scala
```diff
@@ -19,47 +19,77 @@
...
+        .map(_.getValue)
+        .foreach(clientSpan.setTag(REQUEST_ID, _))
+      Option(httpRequest.getFirstHeader(VIA))
+        .map(_.getValue)
+        .foreach(clientSpan.setTag(VIA, _))
+      clientSpan.setTag(ReposeVersion, reposeVersion)
+
+      httpContext.setAttribute("io.opentracing.Span", clientSpan)
+
+      if (tracer.scopeManager.active == null) logger.warn("Current scope is null; possibly failed to start client tracing span.")
```
dmnjohns: Ugh. This is the case where Scala's `Option` always lets me down. I would like to take some action if an `Option` is `None`, but the ways we have to do that aren't the most elegant:
```
Option(tracer.scopeManager.active).getOrElse(logger.warn("Current scope is null; possibly failed to start client tracing span."))
```
```
if (Option(tracer.scopeManager.active).isEmpty) logger.warn("Current scope is null; possibly failed to start client tracing span.")
```
```
Option(tracer.scopeManager.active) match {
  case None => logger.warn("Current scope is null; possibly failed to start client tracing span.")
  case _ => // do nothing
}
```

All of which is debatably better than a straight null check.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/opentracing/httpclient/ReposeTracingResponseInterceptor.scala
```diff
@@ -0,0 +1,53 @@
...
+import java.io.IOException
+
+import com.typesafe.scalalogging.slf4j.StrictLogging
+import io.opentracing.Span
+import io.opentracing.tag.Tags._
+import org.apache.http._
+import org.apache.http.protocol.HttpContext
+
+/**
+  * An [[org.apache.http.HttpResponseInterceptor]] that will enrich Repose HTTP requests made through an
```
dmnjohns: "requests" -> "responses".

repose-aggregator/artifacts/valve/src/config/docs/DEPENDENCIES.txt
```diff
@@ -29,10 +29,9 @@ com.typesafe.akka                                   akka-actor_2.11           2.
...
 com.typesafe.scala-logging                          scala-logging-slf4j_2.11  2.1.2            Apache License Version 2.0  https://github.com/typesafehub/scala-logging
-com.uber.jaeger                                     jaeger-apachehttpclient   0.24.0           Apache License Version 2.0  https://github.com/jaegertracing/jaeger-client-java
-com.uber.jaeger                                     jaeger-core               0.24.0           Apache License Version 2.0  https://github.com/jaegertracing/jaeger-client-java
 io.dropwizard.metrics                               metrics-core              3.2.0            Apache License Version 2.0  https://github.com/dropwizard/metrics/blob/v3.2.0/LICENSE
 io.dropwizard.metrics                               metrics-graphite          3.2.0            Apache License Version 2.0  https://github.com/dropwizard/metrics/blob/v3.2.0/LICENSE
+io.jaegertracing                                    jaeger-core               0.26.0           Apache License Version 2.0  https://github.com/jaegertracing/jaeger-client-java
```
dmnjohns: Looks like the version was actually bumped to `0.27`?

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/opentracing/httpclient/ReposeTracingRequestInterceptor.scala
wdschei: Yep. I can't believe I missed that one. Updated.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/opentracing/httpclient/ReposeTracingRequestInterceptor.scala
wdschei: Updated.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/opentracing/httpclient/ReposeTracingRequestInterceptor.scala
wdschei: I know the `null` check is not the preferred way of doing this, but the only other option that seems to convey the same clarity is the second one. The first is performing the `get` for no reason and makes it fuzzy of what the intent is and the third is just so much extra for no gain. They become convention for convention's sake and make them harder to gain the actual intent.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/opentracing/httpclient/ReposeTracingResponseInterceptor.scala
wdschei: I was taking this from the perspective that it was still the request that was the cause of the enrichment, but yeah. Whatever. Updated.

repose-aggregator/artifacts/valve/src/config/docs/DEPENDENCIES.txt
wdschei: Updated.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/opentracing/httpclient/ReposeTracingRequestInterceptor.scala
dmnjohns: I agree, which is why I dislike the ways that Scala gives us to handle this scenario. I know other Scala folks have used `implicit`s to extend `Option` to provide an `ifNone` (or equivalent) method that invokes some function if the `Option` is `None`, which is something I've been tempted to do...

At any rate, not a big deal, do what seems best here.

PR #1930, opened 2018-05-10
REP-6858 Updated log4j

adrianjgeorge:
This intentionally doesn't go to latest. Turns out latest breaks us in a number of ways that are kind of a pain in the butt. So i went to the minimal version that got us the feature we wanted.


dmnjohns:
Not that it overly matters, but you've got a typo in your commit message.

jog4j -> log4j

repose-aggregator/src/docs/asciidoc/release-notes.adoc
```diff
@@ -4,6 +4,7 @@
 * https://repose.atlassian.net/browse/REP-6862[REP-6862] - Updated dependencies:
 ** Jaeger Core: 0.24.0 → 0.27.0
 *** https://github.com/jaegertracing/jaeger-client-java/blob/master/CHANGELOG.rst#0270-2018-04-18[Jaeger Client Java v0.27.0 change log]
+* https://repose.atlassian.net/browse/REP-6858[REP-6858] - Updated Log4j to a version that allows setting of file permissions
```
dmnjohns: Do we want to be explicit about the version of log4j we upgraded to?

dmnjohns:
Oh, the Java `Files` API does actually give us a way to check ownership and permissions of files (at least in POSIX-compliant environments):
https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#getPosixFilePermissions-java.nio.file.Path-java.nio.file.LinkOption...-

So we can probably write a test to verify this change as part of the QE task.

dmnjohns:
Merge conflict needs to be resolve.

wdschei:
I agree cleaning up the docs around this needs to be done, I'm just not sure that it is in scope of this story.

At the same time, it does seem silly to have a story to do nothing more than add a line or two and clean up some links.

wdschei:
After talking with @adrianjgeorge yesterday, the short term solution is to just updated the outdated links in the docs. I'll finish the review after they are updated.

PR #1931, opened 2018-05-11
REP-6920 Improve Exception Handling in Attribute Mapper

dmnjohns:


repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
```diff
@@ -474,24 +474,20 @@ class SamlPolicyTranslationFilter @Inject()(configurationService: ConfigurationS
           mapper.getFactory.configure(JsonGenerator.Feature.ESCAPE_NON_ASCII, true)
           Option(new ByteArrayInputStream(mapper.writeValueAsBytes(destination)))
         } catch {
-          case jpe: JsonProcessingException =>
-            throw SamlPolicyException(SC_BAD_GATEWAY, "Origin service provided bad response", jpe)
+          case e@(_: AttributeMapperException | _: JsonProcessingException) =>
```
dmnjohns: So this is one change where the exceptions being handled don't quite map one-to-one. I considered changing this to:
```
e@(CausedBy(_: JsonProcessingException) | _: JsonProcessingException)
```

I chose not to because the XML block below catches `SAXException` and `SaxonApiException`, both of which would come from a failed translation. Those were not being caught in this (the JSON) block, so with the updates, I made the two blocks catch the same stuff.

As for the `JsonProcessingException`, that is because we are using a function of `AttributeMapper` that was not covered by the exception handling update. Since we are using that function to parse the JSON into the appropriate data structure, we still need to handle that exception.

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
```diff
@@ -474,24 +474,20 @@ class SamlPolicyTranslationFilter @Inject()(configurationService: ConfigurationS
...
             translatedSamlResponse,
             validate = true,
             XSDEngine.AUTO.toString)
           Option(convertDocumentToStream(destination))
         } catch {
-          case se@(_: SAXException | _: SaxonApiException) =>
-            throw SamlPolicyException(SC_BAD_GATEWAY, "Origin service provided bad response", se)
-        } finally {
-          docBuilder.foreach(XMLParserPool.returnParser)
+          case ame: AttributeMapperException =>
```
dmnjohns: This is another change that does not quite map one-to-one. `AttributeMapperException` is broader than either `SAXException` or `SaxonApiException`, however, I did not want to make things complicated by checking the causes everywhere. Since the intent of this error handling seemed to be to fail when Attribute Mapper fails, I just caught the general exception.

repose-aggregator/components/filters/attribute-mapping-policy-validation-filter/src/main/scala/org/openrepose/filters/attributemapping/AttributeMappingPolicyValidationFilter.scala
```diff
@@ -72,11 +69,14 @@ class AttributeMappingPolicyValidationFilter extends Filter with StrictLogging {
             httpServletResponse.sendError(
               SC_UNSUPPORTED_MEDIA_TYPE,
               ucte.message)
-          case e@(_: SaxonApiException | _: TransformerException | _: JsonProcessingException | _: JsonParseException) =>
+          case e@(_: AttributeMapperException) =>
             logger.debug("Validation failed", e)
             httpServletResponse.sendError(
               SC_BAD_REQUEST,
               "Failed to validate attribute mapping policy in request")
+          case t =>
```
adrianjgeorge: I guess it was for this?

adrianjgeorge: What's the behavior if it goes unmatched?

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlPolicyTranslationFilter.scala
adrianjgeorge: I thought you were going to update those other calls? I'm guessing you didn't.

dmnjohns: I did, actually. The reason we are still catching the `JsonProcessingException` here is that we don't only use the `AttributeMapper` call. We also call `ObjectMapper.writeValueAsBytes` on our own object, which can also throw that exception.

repose-aggregator/components/filters/attribute-mapping-policy-validation-filter/src/main/scala/org/openrepose/filters/attributemapping/AttributeMappingPolicyValidationFilter.scala
```diff
@@ -72,11 +69,14 @@ class AttributeMappingPolicyValidationFilter extends Filter with StrictLogging {
             httpServletResponse.sendError(
               SC_UNSUPPORTED_MEDIA_TYPE,
               ucte.message)
-          case e@(_: SaxonApiException | _: TransformerException | _: JsonProcessingException | _: JsonParseException) =>
+          case e@(_: AttributeMapperException) =>
```
wdschei: Why use the muti-exception pattern here now that it is a single?

dmnjohns: Great question. No reason. Dropping this down to a single exception was one of the last things I did, and it must not have registered that I could simplify this when I dropped the other exception.

Will fix.

adrianjgeorge:
retest this please

PR #1933, opened 2018-05-16
REP-6818 Confirm XML SAML Support

wdschei:
This was just to add/update missing tests around XML support now that it is becoming the more common format due to the ability to add comments.

### DIFF
--- a/repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlIdentityClient.scala
+++ b/repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlIdentityClient.scala
@@ -22,10 +22,10 @@ package org.openrepose.filters.samlpolicy
 import java.nio.charset.StandardCharsets
 import java.time.format.DateTimeFormatter
 import java.time.{ZoneId, ZonedDateTime}
+
 import javax.inject.{Inject, Named}
 import javax.servlet.http.HttpServletResponse.{SC_OK, SC_REQUEST_ENTITY_TOO_LARGE, SC_UNAUTHORIZED}
 import javax.ws.rs.core.{HttpHeaders, MediaType}
-
 import org.openrepose.commons.utils.http.{CommonHttpHeader, ServiceClientResponse}
 import org.openrepose.core.services.serviceclient.akka.{AkkaServiceClient, AkkaServiceClientFactory}
 import org.springframework.web.util.UriUtils
@@ -85,7 +85,7 @@ class SamlIdentityClient @Inject()(akkaServiceClientFactory: AkkaServiceClientFa
     *
     * @param username the username on the account
     * @param password the password on the account
-    * @param traceId an optional identifier to be sent with the request
+    * @param traceId  an optional identifier to be sent with the request
     * @return the token if successful, or a failure if unsuccessful
     */
   def getToken(username: String, password: String, traceId: Option[String]): Try[String] = {
@@ -112,9 +112,8 @@ class SamlIdentityClient @Inject()(akkaServiceClientFactory: AkkaServiceClientFa
         serviceClientResponse.getStatus match {
           case SC_OK =>
             Try {
-              val responseEncoding = serviceClientResponse.getHeaders
-                .find(hdr => HttpHeaders.CONTENT_TYPE.matches(hdr.getName))
-                .map(_.getElements.head.getParameterByName("charset"))
+              val responseEncoding = serviceClientResponse.getHeaderElements(HttpHeaders.CONTENT_TYPE).asScala.headOption
+                .map(_.getParameterByName("charset"))
                 .flatMap(Option.apply)
                 .map(_.getValue)
                 .getOrElse(StandardCharsets.ISO_8859_1.name())
@@ -138,8 +137,8 @@ class SamlIdentityClient @Inject()(akkaServiceClientFactory: AkkaServiceClientFa
   /**
     * Retrieves a IDP-ID from Identity that will be used to fetch the policy.
     *
-    * @param issuer the issuer associated with the Identity provider
-    * @param traceId an optional identifier to be sent with the request
+    * @param issuer     the issuer associated with the Identity provider
+    * @param traceId    an optional identifier to be sent with the request
     * @param checkCache whether or not to use the HTTP request cache
     * @return the IDP ID if successful, or a failure if unsuccessful
     */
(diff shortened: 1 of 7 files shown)


repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/attributemappingvalidation/AttributeMappingPolicyValidationFilterTest.groovy
```diff
@@ -83,7 +83,7 @@ class AttributeMappingPolicyValidationFilterTest extends ReposeValveTest {
         mc.receivedResponse.code.toInteger() == SC_UNSUPPORTED_MEDIA_TYPE
 
         where:
-        contentType << [TEXT_PLAIN]
+        contentType << [TEXT_XML, APPLICATION_ATOM_XML, TEXT_PLAIN, APPLICATION_JSON]
```
dmnjohns: I guess I will wait to see if these pass, but I wouldn't expect them to? The `415` coming out of this filter should only be sent in the case where we know we don't accept the `Content-Type` (e.g., `text/plain`) but not for the case where the content does not match the `Content-Type` (e.g., `application/json`). In the latter case, I would actually expect a `400` response since the request is... well... bad.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/samlpolicy/SamlAttributeMappingTest.groovy
```diff
@@ -414,15 +422,23 @@ class SamlAttributeMappingTest extends ReposeValveTest {
...
         given: "a mapping policy an extended attribute with no value"
         def extUserAttribName = "foo"
         def extUserAttribValues = ["{0}"]
         def remotePath = "()"
         def remoteValue = [[path: remotePath]]
-        def mappingPolicy = createMappingJsonWithValues(
-            userExtAttribs: [(extUserAttribName): extUserAttribValues],
-            remote: remoteValue)
+        def mappingPolicy = contentType == APPLICATION_XML ?
+            createMappingJsonWithValues(
```
dmnjohns: Should be creating mapping XML?

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/samlpolicy/SamlAttributeMappingTest.groovy
```diff
@@ -414,15 +422,23 @@ class SamlAttributeMappingTest extends ReposeValveTest {
...
+                userExtAttribs: [(extUserAttribName): extUserAttribValues],
+                remote: remoteValue) : contentType == APPLICATION_JSON ?
+            createMappingJsonWithValues(
+                userExtAttribs: [(extUserAttribName): extUserAttribValues],
+                remote: remoteValue) :
+            createMappingYamlWithValues(
+                userExtAttribs: [(extUserAttribName): extUserAttribValues],
+                remote: remoteValue)
```
dmnjohns: At this point, I feel like using a standard try/catch block would make the intent and behavior clearer, but if you do that the standard Java way, you end up with some assignment repetition:
```
def mappingPolicy
if (contentType == APPLICATION_XML) {
    mappingPolicy = createMappingXmlWithValues(
                userExtAttribs: [(extUserAttribName): extUserAttribValues],
                remote: remoteValue)
} else if (contentType == APLICATION_JSON) {
    mappingPolicy = createMappingJsonWithValues(
                userExtAttribs: [(extUserAttribName): extUserAttribValues],
                remote: remoteValue)
} else {
    mappingPolicy = createMappingYamlWithValues(
                userExtAttribs: [(extUserAttribName): extUserAttribValues],
                remote: remoteValue)
}
```

Instead, you could use a code block, but Groovy doesn't make that super easy since code blocks and closures share syntax. So... you might actually have to use a closure or a hacky construct:
```
def mapingPolicy = {
    if (contentType == APPLICATION_XML) {
        createMappingXmlWithValues(
            userExtAttribs: [(extUserAttribName): extUserAttribValues],
            remote: remoteValue)
    } else if (contentType == APLICATION_JSON) {
        createMappingJsonWithValues(
            userExtAttribs: [(extUserAttribName): extUserAttribValues],
            remote: remoteValue)
    } else {
        createMappingYamlWithValues(
            userExtAttribs: [(extUserAttribName): extUserAttribValues],
            remote: remoteValue)
    }
}()
```

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/attributemappingvalidation/AttributeMappingPolicyValidationFilterTest.groovy
wdschei: Yes, it should and yes, it did. Updated.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/samlpolicy/SamlAttributeMappingTest.groovy
wdschei: OBE based on the next finding.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/samlpolicy/SamlAttributeMappingTest.groovy
wdschei: Completely agree with extracting this to a method for reuse.

repose-aggregator/components/filters/saml-policy-translation-filter/src/main/scala/org/openrepose/filters/samlpolicy/SamlIdentityClient.scala
```diff
@@ -254,8 +250,7 @@ object SamlIdentityClient {
     (idpId: String) => s"/v2.0/RAX-AUTH/federation/identity-providers/$idpId/mapping"
 
   def buildRetryValue(response: ServiceClientResponse): String = {
-    response.getHeaders
-      .find(hdr => HttpHeaders.RETRY_AFTER.matches(hdr.getName))
+    response.getHeaderElements(HttpHeaders.RETRY_AFTER).asScala.headOption
```
dmnjohns: This does not map 1:1 with the old behavior. I'm slightly worried that since the `Retry-After` header may contain a comma as part of a date, and not to delimit multiple values, this will mess with stuff. Is there a reason you used `getHeaderElements` rather than `getHeaders`?

wdschei: You would be correct sir. Good catch. This one should have only been `getHeaders()`.

PR #1939, opened 2018-06-06
REP-6980 Tagging Performance Test Metrics with Repose Version

dmnjohns:
See also:
https://github.com/rackerlabs/repose-infrastructure-ng/pull/124


repose-aggregator/tests/performance-tests/roles/gatling/tasks/install_gatling.yml
```diff
@@ -49,7 +49,7 @@
     state: absent
 
 - name: Copy Gatling configuration
-  copy:
+  template:
```
wdschei: I like that it uses the section (e.g. copy, template) to determine where the file should be. That way you could copy a template to be interpreted by a different process later. Someone was using their noodle.

PR #1942, opened 2018-06-06
REP-6038 Initial implementation of ReposeFilterChain

adrianjgeorge:


adrianjgeorge:
Docs are still to come, and i think i may add a little more error handling.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/intrafilterlogging/RequestLog.java
```diff
@@ -42,10 +40,10 @@
     String requestBody;
     Map<String, String> headers;
 
-    public RequestLog(HttpServletRequest httpServletRequest, Filter filter) throws IOException {
+    public RequestLog(HttpServletRequest httpServletRequest, String currentFilter) throws IOException {
         preamble = "Intrafilter Request Log";
         timestamp = new DateTime().toString();
-        currentFilter = StringUtils.isEmpty(filter.getId()) ? filter.getName() : filter.getId() + "-" + filter.getName();
+        this.currentFilter = currentFilter;
```
adrianjgeorge: This change may make some functional tests fail, but it doesn't seem like anyone was using the id field anyway.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/intrafilterlogging/ResponseLog.java
```diff
@@ -37,10 +35,10 @@
     String responseBody;
     Map<String, String> headers;
 
-    public ResponseLog(HttpServletResponseWrapper wrappedServletResponse, Filter filter) throws IOException {
+    public ResponseLog(HttpServletResponseWrapper wrappedServletResponse, String currentFilter) throws IOException {
         preamble = "Intrafilter Response Log";
         timestamp = new DateTime().toString();
-        currentFilter = StringUtils.isEmpty(filter.getId()) ? filter.getName() : filter.getId() + "-" + filter.getName();
+        this.currentFilter = currentFilter;
```
adrianjgeorge: Same here.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeFilterChain.scala
```diff
@@ -0,0 +1,135 @@
...
+            })
+          })
+        } else {
+          logger.debug("Skipping filter: {}", head.filterName)
+          runNext(tail, request, response)
+        }
+    }
+  }
+
+  def doIntrafilterLogging(request: HttpServletRequest, response: HttpServletResponse, filter: String, requestProcess: (HttpServletRequest, HttpServletResponse) => Unit): Unit = {
```
adrianjgeorge: I decided to do this and metrics like this so that what we do for logging, metrics and routing are a little more clear. since they are all isolated.

wdschei:
It looks like something got tweaked in the Intrafilter Logging. The two failures are just inverted.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeFilterChain.scala
dmnjohns: A PR with these changes a bit more isolated:
https://github.com/rackerlabs/repose/pull/1943/files

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeFilterChain.scala
```diff
@@ -0,0 +1,135 @@
...
+  def runNext(chain: List[FilterContext], request: HttpServletRequest, response: HttpServletResponse): Unit = {
+    chain match {
+      case Nil =>
+        doIntrafilterLogging(request, response, "origin", (intraRequest, intraResponse) => {
+          doMetrics(intraRequest, intraResponse, "origin", (metricsRequest, metricsResponse) => {
+            logger.debug("End of the filter chain reached")
+            originalChain.doFilter(metricsRequest, metricsResponse)
+          })
+        })
+      case head::tail =>
```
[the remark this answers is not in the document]
adrianjgeorge: Do we? Style didn't ping it. I feel like i do when i'm building a list, but for whatever reason when deconstructing it i like it better without? Weird.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeFilterChain.scala
```diff
@@ -0,0 +1,135 @@
...
+            doMetrics(intraRequest, intraResponse, head.filterName, (metricsRequest, metricsResponse) => {
+              logger.debug("Entering filter: {}", head.filterName)
+              head.filter.doFilter(metricsRequest, metricsResponse, new ReposeFilterChain(tail, originalChain, None, metricsRegistry))
+            })
+          })
+        } else {
+          logger.debug("Skipping filter: {}", head.filterName)
+          runNext(tail, request, response)
+        }
+    }
```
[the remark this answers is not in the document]
adrianjgeorge: A guard is probably reasonable

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeFilterChain.scala
```diff
@@ -0,0 +1,135 @@
...
+    bypassUrlRegex.map(_.r.pattern.matcher(request.getRequestURI).matches()) match {
+      case Some(true) =>
+        logger.debug("Bypass url hit")
+        runNext(List.empty, request, response)
+      case _ =>
+        runNext(filterChain, request, response)
+    }
+  }
+
+  def runNext(chain: List[FilterContext], request: HttpServletRequest, response: HttpServletResponse): Unit = {
```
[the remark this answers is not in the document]
adrianjgeorge: I feel like the recursion while slightly more expensive is super easy to read and follow. While all the other's involve looking at multiple other pieces of code to figure out what they are doing, and i think the logging is actually explicitly part of the acceptance criteria.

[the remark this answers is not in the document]
adrianjgeorge: If we see a performance hit, i'd be willing to switch it, but until then...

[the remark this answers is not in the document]
dmnjohns: After looking at it more, I think I agree with recursion being the easier approach to read and grok.

repose-aggregator/src/docs/asciidoc/architecture/system-model.adoc
```diff
@@ -32,6 +32,7 @@ Filter::
 An HTTP message interceptor that provides a specific piece of business functionality.
 A filter contains a *Repose* specific named reference.
 
+// <<filter-chain.adoc#,Filters (Filter Chain)>>::
```
dmnjohns: Huh? Why is this commented out?

repose-aggregator/src/docs/asciidoc/architecture/filter-chain.adoc
```diff
@@ -0,0 +1,106 @@
...
+.system-model.cfg.xml (partial)
+----
+...
+<filters>
+    <filter name="keystone-v2" uri-regex=".*/foo"/> <!--1-->
+    <filter name="api-validator"/> <!--2-->
+</filters>
+...
+----
+<1> This filter will only be run when the request is to the `/foo` resource.
```
dmnjohns: Technically, this would also be run for the `/test/foo` resource, `/bar/baz/foo` resource, etc. You may want to either reword this callout, or drop the `.*` from the regex.

repose-aggregator/src/docs/asciidoc/architecture/filter-chain.adoc
```diff
@@ -0,0 +1,106 @@
...
+The *Repose* filter chain allows you to determine if a filter should be run while the request is processing.
+
+=== URI Regex ===
+
+[NOTE]
+====
+This method of determining if a filter should be run is *DEPRECATED* and will be removed in *Repose* 10.
+====
+
+You can decide whether or not a filter should run based on it's URI using a https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html[regular expression].
```
dmnjohns: I think technically we only match the `uri-regex` against the path from the URI right now. That might be worth noting so that folks aren't trying to match the protocol, host, or port portion of the URI.

repose-aggregator/src/docs/asciidoc/architecture/system-model.adoc
dmnjohns: OIC. You don't want the filter chain doc to be reachable (at least by navigation) until it goes into effect?

repose-aggregator/src/docs/asciidoc/architecture/filter-chain.adoc
```diff
@@ -0,0 +1,106 @@
+= Filter Chain
```
dmnjohns: We probably also want a link to this page (perhaps commented our for the time being) in the architecture `index.adoc`?

repose-aggregator/src/docs/asciidoc/architecture/filter-chain.adoc
```diff
@@ -0,0 +1,106 @@
...
+<2> This filter will be run everytime.
+
+It's worth noting that if anything changes the request uri while processing the filter chain that will potentially impact whether a filter is run.
+Or in other words the filters to be run is not determined up front, but instead evaluated when the request hits a given filter in the chain.
+
+=== New Fancy Criteria ===
+
+TODO: Put some info here
+
+== Intrafilter Logging ==
```
dmnjohns: We also have a section on Intrafilter Logging in `troubleshooting.adoc`. Having it here seems legit since we can go into more of the implementation details, but I wonder if it's worth linking between the two. Maybe a one way link from here to the user description?

repose-aggregator/src/docs/asciidoc/architecture/system-model.adoc
adrianjgeorge: Correct

repose-aggregator/src/docs/asciidoc/architecture/filter-chain.adoc
adrianjgeorge: I thought we'd eventually add a page that actually described the overall architecture of repose, and it'd be listed there.

dmnjohns: Ah, I see. Actually, I guess I kind of did that with the "Servlet Specification" page. The intent was to detail how Repose interacts with the servlet specification. It totally links to the filter chain page.

Even so, I would certainly not be opposed to having this page in the index in addition to being linked to by that page. Personally, I like complete indices that I can reach all pages from.

PR #1944, opened 2018-06-08
Ability to support x-forwarded-for header for x-pp-groups in ip-user filter

dimtruck:
Currently, only source IP from the calling service is matched against the CIDR
list in ip-user filter.  This is a limitation for those products that use a
load balancer in front of repose since that load balancer's ip is used instead
of the calling service's ip.  This change will mimic the behavior for x-pp-user
header population in the same filter by supporting x-forwarded-for header used
to overwrite the source ip.

### DIFF
--- a/repose-aggregator/components/filters/ip-user-filter/src/test/scala/org/openrepose/filters/ipuser/IpUserFilterTest.scala
+++ b/repose-aggregator/components/filters/ip-user-filter/src/test/scala/org/openrepose/filters/ipuser/IpUserFilterTest.scala
@@ -84,15 +84,98 @@ class IpUserFilterTest extends FunSpec with BeforeAndAfterEach with Matchers {
     }
   }
 
+  describe("setting the request x-pp-groups header") {
+    it("should use the value of the remote IP address of the request when no X-Forwarded-For header is included and group matches") {
+      servletRequest.setRemoteAddr("10.55.66.77")
+
+      ipUserFilter.doFilter(servletRequest, servletResponse, filterChain)
+
+      getPostFilterRequest.getHeader("x-pp-groups") shouldEqual "my-awesome-group" + DefaultQuality
+    }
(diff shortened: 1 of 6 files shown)


repose-aggregator/components/filters/ip-user-filter/src/test/scala/org/openrepose/filters/ipuser/IpUserFilterTest.scala
```diff
@@ -84,15 +84,98 @@ class IpUserFilterTest extends FunSpec with BeforeAndAfterEach with Matchers {
...
+    }
+
+    it("group should be null when there are multiple non-split values in the X-Forwarded-For header - last match") {
+      servletRequest.setRemoteAddr("10.1.2.3")
+      servletRequest.addHeader("X-Forwarded-For", "10.1.1.1")
+      servletRequest.addHeader("X-Forwarded-For", "10.55.66.77")
+
+      ipUserFilter.doFilter(servletRequest, servletResponse, filterChain)
+
+      getPostFilterRequest.getHeader("x-pp-groups") should be (null)
```
adrianjgeorge: why would this be null?

adrianjgeorge: I see you changed the config being tested.

adrianjgeorge:
Release notes also need updating

repose-aggregator/components/filters/ip-user-filter/src/test/scala/org/openrepose/filters/ipuser/IpUserFilterTest.scala
dimtruck: That header doesn't populate if nothing matches.

PR #1948, opened 2018-06-12
REP-7008 The New ReposeServlet with Routing

dmnjohns:


repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeServlet.scala
```diff
@@ -0,0 +1,48 @@
...
+package org.openrepose.powerfilter
+
+import com.typesafe.scalalogging.slf4j.StrictLogging
+import javax.inject.{Inject, Named}
+import javax.servlet.http.HttpServletResponse.SC_INTERNAL_SERVER_ERROR
+import javax.servlet.http.{HttpServlet, HttpServletRequest, HttpServletResponse}
+import org.openrepose.commons.utils.servlet.http.{HttpServletRequestWrapper, HttpServletResponseWrapper, ResponseMode}
+
+@Named
+class ReposeServlet @Inject()(router: PowerFilterRouter) extends HttpServlet with StrictLogging {
```
dmnjohns: Just noting that the `PowerFilterRouter` (or whatever it becomes) will need to be made a Spring-managed bean for this to work.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeServlet.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  override def service(req: HttpServletRequest, resp: HttpServletResponse): Unit = {
+    val wrappedRequest = new HttpServletRequestWrapper(req)
+    val wrappedResponse = new HttpServletResponseWrapper(resp, ResponseMode.MUTABLE, ResponseMode.PASSTHROUGH)
+
+    try {
+      if (wrappedResponse.getStatus() < SC_INTERNAL_SERVER_ERROR) {
```
dmnjohns: I wasn't actually sure whether or not we wanted to keep this check. I left it in since that's the current behavior, but honestly... I don't particularly like it.

wdschei: This just short circuits the request from going to the origin if it is a 5xx.
Although the filters that care about this already short circuit if not in delegation mode.
What were your thoughts/reasoning on not doing this?

dmnjohns: For one, as you pointed out, if a filter wants to reject a request, it will probably do so itself, and not rely on processing from subsequent filters or the filter chain. This conditional would allow for further processing from subsequent filters, which may have some use case that I can't think of, but so does delegation and delegating mode. The difference is, delegation provides more information. Also, it's likely safer, since filters normally don't check the response status code before writing their own, meaning if the first filter set a `500` response, but a later filter set a `400` response without checking the current code, this condition would still be satisfied and the request would be routed. 

Secondly, this conditional only covers `5xx` status codes. What about `4xx` codes? Why only circumvent routing and rewriting the response on `5xx`s? This emphasis how niche this code is.

adrianjgeorge: Yeah, i'm not a fan. If we make it to the servlet send the request on to the OS. We shouldn't be half writing responses and hoping somebody else cleans up the mess. Delegating mode doesn't even do that, it just adds a request header saying here's what i would have done, but you told me not to.

dmnjohns: Sure, delegating mode is a little different, but it seems like both are meant to say, "I would have failed, but I'm putting that failure off to allow for further processing."

Anyway, great! I'll remove this. That makes me happy.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeServlet.scala
adrianjgeorge: Apparently the current bean is a power filter router factory, i have no idea why.

repose-aggregator/artifacts/web-application/src/main/java/org/openrepose/core/ReposeInitializer.java
```diff
@@ -91,5 +91,9 @@ public void onStartup(ServletContext servletContext) throws ServletException {
         servletContext.addServlet("emptyServlet", EmptyServlet.class).addMapping("/*");
         servletContext.addFilter("springDelegatingFilterProxy", new DelegatingFilterProxy("powerFilter"))
                 .addMappingForUrlPatterns(EnumSet.allOf(DispatcherType.class), false, "/*");
+
+        // todo: when switching to the new ReposeFilterChain, replce the previous line with the following line
+        // todo: to wire in the routing servlet
+        // servletContext.addServlet("reposeServlet", rootContext.getBean(ReposeServlet.class)).addMapping("/*");
```
adrianjgeorge: I think this comment is offset by a line. I'm still able to figure it out, but that was confusing for a second.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeServlet.scala
adrianjgeorge: Yeah, my policy in the filter chain and with all the other changes we are making to these bits, is if it seems like it might be stupid raise the flag, so that we can discuss and probably remove it.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeServlet.scala
```diff
@@ -0,0 +1,48 @@
...
+import javax.inject.{Inject, Named}
+import javax.servlet.http.HttpServletResponse.SC_INTERNAL_SERVER_ERROR
+import javax.servlet.http.{HttpServlet, HttpServletRequest, HttpServletResponse}
+import org.openrepose.commons.utils.servlet.http.{HttpServletRequestWrapper, HttpServletResponseWrapper, ResponseMode}
+
+@Named
+class ReposeServlet @Inject()(router: PowerFilterRouter) extends HttpServlet with StrictLogging {
+
+  override def service(req: HttpServletRequest, resp: HttpServletResponse): Unit = {
+    val wrappedRequest = new HttpServletRequestWrapper(req)
```
adrianjgeorge: Why does the router take a wrapped request? it should just wrap it it's own damn self if it want's it wrapped.

dmnjohns: Good question. I do not have an answer to that. In fact, the more I look at the router, the more I want to rewrite it...

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeServlet.scala
```diff
@@ -0,0 +1,48 @@
...
+    val wrappedResponse = new HttpServletResponseWrapper(resp, ResponseMode.MUTABLE, ResponseMode.PASSTHROUGH)
+
+    try {
+      if (wrappedResponse.getStatus() < SC_INTERNAL_SERVER_ERROR) {
+        router.route(wrappedRequest, wrappedResponse)
+      }
+    } catch {
+      case e: Exception =>
+        logger.error("Failed to route the request to the origin service", e)
```
adrianjgeorge: Do we really want to wrap the response? In the chain i mostly checked if the response was committed and did a send error if it wasn't. And always added the log message. I'm not certain which is the better approach here, but i wanted to bring it up for discussion.

repose-aggregator/src/docs/asciidoc/architecture/servlet-spec.adoc
```diff
@@ -0,0 +1,22 @@
...
+
+The https://javaee.github.io/servlet-spec/[Java Servlet Specification] is a document that details the design and API of a Java-based system that can respond to requests.
+To learn more about the specification itself, visit the link near the start of this paragraph.
+
+*Repose* is built on top of the servlet specification.
+However, *Repose* also constrains the specification in certain ways.
+For example, *Repose* can only service HTTP requests.
+
+To understand the architecture of *Repose*, it is vital that one understands the servlet specification.
+Currently, *Repose* complies with version 3.1 of the specification.
```
adrianjgeorge: Pretty sure it's actually 3.0, we have a couple of classes that don't implement the added methods of 3.1.

repose-aggregator/src/docs/asciidoc/architecture/index.adoc
```diff
@@ -11,6 +11,9 @@ Provides detail about the *Repose* container -- the configuration file which out
 == <<packages.adoc#,Packages>>
 Provides details about the packages *Repose* maintains for easy installation by our users.
 
+== <<servlet-spec.adoc#,Servlet Specification>>
```
adrianjgeorge: Probably comment this out for now?

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeServlet.scala
adrianjgeorge: yeah...

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeServlet.scala
dmnjohns: Ah. Somehow I did not notice that. Well that's crappy. Well, the router will change with the system model (since it uses the destinations therein). So ideally, I think the router would still be a bean, but it would have to be a bean that can be updated or replaced. We can probably drop the factory and the router can initialize itself as a bean with the services it needs, plus the cluster and node IDs so that it can determine destinations. Which means the system model comes into play somewhere. Not sure exactly how that looks, but... that should be a separate story anyway.

repose-aggregator/src/docs/asciidoc/architecture/index.adoc
dmnjohns: Oh, yeah, probably so.

repose-aggregator/src/docs/asciidoc/architecture/servlet-spec.adoc
dmnjohns: Ah, I think you are right. Will update.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeServlet.scala
dmnjohns: Fair question. I debated this a bit with myself. I guess the root question is, if an `Exception` is thrown, do we want downstream to be able to write the response, or should the active component be responsible for it?

I chose to put the responsibility on the active component, since if the downstream component threw an exception, processing failed in some way, so I don't trust whatever response was written. In my mind, the exception communicates to the active component that downstream could not handle processing. Admittedly, a competent downstream component would never write a response that should not reach the user. Also, the downstream component may have more context to write a good response. Since the exception is probably some technical problem, though, I'd rather stick whatever context we have in a log, and not send it back to the end user.

adrianjgeorge:
retest this please

PR #1949, opened 2018-06-18
REP-6969 Updated the System Model schema with the new conditional filter processing.

wdschei:


repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/system-model/system-model.xsd
```diff
@@ -275,13 +283,153 @@
...
+                        Deprecated: This attribute is deprecated and will be removed in future major release.
+                        Regex of what uri is allowed access.
+                    </html:p>
                 </xs:documentation>
             </xs:annotation>
         </xs:attribute>
+        <xs:assert vc:minVersion="1.1"
+                   xpathDefaultNamespace="##targetNamespace"
+                   test="not(@uri-regex and uri)"
+                   xerces:message="Cannot define both a deprecated uri-regex attribute and the new uri element"
```
adrianjgeorge: We aren't going to allow uri regex with any of the new criteria.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/system-model/system-model.xsd
```diff
@@ -275,13 +283,153 @@
...
+
+
+    <xs:complexType name="FilterCriterion" abstract="true">
+        <xs:annotation>
+            <xs:documentation>
+                <html:p>Defines the parent of all the conditional filter processing elements.</html:p>
+            </xs:documentation>
+        </xs:annotation>
+    </xs:complexType>
+    <xs:complexType name="Method">
```
adrianjgeorge: Do we want to support a list of methods or make them do an or and an entry for each?

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/system-model/system-model.xsd
```diff
@@ -275,13 +283,153 @@
...
+                                The case insensitive name of the header being tested for.
+                            </html:p>
+                        </xs:documentation>
+                    </xs:annotation>
+                </xs:attribute>
+                <xs:attribute name="value" type="xs:string">
+                    <xs:annotation>
+                        <xs:documentation>
+                            <html:p>
+                                The case sensitive value of the header being tested for.
```
adrianjgeorge: Did we decide whether or not we were going to do regexs?

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/system-model/system-model.xsd
```diff
@@ -275,13 +283,153 @@
...
+        </xs:annotation>
+        <xs:complexContent>
+            <xs:extension base="mod:FilterCriterion">
+                <xs:choice minOccurs="2" maxOccurs="unbounded">
+                    <xs:element name="method" type="mod:Method"/>
+                    <xs:element name="header" type="mod:Header"/>
+                    <xs:element name="uri" type="mod:Uri"/>
+                    <xs:element name="not" type="mod:Not"/>
+                    <xs:element name="or" type="mod:Or"/>
+                </xs:choice>
```
adrianjgeorge: And should be in this list

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/system-model/system-model.xsd
wdschei: For the headers, I thought we decided that it would be explicit.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/system-model/system-model.xsd
```diff
@@ -275,13 +283,153 @@
...
+            </xs:documentation>
+        </xs:annotation>
+        <xs:complexContent>
+            <xs:extension base="mod:FilterCriterion">
+                <xs:choice minOccurs="2" maxOccurs="unbounded">
+                    <xs:element name="method" type="mod:Method"/>
+                    <xs:element name="header" type="mod:Header"/>
+                    <xs:element name="uri" type="mod:Uri"/>
+                    <xs:element name="and" type="mod:And"/>
+                    <xs:element name="not" type="mod:Not"/>
```
adrianjgeorge: Or should be in this list.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/system-model/system-model.xsd
wdschei: As it is right now, it would need to be `Or`'d if more than one was possible.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/system-model/system-model.xsd
wdschei: Ok. I'll update that.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/system-model/system-model.xsd
adrianjgeorge: Actually, i'm not sure it is. i can't think of a case where it would make a difference, other than in how one thinks about it. Same for and i suppose.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/system-model/system-model.xsd
```diff
@@ -275,13 +283,170 @@
...
+    <xs:simpleType name="MethodList">
+        <xs:annotation>
+            <xs:documentation>
+                <html:p>
+                    A list defined as a string enumeration of HTTP method verbs as outlined in
+                    <a href="http://www.ietf.org/rfc/rfc2616.txt">RFC2616 section 9.</a>
+                </html:p>
+            </xs:documentation>
+        </xs:annotation>
+        <xs:list itemType="xs:string"/>
```
adrianjgeorge: Do we not want to limit the values?

wdschei: I was thinking about that. We do in one use of an HTTP method configuration and we don't in another. I don't see what limiting it does other than force us to make an update if a new verb is needed.

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/systemmodel/config/And.java
```diff
@@ -0,0 +1,59 @@
...
+    private List<FilterCriterion> filterCriteria = new ArrayList<>();
+
+    @Override
+    public boolean evaluate(HttpServletRequestWrapper httpServletRequestWrapper) {
+        boolean rtn = getFilterCriteria().stream()
+            .allMatch(criterion -> criterion.evaluate(httpServletRequestWrapper));
+        LOG.trace("{} of the sub-criterion matched the request.", rtn ? "All" : "None");
+        return rtn;
+    }
+    private static final Logger LOG = LoggerFactory.getLogger(And.class);
```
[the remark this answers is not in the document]
wdschei: I had them in the normal place at the stop and the JAXB marshaling stuff got all wonky.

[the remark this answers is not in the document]
wdschei: This is also why the Codacy Review fails since they are not at the top.

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/systemmodel/config/Filter.java
```diff
@@ -0,0 +1,84 @@
...
+        } else {
+            return Optional.<FilterCriterion>ofNullable(methods)
+                .orElseGet(() -> Optional.<FilterCriterion>ofNullable(header)
+                    .orElseGet(() -> Optional.<FilterCriterion>ofNullable(uri)
+                        .orElseGet(() -> Optional.<FilterCriterion>ofNullable(and)
+                            .orElseGet(() -> Optional.<FilterCriterion>ofNullable(not)
+                                .orElseGet(() -> Optional.<FilterCriterion>ofNullable(or)
+                                    .orElse(DEFAULT_FILTER_CRITERION))))));
+        }
+    }
```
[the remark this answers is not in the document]
wdschei: PR #1963 Merged.

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/systemmodel/config/And.java
```diff
@@ -0,0 +1,59 @@
...
+        @XmlElement(name = "not", type = Not.class),
+        @XmlElement(name = "or", type = Or.class)
+    })
+    private List<FilterCriterion> filterCriteria = new ArrayList<>();
+
+    @Override
+    public boolean evaluate(HttpServletRequestWrapper httpServletRequestWrapper) {
+        boolean rtn = getFilterCriteria().stream()
+            .allMatch(criterion -> criterion.evaluate(httpServletRequestWrapper));
+        LOG.trace("{} of the sub-criterion matched the request.", rtn ? "All" : "None");
```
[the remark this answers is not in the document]
wdschei: Updated.

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/systemmodel/config/AndTest.scala
```diff
@@ -0,0 +1,81 @@
...
+
+    Seq(
+      (false, false, false),
+      (true, false, false),
+      (false, true, false),
+      (false, false, true),
+      (true, true, false),
+      (true, false, true),
+      (false, true, true)
+    ).foreach { case (retOne, retTwo, retToo) =>
```
dmnjohns: Man, we use combinatorics a bunch when generating tests, and it's really something that seems to be lacking in Java/Scala standard libraries.

So here we could use the Scala standard library like so:
```
(List.fill(3)(List(true, false))).flatten.combinations(3).map(_.permutations).flatten.toList
```

But that's ugly. Both in terms of readability, and also in terms of efficiency. I am tempted to write a small library to supplement Scala's combinatorics... but not today.

wdschei: The problem with that is there is one case that is added to the final list that can't exist for this test, `(true, true, true)`.
Of course we could filter that out, but I think that takes away from the table like simplicity of what is being tested.

dmnjohns: Ah. I didn't notice that here. Fair enough. This is all good.

Also, in this case, the permutation of combinations is still a mess, so I wouldn't use it anyway.

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/systemmodel/config/And.java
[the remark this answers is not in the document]
dmnjohns: Ah. I think you can get it working by making use of the [@XmlTransient](https://docs.oracle.com/javaee/7/api/javax/xml/bind/annotation/XmlTransient.html|@XmlTransient) annotation.

wdschei:
Retest this please.

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/systemmodel/config/And.java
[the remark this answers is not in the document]
wdschei: That does work. Updated.

repose-aggregator/src/docs/asciidoc/architecture/system-model.adoc
```diff
@@ -119,6 +119,55 @@ If set to localhost, any physical node will match and run the *Repose* node.
...
+      <endpoint id="origin_service", protocol="http", port="80", default="true"/> <!--13-->
+    </destinations>
+  </repose-cluster>
+</system-model>
+----
+<1> Defines a new *Repose* cluster with an ID that must be unique across all defined clusters.
+<2> Defines a new *Repose* node.
+<3> Defines a sequence of filters to run.
+<4> Defines the <<../filters/ip-user.adoc#,IP User Filter>> should be used by the nodes in this cluster.
+<5> Defines a boolean `AND` operator were all direct sub-elements must evaluate to `true`.
```
dmnjohns: `were` -> `where`

repose-aggregator/src/docs/asciidoc/architecture/system-model.adoc
```diff
@@ -119,6 +119,55 @@ If set to localhost, any physical node will match and run the *Repose* node.
...
+<6> Defines the space separated list of case sensitive methods to check the current request against.
+<7> Defines a boolean `OR` operator were at least one direct sub-element must evaluate to `true`.
```
dmnjohns: `were` -> `where`

repose-aggregator/src/docs/asciidoc/architecture/system-model.adoc
```diff
@@ -119,6 +119,55 @@ If set to localhost, any physical node will match and run the *Repose* node.
      The default destination endpoint will be responsible for handling all traffic that is not explicitly routed to a different destination endpoint.
      Remember that a target can be specified rather than an endpoint if you want to use round-robin routing to a service cluster.
 
+=== Boolean Filter Determination
```
dmnjohns: In this case, I wonder if the word `Boolean` doesn't unnecessarily complicate the meaning of this section. Where does the boolean part come in? Because criteria evaluates to a boolean? Isn't that implied by determination? Because criteria must be boolean? That isn't actually the case. Is `Boolean Filter` a filter that needs determination? Well, no.

Just my 2c, but I like leaving `Boolean` out of the header here, but still using it where appropriate in the callouts below.

repose-aggregator/src/docs/asciidoc/architecture/system-model.adoc
```diff
@@ -119,6 +119,55 @@ If set to localhost, any physical node will match and run the *Repose* node.
...
+[source,xml]
+----
```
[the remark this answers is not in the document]
wdschei: We do normally title them if it is a full or explicit configuration. I found a few other places that were using `partial` in the title as well. I also fixed a few other places this wasn't being done so we can be consistent.

Good catch.

repose-aggregator/src/docs/asciidoc/architecture/system-model.adoc
wdschei: Updated.

repose-aggregator/src/docs/asciidoc/architecture/system-model.adoc
wdschei: Updated.

.gitignore
```diff
@@ -22,7 +22,7 @@ nbactions.xml
 *.ipr
 *.iws
 .idea/
-local/
+local
```
adrianjgeorge: Why change this? If i recall the slash indicates you want to ignore the directory, you've changed it to ignoring a single file.

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/systemmodel/config/And.java
```diff
@@ -0,0 +1,60 @@
...
+import org.slf4j.LoggerFactory;
+
+import javax.xml.bind.annotation.*;
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlType(name = "And", propOrder = {
+    "filterCriteria"
```
adrianjgeorge: `propOrder` shouldn't be necessary since there's only one property.

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/systemmodel/config/Cluster.java
```diff
@@ -0,0 +1,46 @@
...
+import lombok.Data;
+
+import javax.xml.bind.annotation.*;
+import javax.xml.bind.annotation.adapters.CollapsedStringAdapter;
+import javax.xml.bind.annotation.adapters.XmlJavaTypeAdapter;
+import java.io.Serializable;
+
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlType(name = "Cluster", propOrder = {
+    "nodes"
```
adrianjgeorge: Another `propOrder`. I'm going to stop pointing them out in case you come back and tell me that it didn't want to work without them.

.gitignore
wdschei: The trailing slash means it *has* to be a directory. This allows it to be either. In my new setup, I have the one master local directory and all the other roots have a link to that directory. This allows for both.

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/systemmodel/config/Filter.java
```diff
@@ -0,0 +1,76 @@
...
+    };
+    @XmlElements({
+        @XmlElement(name = "methods", type = Methods.class),
+        @XmlElement(name = "header", type = Header.class),
+        @XmlElement(name = "uri", type = Uri.class),
+        @XmlElement(name = "and", type = And.class),
+        @XmlElement(name = "not", type = Not.class),
+        @XmlElement(name = "or", type = Or.class)
+    })
+    private FilterCriterion filterCriterion;
```
adrianjgeorge: Why not initialize this to the default? Then you don't have to worry about it ever being null.

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/systemmodel/config/Filter.java
```diff
@@ -0,0 +1,76 @@
...
+    @XmlAttribute(name = "name", required = true)
+    private String name;
+    @XmlAttribute(name = "configuration")
+    @XmlSchemaType(name = "anyURI")
+    private String configuration = "";
+    @XmlAttribute(name = "uri-regex")
+    private String uriRegex;
+
+    public FilterCriterion getFilterCriterion() {
+        // Remove this when the deprecated uri-regex attribute is deleted.
```
adrianjgeorge: Go ahead and add a todo in front of this comment, so that we might find it eventually.

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/systemmodel/config/FilterCriterion.java
```diff
@@ -0,0 +1,34 @@
...
+import org.openrepose.commons.utils.servlet.http.HttpServletRequestWrapper;
+
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlType;
+import java.io.Serializable;
+
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlType(name = "FilterCriterion")
+public abstract class FilterCriterion
```
adrianjgeorge: Would an interface still work?

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/systemmodel/config/FilterCriterion.java
```diff
@@ -0,0 +1,34 @@
...
+    implements Serializable {
+    public abstract boolean evaluate(HttpServletRequestWrapper httpServletRequestWrapper);
```
adrianjgeorge: Do we need it to be the wrapper? It's not necessarily hard to update the chain to match, i'm just wondering if we need to go to it.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/container/bindings.xjb
```diff
@@ -24,8 +24,10 @@
           version="2.0"
           xsi:schemaLocation="http://java.sun.com/xml/ns/jaxb http://java.sun.com/xml/ns/jaxb/bindingschema_2_0.xsd"
           schemaLocation="container-configuration.xsd">
-
     <schemaBindings>
         <package name="org.openrepose.core.container.config"/>
     </schemaBindings>
-</bindings>
\ No newline at end of file
+    <globalBindings>
```
adrianjgeorge: is this change needed, if so why?

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/system-model/ignore.bindings
```diff
@@ -18,17 +18,12 @@
   limitations under the License.
   =_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_=_
   -->
-
-
-<bindings xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://java.sun.com/xml/ns/jaxb"
+<bindings xmlns="http://java.sun.com/xml/ns/jaxb"
```
adrianjgeorge: we just want to delete this file don't we?

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/system-model/system-model.xsd
```diff
@@ -275,13 +283,170 @@
             </xs:annotation>
         </xs:attribute>
 
-        <xs:attribute name="uri-regex" type="xs:string" use="optional" default=".*">
+        <xs:attribute name="uri-regex" type="xs:string" use="optional">
             <xs:annotation>
                 <xs:documentation>
-                    <html:p>Regex of what uri is allowed access.</html:p>
+                    <html:p>
+                        Deprecated: This attribute is deprecated and will be removed in future major release.
```
adrianjgeorge: Maybe mention what's replacing it?

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/systemmodel/config/FilterCriterion.java
wdschei: The header methods in the wrapper are easier to use.

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/systemmodel/config/AndTest.scala
```diff
@@ -0,0 +1,81 @@
...
+import org.scalatest.mock.MockitoSugar
+import org.scalatest.{FunSpec, Matchers}
+import org.springframework.mock.web.MockHttpServletRequest
+
+@RunWith(classOf[JUnitRunner])
+class AndTest extends FunSpec with Matchers with MockitoSugar {
+  describe("evaluate") {
+    it("should return true when all of the sub criteria are true") {
+      val filterCriteriaOne = mock[FilterCriterion]
+      when(filterCriteriaOne.evaluate(any(classOf[HttpServletRequestWrapper]))).thenReturn(true)
```
adrianjgeorge: It seems like it would have been easier to just write a test implementation of `FilterCriterion` that takes a boolean. You could have used it here and in several other tests instead of mocking.

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/systemmodel/config/AndTest.scala
```diff
@@ -0,0 +1,81 @@
...
+      val filterCriteriaTwo = mock[FilterCriterion]
+      when(filterCriteriaTwo.evaluate(any(classOf[HttpServletRequestWrapper]))).thenReturn(true)
+      val filterCriterion = new And()
+      val subCriteria = filterCriterion.getFilterCriteria
```
adrianjgeorge: Couldn't you just call `setFilterCriteria(List(filterCriteriaOne, filterCriteriaTwo).asJava)`?

adrianjgeorge: Heck if you create the test implementation you could inline the definition.

repose-aggregator/src/docs/asciidoc/architecture/system-model.adoc
adrianjgeorge: Filter Activation Evaluation? Filter Activation Determination?

adrianjgeorge:
You didn't update the filter chain docs page.

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/systemmodel/config/Or.java
```diff
@@ -0,0 +1,60 @@
...
+        @XmlElement(name = "uri", type = Uri.class),
+        @XmlElement(name = "not", type = Not.class),
+        @XmlElement(name = "and", type = And.class)
+    })
+    private List<FilterCriterion> filterCriteria = new ArrayList<>();
+
+    @Override
+    public boolean evaluate(HttpServletRequestWrapper httpServletRequestWrapper) {
+        boolean rtn = getFilterCriteria().stream()
+            .anyMatch(criterion -> criterion.evaluate(httpServletRequestWrapper));
```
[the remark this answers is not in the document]
wdschei: It does. The docs say:

> [*"May not evaluate the predicate on all elements if not necessary for determining the result."*](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#anyMatch-java.util.function.Predicate-)

.gitignore
adrianjgeorge: Fair enough

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/systemmodel/config/Filter.java
wdschei: That works.

repose-aggregator/core/repose-core-api/src/main/java/org/openrepose/core/systemmodel/config/Filter.java
wdschei: Updated.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/system-model/system-model.xsd
wdschei: Updated.

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/systemmodel/config/AndTest.scala
wdschei: Could have. Could have also used anonymous classes that returned the desired value. My brain was just using mocks that day I guess.

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/systemmodel/config/AndTest.scala
wdschei: Could.

repose-aggregator/src/docs/asciidoc/architecture/system-model.adoc
wdschei: Updated.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/container/bindings.xjb
wdschei: This ended up being needed by the `DeploymentConfigPatchUtil.deepCopy()` method so the objects are forced to implement `Serializable`.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/system-model/ignore.bindings
adrianjgeorge: Nuke it

wdschei: Gone.

repose-aggregator/src/docs/asciidoc/architecture/filter-chain.adoc
```diff
@@ -40,9 +40,30 @@ Here's a simple example:
...
+
+[NOTE]
+====
+The filter chain is dynamic.
+It is not determined at the start of the request, rather before each filter is executed the request is compared against the configured criteria.
+This means that anything that modifies the request may effect whether a later filter will process or not.
+For example, the <<../filters/body-extractor-to-header.adoc#,Body Extractor to Header Filter>> could add a header based on the request body which is used to determine if a later filter executes.
+====
+
+Refer to <<system-model.adoc#_filter_activation_determination,System Model Filter Activation Determination>> for further details about configuring this capability.
```
adrianjgeorge: I'm not certain which doc we want the details to live on. I agree we should put in one place, and link in the other, but i'm not certain where it should live.

wdschei: Since each configuration file has a page where we put all the configuration permutations, it felt more natural to put these updates on the System Model page.

I'm not sure if the Filter Chain page shouldn't be incorporated into the System Model page when we release v9.

adrianjgeorge: Lol, I was thinking the opposite that we'd eventually abolish the system model page. We'll leave it as you've done it for now.

wdschei: My thinking was we put the `xyz` configs in the `xyz` page, so why in this one case would we put the `abc` configs in the `rst` page.

We probably need to visit this again before we do the actual conversions too.

wdschei:
Retest this please.

wdschei:
All the failures are `TimeoutException`'s, but the addition of the SAML tests to that list is a bit disconcerting.

adrianjgeorge:
The Saml tests are already on my list to potentially kick over to the intense side.

PR #1950, opened 2018-06-25
REP-6932 Collecting GC and Thread Data from Performance Tests

dmnjohns:
Data from a test run can be viewed as various graphs under the `JMX` panel on the [Grafana performance test dashboard](http://grafana.openrepose.org/d/000000005/performance-tests?orgId=1&from=1529677818996&to=1529679467580&var-testName=ScriptingFilterJavascriptSimulation&var-endpoint=Repose&var-requests=allRequests).


wdschei:
The test failures are the new standard `TimeoutException`'s.
They are in in known failure spots, so merging even with failures.

PR #1952, opened 2018-06-25
Adding a router bean stub to satisfy the new servlet

dmnjohns:
The primary purpose of this fix is to allow builds to complete and jobs to pass again.


repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRouter.scala
```diff
@@ -0,0 +1,31 @@
...
+ */
+package org.openrepose.powerfilter
+
+import javax.inject.Named
+import javax.servlet.http.HttpServletResponse
+import org.openrepose.commons.utils.servlet.http.HttpServletRequestWrapper
+
+// fixme: This class will serve as the basis of a rewrite of the routing component for Repose
+
+@Named
```
dmnjohns: One way of actually implementing this router would be to replace the router bean on configuration update and dropping the router factory completely.

I will also note that since this bean is in the `org.openrepose.powerfilter` package, it will exist per-node, and can access other node-level beans (e.g., node ID and cluster ID, which would be used to determine routing destinations).

adrianjgeorge:
retest this please

PR #1953, opened 2018-06-26
REP-5083 Split Header Filter

dmnjohns:


adrianjgeorge:
Config looks good.

adrianjgeorge:
Don't forget to create an example config that we ship that mimics current behavior.

adrianjgeorge:
retest this please

wdschei:
Should we log something if the user configures to split an RFC defined header that isn't supposed to be splittable?
I'm not sure I want to get into the business of defining/maintaining the exhaustive list of what should and shouldn't be split, but things like `User-Agent` are pretty obvious.
The message could be logged on configuration load to prevent it being to annoying, but it puts it out there that they are doing something against the spec.
At the very least, we should probably have a note in the docs to make them aware that there are some headers that they shouldn't split.

repose-aggregator/components/filters/split-header-filter/src/main/scala/org/openrepose/filters/splitheader/SplitHeaderFilter.scala
```diff
@@ -0,0 +1,84 @@
...
+
+    splitHeaders(wrappedHttpResponse, configuration.getResponse)
+
+    wrappedHttpResponse.commitToResponse()
+  }
+
+  private def splitHeaders(wrappedHttpMessage: HeaderInteractor, configuredHeaders: HeaderList): Unit = {
+    Option(configuredHeaders).map(_.getHeader.asScala).getOrElse(Seq.empty) foreach { headerToSplit =>
+      val headerValues = wrappedHttpMessage.getSplittableHeaders(headerToSplit).asScala
+      if (headerValues.size > wrappedHttpMessage.getHeadersList(headerToSplit).size) {
```
adrianjgeorge: Why bother?

repose-aggregator/src/docs/asciidoc/filters/index.adoc
```diff
@@ -95,6 +95,9 @@ Validates calls to an API by checking the URI format and HTTP method and enforci
 == <<slf4j-http-logging.adoc#,SFL4J HTTP Logging Filter>>
 *IN PROGRESS*
 
+== <<split-header.adoc#,Split Header Filter>>
+Splits a single line header with multiple values into multiple header lines each with a single value.
```
adrianjgeorge: They  don't necessarily have to be a single line. you could have two lines with five values

dmnjohns: True, although I think I underspecified here. This filter really will only split single line, multi-valued headers, but it will split _all_ single line, multi-valued headers (that are configured).

I'll think of some better words and update this.

repose-aggregator/src/docs/asciidoc/filters/split-header.adoc
```diff
@@ -0,0 +1,184 @@
...
+=== Required Request Headers
+This filter does not require any request headers.
+
+=== Required Preceding Filters
+This filter has no dependency on any other filter.
+
+=== Request Headers Created
+Changes to request headers vary based on configuration.
+
+Existing response headers may be split across multiple lines.
```
adrianjgeorge: request not response

repose-aggregator/tests/functional-tests/src/integrationTest/configs/features/filters/splitheaders/addHeader/add-header-request.cfg.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<add-headers xmlns="http://docs.openrepose.org/repose/add-header/v1.0">
+    <request>
+        <header name="request-test-1">split</header>
```
adrianjgeorge: why are there values in these elements?

adrianjgeorge: ohh, i see this isn't config for the header splitting filter. man the config looks almost the same.

PR #1959, opened 2018-07-06
WIP: REP-7097 Extract boilerplate from the performance tests.

wdschei:


repose-aggregator/tests/performance-tests/Jenkinsfile
```diff
@@ -57,6 +57,9 @@ stage("Performance Test") {
     // builds the Repose artifacts for performance test jobs to use
     build("repose-gradle-package")
 
+    // builds the performance test library used by all these jobs
+    build("performance-test-library")
```
dmnjohns: Just want to make a note that to get this to work, a build step will need to be added to the `performance-test` job to copy this artifact.

repose-aggregator/tests/performance-tests/roles/gatling/tasks/install_gatling.yml
```diff
@@ -54,3 +54,11 @@
     dest: /root/gatling/current/conf/gatling.conf
     mode: 0644
     backup: yes
+
+- name: Copy the Repose library for Gatling
```
dmnjohns: Once again, for this to work, the Jenkins `performance-test` job must be updated to copy the library JAR.

The relative destination for that JAR should be:
```
repose-aggregator/tests/performance-tests/roles/gatling/files
```

dmnjohns:
I left a commit and revert in the history that was me setting up a Gradle project for the `performance-tests` directory and adding a task to copy the library JAR. I figure if we ever bridge our Gradle/Ansible divide, that may be useful. Right now, Jenkins is acting as that bridge.

repose-aggregator/tests/performance-test-library/build.gradle
```diff
@@ -0,0 +1,4 @@
+dependencies {
+    compile 'io.gatling:gatling-test-framework:2.2.4'
```
dmnjohns: Ah. Didn't see this the first time. Is there any reason not to stick this version in the version properties file?

Also, is there a way we can tie this version number to the version of Gatling we are installing using Ansible? I can't think of one, and it comes back to our lack of a Gradle/Ansible bridge, but it's something that would be nice to have.

repose-aggregator/tests/performance-test-library/src/main/scala/org/openrepose/performance/test/AbstractReposeSimulation.scala
```diff
@@ -0,0 +1,62 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_=_
+ */
+
+package org.openrepose.performance.test
+
+import com.typesafe.config.ConfigFactory
```
dmnjohns: Ah. I don't think this import was removed from all of the individual performance tests, but it probably can be now.

repose-aggregator/tests/performance-test-library/src/main/scala/org/openrepose/performance/test/AbstractReposeSimulation.scala
```diff
@@ -0,0 +1,83 @@
...
+        constantUsersPerSec(rampUpUsers) during (rampUpDuration seconds))
+      .throttle(
+        jumpToRps(throughput), holdFor((warmUpDuration + duration) minutes))
+  }
+
+  // run the scenarios
+  def runScenarios(): Unit = {
+    setUp(
+      populateWarmupScenario,
+      populateMainScenario
```
wdschei: I like the use of these `populate[Warmup|Main]Scenario` methods to extract out even more of the boiler plate.

repose-aggregator/tests/performance-tests/roles/gatling/defaults/main.yml
```diff
@@ -1,6 +1,8 @@
 {
   gatling: {
     install: {
+      # Note: If the version of Gatling changes, it should also be changed in the Gradle versions properties
+      # Note: file so that our tests, test library, and Gatling runtime environment stay in sync.
```
wdschei: :100:

PR #1960, opened 2018-07-09
REP-7009 Using the x-trace-request header to mark requests for logging

dmnjohns:


repose-aggregator/src/docs/asciidoc/architecture/filter-chain.adoc
```diff
@@ -84,10 +84,8 @@ For more troubleshooting help see our <<../troubleshooting.adoc, guide>>.
 If you wish to see these numbers or exfiltrate them to another tool look into the <<../services/metrics.adoc, Metrics Service>>.
 They are listed under `org.openrepose.core.FilterProcessingTime.Delay` with an entry for each filter by name.
 
-If you are curious about the numbers for a specific request you can send in the `X-Trace-Request` header with a value of true.
-This will add a header to the response for each filter with a name of `X-_filter name_-Time` with the time it took that filter to process.
-
-It's worth noting that the reported time is inclusive of the processing time of all filter further down the chain and the origin service.
+If you are curious about the numbers for a specific request you can add the `X-Trace-Request` header to the request.
+See <<../troubleshooting.adoc#,Time Spent in Each Filter>> for more details.
```
adrianjgeorge: Is there not an anchor to directly link to on that page?

adrianjgeorge:
I didn't see where you made sure to propagate the tracing through to akka.

dmnjohns:
The code we wrote to transfer the trace GUID to Akka also transfers the flag I added. In fact, it just transfers the entire contents of the MDC context map:

https://github.com/rackerlabs/repose/blob/master/repose-aggregator/components/services/akka-http-client-service/akka-http-client-service-impl/src/main/java/org/openrepose/core/services/serviceclient/akka/impl/ActorRequest.java#L31-L43

https://github.com/rackerlabs/repose/blob/master/repose-aggregator/components/services/akka-http-client-service/akka-http-client-service-impl/src/main/java/org/openrepose/core/services/serviceclient/akka/impl/ActorRequest.java#L31-L43

repose-aggregator/src/docs/asciidoc/architecture/filter-chain.adoc
dmnjohns: There is, but I could not figure out how to link to it here. I did not want to link to the HTML anchor since that couples this document to the output format, but since the section title has spaces in it, I could not use that.

Actually, after playing with this a little more, I think can get it to work by switching from `URI Regex` to `_uri_regex`. That is the generated header ID, so as long as the way that header IDs are generated doesn't change, it should work. Even if that does change, the link should still be to the page, just not the specific section, which is the current behavior anyway.

adrianjgeorge:
Well that's convenient. Good job future proofing things us.

adrianjgeorge:
Not mergeable, The Cors tests failed on both pr builders suggesting something is wrong, and one of the intrafilter logging tests failed as well.

repose-aggregator/artifacts/src/config/filters/log4j2.xml
```diff
@@ -26,9 +26,23 @@
...
+        <Logger name="intrafilter-logging" level="trace">
+            <ContextMapFilter onMatch="ACCEPT" onMismatch="DENY">
+                <KeyValuePair key="x-trace-request" value="true"/>
+            </ContextMapFilter>
+        </Logger>
+        <Logger name="filter-timing" level="trace">
+            <ContextMapFilter onMatch="ACCEPT" onMismatch="DENY">
+                <KeyValuePair key="x-trace-request" value="true"/>
+            </ContextMapFilter>
+        </Logger>
```
dmnjohns: Ah. This probably needs to change. The problem with this configuration is that if the `x-trace-request` header isn't present, this logger still won't log. That is, we effectively ignore the log level implicit filter.

Using the `DynamicThresholdFilter` we could get around that and see the behavior we want, but unfortunately, it would apply to all loggers (i.e., all loggers would effectively be set to trace if the `x-trace-request` mapping was present).

dmnjohns:
Whelp. I have found the issue with the tests. We have checks in the `PowerFilterChain` to manipulate the HTTP message if a logger's level is set a certain way:
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilterChain.java#L186

Problematically, using filters means that the intrafilter logging logger is always set to trace, so that condition always evaluates to true. So work is done that might not strictly be necessary, and the HTTP messages are changed in ways that certain tests do not account for.

Ideally, we would do that work only if a log event made it past the logger's filters (if not all filters), but I can not find a way to hook in at that point.

I really wish we could use the Log4j `DynamicThresholdFilter` to change the log level certain loggers rather than using filters, but I have not found a way to do so.

wdschei:
Retest this please.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeFilterChain.scala
```diff
@@ -84,7 +85,7 @@ class ReposeFilterChain(val filterChain: List[FilterContext], originalChain: Fil
     var conditionallyWrappedRequest = request
     var conditionallyWrappedResponse = response
 
-    val doLogging = IntrafilterLog.isTraceEnabled
+    val doLogging = IntrafilterLog.isTraceEnabled && Option(MDC.get(TRACE_REQUEST)).isDefined
```
adrianjgeorge: I'm not really a fan of checking if intrafilter is enabled or not. It seems pointless to do. If they sent the header they've expressed that they want the info.

repose-aggregator/artifacts/src/config/filters/log4j2.xml
```diff
@@ -26,9 +26,22 @@
...
+                <ContextMapFilter onMatch="ACCEPT" onMismatch="NEUTRAL">
+                    <KeyValuePair key="x-trace-request" value="true"/>
+                </ContextMapFilter>
+                <ThresholdFilter level="WARN" onMatch="NEUTRAL" onMismatch="DENY"/>
+            </Filters>
+        </Logger>
         <Logger name="org.eclipse.jetty" level="off"/>
         <Logger name="org.springframework" level="warn"/>
-        <Logger name="intrafilter-logging" level="info"/>
+        <Logger name="intrafilter-logging" level="trace"/>
```
adrianjgeorge: I don't like this inconsistency

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeFilterChain.scala
dmnjohns: I look at it from the other side. If we don't check if the logger is set to trace, then we set up the request stream for buffering (which includes copying the entire request body) whenever the header is present. If the header is present and the log level in configured below trace, though, we will never see the information that we set ourselves up to see. So it seems pointless to me to do work that essentially gets dropped on the ground.

That being said, we also add a UUID header as part of the intrafilter setup, so there's some argument to be made about the using sending the header just to get the UUID added... but that seems a little silly.

I can drop that part of the check if you would like (both you and Bill pointed it out), but I do see a reason not to drop it.

adrianjgeorge: Eh, you've told us you want to debug this request, i don't care if we pointlessly add overhead. It just adds complication, especially when in this case maybe they wanted timing numbers, but they won't get them because you made that part of the condition to get into the metrics collection.

repose-aggregator/artifacts/src/config/filters/log4j2.xml
dmnjohns: Yeah, I went back on forth on whether or not to put the filter on this logger. At the moment, the filter isn't strictly necessary since we put the check in code, so I figured I would leave it off to keep the config simpler and cut out the miniscule overhead of having the filter run.

I wanted to include the filter to be explicit in the logging config about the preconditions for intrafilter logging, but then I was worried that if I did add the filter I would effectively create another precondition since there is nothing that would keep the filter check in-sync with the code check.

So ultimately, I opted to keep this logger simple, if inconsistent, and rely on documentation to tell users how to troubleshoot with intrafilter logging. If you would rather I add the filter, I'll go ahead and do that.

adrianjgeorge: You add it in

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeFilterChain.scala
adrianjgeorge: Actually that brings an interesting point, we've added overhead to the filter timing with the pointless intrafilter work. That suggests that code wise we need to be a little more discerning about when we do different things.

dmnjohns: To that point, I would argue that you've also told us you don't want to debug any request by turning the intrafilter logger level down below trace. What's more, if you expect intrafilter logging information simply because you sent the header, you might be surprised when you don't see it.

In fact, dropping the logger level check also adds complication. That takes us from two possible states to three. The current states are: do intrafilter logging with support code or do not do intrafilter logging with support code. The states after this change would be: do not do intrafilter logging with support code, do support code but not intrafilter logging, and do intrafilter logging with support code.

As for timing numbers, I'm not sure I follow. None of the code gated by this check affects the filter timing logging?

dmnjohns: I don't think we have added overhead to the filter timing with the intrafilter work? The timing code just wraps the "requestProcess" code, which is where we call `doFilter`. Then the intrafilter code wraps all of that. Since the intrafilter code is on the outside, though, the timing should not include it, and should still be correct?

Here's an example of function composition to show that:
```
scala> val f: (String) => String = s => s + "f"
f: String => String = <function1>

scala> val g: (String) => String = s => s + "g"
g: String => String = <function1>

scala> f.compose(g).apply("hello")
res2: String = hellogf
```

adrianjgeorge: except the numbers include everything down stream which will have the intrafilter overhead. Also it looked like whether or not to log the timing was blocked in an if that included the intrafilter logging being on as a check. that could just be github misdirecting me

adrianjgeorge: yeah, it was github lying to me. so I think i've come around to leaving it as is.

dmnjohns: Yeah, I'm still not super satisfied that we had to resort to code to get the logging we want, but I wasn't able to come up with any better solutions...

dmnjohns:
retest this please

PR #1962, opened 2018-07-11
REP-7113 Updating our Categories of testing

adrianjgeorge:
The intent is to split the pr builds into a main build that runs everything other than the intense tests, which will get a separate pr build. these two different sets of builds will run on different build slaves. I've also added a category that truly runs them all that we will use on the release build, and tie that build to the same slaves as the intense tests.


adrianjgeorge:
I have no idea why the smoke test failed on both pr builders. I did nothing to impact them code wise

adrianjgeorge:
retest this please

PR #1965, opened 2018-07-13
Updating the docs to bring in the note about Non-Breaking Spaces in role names.

wdschei:


wdschei:
That is a good question.

PR #1966, opened 2018-07-16
REP-7113 Investigate Flaky Tests

dmnjohns:
I will be using the PR during an interview, so it has some intentional flaws.


dmnjohns:
retest this please

dmnjohns:
So this PR is not in an acceptable state, but it did pass on both of the intense jobs. I think that demonstrates the additional stability added by this change, even if it does not solve the flakiness with the `RemoteDatastoreServiceTest` entirely.

PR #1972, opened 2018-08-07
REP-7096 Repose Routing Servlet

wdschei:
This will eventually effectively replace all of the following:
- PowerFilterRouter
- PowerFilterRouterImpl
- PowerFilterRouterFactory
- DispatchPathBuilder
- RoutingService
- RoundRobinRoutingService
- DestinationLocationBuilder
- DestinationLocation
- EndpointUriBuilder
- EndpointUrlBuilder
- RequestHeaderService
- RequestHeaderServiceImpl
- ResponseHeaderService
- ResponseHeaderServiceImpl


wdschei:
Retest this please.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,362 @@
...
+class ReposeRoutingServlet @Inject()(@Value(ReposeSpringProperties.NODE.CLUSTER_ID) clusterId: String,
+                                     @Value(ReposeSpringProperties.NODE.NODE_ID) nodeId: String,
+                                     configurationService: ConfigurationService,
+                                     requestHeaderService: RequestHeaderService,
+                                     responseHeaderService: ResponseHeaderService,
+                                     reportingService: ReportingService,
+                                     optMetricsService: Optional[MetricsService]
+                                    ) extends HttpServlet with UpdateListener[SystemModel] with StrictLogging {
+
+  private val localCluster = new AtomicReference[Option[ReposeCluster]](None)
```
adrianjgeorge: Why all the atomic references? Especially on things that have no ability to change?

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,362 @@
...
+    localCluster.set(Option(interrogator.getLocalCluster(configurationObject).orElse(null)))
+    localNode.set(Option(interrogator.getLocalNode(configurationObject).orElse(null)))
+    defaultDestination.set(Option(interrogator.getDefaultDestination(configurationObject).orElse(null)))
+    destinations.set(localCluster.get().map(_.getDestinations) match {
+      case Some(dests) => getDestinations(dests.getEndpoint.asScala.toList) ++ getDestinations(dests.getTarget.asScala.toList)
+      case None => Map.empty[String, Destination]
+    })
+    domains.set(Option(new Clusters(configurationObject)))
+
+    if (logger.underlying.isDebugEnabled) {
```
adrianjgeorge: This doesn't strike me as being so expensive we need to tuck it into an if. This method isn't getting called aften and this isn't doing a bunch of work as near as i can tell.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,362 @@
...
+  override def isInitialized: Boolean = {
+    initialized
+  }
+
+  override def service(req: HttpServletRequest, resp: HttpServletResponse): Unit = {
+    if (!initialized) {
+      logger.error(NotInitializedMessage)
+      trySendError()
+    } else {
+      logger.trace("{} processing request...", this.getClass.getSimpleName)
```
adrianjgeorge: Why would we include the class name? That's already part of the log message

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,362 @@
...
+        resp.sendError(SC_INTERNAL_SERVER_ERROR, FailedToRouteMessage)
+      } else {
+        logger.error("Failed to Send Error due to response already being committed")
+      }
+    }
+  }
+
+  def doRoute(servletRequest: HttpServletRequestWrapper, servletResponse: HttpServletResponse): Unit = {
+    val attributeDestinations =
+      Option(servletRequest.getAttribute(CommonRequestAttributes.DESTINATIONS))
```
adrianjgeorge: We've talked about taking this out of being an attribute and making it a header like we do basically everything else. We should add a story to take a look at the viability of that. I see that the objects that make up the list have a uri, i wonder how necessary that is to it working?

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,362 @@
...
+        .getOrElse(List.empty[RouteDestination]).asInstanceOf[List[RouteDestination]]
+    val defaultDestinations = {
+      defaultDestination.get() match {
```
adrianjgeorge: Is it possible to not have a default destination?

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,362 @@
...
+          List[RouteDestination] {
+            new RouteDestination(defDest.getId, servletRequest.getRequestURI, -1)
+          }
+        case _ => List.empty[RouteDestination]
+      }
+    }
+    (attributeDestinations ++ defaultDestinations)
+      .sortWith(_.compareTo(_) > 0) // Highest Quality first
+      .headOption
+      .foreach { routeDest =>
```
adrianjgeorge: This will hide the fact that somehow we ended up with no destinations, we should definitely do something about that.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: Unfortunately, yes. That is an optional field in the schema definition with the only restriction being a maximum of 1.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,362 @@
...
+                        servletRequest,
+                        servletResponse,
+                        locationUrl,
+                        dispatchPath,
+                        targetContext,
+                        routeDest,
+                        dispatcher,
+                        destination)
+                    case _ =>
+                      logger.warn(s"No Request Dispatcher found in Servlet Context for $dispatchPath; not forwarding")
```
adrianjgeorge: I'm not certain warn is a high enough log. I could be convinced otherwise though.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,362 @@
...
+    requestHeaderService.setVia(servletRequest)
+    requestHeaderService.setXForwardedFor(servletRequest)
+    logger.debug("Attempting to route to: {}", locationUrl.getPath)
+    logger.debug("  Using dispatcher for: {}", uri)
+    logger.debug("           Request URL: {}", servletRequest.getRequestURL)
+    logger.debug("           Request URI: {}", servletRequest.getRequestURI)
+    logger.debug("          Context path: {}", targetContext.getContextPath)
+    val startTime = System.currentTimeMillis
+    Try {
+      reportingService.incrementRequestCount(routeDest.getDestinationId)
```
adrianjgeorge: What is the reporting service?

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,362 @@
...
+                 targetContext: ServletContext,
+                 routeDest: RouteDestination,
+                 dispatcher: RequestDispatcher,
+                 destination: Destination): Unit = {
+    servletRequest.setScheme(locationUrl.getProtocol)
+    servletRequest.setServerName(locationUrl.getHost)
+    servletRequest.setServerPort(locationUrl.getPort)
+    servletRequest.setRequestURI(locationUrl.getPath)
+
+    requestHeaderService.setVia(servletRequest)
```
adrianjgeorge: Why is this a service instead of a helper of some sort? I know it's not your doing, but we should make a story to change it.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: Since we require at least one destination to be defined, this should probably be a v9 breaking change to require at least one default be defined.

repose-aggregator/core/repose-core/src/test/scala/org/openrepose/powerfilter/ReposeFilterChainTest.scala
```diff
@@ -295,17 +316,19 @@ class ReposeFilterChainTest extends FunSpec with Matchers with MockitoSugar with
...
+        originalChain,
+        None,
+        metricsRegistry,
+        tracer)
 
       filterChain.doFilter(mockRequest, mockResponse)
 
       tracer.finishedSpans should have length 1
-      tracer.finishedSpans.asScala.head.operationName shouldBe "Filter foo"
+      tracer.finishedSpans.asScala.headOption.map(_.operationName) shouldBe Some("Filter foo")
```
dmnjohns: I wonder if failure message here could be improved using a couple more pieces of ScalaTest, like so:

```
import org.scalatest.OptionValues._
...
tracer.finishedSpans.asScala.loneElement should have (operationName ("Filter foo"))
```

The `loneElement` syntax would let you drop the length check, too.

I would actually be very tempted to use `JavaConversions` here to clean that line up even more (especially given that this is test code), but alas, it is deprecated.

Oh, and as a side note, when working with `Option`s, I would recommend using the `Option(...).value shouldBe X` syntax where possible.

References:
http://www.scalatest.org/user_guide/using_matchers#singleElementCollections
http://www.scalatest.org/user_guide/using_matchers#checkingArbitraryProperties
http://doc.scalatest.org/3.0.1/#org.scalatest.matchers.HavePropertyMatcher

http://www.scalatest.org/user_guide/using_matchers#workingWithOptions

repose-aggregator/core/repose-core/src/test/scala/org/openrepose/powerfilter/ReposeRoutingServletTest.scala
```diff
@@ -0,0 +1,230 @@
...
+          val requestDispatcher = mock[RequestDispatcher]
+
+          when(servletContext.getContext(anyString())).thenReturn(targetServletContext)
+          when(targetServletContext.getRequestDispatcher(anyString())).thenReturn(requestDispatcher)
+          when(targetServletContext.getContextPath).thenReturn("")
+          mockServletConfig = new MockServletConfig(servletContext)
+
+          if (dispatchError != null) {
+            when(requestDispatcher.forward(any(classOf[ServletRequest]), any(classOf[ServletResponse]))).thenThrow(dispatchError)
+          }
```
dmnjohns: Oh man do I dislike conditional behavior in tests. It makes the intent of the test more difficult to grasp and expands the scope of the test. It also makes the test more difficult to maintain moving forward.

This may be a place where I find the code a little too DRY, and would prefer it be more DAMP:
https://stackoverflow.com/a/11837973

dmnjohns:
Can you add a note somewhere (perhaps the version "timebomb" test) to remove all of the classes that the `ReposeRoutingServlet` will replace?

repose-aggregator/core/repose-core/src/test/scala/org/openrepose/powerfilter/ReposeRoutingServletTest.scala
wdschei: It is easy enough to add two more boolean values to the test data, but it just seemed redundant to put in the manually calculated value.

wdschei: It would end up with:
```
false | false
false | false
false | false
true  | false
true  | false
true  | false
false | true
false | true
false | true
```

wdschei: For the dispatch error tests that are `null`, then they can be broken out to a separate test. But again, that seemed like overkill.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
dmnjohns: It's a (very old) lookup and tracking service. It basically provides a way to get a `DestinationInfo` object from a `destinationId` String and a way to interact with that `DestinationInfo` object.

The `DestinationInfo` object itself keeps track of number of responses, response codes, and processing time for a specific destination. So basically, rudimentary metrics. It also provides some convenience methods to add more meaning to the data it tracks, like `getThroughput`.

Oh, and of course, there's plenty of abstraction for the sake of abstraction.

wdschei: What else is using it?
Should these be converted to use the JMX metrics?

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: Yeah, those were copy/paste from the `AbstractConfiguredFilter` which was used as the semi-template.
Removed.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: Created the story here:
* https://repose.atlassian.net/browse/REP-7174

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: Story created here:
* https://repose.atlassian.net/browse/REP-7175

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
dmnjohns: I would bet not very necessary. For the default destination, we set that value to the request URI. For other destinations, that value is set by whoever created the destination, which means that the request URI could potentially be ignored (and the destination URI used instead). That seems bad.

So yes, I think we should drop that `uri` field from the `RouteDestination`.

wdschei: We could always remove it during routing.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: Yep. Good call. Updated to a `match` with the default case logging an error and setting the status code.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: This was made a service because it is not static and has to listen to the `system-model.cfg.xml`.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,362 @@
...
+    }
+  }
+}
+
+object ReposeRoutingServlet {
+  final val SystemModelConfigurationFilename = "system-model.cfg.xml"
+  final val NotInitializedMessage = "The ReposeRoutingServlet has not yet been initialized"
+  final val FailedToRouteMessage = "Failed to route the request to the origin service"
+  private val AllEndpoints = "All Endpoints"
+  private val HttpsProtocol = "https"
```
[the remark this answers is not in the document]
wdschei: If a variable is only marked as `val`, then the sub-classes can `override` them.
So the use of the `final` prefix prevents that.
By definition a `private val` is already a `final val` since a `private val` in a companion object can not be seen by sub-classes of it to be `override`'n.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
dmnjohns: That kind of solves my security and transparency concerns. You would still have to worry about forgery, but sure. At that point, though, why use a header?

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
dmnjohns: Are we talking about this attribute?:
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/system-model/system-model.xsd#L513-L517

That seems to read as though exactly one default is required, right?

repose-aggregator/core/repose-core/src/test/scala/org/openrepose/powerfilter/ReposeFilterChainTest.scala
wdschei: I updated this one and added a story to fix the rest:
* https://repose.atlassian.net/browse/REP-7179

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
dmnjohns: Probably. That's why we use a more robust metrics library after all. That would consolidate our metrics, and allow them to be reported to more that _just_ JMX.

Oh, and I think `ReposeReport` is the only thing using it.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: That is what the Header Normalization filter is for.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: It counts all of the ones with the `default` attribute set to `true` and then ensures there is only one of them.
It does not ensure that there is one.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
adrianjgeorge: I'm pretty sure the metrics service can't change, it's flagged optional here because we are sort of heading that way, but any move along those lines will involve this entire context getting remade. You didn't really explain why were are using atomic references here, versus most places? Also at least half the things being referenced would cause the node to be torn down and rebuilt from scratch which would replace the whole servlet.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: So yes. You are right. I updated the wording in the system model and docs as well to be `one and only one`.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
adrianjgeorge: It's just adding unnecessary complexity. i say dump it.

wdschei: Dumped.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
adrianjgeorge: Sounds like we have another story.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: Bad story deleted.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,362 @@
...
+        val destStrings = destinations.get()
+          .filter { case (_, dest) => dest.isInstanceOf[DestinationEndpoint] }
+          .map { case (_, endpoint) => s"${endpoint.getId}-${endpoint.asInstanceOf[DestinationEndpoint].getHostname}" }
+        logger.debug("{}:{} - Destinations for this router: {}", clusterId, nodeId, destStrings)
+      }
+    }
+    initialized = true
+
+    def getDestinations(destList: List[Destination]): Map[String, Destination] = {
+      destList.iterator.map(dest => dest.getId -> dest).toMap
```
dmnjohns: Why the `.iterator` call?

Also, since this code assumes destination IDs are unique (which should be the case, since it's enforced by the schema), you could do:
```
destList.groupBy(_.getId).mapValues(_.head)
```

Hm, and I wonder if we want to leave a comment stating that it is assumed that destination IDs must be unique.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,362 @@
...
+  }
+
+  override def configurationUpdated(configurationObject: SystemModel): Unit = {
+    //Set the current system model, and just update the nodes.
+    val interrogator = new SystemModelInterrogator(clusterId, nodeId)
+    localCluster.set(Option(interrogator.getLocalCluster(configurationObject).orElse(null)))
+    localNode.set(Option(interrogator.getLocalNode(configurationObject).orElse(null)))
+    defaultDestination.set(Option(interrogator.getDefaultDestination(configurationObject).orElse(null)))
+    destinations.set(localCluster.get().map(_.getDestinations) match {
+      case Some(dests) => getDestinations(dests.getEndpoint.asScala.toList) ++ getDestinations(dests.getTarget.asScala.toList)
```
dmnjohns: Why call `getDestinations` twice rather than just joining the lists and calling it once? Also, why convert to a `List` explicitly?

I might recommend:
```
getDestinations(dests.getEndpoint.asScala++ dests.getTarget.asScala)
```

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: Story created:
* https://repose.atlassian.net/browse/REP-7180

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: Wow. This is another one that could be collapsed into this.
Two more classes OBE'd.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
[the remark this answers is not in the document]
wdschei: Ok.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,362 @@
...
+      case None => Map.empty[String, Destination]
+    })
+    domains.set(Option(new Clusters(configurationObject)))
+
+    if (logger.underlying.isDebugEnabled) {
+      localCluster.get().foreach { cluster =>
```
dmnjohns: `().foreach {` could be `() foreach {`. I normally do it the latter way, but I'm not exactly sure why. I guess I've just mentally coupled dot notation with parenthesis. Obviously not something you need to do, just something I thought I'd point out.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: Fair enough.
Updated.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: No biggie.
Updated

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: Smashed down to one readable line.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
dmnjohns: The thought just crossed my mind that `.method {}` is the Groovy way.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: Ok. That made it a couple of lines.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,362 @@
...
+    initialized
+  }
+
+  override def service(req: HttpServletRequest, resp: HttpServletResponse): Unit = {
+    if (!initialized) {
+      logger.error(NotInitializedMessage)
+      trySendError()
+    } else {
+      logger.trace("{} processing request...", this.getClass.getSimpleName)
+      Try(doRoute(new HttpServletRequestWrapper(req), resp)) recover {
```
dmnjohns: I haven't really made up my mind about when to use `Try` vs `try/catch`. In this case, I lean toward using `try/catch` since you aren't doing anything with the return value of `doRoute` (or lack thereof, really). Under the hood, that's basically what `Try` is doing anyway, so this is just a wrapper to use a different syntax. Does that alone improve the code in some way? I'm not sure. It adds another object to the heap, but that's minimal.

🤷‍♂️

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: I like the `().method {` and `} method {` layout.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,362 @@
...
+      defaultDestination.get() match {
+        case Some(defDest) =>
+          List[RouteDestination] {
+            new RouteDestination(defDest.getId, servletRequest.getRequestURI, -1)
+          }
+        case _ => List.empty[RouteDestination]
+      }
+    }
+    (attributeDestinations ++ defaultDestinations)
+      .sortWith(_.compareTo(_) > 0) // Highest Quality first
```
dmnjohns: I think you can use `.sorted.reverse` here instead of this. Might be simpler?

wdschei: That ends up causing the list to be iterated over twice.
I wanted a `.lastOption`, but that doesn't exist.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,362 @@
...
+        .getOrElse(List.empty[RouteDestination]).asInstanceOf[List[RouteDestination]]
+    val defaultDestinations = {
+      defaultDestination.get() match {
+        case Some(defDest) =>
+          List[RouteDestination] {
+            new RouteDestination(defDest.getId, servletRequest.getRequestURI, -1)
+          }
+        case _ => List.empty[RouteDestination]
+      }
+    }
```
dmnjohns: The way this value is determined is gross. Might I suggest:

```
val defaultDestinations = defaultDestination.get()
        .map(defDest => new RouteDestination(defDest.getId, servletRequest.getRequestURI, -1))
        .toList
```

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: All of the uses of `try`/`catch` were flagged by Codacy as use `util.Try`. This seems to be the Scala way, so I did it.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
dmnjohns: So sure, maybe. That's why I called it out, after all. That solves the header coming in the door, and with some work, maybe even the header being added by another filter. I don't think there's ever a time where we take the response from an external service and copy a header onto the request, anyway. But my point is that I have to think about that. Why?

One reason that comes to mind is that we have logging around headers. Not so much around attributes. Of course, we could change that.

To me, per-request routing is functionality internal to Repose. There's no need to expose it to users, so why stick it in an HTTP field meant for transport? Handling it on the backend (as an attribute) works, and keeps me from thinking about the things I mentioned above (as much).

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,362 @@
...
+    destination match {
+      case endpoint: DestinationEndpoint =>
+        Some(buildEndpoint(localhost, localPorts, endpoint, uri, servletRequest))
+      case cluster: DestinationCluster =>
+        // TODO: Remove this when routing to a cluster is removed.
+        domains.get().map(_.getDomain(cluster.getCluster.getId)) match {
+          case Some(clusterWrapper) =>
+            val routableNode = clusterWrapper.getNextNode
+            val port = if (HttpsProtocol.equalsIgnoreCase(cluster.getProtocol)) routableNode.getHttpsPort else routableNode.getHttpPort
+            Some(new URL(cluster.getProtocol, routableNode.getHostname, port, cluster.getRootPath + uri))
```
dmnjohns: Hopefully this goes way, but should `StringUriUtilities.concatUris(cluster.getRootPath, uri)` be used here?

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,380 @@
...
+            }
+          case _ =>
+            logger.warn("No route-able node found; not forwarding")
+        }
+      case id =>
+        val localId = localCluster.map(_.getId) match {
+          case Some(locId) => locId
+          case None => ""
+        }
+        logger.warn("Invalid routing destination specified: {} for cluster: {} ", id, localId)
```
dmnjohns: Huh? `id` must be the `None` object, right? That's probably not what you mean to log. This case should probably be the default case:

``` 
case _ =>
```

... and this log line should log `routeDest.getDestinationId`.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,380 @@
...
+              }
+            }
+          case _ =>
+            logger.warn("No route-able node found; not forwarding")
+        }
+      case id =>
+        val localId = localCluster.map(_.getId) match {
+          case Some(locId) => locId
+          case None => ""
+        }
```
dmnjohns: You may want to replace this with:

```
val localId = localCluster.map(_.getId).getOrElse("")
```

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,380 @@
...
+        servletResponse.setStatus(SC_SERVICE_UNAVAILABLE)
+    }
+  }
+
+  def setVia(request: HttpServletRequestWrapper): Unit = {
+    val builder = new StringBuilder
+    builder.append(HttpServletRequestUtil.getProtocolVersion(request))
+    val requestVia = containerConfigurationService.getRequestVia
+    builder.append(" ")
+    if (requestVia.isPresent) builder.append(requestVia.get)
```
dmnjohns: This line doesn't do anything? It appends to the `StringBuilder`, but then the method returns, and the `StringBuilder` is GC'd without doing anything.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: Sure.
Updated.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: Yep that was a bleed over from all the copy/paste refactor and auto convert from Java to Scala.
Updated.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: Yep.
Updated.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: Yep. Another auto convert that failed.
Updated.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -0,0 +1,376 @@
...
+    dest match {
+      case endpoint: DestinationEndpoint => sb.append(endpoint.getRootPath)
+      case cluster: DestinationCluster => sb.append(cluster.getRootPath)
+      case _ => throw new IllegalArgumentException("Unknown destination type: " + dest.getClass.getName)
+    }
+    sb.toString
+  }
+
+  def markResponseCodeHelper(metricsService: MetricsService, responseCode: Int, component: String): Unit = {
+    if (100 < responseCode && responseCode < 600) {
```
dmnjohns: It's a subtle thing, but I like how `responseCode` variable is kept close together but using it on the RHS of the first expression and the LHS of the second.

repose-aggregator/core/repose-core/src/test/scala/org/openrepose/powerfilter/ReposeRoutingServletTest.scala
```diff
@@ -0,0 +1,230 @@
...
+
+  override def beforeEach(): Unit = {
+    super.beforeEach()
+
+    mockServletConfig = new MockServletConfig()
+    mockConfigurationService = mock[ConfigurationService]
+    mockRequestHeaderService = mock[RequestHeaderService]
+    mockResponseHeaderService = mock[ResponseHeaderService]
+    mockReportingService = mock[ReportingService]
+    mockMetricsService = Optional.empty()
```
dmnjohns: No tests showing that metrics are actually being reported?

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
dmnjohns: That thought did cross my mind, but I figured that this list would be small enough that we wouldn't care.

... but if we do, how about:

```
.sorted(Ordering[RouteDestination].reverse)
```

repose-aggregator/core/repose-core/src/test/scala/org/openrepose/powerfilter/ReposeRoutingServletTest.scala
```diff
@@ -0,0 +1,230 @@
...
+    reposeRoutingServlet = new ReposeRoutingServlet(
+      DefaultClusterId,
+      DefaultNodeId,
+      mockConfigurationService,
+      mockRequestHeaderService,
+      mockResponseHeaderService,
+      mockReportingService,
+      mockMetricsService)
```
dmnjohns: Looks like this constructor call wasn't updated with the constructor. Test will not compile at the moment.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
wdschei: It shouldn't be large, but if it were this would be an obscure thing to find.
Since it is easily avoidable I would rather sort it just once.
The `Ordering.reverse` is in essence doing the same thing, but I agree it is more readable.
Updated.

adrianjgeorge:
Lot of test failures

wdschei:
I evidently didn't run that locally before pushing it the last time after moving the Via code. The mock was returning a `null` instead of an `Optional.empty[String]()`.

PR #1973, opened 2018-08-10
REP-7128 Fully integrate the performance tests

adrianjgeorge:
When this is merged the two jobs that actually do the work will need to be updated with 
- [ ] Add the execution of the new gradle task
- [ ] Change where the packages are copied to match the new destination
- [ ] Change the ansible execution to match the new destination

PR #1974, opened 2018-08-15
REP-7203 Fixing performance tests

dmnjohns:


wdschei:
You might want to recreate this off of the #1973 so the merge conflicts aren't that bad.

dmnjohns:
Psh. Clearly #1973 should be based off this branch! 😉

PR #1981, opened 2018-09-04
REP-3589 HTTP Client Service Rewrite

dmnjohns:


repose-aggregator/components/services/http-client-service/http-client-service-api/src/main/resources/META-INF/schema/config/http-client-service.xsd
```diff
@@ -59,25 +58,46 @@
         </xs:annotation>
 
         <xs:sequence>
-            <xs:element name="pool" type="PoolType" minOccurs="1" maxOccurs="unbounded"/>
+            <xs:element name="client" type="HttpClientConfig" minOccurs="1" maxOccurs="unbounded"/>
```
adrianjgeorge: So the problem with this change is that everything else refers to `connection-pool-id`s.

dmnjohns: Ugh, that's true. If we are breaking things for 9 anyway, I wouldn't be overly opposed to renaming those, but I suppose it's only worth doing if we really care about how they are named. I wanted to change it since `pool` is kind of a misnomer given the configuration options, but we have used that term in a bunch of places now.

adrianjgeorge: Maybe split the difference and call it `clientPool` or `client-pool`

repose-aggregator/components/filters/keystone-v2-filter/src/test/scala/org/openrepose/filters/keystonev2/KeystoneV2FilterRcnTest.scala
```diff
@@ -102,24 +108,39 @@ class KeystoneV2FilterRcnTest extends FunSpec
...
-          anyBoolean()
-        )).thenReturn(new ServiceClientResponse(SC_OK, endpointsResponse()))
-        when(mockAkkaServiceClient.get(
-          mockitoEq(s"$GROUPS_KEY_PREFIX$VALID_USER_ID"),
-          mockitoEq(s"$identityServiceUri${GROUPS_ENDPOINT(VALID_USER_ID)}"),
-          argThat(hasEntry(CommonHttpHeader.AUTH_TOKEN, VALID_TOKEN)),
-          anyBoolean()
-        )).thenReturn(new ServiceClientResponse(SC_OK, groupsResponse()))
+        when(mockHttpClient.execute(any[HttpUriRequest], any[HttpContext]))
+            .thenAnswer(new Answer[CloseableHttpResponse] {
```
dmnjohns: I went to write a custom matcher and found this:
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/commons/commons-test/src/main/scala/org/openrepose/commons/test/MockitoAnswers.scala

At a minimum, I will use that to clean up these answers a bit. Hopefully the custom matcher will work out better, though.

repose-aggregator/components/services/http-client-service/http-client-service-api/src/main/java/org/openrepose/core/services/httpclient/CachingHttpClientContext.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public CachingHttpClientContext(final HttpContext context) {
+        super(context);
+    }
+
+    public CachingHttpClientContext() {
+        super();
+    }
+
+    public Boolean getUseCache() {
```
adrianjgeorge: Why big B? Then we have to worry about NPEs all  over the damn place.

repose-aggregator/components/services/http-client-service/http-client-service-api/src/main/resources/META-INF/schema/config/http-connection-pool.xsd
```diff
@@ -239,28 +229,35 @@
             </xs:annotation>
         </xs:attribute>
 
-        <xs:attribute name="chunked-encoding" type="chunkedEncodingType" use="optional" default="true">
```
adrianjgeorge: You dropped this entirely?

repose-aggregator/components/services/http-client-service/http-client-service-impl/src/main/scala/org/openrepose/core/services/httpclient/HttpClientDecommissioner.scala
```diff
@@ -0,0 +1,97 @@
...
+  * <p>
+  * This system manages its own background processing since it must only
+  * decommission clients when they are no longer in use. For the sake of
+  * consistency and concurrency, synchronization is used.
+  */
+@Named
+class HttpClientDecommissioner extends HttpClientUserManager with StrictLogging {
+
+  logger.debug("Scheduling HttpClientDecommissioner")
+  private val decomDelay: Duration = 5.seconds
```
adrianjgeorge: This also exists as an alternative. 
https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/scheduling/annotation/Scheduled.html

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/system-model/system-model.xsd
```diff
@@ -106,6 +106,28 @@
                     </xs:annotation>
                 </xs:attribute>
 
+                <xs:attribute name="chunked-encoding" type="mod:ChunkedEncoding" use="optional" default="true">
```
adrianjgeorge: I'm sure we talked about it, but why did we move this here again?

repose-aggregator/src/docs/asciidoc/architecture/system-model.adoc
```diff
@@ -170,6 +170,44 @@ If set to localhost, any physical node will match and run the *Repose* node.
...
 
+==== Streaming Request Body to the Origin Service
+Streaming a request body to the origin service is accomplished by using the `chunked` transfer encoding.
+
+[source,xml]
+.partial http-connection-pool.cfg
+----
+<system-model xmlns="http://docs.openrepose.org/repose/system-model/v2.0">
+  <repose-cluster id="repose_cluster1"
+                  chunked-encoding="auto"> <!--1-->
```
adrianjgeorge: So if we do move this to the system model, it makes more sense for it to be on the destinations, not the cluster. It's possible that they may have some destinations that don't support chunking, and if i recall the value is really only for communication to the origin service.

repose-aggregator/src/docs/asciidoc/architecture/system-model.adoc
```diff
@@ -170,6 +170,44 @@ If set to localhost, any physical node will match and run the *Repose* node.
...
+
+    <nodes>
+      <node id="repose_node", hostname="localhost", http-port="8080"/>
```
adrianjgeorge: lol, copied commas

repose-aggregator/src/docs/asciidoc/ver-9-upgrade-notes.adoc
```diff
@@ -1,4 +1,25 @@
 = Upgrade to Version 9 from Version 8 Notes
 
+== Core
```
adrianjgeorge: I'm not certain this section is necessary. We don't normally go into detail about changes in internal APIs. I feel like this is aimed at people who just care about configuring. I'm not against making a specific section for integrators though where we give details like this and we do a guide everytime we bump the first or second digit.

versions.properties
```diff
@@ -66,8 +68,8 @@ org.apache.logging.log4j:log4j-flume-ng=$log4jVersion
 org.apache.logging.log4j:log4j-iostreams=$log4jVersion
 org.apache.logging.log4j:log4j-jul=$log4jVersion
 org.apache.logging.log4j:log4j-slf4j-impl=$log4jVersion
-httpClientVersion=4.5.3
-httpCoreVersion=4.4.6
+httpClientVersion=4.5.6
+httpCoreVersion=4.4.10
```
adrianjgeorge: ugh, they are different? that makes me sad

repose-aggregator/components/services/http-client-service/http-client-service-api/src/main/java/org/openrepose/core/services/httpclient/CachingHttpClientContext.java
dmnjohns: I went back and forth and this one. I ended up on `Boolean` since the context itself is an `Object` store. Therefore, by using `Boolean`, I can be consistent with other attributes. So yes, you do have to deal with the potential for `null`, but you have to do that for all attributes rather than just some. Using `Boolean` also allows me to represent three distinct states: true, false, or not set. If I had used `boolean`, I would have to return a default value of either `true` or `false` when getting this attribute out of the context. Since the context is just an object store, I preferred putting that determination in the place where we actually use the value (thus giving it meaning).

repose-aggregator/components/services/http-client-service/http-client-service-api/src/main/resources/META-INF/schema/config/http-connection-pool.xsd
dmnjohns: I moved it into the system model.

repose-aggregator/components/services/http-client-service/http-client-service-impl/src/main/scala/org/openrepose/core/services/httpclient/HttpClientDecommissioner.scala
dmnjohns: I actually ran across that when I was trying to figure out the best way to handle the asynchronicity of decommissioning, but ended up going this route for... some reason...

Oh. I think it was because you have to explicitly configure Spring to enable scheduling via annotations:
https://spring.io/guides/gs/scheduling-tasks/#_enable_scheduling

I'm pretty sure that would just involve creating a configuration class with some annotations for Spring to pick up when it scans for this service, but I choose not go down that rabbit hole since this seemed to work fine.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/system-model/system-model.xsd
dmnjohns: Chunked encoding, as we call it, is a property of a request, not a client. Therefore, I thought it should be configured for the requestor, not the client.

Configuring it for the client meant that we were setting a parameter on the client itself. To use that parameter, we would get the client when we were about to route a request to the origin service and retrieve the parameter by name. If you take a look at the old code, you can see how indirect that is. Also, the client API deprecated parameters on the client.

repose-aggregator/src/docs/asciidoc/architecture/system-model.adoc
dmnjohns: I thought that too, actually. I originally put it on `destinations` since there is not currently a way to disable chunked encoding for one `destination` but enable it for another. Then I moved it here since you can only have one set of destinations anyway, and this attribute living next to the `rewrite-host-header` attribute made sense to me at the time.

That said, I'm totally cool with dropping this onto `destinations` again. I would actually be happy to drop it onto `destination`, but that would be an enhancement that requires a bit more work.

repose-aggregator/src/docs/asciidoc/architecture/system-model.adoc
dmnjohns: Dang it. I caught it in the `destination`, but not on the `node`, apparently. Will fix.

repose-aggregator/src/docs/asciidoc/ver-9-upgrade-notes.adoc
dmnjohns: I was wondering about that myself. I asked myself WWMD (what would Mario do) and decided to just type everything up since it's easy enough to move or remove.

I'm cool with sticking this in an integrator section.

versions.properties
dmnjohns: They are. It makes me sad, too. I checked multiple times, though, and I am certain that I matched the versions correctly.

repose-aggregator/components/services/http-client-service/http-client-service-impl/src/main/scala/org/openrepose/core/services/httpclient/HttpClientDecommissioner.scala
adrianjgeorge: you could also just add it to our existing application context

repose-aggregator/src/docs/asciidoc/architecture/system-model.adoc
adrianjgeorge: yeah, i was suggesting it should be on the individual destinations.

repose-aggregator/src/docs/asciidoc/ver-9-upgrade-notes.adoc
adrianjgeorge: Yeah, lets go ahead and make a specific section for that

repose-aggregator/components/services/http-client-service/http-client-service-api/build.gradle
```diff
@@ -1,21 +1,10 @@
...
     testCompile "org.scalatest:scalatest_$scalaMajDotMin"
 }
-
-jacocoTestCoverageVerification {
-    violationRules {
-        rules.forEach { rule ->
-            rule.limits.forEach { limit ->
-                if (limit.counter == 'INSTRUCTION' && limit.value == 'COVEREDRATIO') {
-                    // todo: set to the current coverage -- need more coverage, but we don't want the build to fail for the baseline
-                    limit.minimum = 0.02
```
wdschei: This can be increased, but it will need to remain and be at least `0.28`.

dmnjohns: Done. It's a shame it has to be set so low since all of the code that runs is covered...

wdschei: I know. We just need to move everything to Scala. Problem solved.

repose-aggregator/src/docs/asciidoc/architecture/system-model.adoc
dmnjohns: Sure. How do we want to handle per-destination chunking, though?

As of this PR, it would be non-trivial to implement. However, once we drop the routing complications added by supporting the WAR deployment, it becomes much more trivial. Ideally, we would add per-destination chunking as a feature after both of those stories, but I realize I am already messing with the schema here, so...

repose-aggregator/components/services/http-client-service/http-client-service-impl/src/main/scala/org/openrepose/core/services/httpclient/HttpClientDecommissioner.scala
dmnjohns: How would you add it to the existing application context? I thought that would just be adding a class annotated with `@Configuration` and `@EnableScheduling` to the package being scanned to create the core context?

Oh. I think I found the answer in your link while I was writing this comment. It looks like you should also be able to do:
```
coreContext.register(ScheduledAnnotationBeanPostProcessor.class)
```

Sure, I'll look into switching over to this.

dmnjohns:
run performance tests

dmnjohns:
run performance tests

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
```diff
@@ -414,11 +425,10 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
         try {
-            try {
-                // ensures that the method name exists
-                HttpComponentFactory.valueOf(wrappedRequest.getMethod().toUpperCase());
-            } catch (IllegalArgumentException iae) {
-                throw new InvalidMethodException("Request contained an unknown method.", iae);
+            // Ensures that the method name is supported
+            // todo: HTTP request methods are case-sensitive, so this check should not upper case the request method
+            if (!SUPPORTED_HTTP_METHODS.contains(wrappedRequest.getMethod().toUpperCase())) {
+                throw new InvalidMethodException(wrappedRequest.getMethod() + " method not supported");
```
dmnjohns: I ended up leaving this check in for backwards compatibility. This is no longer an Apache HttpClient constraint, but purely a Repose constraint.

I actually attempted to drop this check since we are no longer forced to support a subset of methods, but `CONNECT` gave me some issues. It's a method meant to work with proxies specifically, and Repose certainly does not support it. If desired, I could just blacklist `CONNECT` rather than whitelisting the old list of supported methods.

dmnjohns:
As of writing this comment, the remaining failing tests are either due to the expected version issues or flakiness. That is, the remaining tests do not fail locally.

repose-aggregator/components/filters/valkyrie-authorization-filter/build.gradle
```diff
@@ -20,7 +22,10 @@ dependencies {
     testCompile project(":repose-aggregator:commons:commons-test")
     testCompile "junit:junit"
     testCompile "org.scalatest:scalatest_$scalaMajDotMin"
-    testCompile "org.mockito:mockito-all"
+    testCompile "org.hamcrest:hamcrest-all"
+    testCompile("org.mockito:mockito-all") {
```
adrianjgeorge: Probably don't want all if we are to exclude hamcrest

repose-aggregator/docs/src/asciibinder/architecture/system-model.adoc
```diff
@@ -170,6 +170,44 @@ If set to localhost, any physical node will match and run the *Repose* node.
...
+
+    <filters/>
+
+    <destinations>
+      <endpoint id="origin_service" protocol="http" port="80" default="true"/>
+    </destinations>
+  </repose-cluster>
+</system-model>
+----
+<1> Determines whether or not the request body should be `chunked` when forwarded to the origin service.
```
adrianjgeorge: I thought we decided to put this on the destinations? In fact i thought i saw that in the code? Forget to update the docs?

repose-aggregator/docs/src/asciibinder/welcome/release-notes.adoc
```diff
@@ -3,7 +3,20 @@
 == 9.0.0.0 (2018-??-??)
 <<ver-9-upgrade-notes.adoc, Upgrade Notes>>
 
-* https://repose.atlassian.net/browse/REP-4992[REP-4922] - Removed ambiguous setting for chunked encoding options on <<services/http-connection-pool.adoc, HTTP Client/Connection Pool Service>>.
+* https://repose.atlassian.net/browse/REP-4992[REP-4922] - Removed ambiguous setting for chunked encoding options on <<services/http-client.adoc, HTTP Client/Connection Pool Service>>.
+* https://repose.atlassian.net/browse/REP-3589[REP-3589] - Rewrote the <<services/http-client.adoc#, HTTP Client Service>>.
```
adrianjgeorge: A lot of these bullets seem like they should be in the upgrade notes not here.

repose-aggregator/components/filters/valkyrie-authorization-filter/build.gradle
dmnjohns: Fair enough. I don't think I checked at the time, but the `mockito-all` artifact just combines `mockito-core` with an `org.hamcrest` package and an `org.objenesis` package. Since we aren't using `objenesis` ourselves, I don't think we need it, so I'll try to downgrade this to `mockito-core`.

As a side note, `hamcrest` is clearly not packaged within `mockito-core`. I wonder why I added an exclusion for it even where I switched to `mockito-core` dependencies. I guess it might be a transitive. I'll verify that the exclusion is necessary and remove it if not.

repose-aggregator/tests/functional-tests/src/integrationTest/configs/features/filters/keystonev2/timeout/diffpool/http-connection-pool.cfg.xml
```diff
@@ -1,27 +1,25 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <http-connection-pools xmlns="http://docs.openrepose.org/repose/http-connection-pool/v1.0">
     <pool id="default"
-          default="false"
+          default="true"
```
adrianjgeorge: why were these swapped?

repose-aggregator/tests/functional-tests/src/integrationTest/configs/features/filters/keystonev2/timeout/httpConnTimeout60sec/http-connection-pool.cfg.xml
```diff
@@ -1,27 +1,25 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <http-connection-pools xmlns="http://docs.openrepose.org/repose/http-connection-pool/v1.0">
     <pool id="default"
-          default="true"
+          default="false"
```
adrianjgeorge: these too

repose-aggregator/components/filters/valkyrie-authorization-filter/build.gradle
adrianjgeorge: I'd move away from using all everywhere if we can

repose-aggregator/core/repose-core/src/main/java/org/openrepose/powerfilter/PowerFilter.java
adrianjgeorge: leave the list as is for now.

repose-aggregator/docs/src/asciibinder/architecture/system-model.adoc
dmnjohns: Dang it. Yes, I just forgot to update the docs. Will fix.

repose-aggregator/docs/src/asciibinder/welcome/release-notes.adoc
dmnjohns: I was split on that. I ended up putting them here following the idea a change modifies the code running on a user's machine, and therefore the change should be documented in the release notes. Then I put more details on the upgrade page.

repose-aggregator/tests/functional-tests/src/integrationTest/configs/features/filters/keystonev2/timeout/diffpool/http-connection-pool.cfg.xml
dmnjohns: Hah, I don't think it was swapped. I think GitHub is confused about what file was moved. The "original" file (on master) would be:
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/tests/functional-tests/src/integrationTest/configs/features/filters/keystonev2/akkatimeout/diffpool/http-connection-pool.cfg.xml

repose-aggregator/tests/functional-tests/src/integrationTest/configs/features/filters/keystonev2/timeout/httpConnTimeout60sec/http-connection-pool.cfg.xml
dmnjohns: Same thing as above. I think this is the one that GitHub confused the other for.

wdschei:
Also Codacy doesn't like us using the `scala.util.Random` I see.

dmnjohns:
It doesn't, but since it's in tests I didn't overly care. Scala's `Random` has a few enrichments I kind of like.

PR #1986, opened 2018-09-25
REP-7249 Rebuilt docs using AsciiBinder.

wdschei:


wdschei:
Since the top-level `index.html` requires all of the pre-processing provided by GH-Pages, you can look at the rebuilt versioned docs starting at `./versions/index.html`.

adrianjgeorge:
It appears that a lot of the links on the welcome page are broken?

wdschei:
> It appears that a lot of the links on the welcome page are broken?

That is super weird. It is totally ignoring the `../` in the pathing when the adoc is converted to html.

PR #1994, opened 2018-11-06
REP-6603 Support for multiple capture groups in the URL to header filter

dmnjohns:


repose-aggregator/components/filters/url-extractor-to-header-filter/src/test/scala/org/openrepose/filters/urlextractortoheader/UrlExtractorToHeaderFilterTest.scala
```diff
@@ -118,6 +119,48 @@ class UrlExtractorToHeaderFilterTest extends FunSpec with BeforeAndAfterEach wit
...
+      servletRequest.setRequestURI("/v1/hybrid:45678/entities/96")
+
+      filter.doFilter(servletRequest, servletResponse, filterChain)
+
+      val requestCaptor = ArgumentCaptor.forClass(classOf[HttpServletRequestWrapper])
+      verify(filterChain).doFilter(requestCaptor.capture(), any(classOf[ServletResponse]))
+      requestCaptor.getValue.getHeaders("X-Device-Id").asScala.toStream should contain only "hybrid:45678"
+    }
+
+    it("should add a header with the regex value when the configured regex partially matched the URL") {
```
adrianjgeorge: Is this the behavior we want? It doesn't seem like it is to me.

dmnjohns: I meant to comment on this when I opened the PR. The answer is, I'm not sure. Both a partial or a full match are acceptable solutions, I think, because there are tokens in the regex string that match the start or end of a line as well as tokens that are wildcard matches. In this case, that means a user can effectively force a partial match to act like a full match and vice-versa.

However, changing it would mean that the behavior of this filter will change, potentially forcing even more people to update their configs.

Since I would consider it out-of-scope for this story, I just added the test to call out the behavior explicitly.

adrianjgeorge: If we keep it, it should be called out in the docs.

dmnjohns: I will note that due to constraints on Scala's `Regex` type, fully matching with a number of capture groups that is not predetermined is painful. To make it work at all, we would basically have to rewrite the `unapply` method for `Regex`.

If we did want to switch to full-match, I would probably use the Java `Pattern` API instead. It's almost as clean, but it does expose its own mutability.

Otherwise I'll update the docs.

PR #1997, opened 2018-11-08
REP-7438 Updated some references to repose-valve

adrianjgeorge:


dmnjohns:
It looks like the repose role for the performance tests weren't updated here? At a minimum, the following line will need to be updated:
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/tests/performance-tests/src/performanceTest/resources/roles/repose/tasks/install_repose.yml#L88

repose-aggregator/tests/release-verification/src/scripts/repose_install_deb.sh
```diff
@@ -17,10 +17,10 @@ echo "Installing Repose ${name} package"
...
+  dpkg -i /release-verification/repose*_all.deb /release-verification/repose-filter-bundle*_all.deb /release-verification/repose-extensions-filter-bundle*_all.deb /release-verification/repose-experimental-filter-bundle*_all.deb
   apt-get install -y -f
 else
-  apt-get install -y repose-valve${version} repose-filter-bundle${version} repose-extensions-filter-bundle${version} repose-experimental-filter-bundle${version}
+  apt-get install -y repose${version} repose-filter-bundle${version} repose-extensions-filter-bundle${version} repose-experimental-filter-bundle${version}
 fi
-mkdir -p /etc/systemd/system/repose-valve.service.d && echo "[Service]\nEnvironment=\"JAVA_OPTS=-Xdebug -Xrunjdwp:transport=dt_socket,address=10037,server=y,suspend=n\"" > /etc/systemd/system/repose-valve.service.d/local.conf
-sed -i '/JAVA_OPTS/c\JAVA_OPTS="-Xdebug -Xrunjdwp:transport=dt_socket,address=10037,server=y,suspend=n"' /etc/sysconfig/repose
\ No newline at end of file
+mkdir -p /etc/systemd/system/repose.service.d && echo "[Service]\nEnvironment=\"JAVA_OPTS=-Xdebug -Xrunjdwp:transport=dt_socket,address=10037,server=y,suspend=n\"" > /etc/systemd/system/repose-valve.service.d/local.conf
```
wdschei: The `-valve` needs removed from the directory the `local.conf` file is being written to.

repose-aggregator/tests/release-verification/src/scripts/repose_install_rpm.sh
```diff
@@ -17,9 +17,9 @@ echo "Installing Repose ${name} package"
...
+  yum --nogpgcheck localinstall -y /release-verification/repose-*.noarch.rpm /release-verification/repose-filter-bundle-*.noarch.rpm /release-verification/repose-extensions-filter-bundle-*.noarch.rpm /release-verification/repose-experimental-filter-bundle-*.noarch.rpm
 else
-  yum install -y repose-valve${version} repose-filter-bundle${version} repose-extensions-filter-bundle${version} repose-experimental-filter-bundle${version}
+  yum install -y repose${version} repose-filter-bundle${version} repose-extensions-filter-bundle${version} repose-experimental-filter-bundle${version}
 fi
-mkdir -p /etc/systemd/system/repose-valve.service.d && echo "[Service]\nEnvironment=\"JAVA_OPTS=-Xdebug -Xrunjdwp:transport=dt_socket,address=10037,server=y,suspend=n\"" > /etc/systemd/system/repose-valve.service.d/local.conf
-sed -i '/JAVA_OPTS/c\JAVA_OPTS="-Xdebug -Xrunjdwp:transport=dt_socket,address=10037,server=y,suspend=n"' /etc/sysconfig/repose
\ No newline at end of file
+mkdir -p /etc/systemd/system/repose.service.d && echo "[Service]\nEnvironment=\"JAVA_OPTS=-Xdebug -Xrunjdwp:transport=dt_socket,address=10037,server=y,suspend=n\"" > /etc/systemd/system/repose-valve.service.d/local.conf
```
wdschei: The `-valve` needs removed from the directory the `local.conf` file is being written to.

wdschei:
The package name change from `repose-valve` to just `repose` should probably be in the upgrade guide too.
I know it is in the recipe, but not everyone is going to go look there when they are just upgrading from valve v8.x to v9.x.

adrianjgeorge:
> It looks like the repose role for the performance tests wasn't updated here? At a minimum, the following line will need to be updated:
> https://github.com/rackerlabs/repose/blob/master/repose-aggregator/tests/performance-tests/src/performanceTest/resources/roles/repose/tasks/install_repose.yml#L88

I'm not certain i see what this line is doing, so i wouldn't mind a hand fixing it

dmnjohns:
> > It looks like the repose role for the performance tests wasn't updated here? At a minimum, the following line will need to be updated:
> > https://github.com/rackerlabs/repose/blob/master/repose-aggregator/tests/performance-tests/src/performanceTest/resources/roles/repose/tasks/install_repose.yml#L88
> 
> I'm not certain i see what this line is doing, so i wouldn't mind a hand fixing it

The gist of it is that line is looking for the `repose-valve.deb` file in a specific directory. The task then installs that package. The way it locates the correct package file is to get all the files in the directory (`repose_packages.files`), look at the path of each file (`map(attribute='path')`), select the paths matching a pattern (`select('match', '.*valve.*')`), creating a list of the results (`list`), taking the first item in that list since we assume there is only one item (`first`), then stripping the path of everything except the filename (`basename`).

Oh, and it is using Jinja2 template processing. That's where the Jinja2 filters (e.g., `select(..., ...)`) are coming from.

From a view further out, that task is part of installing Repose from a local package, which is part of the process to build Repose once, pass around the artifacts, and install on all performance test nodes.

PR #2000, opened 2018-11-13
REP-4990 Remove Identity v3 attributes

wdschei:


dmnjohns:
I believe most if not all of the remaining test failures for the Identity v3 filter are due to this branch not being based on `REP-7231`.

repose-aggregator/components/filters/openstack-identity-v3-filter/src/main/scala/org/openrepose/filters/openstackidentityv3/utilities/OpenStackIdentityV3API.scala
```diff
@@ -56,11 +55,17 @@ class OpenStackIdentityV3API(config: OpenstackIdentityV3Config, datastore: Datas
   private val identityServiceUri = config.getOpenstackIdentityService.getUri
   private val timeouts = Option(config.getCache).flatMap(cache => Option(cache.getTimeouts))
   private val cacheOffset =
-    timeouts.flatMap(timeouts => Option(timeouts.getVariance)).getOrElse(config.getCacheOffset)
+    timeouts.flatMap(timeouts => Option(timeouts.getVariance)).getOrElse(0)
```
dmnjohns: I really don't love having the second source of truth here (using `getOrElse(0)`). I'm tempted to suggest using David's old approach and fix the config in `configurationUpdated` to always have a `timeouts` field. That way, if the user does not provide a value we can still use the default value from the configuration. Not that I particularly like that solution, either...

dmnjohns: Also, I would consider replacing this line with the following:
```
timeouts.map(_.getVariance).flatMap(Option.apply).getOrElse(0)
```

In my opinion, it reads more fluidly that way.

repose-aggregator/components/filters/openstack-identity-v3-filter/src/main/scala/org/openrepose/filters/openstackidentityv3/utilities/OpenStackIdentityV3API.scala
```diff
@@ -56,11 +55,17 @@ class OpenStackIdentityV3API(config: OpenstackIdentityV3Config, datastore: Datas
...
+  // TODO: The Token and Group Cache TTL defaults (i.e. 600) will need moved to the XSD when the deprecated elements
+  // TODO: are removed and the attributes can directly have defaults.
   private val tokenCacheTtl =
-    timeouts.flatMap(timeouts => Option(timeouts.getToken).map(_.intValue())).getOrElse(config.getTokenCacheTimeout)
+  timeouts.flatMap(timeouts => Option(timeouts.getTokenElement).map(_.intValue()))
+    .getOrElse(timeouts.flatMap(timeouts => Option(timeouts.getToken).map(_.intValue()))
+      .getOrElse(600))
```
dmnjohns: Since the attributes are a new addition, it makes sense to me to give priority to the attributes over the elements (which are deprecated by this PR).

dmnjohns: Also, since `intValue()` does not mutate any state, I would recommend dropping the unnecessary parenthesis. So `.map(_.intValue)`.

dmnjohns: Also, I would do the same as before to simplify, and I would flatten out the structure. Like so:
```
timeouts.map(_.getToken).flatMap(Option.apply).map(_.intValue)
```

dmnjohns: Oh, and to further flatten out  the structure, I would make use of the `orElse` method.

Ultimately, I came up with:
```
timeouts.map(_.getToken).flatMap(Option.apply)
  .orElse(timeouts.map(_.getTokenElement).flatMap(Option.apply))
  .map(_.intValue)
  .getOrElse(600)
```

Or if you really wanted to minimize redundancy:
```
timeouts.flatMap(t => Option(t.getToken).orElse(Option(t.getTokenElement))
  .map(_.intValue)
  .getOrElse(600)
```

repose-aggregator/components/filters/openstack-identity-v3-filter/src/test/scala/org/openrepose/filters/openstackidentityv3/config/OpenStackIdentityV3SchemaTest.scala
```diff
@@ -32,53 +33,151 @@ class OpenStackIdentityV3SchemaTest extends ConfigurationTest {
...
+               |</openstack-identity-v3>""".stripMargin
+          val
+          exception = intercept[SAXParseException] {
+            validator.validateConfigString(config)
+          }
+          exception.getLocalizedMessage should include(s"Cannot define both a deprecated $cacheTimeout element and the new $cacheTimeout attribute")
+        }
+      }
+
+      cacheTimeouts.permutations.map(_.take(2)) foreach { timeouts =>
```
dmnjohns: Why the `_.take(2)` here? It doesn't seem necessary?

dmnjohns: Also, for the sake of consistency, maybe make the ` foreach {` call `.foreach {`.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/identityv3/cache/IdentityV3CacheOffSetOldTest.groovy
```diff
@@ -1,153 +0,0 @@
...
-
-        then: "All calls should hit identity"
-        //since we are talking about time based testing, we cannot always validate against a concrete number.  This is testing a range of requests.
-        fakeIdentityV3Service.validateTokenCount == uniqueUsers
-
-        where:
-        uniqueUsers | initialCallsPerUser
-        50          | 1
-    }
-}
```
dmnjohns: Should we keep this test and the other deleted test? It looks like they were testing the deprecated configuration style. While we got rid of the previously deprecated style, we introduced a newly deprecated style, so it seems like these would serve the same function they have been serving.

repose-aggregator/components/filters/openstack-identity-v3-filter/src/main/scala/org/openrepose/filters/openstackidentityv3/utilities/OpenStackIdentityV3API.scala
wdschei: Updated.

repose-aggregator/components/filters/openstack-identity-v3-filter/src/main/scala/org/openrepose/filters/openstackidentityv3/utilities/OpenStackIdentityV3API.scala
wdschei: Updated.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/identityv3/cache/IdentityV3CacheOffSetOldTest.groovy
wdschei: We don't seem to consistently test this way and since the schemas are checked for this capability I went ahead and just removed the old configs.

PR #2002, opened 2018-11-14
REP-7428 Removing uri-extraction-regex attribute from the Keystone v2 filter configuration

dmnjohns:


repose-aggregator/components/filters/keystone-v2-filter/src/main/resources/META-INF/schema/config/keystone-v2-common.xsd
```diff
@@ -105,8 +102,7 @@
                 <xs:complexType>
                     <!-- TODO: Does the authorization filter need a default-tenant-quality and roles-tenant-quality? -->
                     <xs:attribute name="default-tenant-quality" type="QualityType" default="0.9" use="optional"/>
-                    <!-- TODO: Should this be renamed to matching-tenant-quality? Not backwards compatible. -->
-                    <xs:attribute name="uri-tenant-quality" type="QualityType" default="0.7" use="optional"/>
+                    <xs:attribute name="validated-tenant-quality" type="QualityType" default="0.7" use="optional"/>
```
dmnjohns: I went ahead and renamed this attribute as well since it no longer makes sense (after this PR) and this will be pulled in for a major release.

As far as the name goes, I settled on `validated-tenant-quality` to match the `validate-tenant` element name. However, I also considered the following names:

- `matching-tenant-quality`
- `required-tenant-quality`
- `expected-tenant-quality`
- `validated-tenant-quality`

PR #2003, opened 2018-11-15
REP-6663 Forwarding the reason phrase from the origin service

dmnjohns:


repose-aggregator/docs/src/asciibinder/welcome/ver-9-upgrade-notes.adoc
```diff
@@ -3,6 +3,8 @@
 == Core Behavior Changes
 . Headers will no longer be split unnecessarily at the boundaries of *Repose*.
   If you were counting on this behavior for some reason see <<../recipes/header-splitting.adoc, Pre 9.0 Header Splitting>>.
+. The response status line reason phrase from the origin service will now be forwarded from the origin service.
+  If you were expecting *Repose* to normalize the reason phrase, that will no longer be the case.
```
wdschei: There should probably be a note here like for the Header Splitting above that points to the recipe for the Scripting filter to get the old behavior.

dmnjohns: Oops. Done!

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/powerfilter/ReposeRoutingServlet.scala
```diff
@@ -268,13 +268,8 @@ class ReposeRoutingServlet @Inject()(@Value(ReposeSpringProperties.CORE.REPOSE_V
...
     // Translate the HTTP client response to a servlet response
-    val responseCode = clientResponse.getStatusLine.getStatusCode
-    val responseProcessor = new HttpComponentResponseProcessor(clientResponse, servletResponse, responseCode)
-    if (responseCode >= HttpServletResponse.SC_MULTIPLE_CHOICES && responseCode < HttpServletResponse.SC_NOT_MODIFIED) {
-      responseProcessor.sendTranslatedRedirect()
-    } else {
-      responseProcessor.process()
-    }
+    val responseProcessor = new HttpComponentResponseProcessor(clientResponse, servletResponse)
+    responseProcessor.process()
```
adrianjgeorge: This feels kind of stupid. why are we making a thing, and then calling a no argument method on it to mutate something we passed in the constructor?

dmnjohns: In this case, I just didn't change it. That's how we have been doing it.

When I was rewriting parts of the routing code, I strongly considered rewriting the request and response processors to be collections of static functions, I just never went ahead and did it.

Of course, internal to the processors being an object means that processing can be split across multiple methods with shared state rather than requiring that shared state to be passed around as arguments. Which is not a good reason to continue doing things that way, it's just an observation, mostly.

dmnjohns: That was all to say, I agree, but my question is whether to make that change as part of this PR, or leave it for a different story?

adrianjgeorge: you already changed basically every line in the class being called. Just go ahead and remove the stupid, on this one at least.

dmnjohns: Done.

PR #2005, opened 2018-11-28
REP-7427 Adding tests to verify that Repose can handle header values with characters allowed by HTTP specification

dmnjohns:
Based off of #1998 since this test was affected by those changes.

There is only a single test failure. That failure is caused by Repose dropping request headers with an empty value. This behavior should not be accepted, but it is a known issue.


dmnjohns:
Ah. I should add negative tests as well...

PR #2006, opened 2018-12-05
Resolving the remote datastore health check service issue

dmnjohns:
This change will make it so that if the remote datastore service is removed from the services in the system model, its health check issues will also be removed. Without this change, Repose would continue returning 503s because of the reported health check issue even though the service is not being used.

This PR needs a story. I have opened a PR first since I already had the change in place from looking into the issue.

Note that services as first class citizens may also solve this issue (assuming health check issues are resolved when services are unloaded due to removal from the system model).

**EDIT:** The `dist-datastore` is also affected by this issue.


dmnjohns:
OBE'd by #2067

PR #2008, opened 2019-01-11
REP-7314 Remove Multi-Cluster Support

dmnjohns:


dmnjohns:
Since this is not being merged into master, Codacy is not running a check against it. If this PR is merged before #1998 , the Codacy check will run on #1998. If #1998 is merged before this PR, I will update this PR to point to master and the Codacy check will run here.

repose-aggregator/docs/src/asciibinder/architecture/system-model.adoc
```diff
@@ -73,101 +61,94 @@ There are several ways to go about getting around this, but one of the most gene
...
      The default destination endpoint will be responsible for handling all traffic that is not explicitly routed to a different destination endpoint.
-     Remember that a target can be specified rather than an endpoint if you want to use round-robin routing to a service cluster.
 
 === Filter Activation Determination
 [source,xml]
 .system-model.cfg.xml
 ----
 <system-model xmlns="http://docs.openrepose.org/repose/system-model/v2.0">
-  <repose-cluster id="repose_cluster1"> <!--1-->
     <nodes>
```
adrianjgeorge: It looks like you didn't reset the tabbing on this element.

adrianjgeorge: Maybe even all of them?

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/systemmodel/config/SystemModelSchemaTest.scala
```diff
@@ -180,19 +180,6 @@ class SystemModelSchemaTest extends ConfigurationTest {
       validator.validateConfigString(config)
     }
 
-    it("should reject the config if there are no defined destination endpoints nor targets") {
```
wdschei: Why was this test removed?

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/system-model/system-model.xsd
```diff
@@ -528,6 +464,8 @@
                 </xs:documentation>
             </xs:annotation>
         </xs:attribute>
+        <xs:attribute name="default" type="xs:boolean" use="optional" default="false">
```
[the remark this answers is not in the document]
wdschei: It didn't have it in the original either, so no biggie.

wdschei:
Why are we not updating the Linter for these changes?

adrianjgeorge:
Seems to be a compilation error in the versioning filter tests

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/systemmodel/config/SystemModelSchemaTest.scala
dmnjohns: Since getting rid of clusters means that we can also get rid of `target` destinations we no longer need an `assert` to verify that at least one `endpoint` or `target`. Instead, that requirement is part of the schema itself (encoded as the `minOccurs="1"` attribute and value).

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/system-model/system-model.xsd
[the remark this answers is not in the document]
dmnjohns: Yeah, I was trying to limit the scope so I didn't end up shaving yaks for the rest of my life. This is one of those things I just didn't address. In truth, we are inconsistent about XML schema documentation in general, this is just an instance of that.

Still, I'll go ahead and add documentation here.

repose-aggregator/core/repose-core-api/src/test/scala/org/openrepose/core/systemmodel/config/SystemModelSchemaTest.scala
wdschei: We normally test schema assertions/expectations just to ensure they aren't accidentally changed in the future.

repose-aggregator/core/repose-core-api/src/main/resources/META-INF/schema/system-model/system-model.xsd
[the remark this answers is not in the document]
wdschei: Yep. I was just pointing it out since I was there.

repose-aggregator/docs/src/asciibinder/architecture/system-model.adoc
dmnjohns: Gah, missed this one.

On it.

adrianjgeorge:
I have a better question. Does the linter in it's current incarnation even need to still exist? It was written so customers could check their configs for compatibility with a product i don't believe exists anymore.

dmnjohns:
> Why are we not updating the Linter for these changes?

Ah. I probably should. 

Of course, we will still have test configurations for the Repose Lint tool that have clusters since the Lint tool must be able to process old configs.

Technically, the Repose Lint tool does not have to be tied to the Repose version since it's a separate artifact that just inspects configuration files. However, we do currently version and ship the Repose Lint tool at the same time and in the same way as our main Repose artifacts, so that's not really an excuse.

Which brings us to...

> I have a better question. Does the linter in it's current incarnation even need to still exist? It was written so customers could check their configs for compatibility with a product i don't believe exists anymore.

Probably not? I actually wonder if it was ever used by anyone.

PR #2011, opened 2019-01-16
REP-7387 IP User X-Forwarded-For bug fix

adrianjgeorge:


repose-aggregator/components/filters/ip-user-filter/src/main/scala/org/openrepose/filters/ipuser/IpUserFilter.scala
```diff
@@ -77,16 +77,21 @@ class IpUserFilter @Inject()(configurationService: ConfigurationService) extends
...
-      logger.trace("IP User filter passing request...")
-      filterChain.doFilter(request, servletResponse)
+        //Always set the user header name to the current IP address
+        request.addHeader(userHeaderName, clientIpAddress, userHeaderQuality)
 
+        logger.trace("IP User filter passing request...")
+        filterChain.doFilter(request, servletResponse)
+      } catch {
+        case uhe: UnknownHostException =>
+          logger.debug(s"Bad X-Forwarded-For header: $clientIpAddress", uhe)
```
dmnjohns: Technically, this exception could be thrown either by a bad `X-Forwarded-For` header or by a bad value returned by `getRemoteAddr`.

repose-aggregator/components/filters/ip-user-filter/src/test/scala/org/openrepose/filters/ipuser/IpUserFilterTest.scala
```diff
@@ -159,6 +159,17 @@ class IpUserFilterTest extends FunSpec with BeforeAndAfterEach with Matchers {
...
 
+  describe("when presented with non-sense") {
+    it("should return a 400 when X-Forwarded-For is malformed") {
+      servletRequest.setRemoteAddr("10.1.2.3")
+      servletRequest.addHeader("X-Forwarded-For", "banana-phone")
+
+      ipUserFilter.doFilter(servletRequest, servletResponse, filterChain)
+
+      servletResponse.getStatus shouldBe 400
+    }
```
dmnjohns: Perhaps add a test to cover if `setRemoteAddr` returns a bad value? That could happen if a bad wrapper is used.

repose-aggregator/components/filters/ip-user-filter/src/test/scala/org/openrepose/filters/ipuser/IpUserFilterTest.scala
```diff
@@ -159,6 +159,17 @@ class IpUserFilterTest extends FunSpec with BeforeAndAfterEach with Matchers {
...
   }
 
+  describe("when presented with non-sense") {
+    it("should return a 400 when X-Forwarded-For is malformed") {
+      servletRequest.setRemoteAddr("10.1.2.3")
+      servletRequest.addHeader("X-Forwarded-For", "banana-phone")
+
+      ipUserFilter.doFilter(servletRequest, servletResponse, filterChain)
+
+      servletResponse.getStatus shouldBe 400
```
dmnjohns: The `400` could be replaced by `HttpServletResponse.SC_BAD_REQUEST`. However, I'm not sure I care since this is in a test. In fact, I wonder if using `400` was intentional to match the test name.

repose-aggregator/docs/src/asciibinder/filters/ip-user.adoc
```diff
@@ -51,7 +51,7 @@ This filter does not modify the response body.
 This filter does not create/modify any response headers.
 
 === Response Status Codes
-This filter does not modify the response code.
+* Returns a `400` when the `X-Forwarded-For` header is improperly formatted.
```
dmnjohns: Pointing this out to bring up the `getRemoteAddr` thing again.

repose-aggregator/components/filters/ip-user-filter/src/main/scala/org/openrepose/filters/ipuser/IpUserFilter.scala
adrianjgeorge: Yes, but I don't see how we'd get a bad value out of `getRemoteAddr`

dmnjohns: Oh, it's quite easy to get a bad value since the return value is just a `String`. I think my example below is the easiest to see; an improperly implemented request wrapper could, say, return `/` which would lead to this exception being thrown.

Of course, we _should not_ get a bad value from `getRemoteAddr`, but since Repose is extensible and operators can inject their own code we _could_ get a bad value from `getRemoteAddr`.

So it's a bit of defensive coding to account for that, and stating expected behavior if that were to happen.

adrianjgeorge: Yes, a bad wrapper will do it, but it will never be because of the user. And if we really want to cover that it's much more complicated. we have to do our own tests at different times so we know which one we were looking at and which one failed because because the behavior should be a 400 for the header check, but 500 for the remote address.

adrianjgeorge: To be clear, I felt the likelihood of the other case was not worth the added complexity.

dmnjohns: Fair enough. Something about conflating the sources of the IP address bothers me, but I see how accounting for those sources adds complexity for an unlikely case.

repose-aggregator/components/filters/ip-user-filter/src/test/scala/org/openrepose/filters/ipuser/IpUserFilterTest.scala
dmnjohns: Since tests are not exposed to the user in any way, I think a test for this is still valuable. It shows that we considered the case in question and this is the behavior that was decided on.

PR #2013, opened 2019-01-23
REP-7486 Docs update for removal of RMS

adrianjgeorge:


repose-aggregator/docs/src/asciibinder/welcome/release-notes.adoc
```diff
@@ -26,6 +26,7 @@
 * https://repose.atlassian.net/browse/REP-6663[REP-6663] - Added support for forwarding the response status line reason phrase from the origin service.
 * https://repose.atlassian.net/browse/REP-7428[REP-7428] - Removed support for URI-based tenant validation in the <<../filters/keystone-v2-authorization.adoc#, Keystone v2 Authorization Filter>> and <<../filters/keystone-v2.adoc#, Keystone v2 Filter>>.
 * https://repose.atlassian.net/browse/REP-7387[REP-7387] - The <<../filters/ip-user.adoc#, IP User>> filter now returns a `400` when the X-Forwarded-For header contains a bad value.
+* https://repose.atlassian.net/browse/REP-7486[REP-7486] - Documented removal and replacement solution for the Remote Messaging Service.
```
dmnjohns: `Remote` should be `Response`.

repose-aggregator/docs/src/asciibinder/welcome/ver-9-upgrade-notes.adoc
```diff
@@ -5,6 +5,39 @@
   If you were counting on this behavior for some reason see <<../recipes/header-splitting.adoc#, Pre 9.0 Header Splitting>>.
 . The response status line reason phrase from the origin service will now be forwarded from the origin service.
   If you were expecting *Repose* to normalize the reason phrase, see the <<../recipes/reason-phrase-normalization.adoc#, Reason Phrase Normalization recipe>>.
+. The Remote Messaging Service has been removed.
```
dmnjohns: `Remote` should be `Response`.

dmnjohns:
> Looking at it, it looks like it is going to be a five minute job to delete it

I mean... I'm not going to stop you...

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/configloadingandreloading/StartWithMissingConfigsTest.groovy
```diff
@@ -55,14 +55,14 @@ class StartWithMissingConfigsTest extends ReposeValveTest {
                 waitOnJmxAfterStarting: false)
         repose.waitForNon500FromUrl(reposeEndpoint)
 
-
+//todo: do something with this test?
```
dmnjohns: Do we want to do something with this test now? A quick refactor would allow us to drop the `@Unroll` and even the`@Category(Slow.class)` if we wanted to. However, this test only tells us the behavior of Repose when missing a metrics config, ignoring all of the other configs we rely on. So that's not terrific. And for that matter, all of these config tests only test a subset of all of our configs (perhaps just because we haven't kept them up-to-date.).

adrianjgeorge: I already updated it, i forgot to drop the todo. i swapped it to metrics so that it would continue working.

dmnjohns: Ah, I see. Fair enough.

The test could be cleaned up a bit, or expanded, but that it outside of the scope of this story.

PR #2014, opened 2019-01-23
REP-7519 Adding proxy support to the HTTP Client Service

dmnjohns:


repose-aggregator/components/services/http-client-service/http-client-service-api/src/main/resources/META-INF/schema/config/http-connection-pool.xsd
```diff
@@ -171,6 +171,30 @@
             </xs:annotation>
         </xs:attribute>
 
+        <xs:attribute name="http.route.default-proxy" type="xs:string" use="optional">
```
adrianjgeorge: I'm a little concerned about using an attribute here. If we ever need to support a proxy that requires authentication this becomes messy.

dmnjohns: That's a fair concern. Did you have a specific example of that messiness in mind?

I chose to use an attribute to remain consistent with other configuration of an HTTP client. Since named parameters were used prior to Apache HTTPClient 4.3, that's the pattern we have  followed. Any relationship between parameters would be a loose one, and the only indication of a relationship is through the parameter name.

That being said, I'm not opposed to making this an element.

adrianjgeorge: Is this the same name as what parameter use to be in the http client? If so, ship it.

dmnjohns: It is, yes.

PR #2015, opened 2019-01-25
REP-7490 Updated a number of log messages to reduce log noise

dmnjohns:


repose-aggregator/commons/commons-configuration/src/main/java/org/openrepose/commons/config/parser/jaxb/JaxbConfigurationParser.java
```diff
@@ -48,10 +48,6 @@ public JaxbConfigurationParser(Class<T> configurationClass, JAXBContext jaxbCont
         super(configurationClass);
         if (xsdStreamSource == null) {
             LOG.warn("Creating a JAXB Parser Pool without any schema to validate for {}", configurationClass);
-            if (LOG.isDebugEnabled()) {
-                Exception tracer = new Exception("Repose Devs might care about this trace");
-                LOG.debug("Logging the current stack to find where a parser pool is created without a validator", tracer);
-            }
```
dmnjohns: We really should not care about this. There are plenty of cases where we listen to a configuration without validating it; after all, a configuration really only needs to be validated once (as long as it is done in the right place). Additionally, not validating a configuration is an explicit choice by the developer. This log message seemed to be added to help track down cases where we may have wanted to validate the configuration but did not do so. Presumably those cases were addressed. In other words, this log message seemed to be a debugging tool for a developer, not the user. In the end, the user can't change anything, so why show them this message? As for the developer, well, we get the `warn` message from above which should be enough information to fix any issues.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/opentracing/httpclient/ReposeTracingRequestInterceptor.scala
```diff
@@ -81,8 +81,6 @@ class ReposeTracingRequestInterceptor(tracer: Tracer, reposeVersion: String, uri
 
       httpContext.setAttribute(OpenTracingSpan, clientSpan)
 
-      if (Option(tracer.scopeManager.active).isEmpty) logger.warn("Current scope is null; possibly failed to start client tracing span.")
```
dmnjohns: This log message would indicate that there was no active span when we create a new span for outbound requests.

Unfortunately, that is always the case when the OpenTracing Service is not configured for use.

Really, a failure to create/activate a span should be logged wherever that span is created/activated, not when we are making this new span. After all, what if deactivating the active span was intentional? Say the `ReposeFilter` started a new span for the request, but a filter in the filter chain intentionally deactivated that span. My point is that this method does not have enough data to determine whether or not there should be an active span. Plus, you could always check the OpenTracing collection service to figure out whether or not a span has a parent.

adrianjgeorge:
Fixes the conflicts, and I'll bring it in

PR #2018, opened 2019-01-29
Adding icons to the Release Notes.

wdschei:


dmnjohns:
Here is what the proposed changes would look like:
screen shot 2019-01-29 at 4 50 47 pm

dmnjohns:
After thinking about it some more, I think I prefer SVG images with alt text and titles over CSS styling. Specifically for the alt text. Well, and SVG gives you a lot of the same flexibility as CSS would. I just need to get used to working with SVG images.

PR #2019, opened 2019-01-29
Adding CSS styling for release note tags

dmnjohns:
I opted to continue using words over symbols here, but that would be an easy change. Colors would also be an easy change.

Ideally, the tags would be aligned, but that is a less easy change.

I like that CSS allows us to easily change the look and feel of these tags (e.g., size, color, symbols vs words). It's all just text. That also plays well with version control. Of course, so would SVG images, but not image formats that are stored as binary data.

Here is what the current changes would look like:
screen shot 2019-01-29 at 4 41 51 pm

### DIFF
--- a/repose-aggregator/docs/src/asciibinder/_templates/page.html.erb
+++ b/repose-aggregator/docs/src/asciibinder/_templates/page.html.erb
@@ -11,6 +11,7 @@
   <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.5/css/bootstrap-theme.min.css">
   <link rel="stylesheet" href="/_stylesheets/asciibinder.css" >
   <link rel="stylesheet" href="/_stylesheets/repose.css" >
+  <link rel="stylesheet" href="/_stylesheets/release-note-categories.css" >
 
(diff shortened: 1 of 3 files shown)


dmnjohns:
I am wary about relying on color to convey meaning (what if a user is colorblind?). I'm also not sure how well CSS interacts with screen readers. Both are concerns over accessibility.

Also, a legend would be necessary to describe what each color means.

repose-aggregator/docs/src/asciibinder/_templates/page.html.erb
```diff
@@ -11,6 +11,7 @@
   <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.5/css/bootstrap-theme.min.css">
   <link rel="stylesheet" href="/_stylesheets/asciibinder.css" >
   <link rel="stylesheet" href="/_stylesheets/repose.css" >
+  <link rel="stylesheet" href="/_stylesheets/release-note-categories.css" >
```
dmnjohns: Doing this here links this CSS to every generated page. Ideally, we would only link it to the release notes, but I could not find a way to get Asciidoctor to do that for me.

PR #2020, opened 2019-01-30
Test Re-categorization

adrianjgeorge:
I haven't changed the build yet, but it'll be inbound soon. Once the build is updated the jobs will need update/creation.


repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/services/datastore/DistDatastoreServiceBurstTest.groovy
```diff
@@ -22,11 +22,11 @@ package features.services.datastore
 import org.junit.experimental.categories.Category
 import org.openrepose.framework.test.PortFinder
 import org.openrepose.framework.test.ReposeValveTest
-import scaffold.category.Bug
 import org.rackspace.deproxy.Deproxy
+import scaffold.category.Bug
 
 @Category(Bug.class)
-//@Category(Slow.class)
+//@Category(Services)
```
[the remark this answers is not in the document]
adrianjgeorge: There's a possibility that if we assign it a second category besides bug that it'll run when it shouldn't.

adrianjgeorge:
So you are correct that exceptional status takes precedence over what is roughly package. We could apply multiples, but that actually makes it harder to achieve what we are attempting here, because then we get super complicated lists of includes and excludes. For the time being I'm taking the position that we give each test a singular category. This defaults to being along package lines, but isn't strictly necessary. Flag something as exceptional when needed, or even if you find a test that is more a recipe but don't want it in that package feel free to change it's category. 
I considered actually implementing with package filtering and leaving the current categories, but I didn't like the potential complication of having two different means of filtering going on in the build configuration. I thought it was cleaner to just map the one criterion over what was the second.

dmnjohns:
> So you are correct that exceptional status takes precedence over what is roughly package. We could apply multiples, but that actually makes it harder to achieve what we are attempting here, because then we get super complicated lists of includes and excludes. For the time being I'm taking the position that we give each test a singular category. This defaults to being along package lines, but isn't strictly necessary. Flag something as exceptional when needed, or even if you find a test that is more a recipe but don't want it in that package feel free to change it's category.
> I considered actually implementing with package filtering and leaving the current categories, but I didn't like the potential complication of having two different means of filtering going on in the build configuration. I thought it was cleaner to just map the one criterion over what was the second.

That seems reasonable to me. Can we document this approach in some way so that it is always understood and never lost? There is actually a section for test categories already written out in the `README.md` for the `functional-tests` project, but the file is a bit out-of-date. Hilariously, that file also goes into many of the same things we have just discussed here.

repose-aggregator/tests/functional-tests/build.gradle
```diff
@@ -108,16 +108,28 @@ processIntegrationTestResources {
...
-            case "all":
-                excludeCategories "scaffold.category.Bug"
+            case "recipe":
+                includeCategories "scaffold.category.Recipe"
+                break
+            case "services":
+                includeCategories "scaffold.category.Services"
+                break
+            case "xml":
+                includeCategories "scaffold.category.XmlParsing"
```
dmnjohns: No case for `bug`?  I get that tests in that category are expected to fail and so should not run in most cases, but not providing a case prevents those tests from ever being run by the Gradle build. It seems like we may as well enable those tests to be run even if just to be able to see what tests are in the `bug` category.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/proxy/PassThruTest.groovy
```diff
@@ -244,12 +238,11 @@ class PassThruTest extends ReposeValveTest {
         'Cache-Control'       | 'something/1.0; q=0.5, another/0.8 ;q=0.8=0.3 ; a=b=c ; abc'
         'Content-Encoding'    | 'something/1.0; q=0.5, another/0.8 ;q=0.8=0.3 ; a=b=c ; abc'
         'Content-Language'    | 'something/1.0; q=0.5, another/0.8 ;q=0.8=0.3 ; a=b=c ; abc'
-        'Content-Length'      | 'something/1.0; q=0.5, another/0.8 ;q=0.8=0.3 ; a=b=c ; abc'
         'Content-Location'    | 'something/1.0; q=0.5, another/0.8 ;q=0.8=0.3 ; a=b=c ; abc'
         'Content-MD5'         | 'something/1.0; q=0.5, another/0.8 ;q=0.8=0.3 ; a=b=c ; abc'
         'Content-Range'       | 'something/1.0; q=0.5, another/0.8 ;q=0.8=0.3 ; a=b=c ; abc'
         'Content-Type'        | 'something/1.0; q=0.5, another/0.8 ;q=0.8=0.3 ; a=b=c ; abc'
-        'Date'                | 'something/1.0; q=0.5, another/0.8 ;q=0.8=0.3 ; a=b=c ; abc'
+//        'Date'                | 'something/1.0; q=0.5, another/0.8 ;q=0.8=0.3 ; a=b=c ; abc' //This one does go through, but a second date is applied
```
dmnjohns: It looks like Jetty adds the second `Date` header. Jetty can be configured to add that header or not to add it, but I'm not sure if there is a way to tell Jetty to add the header only if it does not already exist.

PR #2023, opened 2019-02-04
REP-7582 Test Categorization

adrianjgeorge:


adrianjgeorge:
The one job is broken everywhere due to a bad plugin on jenkins, there's a separate pr to fix that. The other i believe is a flakey test.

dmnjohns:
Fair enough.

PR #2026, opened 2019-02-04
REP-7539 Build Docker Artifacts Using Local Rather Than Published Release Artifacts

dmnjohns:
With this change, Docker images may now be built from local packages in addition to remote packages.

Docker image releases will now be handled as part of the `release` task, gated on the same tests as other publication tasks. This means that Docker images will now be tied to the same Gradle run as other publishing tasks and will be available sooner in the release process.

Since it is no longer necessary (after this change is merged), I have disabled (with intent to delete after merge) the Docker image release job in Jenkins:
https://jenkins.openrepose.org/view/Repose/job/repose-gradle-release-docker-image/


repose-aggregator/artifacts/docker/build.gradle
```diff
@@ -46,28 +96,28 @@ docker {
...
     task("remove${baseImage}Image", type: DockerRemoveImage) {
         dependsOn "build${baseImage}Image"
-        mustRunAfter "tag${baseImage}ImageVersion", "tag${baseImage}ImageLatest", "push${baseImage}ImageVersion", "push${baseImage}ImageLatest"
+        mustRunAfter "tag${baseImage}ImageVersion", "tag${baseImage}ImageLatest", "push${baseImage}ImageVersion", "push${baseImage}ImageLatest", publish
         targetImageId { tasks["build${baseImage}ImageLocal"].getImageId() }
         force = true
     }
 
-    tasks["push${baseImage}ImageLatest"].onlyIf { isLatestDockerHubVersion(reposeVersion) }
+    publish.dependsOn "push${baseImage}Image"
```
adrianjgeorge: I'm not certain what this will do, this project generally doesn't interact with the publishing plugin (yes, i know technically it's defined).

dmnjohns: You're right, we already apply the publishing plugin to all sub-projects and wire every sub-project's `publish` task into the `release` task chain. Which means the `publish` task was already running on this sub-project as part of the `release` task chain.

I did question doing the wiring this way since we do not directly rely on the publishing plugin for this project, but since the task already runs and we have already wired the `publish` task correctly for all sub-projects, I figured this approach was more practical if not as relationally pure.

As an aside, part of the reason I ended up not using a source set to copy things was that I did not want to rely on the Java plugin which sets up the source set and source set directory I would have used.

dmnjohns: Thinking about it now, I do think I missed a step. I don't think I added any wiring to guarantee that the push image tasks run after `buildAll`. So while what I did added the task to the release task chain, it did not necessarily order it correctly.

In which case, I'll modify things a bit to not rely on the `publish` task and to instead insert the push image tasks between `buildAll` and `release`.

adrianjgeorge: `publish` itself depends on `buildAll`

adrianjgeorge:
I think the last commit might be unnecessary, because of the already existing dependencies between `release`, `publish`, and `buildAll`.

repose-aggregator/artifacts/docker/build.gradle
dmnjohns: Sure, but prior to the last commit, I just had `publish` depending on `push${baseImage}Image`. So `publish` would depend on both `buildAll` and `push${baseImage}Image`, but `push${baseImage}Image` would not depend on `buildAll`. Thus, `push${baseImage}Image` had the potential to run before `buildAll`.

dmnjohns:
> I think the last commit might be unnecessary, because of the already existing dependencies between `release`, `publish`, and `buildAll`.

I am aware of the existing dependencies between those tasks, however I believe the last commit serves two purposes:

1. Fixing the ordering between `buildAll` and `push${baseImage}Image` as described in a previous comment.
2. Removing the reliance on the `publish` task. Instead, tasks that we have explicitly defined as part of the release process are being used. This doesn't make much practical different, it just means that we have the ability to **not** apply the `maven-publish` plugin to the `docker` sub-project anymore (without breaking anything). In other words, I chose to rely on explicitly defined Repose release tasks rather than a task being defined by a plugin not being used in this project.

repose-aggregator/artifacts/docker/src/docker/resources/file/ubuntu/Dockerfile
```diff
@@ -0,0 +1,50 @@
+# Use Ubuntu as a base, since it works and is familiar.
+# The version should be the latest tested, LTS version.
+FROM ubuntu:16.04
```
wdschei: I wonder when it is time to move to 18.04?

repose-aggregator/docs/src/asciibinder/developer/docker.adoc
```diff
@@ -23,25 +23,50 @@ They can be passed in as build properties or the easier way is to add them to yo
...
 Then try to build a quick *Repose Docker* image from the root of the *Repose* project.
 
-[source,bash,subs="verbatim,attributes"]
+[source,bash]
 ----
 ./gradlew \
-    :repose-aggregator:artifacts:docker:buildUbuntuImageLocal \
-    -Prepose-version={project-version}
+    :repose-aggregator:artifacts:docker:tagUbuntuImageVersion \ # <1>
+    -x removeUbuntuImage # <2>
```
wdschei: We need to add a note in here that this works for Repose v9.0.0.0 and forward due to the package name changing from `repose-valve` to just `repose`.

repose-aggregator/artifacts/docker/src/docker/resources/file/ubuntu/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+# This environment variable is used to set command-line options.
+# The user can manually set these options using the "-e" flag with the Docker run command.
+ENV JAVA_OPTS=
+
+# Copy local Repose packages
+COPY repose_*.deb ${PKG_DIR}
+COPY repose-filter-bundle_*.deb ${PKG_DIR}
+COPY repose-extensions-filter-bundle_*.deb ${PKG_DIR}
+COPY repose-experimental-filter-bundle_*.deb ${PKG_DIR}
```
wdschei: These `COPY`'s should probably move below the `RUN apt-get...` so that one more possibly cached point can be created/used.
This will prevent having to download and install OpenJDK every time.

repose-aggregator/artifacts/docker/src/docker/resources/file/centos/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+# This environment variable is used to set command-line options.
+# The user can manually set these options using the "-e" flag with the Docker run command.
+ENV JAVA_OPTS=
+
+# Copy local Repose packages
+COPY repose-*.rpm ${PKG_DIR}
+COPY repose-filter-bundle-*.rpm ${PKG_DIR}
+COPY repose-extensions-filter-bundle-*.rpm ${PKG_DIR}
+COPY repose-experimental-filter-bundle-*.rpm ${PKG_DIR}
```
wdschei: Same here.

wdschei: And in the `repo` versions too.

wdschei: Never mind. That is way wrong. I was dealing with a locally modified version of the repository Docker files that is now similar to the local flavor.

repose-aggregator/artifacts/docker/src/docker/resources/file/ubuntu/Dockerfile
dmnjohns: I suppose it doesn't matter too much since Repose will run in this environment and Docker isolates the environment to some extent. Even so, I don't believe there is any reason to keep ourselves on 16.04, so we may as well just upgrade for whatever minimal improvements we get from doing so.

I'm tempted to just build the Docker images off of `ubuntu` and `centos` letting the `latest` tag be used. That introduces the  possibility that our build will break, but it's fairly unlikely considering the content of our Dockerfiles.

Well, I'll go ahead and bump to Ubuntu 18.04 for this PR.

repose-aggregator/docs/src/asciibinder/developer/docker.adoc
dmnjohns: Fair point.

dmnjohns: Considering that, I am much more tempted to drop support for building a Docker image running a specific version of Repose. I almost did so on my first pass, but ended up leaving it in for the sake of convenience. However, there is no guarantee that future `Dockerfiles` will support installing past versions (as was the case here). Your best bet would be to check out the tag for the version you want to build an image for. If that's the approach you take, though, then there is not much point in maintaining the version selection feature here.

Still, it offers a modicum of convenience for certain cases. Hm. I guess I will leave it for now unless anyone else has an opinion on it one way or the other.

repose-aggregator/artifacts/docker/src/docker/resources/file/centos/Dockerfile
dmnjohns: Nope, you are absolutely right. The `ARG` instruction would be given a different value which would force all subsequent instructions, including the installation of the JDK, to run. I'll go ahead and move that down as well, for the sake of efficiency.

PR #2027, opened 2019-02-06
REP-7570 Update build to publish to the new package repository

dmnjohns:
Based on REP-7539 to leverage some build changes made there. This PR should be rebased onto master after #2026 is merged.


repose-aggregator/artifacts/build.gradle
```diff
@@ -0,0 +1,75 @@
...
+import io.github.httpbuilderng.http.HttpTask
+
+import static groovyx.net.http.NativeHandlers.Encoders
+
+apply plugin: 'io.github.http-builder-ng.http-plugin'
+
+// Informs Gradle that child projects must be evaluated (i.e., configured) before this project.
+// Doing so is necessary in order to use the configurations from child projects to generate
+// the requests for publication (via configuration of the HttpTask).
+evaluationDependsOnChildren()
```
dmnjohns: I am not a fan of having to use this instruction. Unfortunately, since I opted to use the `http-plugin`, I need to know what artifacts will be pushed at configuration time which means that the child projects to this project must be evaluated before this project.

Of course, this couples this project to its child projects. Which is why I opted to create a new project to contain the package publication task.

A different approach would be to not use the `http-plugin` but to use `http-builder` directly. Then we could define the package publication task ourselves rather than configuring an `HttpTask`. As a result, the package publication task would not need to know which package(s) to publish until execution time. That is, the request generation would occur in the `doLast` closure. Building the requests themselves would even look largely the same, but with a small amount more boilerplate. And of course, switching to a different HTTP client would be more difficult than it is with the plugin (theoretically).

build.gradle
```diff
@@ -84,7 +77,56 @@ task('exportVersion', description: 'Write the current build version out to a pro
...
+
+        request.uri = 'https://nexus.openrepose.org'
+        request.auth.basic getProperty('nexus.openrepose.username') as String, getProperty('nexus.openrepose.password') as String
+        request.contentType = ContentTypes.BINARY[0]
+        request.encoder ContentTypes.BINARY[0], Encoders.&binary
+
+        // The requests are configured here (inside of the config closure) so that they are evaluated and
+        // configured during the execution build phase rather than the configuration build phase.
+        // This allows us to dynamically create requests to upload files that are not determined until after the
+        // configuration phase for certain projects without having to dictate evaluation order between the projects.
```
dmnjohns: This change solves the issues I mentioned in previous comments. It does kind of rely on the plugin to execute the config closure at an expected time, but... this seems better than the alternative.

build.gradle
```diff
@@ -84,7 +77,56 @@ task('exportVersion', description: 'Write the current build version out to a pro
     }
 }
 
-task('release', description: 'Release the project and publish to all repos.', group: 'release', dependsOn: [exportVersion, tagVersion/*, publishToPackageRepo*/])
+task('publishToPackageRepo', type: HttpTask, group: 'publishing') {
+    description 'Publishes the DEB and RPM packages to the package repository'
+    dependsOn configurations.debBundle, configurations.rpmBundle, buildAll
+    mustRunAfter tagVersion
```
adrianjgeorge: Why?

repose-aggregator/artifacts/cli-utils/build.gradle
```diff
@@ -3,6 +3,8 @@ apply plugin: 'nebula.ospackage'
 
 configurations {
     fatJar
+    deb
```
adrianjgeorge: Forgot to clean up with the others?

build.gradle
dmnjohns: The `mustRunAfter` instruction, you mean?

It isn't really necessary, but my thinking was that if the `tagVersion` fails it is easier to clean up that this task. So I would rather `tagVersion` run first. In general, I would prefer easy-to-cleanup publication tasks to run before other publication tasks.

Of course, this does add some minimal complication and restrains what Gradle can do task-ordering wise.

I'm happy to remove it.

repose-aggregator/artifacts/cli-utils/build.gradle
dmnjohns: Yeah. Well, forgot to add these configurations to this project when I did the others... so it's here now.

dmnjohns:
run perf test filters/scripting/javascript

build.gradle
adrianjgeorge: While I get what your saying, it makes the task graph a little less intuitive as well as making this particular publishing task different from all the others. I'd rather they all match.

adrianjgeorge: To be clear that doesn't mean make them all depend on `tagVersion` because as far as i'm concerned that's also a publishing task.

dmnjohns: Got it. I'll remove this line.

PR #2028, opened 2019-02-12
Gradle Build Release Property Fix

dmnjohns:
This fixes Gradle configuration issues that arise when running with the `release` property.

The change itself no longer assumes that projects have a single signature for publication, and instead will publish every signature available. This accounts for the possibility of `configurations` being added to projects and `artifacts` being associated with them. A choice was made here to allow all artifacts to be signed and to publish all signatures. Note, however, that this configuration is merely the default publishing configuration, and that some projects (e.g., `valve`) override it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -400,7 +400,9 @@ subprojects {
         publications {
             nebula(MavenPublication) {
                 if(project.hasProperty('release')) {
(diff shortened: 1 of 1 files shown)


dmnjohns:
run perf test filters/scripting/javascript

PR #2029, opened 2019-02-18
REP-7150 Perform Automated OWASP Security Checks

dmnjohns:
See #1849 for history including some helpful commands to analyze output.


build.gradle
```diff
@@ -384,6 +386,28 @@ subprojects {
...
+            'runtimeClasspath',
+            'default',
+        ]
+        analyzers {
+            assemblyEnabled = false
+        }
+        suppressionFile = file("${rootDir}/OWASP-Suppression.xml")
+
+        // Configures the dependencyCheck plugin to fail the build if any vulnerability with a CVSS score over the configured value is found
+        failBuildOnCVSS = 5
```
dmnjohns: I'm not sure that this is the appropriate level to fail on. The scale is 1-10, so if we put 1 here we will always fail.

build.gradle
```diff
@@ -384,6 +386,28 @@ subprojects {
...
+    }
+
+    check.dependsOn dependencyCheckAnalyze
```
dmnjohns: Wiring this task into the standard build adds a fair amount of time to the build. Of course, you can always exclude the task so it should be fine.

versions.properties
```diff
@@ -1,12 +1,12 @@
-jacksonVersion=2.8.9
+jacksonVersion=2.8.10
```
dmnjohns: This may be the best solution we have at the moment. `2.9.8` is the latest version and fixes a number of vulnerabilities, but `api-checker` prevents up from moving past `2.8.10` unless we upgrade Jackson in `api-checker`.

See:
https://github.com/rackerlabs/repose/pull/1849#discussion_r257789007

wdschei: I'm good with only bumping it to what we can for the current `api-checker` and then adding a story to bump the version used in the dependency.

The main goal of adding the OWASP checking was to make us and our customers aware of potential vulnerabilities so we could appropriately guard against them.

build.gradle
dmnjohns: Here are the timings on my machine with and without the `dependencyCheckAnalyze` task.

Without, the build took 5m 46s:
```
BUILD SUCCESSFUL in 5m 46s
985 actionable tasks: 841 executed, 144 up-to-date

~/dev/repose REP-7150* 6m 2s
❯ gradlew clean && gradlew build -x dependencyCheckAnalyze -x test -x linkchecker
```

With, the build took 16m 16s:
```
BUILD SUCCESSFUL in 16m 16s
1092 actionable tasks: 948 executed, 144 up-to-date

~/dev/repose REP-7150* 16m 26s
❯ gradlew clean && gradlew build -x test -x linkchecker
```

dmnjohns: With those numbers being what they are, I am going to remove the wiring of this check into the main build task chain. Instead, this task can be invoked explicitly.

build.gradle
dmnjohns: Turns out the number to always fail when a vulnerability found is actually 0. That being the case, I have set this to 0.

dmnjohns:
I need to update the nightly Jenkins job to run the new task after this is merged. That will be tracked elsewhere.

build.gradle
wdschei: That makes sense. We can easily add an additional check on the nightly that invokes just the `dependencyCheckAnalyze` task.

adrianjgeorge: Yeah, I'd say add it into the nightly as well as giving it it's pr job.

adrianjgeorge:
retest this please

adrianjgeorge:
The build that's running suggests that it's currently wired into the main build still?

dmnjohns:
> The build that's running suggests that it's currently wired into the main build still?

Whoops! I apparently did not push my changes from yesterday. Pushed just now.

adrianjgeorge:
retest this please

OWASP-Suppression.xml
```diff
@@ -0,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.2.xsd">
+    <suppress>
+        <notes><![CDATA[
+      file name: echo-filter.jar (project :repose-aggregator:components:filters:echo-filter)
```
adrianjgeorge: What?

build.gradle
```diff
@@ -384,6 +386,26 @@ subprojects {
...
+            'compileClasspath',
+            'runtime',
+            'runtimeOnly',
+            'runtimeClasspath',
+            'default',
+        ]
+        analyzers {
+            assemblyEnabled = false
+        }
+        suppressionFile = file("${rootDir}/OWASP-Suppression.xml")
```
adrianjgeorge: I feel like this should be somewhere other than the root. maybe in the `repose-aggregator\src\config` directory

repose-aggregator/components/services/atom-feed-service/atom-feed-service-api/build.gradle
```diff
@@ -10,3 +10,7 @@ jacocoTestCoverageVerification {
         }
     }
 }
+
+dependencyCheck {
+    skip = true
```
adrianjgeorge: Why?

OWASP-Suppression.xml
dmnjohns: Why is our echo filter being flagged, you mean? It looks like it's a false-positive. The plugin sees some properties of this dependency contain "echo" and thinks the dependency is something that it isn't. So here we tell the plugin that our echo filter is not the echo filter it thinks it is.

build.gradle
dmnjohns: Easy enough. I'll go ahead and move it there assuming @wdschei has no issue with it.

wdschei: I'm good with it.
That was just the easiest place to put it for the original prototype.

repose-aggregator/components/services/atom-feed-service/atom-feed-service-api/build.gradle
dmnjohns: I'm not actually sure on this one. @wdschei?

adrianjgeorge:
retest this please

repose-aggregator/components/filters/scripting-filter/src/main/resources/META-INF/schema/config/scripting.xsd
```diff
@@ -50,8 +50,6 @@
             <xs:enumeration value="luaj"/>
             <xs:enumeration value="python"/>
             <xs:enumeration value="jython"/>
-            <xs:enumeration value="ruby"/>
-            <xs:enumeration value="jruby"/>
```
wdschei: Yeah!!! It is gone.

repose-aggregator/artifacts/extensions-filter-bundle/src/config/docs/DEPENDENCIES.txt
```diff
@@ -17,17 +17,17 @@ outlined in CONTRIBUTORS.txt.
...
 com.rackspace.identity.components.attribute-mapper  mapper-core               2.2.1            Apache License Version 2.0  https://github.com/rackerlabs/attributeMapping
-com.rackspace.papi.components.api-checker           checker-core              2.6.1            Apache License Version 2.0  https://github.com/rackerlabs/api-checker/blob/api-checker-2.5.1/LICENSE.txt
-com.rackspace.papi.components.api-checker           checker-util              2.6.1            Apache License Version 2.0  https://github.com/rackerlabs/api-checker/blob/api-checker-2.5.1/LICENSE.txt
-com.typesafe.akka                                   akka-actor_2.11           2.3.12           Apache License Version 2.0  https://github.com/akka/akka
+com.rackspace.papi.components.api-checker           checker-core              2.7.0            Apache License Version 2.0  https://github.com/rackerlabs/api-checker/blob/api-checker-2.5.1/LICENSE.txt
+com.rackspace.papi.components.api-checker           checker-util              2.7.0            Apache License Version 2.0  https://github.com/rackerlabs/api-checker/blob/api-checker-2.5.1/LICENSE.txt
```
wdschei: The license link should probably be bumped to the same version we reference.

repose-aggregator/artifacts/filter-bundle/src/config/docs/DEPENDENCIES.txt
```diff
@@ -52,24 +52,23 @@ org.apache.logging.log4j                            log4j-flume-ng            2.
...
 org.luaj                                            luaj-jse                  3.0.1            MIT License                 http://luaj.sourceforge.net/
 org.openrepose                                      http-delegation           4.0.0            Apache License Version 2.0  https://github.com/rackerlabs/repose
-org.python                                          jython-standalone         2.7.0            Jython Specific License     http://www.jython.org/license.html
+org.python                                          jython-standalone         2.7.1            Jython Specific License     http://www.jython.org/license.html
 org.scala-lang.modules                              scala-xml_2.11            1.0.5            Scala Specific License      http://www.scala-lang.org/license.html
-org.scala-lang                                      scala-library             2.11.8           Scala Specific License      http://www.scala-lang.org/license.html
+org.scala-lang                                      scala-library             2.11.12          Scala Specific License      http://www.scala-lang.org/license.html
```
wdschei: Since we have added the hard dependency on `scala-reflect` to force a version, do we need to add it here too?

repose-aggregator/docs/src/asciibinder/welcome/release-notes.adoc
```diff
@@ -3,6 +3,19 @@
 == In Progress
 * https://repose.atlassian.net/browse/REP-7391[REP-7391] - Updated dependencies:
 ** Jetty: 9.4.8.v20171121 → 9.4.14.v20181114
+* https://repose.atlassian.net/browse/REP-7150[REP-7150] - Added OWASP security dependency checking to the build.
+** Updated dependencies:
+*** Akka: 2.3.12 → 2.4.20
+*** Apache Commons: 3.3.2 → 3.8.1
+*** Apache Thrift: 0.9.2 → 0.10.0
+*** API Checker: 2.6.1 → 2.7.0
+*** Groovy: 2.4.3 → 2.41.5
```
wdschei: s/2.41.5/2.4.15/

dmnjohns: Done.

repose-aggregator/artifacts/filter-bundle/src/config/docs/DEPENDENCIES.txt
dmnjohns: Yes, and done.

repose-aggregator/artifacts/extensions-filter-bundle/src/config/docs/DEPENDENCIES.txt
dmnjohns: Good catch. Done.

PR #2032, opened 2019-02-21
REP-7630 Decom Sonar

adrianjgeorge:


dmnjohns:
We can probably also delete this old job that I suspect handled Sonar submissions back when we used Maven:
https://jenkins.openrepose.org/job/sonar/

PR #2033, opened 2019-03-01
REP-7663 Fixing EAR unpacking

dmnjohns:
This change sets our `FileOutputStream` to appending mode. On instantiation, a `FileOutputStream` will truncate its underlying file unless it is in append mode. Since it was not in append mode, there may have been times where the file on disk was assumed to be valid when actually it was in a partially written state. I expect this is the change that will solve the issue with the `RemoteDatastoreServiceTest`.

See also:
#2041
#2042

### DIFF
--- a/repose-aggregator/core/repose-core/src/main/scala/org/openrepose/commons/utils/classloader/EarClassProvider.scala
+++ b/repose-aggregator/core/repose-core/src/main/scala/org/openrepose/commons/utils/classloader/EarClassProvider.scala
@@ -19,7 +19,7 @@
  */
 package org.openrepose.commons.utils.classloader
 
-import java.io.{File, FileInputStream, FileOutputStream, IOException}
+import java.io._
 import java.net.{URL, URLClassLoader}
 import java.nio.file.attribute.BasicFileAttributes
 import java.nio.file.{FileVisitResult, Files, Path, SimpleFileVisitor}
@@ -27,7 +27,6 @@ import java.util.zip.{ZipFile, ZipInputStream}
 
 import com.oracle.javaee6.{ApplicationType, FilterType, ObjectFactory, WebFragmentType}
 import javax.xml.bind.JAXBContext
-import org.apache.commons.io.FileUtils.checksumCRC32
 import org.openrepose.commons.config.parser.jaxb.JaxbConfigurationParser
 import org.openrepose.commons.config.resource.impl.BufferedURLConfigurationResource
 import org.slf4j.LoggerFactory
@@ -133,7 +132,7 @@ class EarClassProvider(earFile: File, val outputDir: File) {
       }
 
       //Make sure it's actually a zip file, so we can fail with some kind of exception
-      new ZipFile(earFile)
+      new ZipFile(earFile).close()
 
       val zis = new ZipInputStream(new FileInputStream(earFile))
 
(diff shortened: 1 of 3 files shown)


dmnjohns:
To further test this fix, I ran a test job which was shown to fail against this branch for a while. The results can be viewed at:
https://jenkins.openrepose.org/job/TEST-remote-datastore-failure/

Builds 17 and 29 on the job had previously failed. We seemed to get a failure every 10-15 builds. As you can see from all of the runs on March 1st, 16 builds in a row passed which makes me fairly confident that this change does, in fact, fix our issue.

dmnjohns:
Also, considering this change, I need to add do some documentation and release notes work. I'll add a task for that.

dmnjohns:
> File size is an extremely poor check that can give a false positive way more often then a checksum.

True, a checksum is a much stronger integrity check than file size. However, it also requires more computation. So in theory, there is a trade-off between speed and certainty here.

My thinking was that a byte check would be sufficient to identify invalid files in most cases. The reason being that if the source changed (i.e., the EAR file), the deployment directory would also change, so there is no real potential for invalid files in that case. If the file was written to disk incorrectly (say the process was terminating before fully writing the file to disk), the byte check would catch it. The byte check would not, however, catch certain kinds of data corruption (like those that may occur in storage), but most of these are handled by modern file systems (not that we can assume that any particular file system is in use). The byte check would also not catch intentionally malicious modification to artifacts, though if an attacker is able to modify unpacked artifacts on disk, that may be a moot point..

The reason I made this change was because I was curious what the impact of switching to a byte check would be on startup times and our functional test runs. Looking at the PR jobs for this PR, it does not appear that there was any significant gain. As such, I will switch back to performing a CRC.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/commons/utils/classloader/EarClassProvider.scala
```diff
@@ -143,23 +142,26 @@ class EarClassProvider(earFile: File, val outputDir: File) {
...
-              log.trace("Releasing file lock on: {}", entryFile)
-              lock.release()
+            entryFileChannel.lock()
+
+            if (entry.getSize != entryFile.length) {
+              log.trace("Unpacking: {}", entryFile)
+              // Clear the current contents of the file
+              entryFileChannel.truncate(0)
+              // Write the zip entry contents to the file
+              Stream.continually(zis.read(buffer)).takeWhile(_ != -1).foreach(count => entryFileOut.write(buffer, 0, count))
```
dmnjohns: At one point I actually replaced the `truncate` and `write` lines with simply:
```
Files.copy(earInputStream, entryFile.toPath, REPLACE_EXISTING)
```

While that does clean things up a bit, it is not exactly safe or clear to use. `Files.copy` deletes the file on disk before writing a new file. While this does not break locking on my machine, it may on others. Furthermore, writing to the `FileOutputStream` or its `Channel` after calling `Files.copy` will not have any effect on the file that was written by `Files.copy`.

repose-aggregator/core/repose-core/src/main/scala/org/openrepose/commons/utils/classloader/EarClassProvider.scala
```diff
@@ -143,23 +142,26 @@ class EarClassProvider(earFile: File, val outputDir: File) {
         if (entry.isDirectory) {
           entryFile.mkdir()
         } else {
-          val ofs = new FileOutputStream(entryFile)
+          val entryFileOut = new FileOutputStream(entryFile, true)
```
dmnjohns: From https://docs.oracle.com/javase/8/docs/api/java/io/FileOutputStream.html  :

> Some platforms, in particular, allow a file to be opened for writing by only one FileOutputStream (or other file-writing object) at a time. In such situations the constructors in this class will fail if the file involved is already open.

So that's an issue...

PR #2038, opened 2019-03-06
REP-7608 Fixed the bug with empty bodies

adrianjgeorge:


repose-aggregator/core/repose-core/src/main/java/org/openrepose/nodeservice/httpcomponent/HttpComponentRequestProcessor.java
```diff
@@ -20,6 +20,7 @@
 package org.openrepose.nodeservice.httpcomponent;
 
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.ArrayUtils;
```
dmnjohns: You probably want to import from the `lang3` package. This imports a previous version which is not controlled by our build. I suppose it must be transitive, but since the artifact ID changed to `commons-lang3`, we end up with multiple versions of this library. Still the classes themselves have different packages, so we can specify that we want the newer one.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/nodeservice/httpcomponent/HttpComponentRequestProcessor.java
```diff
@@ -85,7 +88,13 @@ public static HttpUriRequest process(HttpServletRequest servletRequest,
         final RequestBuilder requestBuilder = RequestBuilder.create(servletRequest.getMethod())
             .setUri(getUri(servletRequest, target));
         if (!NO_ENTITY_METHODS.contains(servletRequest.getMethod())) {
-            requestBuilder.setEntity(getEntity(servletRequest, chunkedEncoding));
+            byte[] bytes = new byte[8];
+            PushbackInputStream pushbackInputStream = new PushbackInputStream(servletRequest.getInputStream(), 8);
+            int bytesRead = pushbackInputStream.read(bytes, 0, 8);
```
dmnjohns: Perhaps replace the `8` with `bytes.length`? One less thing to keep in sync.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/nodeservice/httpcomponent/HttpComponentRequestProcessor.java
adrianjgeorge: Will do.

repose-aggregator/core/repose-core/src/main/java/org/openrepose/nodeservice/httpcomponent/HttpComponentRequestProcessor.java
adrianjgeorge: Legit

PR #2040, opened 2019-03-06
REP-7694 Limiting the size of the Rackspace logo

dmnjohns:


_includes/foot.html
```diff
@@ -3,7 +3,7 @@
         <div class="medium-12 columns text-center">
             <div class="row">
                 <div class="medium-12 columns small-centered">
-                    <p><a href="https://rackspace.com"><img src='/img/Rackspace_Wordmark_Red.png'></a></p>
+                    <p><a href="https://rackspace.com"><img src='/img/Rackspace_Wordmark_Red.png' width="225"></a></p>
```
dmnjohns: Opted to use the `width` attribute in the `img` tag since we know the size of the logo ahead of time. It's also closer to what was being done before, which was just relying on the raw image size.

This could have been accomplished by modifying the CSS, and in fact still could to account for mobile users.

wdschei: Or use a percentage instead of a had value.

wdschei: But those are tweaks for another day.

dmnjohns: Sure, though in this case I think I prefer using the hard value, again since we know the actual image size to begin with.

But this looks good enough on my laptop and phone, so moving on.

PR #2045, opened 2019-03-13
Random Cleanup of some root files

adrianjgeorge:


README.adoc
```diff
@@ -0,0 +1,103 @@
...
+*Repose* also makes it easy to create your own custom components.  Check out
+our https://github.com/rackerlabs/repose-hello-world[example custom filter] for more details.
+
+
+== Installation
+You can install *Repose* using the following methods:
+
+* http://www.openrepose.org/versions/9.0.0.0/recipes/valve-installation.html#yum-based-linux-distributions[yum]
+* http://www.openrepose.org/versions/9.0.0.0/recipes/valve-installation.html#apt-based-linux-distributions[apt]
+* http://www.openrepose.org/versions/9.0.0.0/recipes/valve-installation.html#yum-based-linux-distributions[Docker]
```
dmnjohns: This link should probably be to the Docker page?
http://www.openrepose.org/versions/9.0.0.0/recipes/docker.html

dmnjohns: Also, for these links and the ones above, would we want to use the `latest` version rather than `9.0.0.0`? So that we don't have to update this README every release.

adrianjgeorge: yeah, some how my copy didn't take on that one. and i meant to do a find replace when i was done

adrianjgeorge: i fix

PR #2049, opened 2019-03-21
REP-7721: Cleaning up sleeps in functional tests

pglass:
This cleans up nearly all of the unnecessary sleeps in the functional tests.

- Sleeps that waited for repose to start or reload a config, or just to wait for some thing to happen, are replaced with log searches or other wait logic where possible
- A bunch of sleeps of 500ms were waiting for the repose client cache to clear. Since these are short, I kept the sleeps but moved them into a `waitForHttpClientRequestCacheToClear` that is more descriptive
- Various tests need to sleep and I left these alone, like the rate limit tests and various tests of socket timeouts and cache expiration

### DIFF
--- a/repose-aggregator/tests/functional-test-framework/src/main/groovy/org/openrepose/framework/test/ReposeValveTest.groovy
+++ b/repose-aggregator/tests/functional-test-framework/src/main/groovy/org/openrepose/framework/test/ReposeValveTest.groovy
@@ -127,6 +127,16 @@ abstract class ReposeValveTest extends Specification {
 
     }
 
+    def waitForHttpClientRequestCacheToClear(long timeoutMilliseconds = 500) {
+        // Some tests reuse a single repose process across multiple test methods.
+        // In these tests, the repose http client will cache requests from one test
+        // method into the next test method.
+        //
+        // Each test _may_ have a different setting for the Repose client cache,
+        // but this is typically only 500ms.
+        sleep timeoutMilliseconds
+    }
+
     // Helper methods to minimize refactoring in all test classes
     def getReposeEndpoint() {
         return properties.getReposeEndpoint()
--- a/repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/powerfilter/GraphiteTest.groovy
+++ b/repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/powerfilter/GraphiteTest.groovy
@@ -20,11 +20,14 @@
 package features.core.powerfilter
 
 import org.junit.experimental.categories.Category
+import org.linkedin.util.clock.SystemClock
 import org.openrepose.framework.test.PortFinder
 import org.openrepose.framework.test.ReposeValveTest
 import org.openrepose.framework.test.mocks.MockGraphite
 import org.rackspace.deproxy.Deproxy
 import scaffold.category.Core
+
+import static org.linkedin.groovy.util.concurrent.GroovyConcurrentUtils.waitForCondition
 
 @Category(Core)
 class GraphiteTest extends ReposeValveTest {
@@ -67,12 +70,13 @@ class GraphiteTest extends ReposeValveTest {
         def mc1 = deproxy.makeRequest(url: reposeEndpoint + "/endpoint")
         def mc2 = deproxy.makeRequest(url: reposeEndpoint + "/endpoint")
         def mc3 = deproxy.makeRequest(url: reposeEndpoint + "/cluster")
-        sleep(2000)
 
         then:
         mc1.receivedResponse.code == "200"
         mc2.receivedResponse.code == "200"
         mc3.receivedResponse.code == "200"
-        lastCount == 3
+        waitForCondition(new SystemClock(), '2s', '250', {
+            lastCount == 3
+        })
     }
 }
--- a/repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/powerfilter/GraphiteTest.groovy
+++ b/repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/powerfilter/GraphiteTest.groovy
@@ -75,7 +75,7 @@ class GraphiteTest extends ReposeValveTest {
         mc1.receivedResponse.code == "200"
         mc2.receivedResponse.code == "200"
         mc3.receivedResponse.code == "200"
-        waitForCondition(new SystemClock(), '2s', '250', {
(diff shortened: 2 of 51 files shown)


pglass:
These tests were failing in Jenkins after replacing sleeps, so I reverted those changes.

- repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/identityv3/cache/InvalidateV3CacheUsingAtomFeedTest.groovy
- repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/identityv3/cache/InvalidateV3CacheUsingAuthenticatedFeedTest.groovy

pglass:
Oh, actually the keystone v2 versions of those tests were failing as well. I'll wait for Jenkins to finish and revert those as well (curious if they still fail on a retest).

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/core/powerfilter/GraphiteTest.groovy
```diff
@@ -67,12 +70,13 @@ class GraphiteTest extends ReposeValveTest {
...
         def mc2 = deproxy.makeRequest(url: reposeEndpoint + "/endpoint")
         def mc3 = deproxy.makeRequest(url: reposeEndpoint + "/cluster")
-        sleep(2000)
 
         then:
         mc1.receivedResponse.code == "200"
         mc2.receivedResponse.code == "200"
         mc3.receivedResponse.code == "200"
-        lastCount == 3
+        waitForCondition(new SystemClock(), '2s', '250', {
```
dmnjohns: Being that this is a `ReposeValveTest`, you have access to a `SystemClock` instance via the `repose.clock` (i.e., from the `ReposeValveLauncher` instance). It doesn't make much difference, but it may be preferable to use that clock instance rather than creating a new one.

In fact, we should probably update `ReposeValveLauncher` to use the static `SystemClock.instance()` or `SystemClock.INSTANCE` rather than creating a new instance ourselves.

repose-aggregator/tests/functional-test-framework/src/main/groovy/org/openrepose/framework/test/ReposeValveTest.groovy
```diff
@@ -127,6 +127,16 @@ abstract class ReposeValveTest extends Specification {
 
     }
 
+    def waitForHttpClientRequestCacheToClear(long timeoutMilliseconds = 500) {
```
dmnjohns: Love the method to document what this sleep is for.

dmnjohns: I will note that as of Repose 9, caching of shared requests can be disabled. To do so, set the `cache.ttl` attribute value to `0` in the `http-connection-pool.cfg.xml` file used for the test.

With that in mind, I would probably say that we should change the default `http-connection-pool.cfg.xml` file used in tests to disable shared request caching. Then for any tests that need that caching, we could set the `cache.ttl` attribute to a reasonable value. This is probably easiest to do by templating that value and giving it a default value of `0`. However, it would require identifying tests that need shared request caching. So that is probably outside the scope of this PR.

pglass: Another thought I had was to have this method check the repose config file for the `cache.ttl` value and sleep only if the cache is enabled (e.g. cache.ttl > 0). But that would only save 500ms per call, and there can be multiple pools to slightly complicate things, so I didn't bother with that idea for now.

repose-aggregator/tests/functional-tests/src/integrationTest/groovy/features/filters/urinormalization/UriNormalizationFilterTest.groovy
```diff
@@ -296,4 +332,20 @@ class UriNormalizationFilterTest extends ReposeValveTest {
...
     }
 
+    // This is a hack to avoid applying configs more than once, which saves on waiting time.
+    //
+    // Some test methods in this class with multiple iterations test against the same repose config file.
+    // On the first iteration of a test method, the config file is reloaded and tests wait for repose to
+    // reload. Since the config file is static, there's no point reapplying and waiting again. So we've
+    // got this helper to only apply configs and wait on the first iteration.
+    static def appliedConfigs = []
+    def applyConfigsOnlyFirstTime(String path, Map params, Closure closure) {
```
dmnjohns: This method makes me nervous. It relies on test execution order. If, for example, the first test applies a set of configs, the second test applies a different set of configs, and then a third test requires the set of configs applied by the first test, the configs are not reapplied and we have an issue. So my questions are:

- Does Spock ensure that test methods are run in any particular order? What if the `@Unroll` annotation is used?
- Does Spock support parallel test execution within the same test class (or, really, within a `Specification`)?

We probably don't care too much about parallel test execution within a `Specification`. We already have reasons prevent us from doing that, and I question how valuable it would be anyway. Run order is a concern, though. This `Specification` is running fine now, but I worry that upgrading Spock could change that.

You could check the contents of the configuration files rather than their past application using a checksum function, then conditionally apply new configs if they don't match the currently applied configs.

Or to reduce complexity, you could just separate this class into multiple `Specification`s (i.e., `Class`es). You could even keep all of these tests in the same test file if you wanted, though you don't have to. Each `Specification` would apply the configs it needs for the tests it runs. This is basically the approach we use for all of our tests, so it's a natural choice, the only new information I am providing is that I believe it is possible to keep multiple `Specification`s in the same file.

pglass: Yeah, agreed. 

There's only one method in this class that actually uses multiple iterations (i.e. multiple entries in the `where` table), so I think I'll move that into another class, and leave the others as-is (after removing the `applyConfigsOnlyFirstTime`).

pglass:
There's also a `features.core.valveselfconfigure.RuntimeSysmodChangesTest` that seems flaky with my new changes. It's failed twice in the past 20 runs or so, and both failures were on this PR: https://jenkins.openrepose.org/job/repose-gradle-pull_request-core/86/testReport/junit/features.core.valveselfconfigure/RuntimeSysmodChangesTest/history/

I'll revert changes for that test for now, and I'll make a note on the JIRA about the tests where my log searches didn't quite work out.

pglass:
retest this please

PR #2050, opened 2019-03-26
REP-7720: Specify repose heap size for tests

InblEric:
This PR adds the ability to specify the repose heap size for individual tests.

- Decreased the default values in ReposeValveLauncher from 1536M and 1024M to 1024M and 512M
- To specify the heap size, replace 
`repose.start()` with 
`repose.start([reposeXmx:"<xmx>",reposeXms:"<xms>"])` 
where xmx is the max heap size, and xms is the min heap size. 
Ex: `repose.start([reposeXmx:"1536M",reposeXms:"1024M"])`
- Pull the values from the params to repose start if they are there, and insert them into the command to start repose.

Right now all the tests still use the default values.

### DIFF
--- a/repose-aggregator/tests/functional-test-framework/src/main/groovy/org/openrepose/framework/test/ReposeValveLauncher.groovy
+++ b/repose-aggregator/tests/functional-test-framework/src/main/groovy/org/openrepose/framework/test/ReposeValveLauncher.groovy
@@ -47,6 +47,8 @@ class ReposeValveLauncher extends ReposeLauncher {
     def debugPort = null
     def classPaths = []
     def additionalEnvironment = [:]
+    def xmx = "1536M"
+    def xms = "1024M"
 
     ReposeConfigurationProvider configurationProvider
 
@@ -86,6 +88,12 @@ class ReposeValveLauncher extends ReposeLauncher {
         boolean waitOnJmxAfterStarting = true
         if (params.containsKey("waitOnJmxAfterStarting")) {
             waitOnJmxAfterStarting = params.waitOnJmxAfterStarting
+        }
+        if (params.containsKey("reposeXmx")) {
+            xmx = params.reposeXmx
+        }
+        if (params.containsKey("reposeXms")) {
+            xms = params.reposeXms
         }
 
         String nodeId = params.get('nodeId', "")
@@ -156,7 +164,7 @@ class ReposeValveLauncher extends ReposeLauncher {
         //TODO: possibly add a -Dlog4j.configurationFile to the guy so that we can load a different log4j config for early logging
 
         //Prepended the JUL logging manager from log4j2 so I can capture JUL logs, which are things in the JVM (like JMX)
-        def cmd = "java -Djava.util.logging.manager=org.apache.logging.log4j.jul.LogManager -Xmx1536M -Xms1024M -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/tmp/dump-${debugPort}.hprof $classPath $debugProps $jmxprops $jacocoProps $javaArgs -jar $reposeJar -c $configDir"
+        def cmd = "java -Djava.util.logging.manager=org.apache.logging.log4j.jul.LogManager -Xmx${xmx} -Xms${xms} -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/tmp/dump-${debugPort}.hprof $classPath $debugProps $jmxprops $jacocoProps $javaArgs -jar $reposeJar -c $configDir"
         println("Starting repose: $cmd")
 
         def th = new Thread({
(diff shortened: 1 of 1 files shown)


repose-aggregator/tests/functional-test-framework/src/main/groovy/org/openrepose/framework/test/ReposeValveLauncher.groovy
```diff
@@ -47,6 +47,8 @@ class ReposeValveLauncher extends ReposeLauncher {
     def debugPort = null
     def classPaths = []
     def additionalEnvironment = [:]
+    def xmx = "1024M"
+    def xms = "512M"
```
pglass: Hmm, I think we should move these to be default values on the `start` method? Something like,

```groovy
start(boolean killOthersBeforeStarting, boolean waitOnJmxAfterStarting,
      String nodeId, String xmx = "1024M", String xms = "512M") {
```

repose-aggregator/tests/functional-test-framework/src/main/groovy/org/openrepose/framework/test/ReposeValveLauncher.groovy
```diff
@@ -87,6 +89,12 @@ class ReposeValveLauncher extends ReposeLauncher {
         if (params.containsKey("waitOnJmxAfterStarting")) {
             waitOnJmxAfterStarting = params.waitOnJmxAfterStarting
         }
+        if (params.containsKey("reposeXmx")) {
+            xmx = params.reposeXmx
```
pglass: Following the other comment, then convert these to a local variable and pass along to the start method, similar to `killOthersBeforeStarting` and `waitOnJmxAfterStarting`, so we're not changing those instance variables each call.

InblEric: If I'm always passing those variables to start though, the default value wouldn't ever be used right?

InblEric: I guess I could just check if those values are in "params" and if so do this 

`start(killOthersBeforeStarting, waitOnJmxAfterStarting, nodeId, xmx, xms)`

else do `start(killOthersBeforeStarting, waitOnJmxAfterStarting, nodeId)`

pglass: Well, the defaults on the start method would get used any time someone bypasses this method and calls `start(false, true, "some-node")` directly (which I'm pretty sure I've seen in some places?)

InblEric: A lot of the tests right now just call `repose.start()` which call the other start method `void start(Map params)`

This one calls `start(killOthersBeforeStarting, waitOnJmxAfterStarting, nodeId)`, but if a test calls it like this: `repose.start([reposeXmx:"1536M",reposeXms:"1024M"])` then we'd need to pass those values to the other start:

`start(killOthersBeforeStarting, waitOnJmxAfterStarting, nodeId, xmx, xms)` which overrides the default values. But we'd still want `repose.start()` to not override those values, so we'd need a conditional so that when the values aren't passed, it just calls `start(killOthersBeforeStarting, waitOnJmxAfterStarting, nodeId)`. 

Or I could move the default values to the `void start(Map params)` method instead of having them come in through params.

`void start(Map params, xmx="1024M", xms="512M")`

I think I might like that more actually.

pglass: Oh, I see. I'm fine leaving it the way it is (just nitpicking).

I think I'd do this.

```groovy
// Put these on the class
static def DEFAULT_REPOSE_XMX = "1024M"
static def DEFAULT_REPOSE_XMS = "512M"

// Update this with "get or default" logic
def start(Map params) {
    ...
    def reposeXmx = params.get('reposeXmx', DEFAULT_REPOSE_XMX)
    def reposeXms = params.get('reposeXms', DEFAULT_REPOSE_XMS)
    start(killOthersBeforeStarting, waitOnJmxAfterStarting, nodeId, reposeXmx, reposeXms)
}

// Update this to have the defaults
start(boolean killOthersBeforeStarting, boolean waitOnJmxAfterStarting, String nodeId, 
      String xmx = DEFAULT_REPOSE_XMX, String xms = DEFAULT_REPOSE_XMS)
```

The `start([:])` method is still overriding the default values in the other start method, but they at least reference all the same constants so it will be the right values.

PR #2055, opened 2019-04-30
REP-7786 HTTP Logging Service

dmnjohns:
https://repose.atlassian.net/browse/REP-7786


dmnjohns:
Configuration is ready for review.

repose-aggregator/docs/src/asciibinder/services/http-logging.adoc
```diff
@@ -0,0 +1,310 @@
...
+</http-logging>
+----
+<1> Declares a message to be logged for every HTTP interaction.
+    The value of this element serves as the content of the message.
+    The value will be processed as a template.
+    See <<Templating>> for more information.
+<2> Specifies the name of the logger to which messages should be logged.
+    See <<../architecture/logging.adoc#, Logging>> for more information.
+<3> Specifies the format of the content.
+    The format determines if and how to validate the content and how to escape values populated by the templating engine.
```
wdschei: I was looking for the `Default` and `Options` for the `format` attribute.
I realize they are in the schema, but this is something that has been being put here as well since converting to the new versioned docs.
The main concern here is that it is another change and we haven't finished fully converting all of the old docs yet.
Maybe something to talk about.

wdschei: Something like these:
- Callout **4** [here](http://www.openrepose.org/versions/9.0.0.0/services/atom-feed-consumption.html#full-configuration)
- Callout **14** & **15**  [here](http://www.openrepose.org/versions/9.0.0.0/filters/api-validator.html#all-the-features-of-a-validator-element)

dmnjohns: Quoted most of what I wrote on Slack:

> On my first pass I included the default and options, then started adding multiplicity for the `message` element. As I was doing so, I started thinking that listing all of that in the example configuration docs was quite verbose and technically redundant with the schema. So I ended up dropping all of that information with the intent of making examples which would serve to show what can be done with the configuration, and if a user needs full details for some reason, the schema is linked on the page anyway. 

There is also the issue of having multiple sources of truth: one in the schema, and one in the documentation. 

However, I agree that this is inconsistent with existing docs, which does not make for a good user experience. Additionally, if/when we move away from XML for configuration we may lose schemas anyway, in which case we still need a way to give users a complete view of acceptable configuration.

So point taken. I will add the default and valid values where applicable to remain consistent with our current approach, but I am glad to have tried this out and to have discussed  it.

repose-aggregator/components/services/http-logging-service/http-logging-service-impl/src/main/scala/org/openrepose/core/services/httplogging/HttpLoggingConfigListener.scala
```diff
@@ -0,0 +1,111 @@
...
+      .toList
+      .flatMap { message =>
+        Option(HttpLoggingEnvironmentConfiguration(message.getFormat))
+          .filter(
+            validateFormat(message, _)
+              .recover { case e: JsonParseException =>
+                logger.warn("Unable to validate JSON for {}", message, e)
+              }
+              .isSuccess)
+          .map(JtwigTemplate.inlineTemplate(message.getValue, _))
```
wdschei: We should probably make a note that leading and trailing whitespace is preserved from the config so extra lines will be in the log.
Or we could do a .trim here.

dmnjohns: I wrote up my thoughts on this in a less-public comment on the story:
https://repose.atlassian.net/browse/REP-7786?focusedCommentId=38519&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-38519

To be clear, leading and trailing whitespace is not preserved in all cases. In general, that is determined by the XML processor, the `xml:space` attribute, and the use of `<![CDATA[...]]>`. For our usage, we theoretically know what XML process is being used and how it is configured, so we should only have to worry about `xml:space` and `<![CDATA[...]]>`. Correct me if I'm wrong, but I believe it is that case that if `CDATA` is not used, leading and trailing whitespace will be removed.

Of course, there are many mechanisms already in place for dealing with whitespace. There are those described in the previous paragraph, but there is also [JTwig whitespace control](http://jtwig.org/documentation/reference/syntax/code-islands#whitespace-control) and the [JTwig trim function](http://jtwig.org/documentation/reference/functions#trim). And of course, the operator can always write message content using exactly the whitespace that they desire.

With all of those tools existing to handle whitespace and with outstanding questions about how to handle whitespace compression or elimination, I did not feel the need to either `trim` here or include a `crush` feature on the message content as part of this story.

repose-aggregator/components/services/http-logging-service/http-logging-service-impl/src/main/scala/org/openrepose/core/services/httplogging/HttpLoggingContextMap.scala
```diff
@@ -0,0 +1,101 @@
...
+import org.openrepose.commons.utils.logging.TracingHeaderHelper
+
+import scala.collection.JavaConverters._
+import scala.language.implicitConversions
+import scala.util.Try
+
+// TODO: This utility is not strictly necessary. Since JTwig allows access to
+// TODO: Java objects via the selection operator, we could pass the context
+// TODO: object itself into the template. That would help remove some redundancy
+// TODO: but may or may not improve user experience.
```
dmnjohns: I just wanted to call attention to this note to get opinions. I ended up going with the context map because it's probably easier for an operator to use since the names and navigation are simplified. That being said, I love less code.

repose-aggregator/components/services/http-logging-service/http-logging-service-impl/src/main/scala/org/openrepose/core/services/httplogging/HttpLoggingContextMap.scala
```diff
@@ -0,0 +1,101 @@
...
+    "outboundResponseContentLength" -> (_.getOutboundResponseContentLength),
+    "outboundResponseBytesWritten" -> (_.getOutboundResponseBytesWritten),
+    "timeRequestReceived" -> (_.getTimeRequestReceived),
+    "timeToHandleRequest" -> (ctx => Duration.between(ctx.getTimeRequestReceived, ctx.getTimeRequestCompleted)),
+    "timeInOriginService" -> (_.getTimeInOriginService),
+    "localIpAddress" -> (_.getInboundRequest.getLocalAddr),
+    "remoteIpAddress" -> (_.getInboundRequest.getRemoteAddr),
+    "remoteHost" -> (_.getInboundRequest.getRemoteHost),
+    "traceId" -> (ctx => TracingHeaderHelper.getTraceGuid(ctx.getOutboundRequest.getHeader(CommonHttpHeader.TRACE_GUID))),
+    "extensions" -> (_.getExtensions)
```
dmnjohns: I'm just pointing out that these names are not single-character keys. I am personally a fan of the more descriptive keys, especially since JTwig seems built to use them (i.e., expressions are fully enclosed `{{ example }}` rather than being prefixed `%I`).

repose-aggregator/docs/src/asciibinder/services/http-logging.adoc
```diff
@@ -0,0 +1,322 @@
...
+| remoteHost
+| String
+| The fully qualified name of the client or the last proxy that sent the inbound request.
+
+| traceId
+| String
+| The ID associated with the inbound request used for tracing purposes.
+This value will match the value from the decoded `X-Trans-Id` header.
+
+| extensions
```
adrianjgeorge: Where are all the auth values? They were supposed to be explicitly exposed.

repose-aggregator/components/services/http-logging-service/http-logging-service-impl/src/main/scala/org/openrepose/core/services/httplogging/HttpLoggingConfigListener.scala
```diff
@@ -0,0 +1,115 @@
...
+import play.api.libs.json.Json
+
+import scala.collection.JavaConverters._
+import scala.util.{Success, Try}
+
+/**
+  * Handles configuration for the [[HttpLoggingService]].
+  */
+@Named
+class HttpLoggingConfigListener extends UpdateListener[HttpLoggingConfig] with StrictLogging {
```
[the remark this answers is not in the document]
dmnjohns: Indeed. I messed around with the organization here for a while, but ultimately I think this is what works best. Not only because of the size of the class, but also (and relatedly) because I can use dependency injection to unit test both this and the service.

repose-aggregator/components/services/http-logging-service/http-logging-service-impl/src/main/scala/org/openrepose/core/services/httplogging/HttpLoggingServiceImpl.scala
```diff
@@ -0,0 +1,90 @@
...
+    logger.trace("Destroyed HTTP Logging Service")
+  }
+
+  override def open(): HttpLoggingContext = {
+    val httpLoggingContext = new HttpLoggingContext()
+    logger.trace("Opened a new HTTP logging context {}", s"${httpLoggingContext.hashCode()}")
+    httpLoggingContext
+  }
+
+  override def close(httpLoggingContext: HttpLoggingContext): Unit = {
```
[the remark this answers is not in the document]
dmnjohns: I did. Doing so was an immediate solution to the problem with Jetty recycling the request object, though there were other options to solve that problem. Even so, the async bit made calling this method unpredictable, as async processing is, so I removed it.

The key points here are:

- Even without the async processing in this method, we should get away without adding any overhead to the response time. The reason is that the Jetty hook won't call this until the response has already been sent to the requester. That being said, with this approach we do tie up a Jetty processing thread rather than a thread from our own thread pool, but I don't suspect that will have much of an impact. If it does, we can always address it later.
- If a user of this API desires async processing, they can always implement it on their end. This way the user can decide whether or not to log asynchronously.

repose-aggregator/artifacts/valve/src/main/scala/org/openrepose/valve/Valve.scala
```diff
@@ -127,10 +127,9 @@ class Valve {
 
         //Set up all the valve context, it's lazy, so it won't get instantiated until here as well.
         valveContext.setParent(coreContext)
-        valveContext.scan("org.openrepose.valve.spring") //TODO: config file?
+        valveContext.scan("org.openrepose.valve") //TODO: config file?
```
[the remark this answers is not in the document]
dmnjohns: To pick up the class in `org.openrepose.valve` (the Jetty hooks that call into the logging service). Granted, I could have nested those classes somewhere in the `spring` package, but that didn't feel quite right to. Especially since the `org.openrepose.valve.jetty` package already exist.

Actually, I was a little careless here. This change will also cause this scan to pick up `ValvePortMXBeanImpl` which seems like it is being created manually currently (despite being annotated) when running in `test-mode`. However, the code to set up that bean includes the following comment:
> //If I'm in test mode, set up the port mbean... Heck, can probably always set it up

Oddly, a quick test shows that scanning this bean does not in fact set it up with the JMX. I am really not sure why that is the case, but I suspect we were having problems getting Spring to set up this JMX bean and that is why we ended up doing manually. That being the case, scanning this bean likely only creates a nearly empty object and nothing else. Still not ideal, but I can live with it. Especially since it would probably be more ideal to manage this bean with Spring (in which case scanning it would be warranted), we seemingly just did not get that working.

repose-aggregator/artifacts/valve/src/test/scala/org/openrepose/valve/jetty/HttpLoggingServiceChannelListenerTest.scala
```diff
@@ -0,0 +1,85 @@
...
+import org.junit.runner.RunWith
+import org.mockito.Mockito.{verify, when}
+import org.openrepose.commons.utils.http.CommonRequestAttributes
+import org.openrepose.core.services.httplogging.{HttpLoggingContext, HttpLoggingService}
+import org.scalatest.junit.JUnitRunner
+import org.scalatest.mock.MockitoSugar
+import org.scalatest.{BeforeAndAfterEach, FunSpec, Matchers}
+
+@RunWith(classOf[JUnitRunner])
+class HttpLoggingServiceChannelListenerTest extends FunSpec with BeforeAndAfterEach with MockitoSugar with Matchers {
```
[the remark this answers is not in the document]
dmnjohns: Do we? I mean, I will if it gets too long, but otherwise I don't bother. I am also not opposed to doing so.

I'll just go ahead and split the line.

repose-aggregator/commons/commons-utilities/src/main/scala/org/openrepose/commons/utils/logging/HttpLoggingContextHelper.scala
```diff
@@ -0,0 +1,45 @@
...
+import org.openrepose.commons.utils.http.CommonRequestAttributes
+import org.openrepose.core.services.httplogging.HttpLoggingContext
+
+object HttpLoggingContextHelper extends StrictLogging {
+
+  /**
+    * @param request the [[HttpServletRequest]] from which to extract a [[HttpLoggingContext]]
+    * @return the [[HttpLoggingContext]] for this request, or null if not available
+    */
+  def extractFromRequest(request: HttpServletRequest): HttpLoggingContext = {
```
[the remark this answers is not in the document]
dmnjohns: So yes, my concern was with writing this method to be friendly whether it is called from Java or Scala code. That mostly seems like a reasonable concern given that this lives in the `commons-utilities` project.

My thinking was that returning `null` is the most language-agnostic way of handling the possibility that the object does not exist on the request. For one, the `getAttribute` method on the request would return `null` in that case, so this matches that API in that way. For two, returning `Optional` or `Option` would lend itself to either Java or Scala, respectively. I have to say, `Optional` is pretty gross to work with in Scala 2.11 (though would be slightly improved with Scala 2.12. For three, the caller of this method can handle the return value in whatever way is most natural for the language they are working in; `Optional` for Java, `Option` for Scala, or even nullable types and safe calls with Kotlin.

I also dislike returning `null`, but in this case it seemed like the right compromise to me. So I did that and documented it.

repose-aggregator/components/services/http-logging-service/http-logging-service-impl/src/main/scala/org/openrepose/core/services/httplogging/HttpLoggingConfigListener.scala
```diff
@@ -0,0 +1,115 @@
...
+      case _ => super.convert(input)
+    }
+  }
+
+  private final def jsonValidationEnvConf(envConf: EnvironmentConfiguration): EnvironmentConfiguration = {
+    // @formatter:off
+    new EnvironmentConfigurationBuilder(envConf)
+        .value()
+          .withStringConverter(JsonValidationStringConverter)
+        .and()
```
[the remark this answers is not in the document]
dmnjohns: It is correct. The configuration object itself is hierarchical; traversal of that hierarchy is achieve by dropping into the relevant configuration (e.g., `value`) then returning to the parent configuration (via `and`) so that you can then drop into another configuration. That's just how the API is written, so I went with it.

repose-aggregator/components/services/http-logging-service/http-logging-service-impl/src/main/scala/org/openrepose/core/services/httplogging/HttpLoggingConfigListener.scala
```diff
@@ -0,0 +1,115 @@
...
+object HttpLoggingConfigListener extends StrictLogging {
+  private final val JsonValidationStringConverter: StringConverter = new DefaultStringConverter {
+    override def convert(input: Any): String = input match {
+      case Undefined.UNDEFINED => "123"
+      case _ => super.convert(input)
+    }
+  }
+
+  private final def jsonValidationEnvConf(envConf: EnvironmentConfiguration): EnvironmentConfiguration = {
+    // @formatter:off
```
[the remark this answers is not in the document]
dmnjohns: It is, yes. I think it's fair to include since our formatting is performed by IntelliJ (though not necessarily IntelliJ specific). We have used it in a few other places, too.

repose-aggregator/components/services/http-logging-service/http-logging-service-impl/src/main/scala/org/openrepose/core/services/httplogging/jtwig/HttpLoggingEnvironmentConfiguration.scala
```diff
@@ -0,0 +1,55 @@
...
+  def apply(format: Format): EnvironmentConfiguration = {
+    val initialEscapeEngine = EscapeEnginesByFormat.getOrElse(format, NoopEscapeEngineName)
+
+    // @formatter:off
+    EnvironmentConfigurationBuilder
+      .configuration()
+        .parser()
+          .syntax()
+            .withStartCode(StartCodeTag).withEndCode(EndCodeTag)
+          .and()
```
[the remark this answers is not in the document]
dmnjohns: Just for reference: https://github.com/rackerlabs/repose/pull/2055#discussion_r288160302

repose-aggregator/docs/src/asciibinder/architecture/logging.adoc
```diff
@@ -0,0 +1,3 @@
+= Logging
+
+include::./_includes/in-progress.adoc[]
```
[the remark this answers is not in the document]
dmnjohns: I created this page so that I could link to it from the service documentation. This is a page I would like for us to have that describes how we use SLF4J backed by Log4j2 as our logging solution with some details on how to customize and configure Repose logging, and perhaps even call out important named loggers. I wouldn't want to duplicate the documentation that exists for the log4j2 project, but we can certainly link to it.

I did not, however, complete this page since I believe it is outside the scope of this story.

Also, I did not add this page to the index or topic map. That was oversight, so I'll add those references now.

repose-aggregator/docs/src/asciibinder/services/http-logging.adoc
dmnjohns: Ah, I seem to have missed that. I see the `username` called out in the acceptance criteria, but are there any other auth values you want to expose explicitly? I will add whatever is missing but desired.

adrianjgeorge: I think we wanted impersonator info as well

repose-aggregator/docs/src/asciibinder/services/http-logging.adoc
```diff
@@ -353,7 +353,7 @@ Notably, this service does not manage:
 
 Management of those facets is left to developers intent on interacting with the logging system supported by this service.
 
-There many be instances when it is desired to add data to the logging context that is not part of the predefined data set.
+There may be instances when it is desired to add data to the logging context that is not part of the predefined data set.
```
dmnjohns: Oops. Nice catch.

repose-aggregator/components/services/http-logging-service/http-logging-service-impl/src/main/scala/org/openrepose/core/services/httplogging/HttpLoggingContextMap.scala
adrianjgeorge: The map is definitely more user friendly, so it's the right choice.

repose-aggregator/components/services/http-logging-service/http-logging-service-impl/src/main/scala/org/openrepose/core/services/httplogging/HttpLoggingContextMap.scala
adrianjgeorge: I concur

repose-aggregator/docs/src/asciibinder/services/http-logging.adoc
```diff
@@ -0,0 +1,359 @@
...
+| Integer
+| The size of the outbound response body in bytes.
+This value is taken from the `Content-Length` header of the response.
+For example, for a response with the header line `Content-Length: 1234`, this value would be `1234`.
+
+| outboundResponseBytesWritten
+| Integer
+| The number of bytes of the outbound response body written.
+
+| userId
```
adrianjgeorge: These values shouldn't be from the outbound request. They should be written directly by the keystone filter. In fact this implies if the keystone filter rejects the request for authorization reasons these wouldn't get populated, which is super not the behavior we want.

dmnjohns: While I agree in principle that we want to capture Identity data as soon as possible, I have a number of concerns about the implementation as you have described it:

- To my understanding, it has been the plan to moved authorization out of the Keystone v2 filter. In the case, the reasons for a request being rejected from the Keystone v2 filter should be limited to a failure to retrieve data, in which case we don't have data to populate anyway. To split hairs, it's true that we could succeed on the validation call and fail on a different call, though I believe we actually handle that right now not by rejecting the request, but by not populating the data we could not acquire. In other words, if the Keystone v2 filter will only reject a request if it cannot retrieve user data from Identity, then we don't have to worry about this issue, and that may be the case once we remove authorization from that filter.
- The difference in granularity is minimal (in-filter versus after-filter). Since we have a general way of capturing data after-filter, I certainly prefer relying on that over a filter-specific solution. Doing so reduces the amount of code, the code complexity, and the duplication of data.
- We can always change what data we capture and where we capture it in the future. Which is to say, I believe this implementation may fall outside the scope of this story. Certainly it is not specified in the acceptance criteria.

I understand that the primary concern is the case where the Keystone v2 filter rejects a request, but keeping in mind the points above, I believe that concern is addressed by the (temporarily) continued existence of delegation and by the plan to move authorization logic. When we remove delegation, we will need to make sure we have a way of reliably capturing Identity data, but I see that as work outside the scope of this story. Personally, I am not a fan of how inflated some of our initial implementation stories get, and this is one instance where I was trying to trim the fat. And in truth, I would like more time to consider this issue so that I can hopefully devise an acceptable general solution that does not involve instrumenting every data-producing filter that may reject a request.

dmnjohns: If we do decide to move forward with capturing user data as described, we would likely do so here:
https://github.com/rackerlabs/repose/blob/master/repose-aggregator/components/filters/keystone-v2-filter/src/main/scala/org/openrepose/filters/keystonev2/KeystoneV2Filter.scala#L379

Basically, as soon as we we have validated the token, grab the logging context and populate user data.

repose-aggregator/components/services/http-logging-service/http-logging-service-impl/src/main/scala/org/openrepose/core/services/httplogging/HttpLoggingContextMap.scala
```diff
@@ -0,0 +1,98 @@
...
+
+  def from(httpLoggingContext: HttpLoggingContext): Map[String, AnyRef] = {
+    // Removes mappings where a value could not be generated.
+    Generators
+      .mapValues(generator => Try(generator(httpLoggingContext)))
+      .mapValues(tryValue => tryValue.map(Option.apply).getOrElse(None))
+      .collect({ case (key, Some(value: AnyRef)) => key -> value })
+  }
+
+  private def headerMap(request: HttpServletRequest): util.Map[String, _ <: util.List[String]] = {
```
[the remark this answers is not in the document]
dmnjohns: While this would be a useful method to have on the wrappers, even if we did, we would still have to lowercase the keys here. That would probably look like:
```
request.getHeaderMap.asScala
      .map({ case (name, value) => name.toLowerCase -> value })
      .asJava
```

So while it's a useful thing in general, it probably wouldn't clean this code up much. Still, it would decrease duplicated code which is nice.

repose-aggregator/docs/src/asciibinder/services/http-logging.adoc
adrianjgeorge: SO to me this auth information was always one of the key pieces for acceptance. If it didn't get down into the acceptance criteria that was an oversight, because it was in the list originally. also despite the fact that we are splitting them in the future the fact remains that most of our users are on the single filter solution right now, and that use case has to work, especially since this work is meant to be a replacement for the existing http slf4j filter. Which actually reminds me that it should be deprecated as well.

dmnjohns: Wait, what? I thought we wanted to keep the `Slf4jHttpLoggingFilter` to be able to capture data at certain points in the filter chain? This service only keeps the latest inbound and outbound messages, while the SLF4J filter allows inspection/logging of messages at designated points in the filter chain.

As for the auth information, to be clear, we are providing a way to capturing and expose that data as it stands. What we are talking about here is really making it easier for the user to capture that data in the case where the user is using the Keystone v2 filter to perform authorization (i.e., not forcing the user to use delegation). However, given that authorization in Keystone v2 is already deprecated, it seems odd to me to write code to support that use case. As soon as Keystone v2 authorization goes away, that code is no longer necessary. As long as delegation exists, that code is not necessary. If delegation and Keystone v2 authorization are both being removed in `10.0.0.0`, then we don't have a problem.

As a quick aside, the user name was certainly part of the acceptance criteria, and it was reasonable to add user ID, impersonator ID, and impersonator name.

adrianjgeorge: SO yes, its usage has been deprecated, but while it hasn't been removed we still have to support the use case. This work is on master and will roll out on repose 9.x. I really don't expect anyone to move off of it till we force the issue in 10. Are we keeping the other filter? What purpose does that serve? The only real use case we know of it has been completely superseded by this. While there is a theoretical use for it, i don't see it as one people will actually use.

I totally get that it's unfortunate to add code we know will go away soon, but there's only so much change we want to push on the users at once.

dmnjohns: I guess in my view we are adding a feature with this service, and a new feature does not necessarily have to accommodate a deprecated feature. Keystone v2 filter authorization is a deprecated feature, as is delegation, and in conjunction they will continue to work until they are removed. Additionally, delegation allows the new HTTP Logging Service to work as expected alongside Keystone v2 filter authorization. So if the user is currently configured to do audit logging (using delegation), that will continue working. If they want to use the new service, that will also work (allowing them to drop the HERP filter, though the DeRP filter would still be necessary). If they wanted to drop delegation and continue using Keystone v2 filter authorization that is where trouble arises, and that is the use case we would not support (where the user is trying to move off of some deprecated functionality but not other deprecated functionality). Of course, at that point, the user is also updating their configuration, so it would likely be just as easy to drop all deprecated functionality (i.e., Keystone v2 filter authorization and delegation) and move to current functionality (i,.e., Keystone v2 Authorization filter and HTTP Logging Service).

So to my mind, this is less forcing a change on the user, and more incentivizing the user to stop relying on deprecated features to make using new features easier.

At any rate, I've finished implementing what you've asked for and will push it shortly.

wdschei: So the missing FIT case is using the Keystone v2 filter without delegation where authentication succeeds and authorization fails, then the `user`/`impersonator` `id`/`name` should all still be logged?

repose-aggregator/components/services/http-logging-service/http-logging-service-api/src/main/java/org/openrepose/core/services/httplogging/HttpLoggingContext.java
```diff
@@ -0,0 +1,62 @@
...
+ * indicate mutability and keep the behavior of methods clear.
+ * <p>
+ * This class should only be constructed by the {@link HttpLoggingService}.
+ * By only allowing the {@link HttpLoggingService} to construct this class,
+ * stronger assertions can be made about the tracking and integrity of any
+ * instances. For that reason, constructor access is limited to the package.
+ */
+@Data
+@RequiredArgsConstructor(access = AccessLevel.PACKAGE)
+public class HttpLoggingContext {
```
dmnjohns: This has remained a design problem for me. On the one hand, we need to pass a `Map` to JTwig. A `Map` is a good structure for associating tokens to data. On the other hand, to accommodate the breadth of data we capture, that `Map` would have to be fairly general (e.g., `Map<String, Object>`) which means we would lose type safety which makes working with this context a little more painful.

The current solution was to use a type-safe context object to capture data, then to convert that object to a `Map` to render a template. 

I believe that perhaps a better solution would be to have this context `implements Map<String, Object>`. The context would then delegate to an underlying `HashMap`. What about the type-safety, you ask? The context would add methods to access named, typed data (e.g., `public String getInboundResponseProtocol()`, `public void setInboundResponseProcotol(String protocl)`). That was we don't have to go through a conversion process, we simply build the correct `Map` instead. The context would also perform any necessary mapping (i.e., a method would take a request object, and map it to the corresponding fields like query string, headers, etc.). Extensions could be made directly against the underlying map, or to protect the data we are populating, we could continue to map the key `extensions` to a sub-map and mutate that.

repose-aggregator/docs/src/asciibinder/services/http-logging.adoc
dmnjohns: @wdschei That sounds correct to me.

repose-aggregator/components/services/http-logging-service/http-logging-service-api/src/main/java/org/openrepose/core/services/httplogging/HttpLoggingContext.java
adrianjgeorge: I'm on board but would probably suggest composition instead of inheritance in this case.

adrianjgeorge: I'm on board but would probably suggest composition instead of inheritance in this case.

dmnjohns: Funny enough, the more I think about that approach, the more it starts to look like what we already have. I'm not too concerned with making this change right now unless someone has a clear vision of a better solution.

PR #2057, opened 2019-05-16
REP-7776 Upgrading okhttp to resolve CVE-2018-20200

dmnjohns:
The dependency check plugin is flagging an issue with our transitive dependency on `okhttp` (via `jaeger-core`).

Unfortunately, upgrading `jaeger-core` does not resolve this issue for us at this time. So instead we upgrade the library directly, preferring that to suppressing the issue.


wdschei:
Retest this please.

PR #2060, opened 2019-06-19
REP-7875 Failing Performance test investigation.

wdschei:
The end result is that something in [Telegraf v1.11.0](https://github.com/influxdata/telegraf/releases/tag/1.11.0) breaks the performance tests.
The Ansible Role being used to install Telegraf, [rossmcdonald.telegraf](https://github.com/rossmcdonald/telegraf#role-variables), explicitly states it only supports installing stable/latest.
This PR uses a different Ansible Role, [juliendufresne.telegraf](https://github.com/juliendufresne/ansible-role-telegraf), to install the previous version of Telegraph, but this role seems to be abandoned.
They both use the MIT license and it should not be hard to add the functionality in [juliendufresne.telegraf](https://github.com/juliendufresne/ansible-role-telegraf) that is missing from [rossmcdonald.telegraf](https://github.com/rossmcdonald/telegraf).
Specifically, this would require:
* moving in [rossmcdonald.telegraf](https://github.com/rossmcdonald/telegraf) the [`/tasks/install-debian.yml`](https://github.com/rossmcdonald/telegraf/blob/master/tasks/install-debian.yml) to `/tasks/install-debian-packaged.yml`
* replace it with the [`/tasks/install-debian.yml`](https://github.com/juliendufresne/ansible-role-telegraf/blob/master/tasks/install-debian.yml) from [juliendufresne.telegraf](https://github.com/juliendufresne/ansible-role-telegraf).
* add the [`/tasks/install-debian-custom.yml`](https://github.com/juliendufresne/ansible-role-telegraf/blob/master/tasks/install-debian-custom.yml) from [juliendufresne.telegraf](https://github.com/juliendufresne/ansible-role-telegraf) to [rossmcdonald.telegraf](https://github.com/rossmcdonald/telegraf)
* open a PR for [rossmcdonald.telegraf](https://github.com/rossmcdonald/telegraf)

Should be easy enough, but you never know.
Feel free to close this PR until things are where they need to be.

For reference, a Jenkins job where this issue manifested has been marked to be kept forever. When this issue is resolved, it should be released:
https://jenkins.openrepose.org/job/performance-test/20733/

### DIFF
--- a/repose-aggregator/tests/performance-tests/src/performanceTest/resources/group_vars/all.yml
+++ b/repose-aggregator/tests/performance-tests/src/performanceTest/resources/group_vars/all.yml
@@ -45,7 +45,7 @@
       port: 2003,
     },
   },
-  telegraf_install_version: "1.10.4",
+  telegraf_install_version: "1.10.4-1",
   telegraf_influxdb_urls: [
     "udp://influxdb.performance.openrepose.org:8089",
   ],
--- a/repose-aggregator/tests/performance-tests/src/performanceTest/resources/group_vars/all.yml
+++ b/repose-aggregator/tests/performance-tests/src/performanceTest/resources/group_vars/all.yml
@@ -45,7 +45,7 @@
       port: 2003,
     },
   },
-  telegraf_install_version: "1.10.4-1",
+  telegraf_install_url: "https://dl.influxdata.com/telegraf/releases/telegraf_1.10.4-1_amd64.deb",
   telegraf_influxdb_urls: [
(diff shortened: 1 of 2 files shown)


dmnjohns:
The release of [Telegraf 1.11.1](https://github.com/influxdata/telegraf/releases/tag/1.11.1) fixes the bug that we were encountering.

It still is not ideal that we are unable to easily set the version of Telegraf we are using with the current plugin, but that's a concern for another day.

PR #2061, opened 2019-06-21
REP-7785 Hazelcast Datastore Service Configuration

dmnjohns:
Just the configuration for the Hazelcast Datastore Service for review. It should be safe to merge it, but it is not necessary since the work is already on my main working branch.

### DIFF
--- a/repose-aggregator/components/services/datastore-service/datastore-service-api/src/main/resources/META-INF/schema/config/hazelcast-datastore.xsd
+++ b/repose-aggregator/components/services/datastore-service/datastore-service-api/src/main/resources/META-INF/schema/config/hazelcast-datastore.xsd
@@ -1,0 +1,222 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  _=_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_=
+  Repose
+  _-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-
+  Copyright (C) 2010 - 2015 Rackspace US, Inc.
+  _-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+  
+       http://www.apache.org/licenses/LICENSE-2.0
+  
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+  =_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_=_
+  -->
+
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           xmlns:html="http://www.w3.org/1999/xhtml"
+           xmlns="http://docs.openrepose.org/repose/hazelcast-datastore/v1.0"
+           targetNamespace="http://docs.openrepose.org/repose/hazelcast-datastore/v1.0"
+           attributeFormDefault="unqualified"
+           elementFormDefault="qualified">
+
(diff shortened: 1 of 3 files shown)


repose-aggregator/components/services/datastore-service/datastore-service-api/src/main/resources/META-INF/schema/config/hazelcast-datastore.xsd
```diff
@@ -0,0 +1,222 @@
...
+        <xs:annotation>
+            <xs:documentation>
+                <html:p>
+                    Hazelcast Datastore settings.
+                </html:p>
+            </xs:documentation>
+        </xs:annotation>
+
+        <xs:choice>
+            <xs:element name="standard" type="StandardConfig"/>
```
adrianjgeorge: I don't really like this name. Maybe full or custom?

repose-aggregator/components/services/datastore-service/datastore-service-api/src/main/resources/META-INF/schema/config/hazelcast-datastore.xsd
```diff
@@ -0,0 +1,222 @@
...
+    </xs:complexType>
+
+    <xs:complexType name="Port">
+        <xs:annotation>
+            <xs:documentation>
+                <html:p>
+                    The ports which Hazelcast will use to communicate between cluster members. Its default value is
+                    5701.
+                </html:p>
+                <html:p>
```
adrianjgeorge: We'd normally put these notes on the attribute itself. What do these actually do? Can it not run off a single port?

repose-aggregator/components/services/datastore-service/datastore-service-api/src/main/resources/META-INF/schema/config/hazelcast-datastore.xsd
```diff
@@ -0,0 +1,222 @@
...
+                        build up the cluster. Its default value is 5.
+                    </html:p>
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+    </xs:complexType>
+
+    <xs:complexType name="Multicast">
+        <xs:all>
+            <xs:element name="multicast-group" type="xs:string" minOccurs="0" maxOccurs="1" default="224.2.2.3">
```
adrianjgeorge: These seem like they should be optional attributes instead of elements.

dmnjohns: While I would agree and can make that change, the reason they are `element`s is that is how they are defined in Hazelcast's configuration. That is, I left them as `element`s to make it as easy as possible to transition from our configuration to a full Hazelcast configuration.

repose-aggregator/components/services/datastore-service/datastore-service-api/src/main/resources/META-INF/schema/config/hazelcast-datastore.xsd
dmnjohns: It can run off a single port. I believe these attributes just enable Hazelcast to bind to more than one port. I know one case where it will do so is if the selected port is unavailable. I am not sure what cases would cause Hazelcast to bind to multiple ports.

See:
https://docs.hazelcast.org/docs/3.12/manual/html-single/index.html#port

repose-aggregator/components/services/datastore-service/datastore-service-api/src/main/resources/META-INF/schema/config/hazelcast-datastore.xsd
dmnjohns: I chose "standard" since this choice enables the usage of one of the official, standard methods of configuring Hazelcast. I grant you, however, that a standard Hazelcast configuration is not necessarily a standard Hazelcast Datastore configuration.

As for alternatives, I dislike "custom" very much in this case. Any configuration makes for a custom deployment. I suppose I'm fine with "full". Or maybe "traditional"?

repose-aggregator/components/services/datastore-service/datastore-service-api/src/main/resources/META-INF/schema/config/hazelcast-datastore.xsd
dmnjohns: Oh, as for the notes, that was just the result of copy-pasting the type definition. I will move them.

adrianjgeorge: yeah, i think single port will do.

repose-aggregator/components/services/datastore-service/datastore-service-api/src/main/resources/META-INF/schema/config/hazelcast-datastore.xsd
adrianjgeorge: unfortunate but understandable.

dmnjohns:
Since I think I received the feedback I was looking for with this PR, I am going to close it and open a PR on the main branch instead.

PR #2064, opened 2019-06-27
REP-7920 Signature publishing fix

adrianjgeorge:


repose-aggregator/src/config/dependency-check/OWASP-Suppression.xml
```diff
@@ -26,8 +26,22 @@
             Therefore the transitive dependency on an old Jetty version with a known vulnerability that is used to support polling does not expose the potential DOS attack vector in this instance.
         ]]></notes>
         <gav regex="true">^org\.mortbay\.jetty:jetty:.*$</gav>
+        <cve>CVE-2009-1523</cve>
         <cve>CVE-2011-4461</cve>
     </suppress>
+    <suppress>
+        <notes><![CDATA[
+            file name: jetty-6.1.26.jar
+            This vulnerability is for version 9.x but is false positiving on 6.1.26.
```
dmnjohns: I'm not sure that this note is correct.

`repose-core` has a dependency on `flume-ng-sdk` which leads to the following dependency chain:
```
+--- org.apache.flume:flume-ng-sdk -> 1.9.0
|    |    |    +--- org.slf4j:slf4j-api:1.7.25
|    |    |    +--- org.apache.avro:avro:1.7.4
|    |    |    |    +--- org.codehaus.jackson:jackson-core-asl:1.8.8 -> 1.9.13
|    |    |    |    +--- org.codehaus.jackson:jackson-mapper-asl:1.8.8 -> 1.9.13
|    |    |    |    |    \--- org.codehaus.jackson:jackson-core-asl:1.9.13
|    |    |    |    +--- com.thoughtworks.paranamer:paranamer:2.3 -> 2.8
|    |    |    |    +--- org.xerial.snappy:snappy-java:1.0.4.1
|    |    |    |    +--- org.apache.commons:commons-compress:1.4.1
|    |    |    |    |    \--- org.tukaani:xz:1.0
|    |    |    |    \--- org.slf4j:slf4j-api:1.6.4 -> 1.7.25
|    |    |    +--- org.apache.avro:avro-ipc:1.7.4
|    |    |    |    +--- org.apache.avro:avro:1.7.4 (*)
|    |    |    |    +--- org.codehaus.jackson:jackson-core-asl:1.8.8 -> 1.9.13
|    |    |    |    +--- org.codehaus.jackson:jackson-mapper-asl:1.8.8 -> 1.9.13 (*)
|    |    |    |    +--- org.mortbay.jetty:jetty:6.1.26
|    |    |    |    |    \--- org.mortbay.jetty:jetty-util:6.1.26
|    |    |    |    +--- org.mortbay.jetty:jetty-util:6.1.26
|    |    |    |    +--- org.apache.velocity:velocity:1.7
|    |    |    |    |    +--- commons-collections:commons-collections:3.2.1 -> 3.2.2
|    |    |    |    |    \--- commons-lang:commons-lang:2.4
|    |    |    |    \--- org.slf4j:slf4j-api:1.6.4 -> 1.7.25
|    |    |    +--- io.netty:netty:3.10.6.Final
|    |    |    \--- org.apache.thrift:libthrift:0.9.3 -> 0.12.0 (*)
```

The reason that the transitive dependency isn't being overridden is that it is a dependency on `org.mortbay.jetty` while Repose directly depends on `org.eclipse.jetty`.

With that in mind, the suppression is still probably valid since I don't believe a Jetty server is being exposed, but we may want to update the note. Or we can mess with our transitive dependencies, but there's a good chance that would get messy.

build.gradle
```diff
@@ -448,7 +448,7 @@ subprojects {
         publications {
             nebula(MavenPublication) {
                 if(project.hasProperty('release')) {
-                    signArchives.signatures.each {
+                    signArchives.signatures.findAll({!it.toSign.canonicalPath.contains("deb") && !it.toSign.canonicalPath.contains("rpm")}).each {
```
dmnjohns: Do we want a story to look at this in the future? I feel like there has to be a way to publish multiple signatures.

repose-aggregator/src/config/dependency-check/OWASP-Suppression.xml
adrianjgeorge: if you look at the actual vulnerabilities they say they only apply to version 9.x of jetty, but the report was pinging them for 6.1.26. It's worth noting that the maven coordinates changed for jetty in-between, so this suppression should just turn them off for 6.1.26 where they are a false positive.

build.gradle
adrianjgeorge: You can but it requires doing a bunch of crap with classifiers, because we do in fact publish multiples where we are using classifiers (the fat jar). these artifacts are already internally signed, so i felt publishing a completely secondary signature for them was unnecessary. Note that this change would allow any other properly classified artifacts we add later to just work.

repose-aggregator/src/config/dependency-check/OWASP-Suppression.xml
```diff
@@ -77,4 +91,42 @@
...
+        <gav regex="true">^com\.sleepycat:je:.*$</gav>
+        <cpe>cpe:/a:oracle:berkeley_db</cpe>
+        <cpe>cpe:/a:oracle:oracle_berkeley_db</cpe>
+    </suppress>
+    <suppress>
+        <notes><![CDATA[
+            file name: jetty-util-6.1.26.jar
+            NOTE: Misidentified as Jetty itself.
+        ]]></notes>
+        <gav regex="true">^org\.mortbay\.jetty:jetty-util:.*$</gav>
```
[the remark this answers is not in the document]
adrianjgeorge: Lol, this was your contribution.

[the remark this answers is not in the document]
dmnjohns: Heh, I know. I switched to `gav` because I forgot to update the namespace and so `packageUrl` wasn't working. Which works, but so does updating the namespace, so we may as well be consistent.

repose-aggregator/src/config/dependency-check/OWASP-Suppression.xml
dmnjohns: I see the following in the report descriptions, all of which seem like they would cover Jetty 6. Although some specify "Eclipse Jetty", so I'm not certain those would apply to Mortbay Jetty.

> In Eclipse Jetty, versions 9.2.x and older,
> In Eclipse Jetty Server, versions 9.2.x and older,
> Jetty through 9.4.x
> In Eclipse Jetty version 7.x, 8.x, 9.2.27 and older,

I think pinging for Jetty 6.1.26 is legit. I also think the suppression is legit, but that's because I don't believe we expose any instances of Jetty running that version. I would just suggest that the note say something about where Jetty 6.1.26 is coming from and why we can ignore it.

Ideally, we would cut this Jetty dependency out since it should be unused bloat, but that may not be easy to accomplish. I'm not sure what `avro-ipc` is doing with Jetty, exactly. I assume it's used if you want to listen for Avro data, but we should only ever be transmitting, so I doubt we need it.

build.gradle
dmnjohns: Fair enough.

repose-aggregator/src/config/dependency-check/OWASP-Suppression.xml
adrianjgeorge: it's not eclipse jetty in version 6. Additionally note how they call out the major versions but leave out 6. When compared to vulnerabilities that do apply to 6 you can tell that this verbiage is not meant to apply there. And we do have a suppressed vulnerability right above this that explains how this transitive is not exposed.

